PR #25, opened 2016-09-23
Add HEALTHCHECK instruction

### LINKED ISSUE #23: Add support for new HEALTHCHECK instruction
I added a HEALTHCHECK instruction to all of my Dockerfiles recently and it didn't show up as a suggestion or highlight properly so I'm guessing the support just isn't there at all yet.  This is not surprising since it's new for 1.12

You can find more details about the HEALTHCHECK instruction at: https://docs.docker.com/engine/reference/builder/#/healthcheck

audioboxer217:
This update should only add in support for the new "HEALTHCHECK" instruction introduced in Docker 1.12.  This will close #23 that I opened.

### DIFF
--- a/snippets/dockerfile.json
+++ b/snippets/dockerfile.json
@@ -114,7 +114,7 @@
     "HEALTHCHECK": {
         "prefix": "HEALTHCHECK",
         "body": [
-            "HEALTHCHECK --interval=30s --timeout=30s --retries=3 CMD ${command}"
+            "HEALTHCHECK CMD ${command}"
             "$0"
         ],
         "description": "HEALTHCHECK"
--- a/snippets/dockerfile.json
+++ b/snippets/dockerfile.json
@@ -110,5 +110,14 @@
             "$0"
         ],
         "description": "ONBUILD"
+    },
+    "HEALTHCHECK": {
+        "prefix": "HEALTHCHECK",
+        "body": [
+            "HEALTHCHECK --interval=30s --timeout=30s --retries=3 CMD ${command}"
+            "$0"
+        ],
+        "description": "HEALTHCHECK"
     }
+}
 }


snippets/dockerfile.json
```diff
@@ -110,5 +110,14 @@
             "$0"
         ],
         "description": "ONBUILD"
+    },
+    "HEALTHCHECK": {
+        "prefix": "HEALTHCHECK",
+        "body": [
+            "HEALTHCHECK --interval=30s --timeout=30s --retries=3 CMD ${command}"
```
saurabsa: >--interval=30s --timeout=30s --retries=3 [](start = 25, length = 40)

Do we need to provide these defaults? Does any of the above Keywords have any defaults?

audioboxer217:
Those actually are the default values for those options.  I wasn't sure of the best way to handle including them for reference so I went with that. However, I'm open to suggestions.  None of the other entities had similar situations for me to look at as a reference so I wasn't sure if `--interval=${interval} --timeout=${timeout} --retries=${retries}` would work.

snippets/dockerfile.json
audioboxer217: Those actually are the default values for those options. I wasn't sure of the best way to handle including the options for reference so I went with that. However, I'm open to suggestions. 

None of the other entities had similar situations for me to look at as a reference so I wasn't sure if `--interval=${interval} --timeout=${timeout} --retries=${retries}` would work, but I also didn't want to leave them out completely because then a user would have to go look them up if they needed them which seems to defeat the purpose of a helper.

audioboxer217: Any thoughts on this @saurabsa? As I mentioned, I'm open to whatever works best.  The way I see it there are a few options.

1. list them with their default values as I did in my original commit
2. list them with variables as I mentioned in my last post (e.g. --interval=${interval})
3. leave them out completely.  As I mentioned, my only issue with this is that there would then be no reference to these options in the helper, however, it may be that that is the best option.

If there are any other options please let me know.

Thank you for your considerations.

lostintangent:
@audioboxer217 Thanks for the contribution!

PR #30, opened 2016-10-24
Chrisdias/packagejson

chrisdias:


commands/start-container.ts
```diff
@@ -19,4 +19,11 @@ export function startContainer() {
 
 export function startContainerInteractive() {
     doStartContainer(true);
+}
+
+export function startAzureCLI() {
+    let terminal: vscode.Terminal = vscode.window.createTerminal('Azure CLI');
+    terminal.sendText(`docker run -it --rm azure-xplat-cli:latest`);
```
lostintangent: Do we want to use the `microsoft/azure-cli` image? I tried running this command as-is and the `azure-xplat-cli` couldn't be found.

chrisdias:
good catch, i had a local image i was playing with.  fixed with https://github.com/Microsoft/vscode-docker/commit/27fe9d82bb73d6367ebefa9b75efe30d3e9536e3

commands/start-container.ts
```diff
@@ -19,4 +19,11 @@ export function startContainer() {
...
+    terminal.show();
+    terminal.sendText(`azure login`);
```
lostintangent: Did you want to prompt for login via `az login`? Maybe the login sessions between the 1.0 and 2.0 CLI versions are interoperable, and so it doesn't matter, but I just thought I'd double-check.

Issue #17, opened 2016-03-24
Support docker-compose V2 syntax?


tkvangorder:
Are there any plans to support the V2 syntax for docker compose? 

https://docs.docker.com/compose/compose-file/#versioning

SteveLasker:
Hi Tyler,
Yes, this is something we've been looking at with our other docker tools and need to round back here. 
Is there specific features your looking to scaffold out?

tkvangorder:
Just looking for intellisense for a V2 file. Thanks!

SteveLasker:
Ahh, ok. So your looking for an updated language service. Not just updated scaffolding from yo docker.

miguelchico:
Any update about V2 support?

SteveLasker:
Unfortunately, the language service isn't something we've had a chance to prioritize. We've been working on our Windows container support, perf and consolidating the experience between vs code and vs. 
as an open source project, we would be very interested if someone wanted to propose some updates, have a quick review and take a pull request.

lostintangent:
@tkvangorder This has been resolved via #34. If you want, you can pull the latest bits from this repo, and build/install the extension locally. Otherwise, it will get picked up in the next marketplace release. Thanks so much for the feedback!

PR #34, opened 2016-10-27
Adding completion/hover support for compose v2 files

lostintangent:
This addresses #17 by updating the completion provider for compose files in the following ways:
1. Updates the set of recommended properties to match the v1 spec (e.g. `hostname`, `cpu_shares`). 
2. Adds support for the new properties in v2 files, and provides the correct completion results depending on the version of the compose file being edited (e.g. If you're editing a v1 file, you won't see a `log_driver` keyword recommendation).

screen shot 2016-10-27 at 12 31 12 pm

screen shot 2016-10-27 at 12 31 30 pm

Additionally, this updates the hover provider to display the help text for both v1 and v2 keys. 

Longer term, we should investigate improving the way that auto-completion works, so that we can display context-based recommendations, which would make this experience much better and less error-prone. I logged #35 to track this enhancement.

Additionally, it would be great if we could base the completion/hover behavior on the Docker Compose JSON schema files, as opposed to needing to define the schema ourselves. Once we did that, we could also introduce linting support and other nice features.

_NOTE: I also took the liberty of alpha-sorting the schema definition for compose files, which is the cause of a bit of churn in this PR._

### DIFF
--- a/dockerCompose/dockerComposeCompletionItemProvider.ts
+++ b/dockerCompose/dockerComposeCompletionItemProvider.ts
@@ -6,7 +6,7 @@
 
 import {TextDocument, Position, CancellationToken, CompletionItem, CompletionItemProvider, CompletionItemKind, Uri} from 'vscode';
 import helper = require('../helpers/suggestSupportHelper');
-import {DOCKER_COMPOSE_KEY_INFO} from './dockerComposeKeyInfo';
+import {DOCKER_COMPOSE_V1_KEY_INFO, DOCKER_COMPOSE_V2_KEY_INFO} from './dockerComposeKeyInfo';
 import hub = require('../dockerHubApi');
 
 export class DockerComposeCompletionItemProvider implements CompletionItemProvider {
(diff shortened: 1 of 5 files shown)


dockerCompose/dockerComposeCompletionItemProvider.ts
```diff
@@ -55,12 +59,14 @@ export class DockerComposeCompletionItemProvider implements CompletionItemProvid
         return Promise.resolve([]);
     }
 
-    private suggestKeys(word: string): CompletionItem[] {
-        return Object.keys(DOCKER_COMPOSE_KEY_INFO).map(ruleName => {
+    private suggestKeys(word: string, useV2Schema: boolean = false): CompletionItem[] {
```
sparty02: Can I suggest passing in a version (string or enum) instead of the ```useV2Schema``` boolean?  Then you could do a switch below, vs a ternary to get the appropriate keyset.  This will make this more amenable to future versions of the docker-compose.yml schema.

lostintangent: We could definitely do that, but I generally think the way that auto-completion is implemented isn't the best long-term solution, so I just went with a simple fix for getting v2 support added. I filed #36 to track adding support for Compose v2.1 files, which would definitely necessitate tweaking how this works, and would be pretty trivial to do once needed.

sparty02: I opened a [PR](https://github.com/lostintangent/vscode-docker/pull/1) to address this, but I'll let you decide if you want it or not!

lostintangent: I actually just sent out a new commit that addressed your feedback, and cleaned up the way that the version keys are composed, that I think makes it a little easier to read. Let me know what you think!

lostintangent:
@chrisdias Any thoughts on this?

PR #42, opened 2016-12-01
Chrisdias/linting

chrisdias:
adds linting for dockerfiles using dockerfile_lint library


linting/dockerLinting.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    return;
+}
+
+function doValidate(document: vscode.TextDocument) {
+
+    let diagnostics: vscode.Diagnostic[] = [];
+
+    let linterRuleFile = configOptions.get('linterRuleFile', 'basic_rules.yaml');
+
```
lostintangent: It looks like this lib will actually default to using this exact rules file if you specific a null/empty rules file path, so we could choose not to ship any of these sample rules file, and allow either turning on the default rules or specifying a custom one.

.gitignore
```diff
@@ -1,2 +1,3 @@
-out
+out/**/*.js
+out/**/*.map
```
chrisdias: actually it does use the .yaml files, they are the linting rules.  i had to change gitignore to persist them. happy if you have another way to do this.

linting/dockerLinting.ts
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    if (result.warn.count > 0) {
+        for (let i = 0; i < result.warn.count; i++) {
+            diagnostics.push(new vscode.Diagnostic(new vscode.Range(0, 0, 0, 0), result.warn.data[i].message + ': ' + result.warn.data[i].description, vscode.DiagnosticSeverity.Warning));
+        }
+    }
+
+    diagnosticCollection.set(document.uri, diagnostics);
```
chrisdias: also forgot to do this :)

linting/rules/basic_rules_atomic.yaml
```diff
@@ -0,0 +1,287 @@
...
+        - 
+          label: "is_latest_tag"
+          regex: /latest/
+          level: "error"
+          message: "base image uses 'latest' tag"
+          description: "using the 'latest' tag may cause unpredictable builds. It is recommended that a specific tag is used in the FROM line or *-released which is the latest supported release."
+          reference_url: 
+            - "https://docs.docker.com/reference/builder/"
+            - "#from"
+          label: "no_tag"
```
chrisdias: these came with dockerfile_lint so i don't know.  but, we can remove all these per the above comments.

chrisdias:
great feedback!  i've still got work to do, will make changes and then resubmit.

dockerExtension.ts
```diff
@@ -54,4 +58,16 @@ export function activate(ctx: vscode.ExtensionContext): void {
     ctx.subscriptions.push(vscode.commands.registerCommand('vscode-docker.container.open-shell', openShellContainer));
     ctx.subscriptions.push(vscode.commands.registerCommand('vscode-docker.compose.up', composeUp));
     ctx.subscriptions.push(vscode.commands.registerCommand('vscode-docker.compose.down', composeDown));
+
+    diagnosticCollection = vscode.languages.createDiagnosticCollection('docker-diagnostics');
```
chrisdias: fixed

linting/rules/basic_rules_atomic.yaml
chrisdias: removed

linting/dockerLinting.ts
chrisdias: fixed

linting/dockerLinting.ts
chrisdias: "fixed" by not relying on file...

.gitignore
chrisdias: fixed

lostintangent:
LGTM!

Issue #36, opened 2016-10-27
Add completion support for compose v2.1 files


lostintangent:
[Docker Compose v1.9.0](https://github.com/docker/compose/releases/tag/1.9.0-rc1) introduces a new v2.1. compose file format, so we should introduce completion/hover/etc. support for the new properties it includes.

lostintangent:
Resolved via #50

PR #50, opened 2016-12-20
Adding auto-complete support for compose v2.1 files

lostintangent:
This addresses #36 by enabling auto-complete support for the six new properties that were added to the compose v2.1 format (that was released with Docker Compose v1.9), and ensuring that they are only suggested for files that are specifically using that version.

*Note: I also deleted the vsix from the top-level of the repo since we didn't need to version control it.*

CC @chrisdias


lostintangent:
@chrisdias I believe this PR is already removing the `.vsix` file. Should I revert this change? I wasn't sure whether this file was needed so I just decided to delete it.

chrisdias:
Doh! I read it backwards ☺  i didn’t realize there was  vsix in the repo already so I automatically assumed this was a small oversight.

Sorry!

lostintangent:
@chrisdias Cool :) Are you good with this PR then?

PR #53, opened 2017-01-04
Chrisdias/51v2

chrisdias:
Made a better fix for #51 and bumped the version number so I can publish.


chrisdias:
I found another path problem with the fix for 51 which is why this is "better" :)

lostintangent:
The code is cleaner is well, so it's a nice change all around 👍

PR #58, opened 2017-01-30
Add Windows support for shell attach

lostintangent:
This change updates the `Docker: Attach Shell` command in order to properly support Windows containers. Previously, this command assumed that it could run `/bin/sh` in the selected image, whereas this change will detect a Windows container, and run `cmd` instead.

Beyond this change, we should allow configuring the default shell use, so that Windows users could toggle PowerShell and Linux users could use Bash/etc.

### DIFF
--- a/commands/open-shell-container.ts
+++ b/commands/open-shell-container.ts
@@ -1,12 +1,20 @@
-import vscode = require('vscode');
+import * as vscode from 'vscode';
 import {ContainerItem, quickPickContainer} from './utils/quick-pick-container';
+import {DockerEngineType, docker} from './utils/docker-endpoint';
+
+const engineTypeShellCommands = {
+    [DockerEngineType.Linux]: "/bin/sh",
+    [DockerEngineType.Windows]: "cmd"
+}
 
 export function openShellContainer() {
-    quickPickContainer().then(function (selectedItem: ContainerItem) {
+    quickPickContainer().then((selectedItem: ContainerItem) => {
         if (selectedItem) {
-            let terminal = vscode.window.createTerminal(`sh ${selectedItem.label}`);
-            terminal.sendText(`docker exec -it ${selectedItem.ids[0]} /bin/sh`);
-            terminal.show();
+            docker.getEngineType().then((engineType: DockerEngineType) => {
+                const terminal = vscode.window.createTerminal(`Shell: ${selectedItem.label}`);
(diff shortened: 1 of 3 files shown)


commands/open-shell-container.ts
```diff
@@ -1,12 +1,20 @@
-import vscode = require('vscode');
+import * as vscode from 'vscode';
 import {ContainerItem, quickPickContainer} from './utils/quick-pick-container';
+import {DockerEngineType, docker} from './utils/docker-endpoint';
+
+const engineTypeShellCommands = {
+    [DockerEngineType.Linux]: "/bin/sh",
+    [DockerEngineType.Windows]: "cmd"
```
lostintangent: Should this default to Powershell instead? I'd love to hear feedback here.

heaths: I keep hearing rumor (I have no inside knowledge one way or another..even though I couldn't share it anyway) that Windows is making the switch themselves, but I question whether that would be accepted in general. I think as long as this was an option for the extension, though, you could default to whatever you want.

chrisdias: here is the blog post: https://blogs.windows.com/windowsexperience/2016/11/17/announcing-windows-10-insider-preview-build-14971-for-pc/#hjIQ7kQcdSPrzb1z.97

Given this, i suggest we default to PowerShell instead of cmd.

chrisdias:
@lostintangent all changes look good, but lets default to PS instead of CMD (see comment above).

commands/open-shell-container.ts
lostintangent: @chrisdias Sounds good to me! I'll make this change.

heaths:
@chrisdias understandable (glad you found the link; I remember hearing something along those lines), but still seems it should be a user option, because:

1. The terminal type for platforms is already an option for VSCode generally, and
2. Even people on Linux may not want sh, but bash csh, ec.

I am partial to PowerShell myself, but just thinking about what others may want. The default shell for Windows containers is cmd, so users may still want to default that.

chrisdias:
@heaths I created new issue https://github.com/Microsoft/vscode-docker/issues/60 to track making the shell configurable.

lostintangent:
Merging this in and will address the shell configuration in a separate PR. Thanks for the review!

Issue #56, opened 2017-01-28
Debug Container


galvesribeiro:
Hello guys,

I would like to know if this extension would accept a PR adding debugging capabilities for a container...

Initially the idea is to make it debug .Net Core applications that are hosted in (multiple) containers and the debugger could be eventually extended to support other types of applications like node.js for example...

If that is the case, I can work on the .Net Core part initially and later on extend it for other techs.

The idea is to make something similar to the experience we have on VS2017 and docker tooling there where you can add the initial docker files (this extension does the very basic already), create the whole infra from the docker (compose) files and then attach a debugger to it.

How does it sound?

Thanks

lostintangent:
@galvesribeiro PRs are always welcome! I'm not too familiar with the VS2017 experience, so any feedback/thoughts you have on how to improve this extension would be awesome.

galvesribeiro:
Hello @lostintangent, thanks for the promptly reply.

Have a look on this post https://blogs.msdn.microsoft.com/webdev/2016/11/16/new-docker-tools-for-visual-studio/

It show how it work on Visual Studio. Basically you right click a project or solution and add Docker support to it. Once it is added, just hit F5 and all the "magic" happens and you have a debugger attached to your target container. On VS2017 it even support multiple containers which are running on `docker-compose` scenarios so you can debug a bunch of containers part of same compose file at same time...

chrisdias:
For Node, the extension generates a docker-compose.debug.yml file that will open both the site and debugger ports and run node with the debug switch. you could automate the debug process by running docker-compose up on the yml file as a pre launch task and then VS Code should be able to attach to it.

For C#, the same pattern could be used... generate the right task file, docker-compose.debug.yml, and then connecting them in launch.json.

spboyer:
@lostintangent look at my repo here: https://github.com/spboyer/dockerdebugapp
This has the base images and proper debug compose images to attach the debugger to VS Code. There are changes needed to the extension to generate these files for .NET Core.

chrisdias:
Cool!

galvesribeiro:
Guys I just illustrated the docker-compose as an example. Your approaches work well for a single project, without a solution and that has a single container to attach. The idea is to press F5 and attach a debugger to **all** services inside docker-compose file.

After multiple try and error, I found a better way that doesn't rely on scripts/hacks to suppress the lack of the feature in VSCode.

This is a sample `docker-compose.yml` at the solution level:

```yaml
[...]

PR #61, opened 2017-02-08
Update Docker templates for .NET Core to bring parity with Visual Studio 2017

spboyer:
Partially addresses #56 

There are 2 pieces that need to be added.

1. launch.json needs to be created or inserted (if exists)
2. tasks.json needs to be created or inserted (if exists)

tasks.json - https://github.com/spboyer/dockerdebugapp/blob/master/.vscode/tasks.json
launch.json - https://github.com/spboyer/dockerdebugapp/blob/master/.vscode/launch.json

Need API help for 'Add configuration' for these.

@chrisdias @SteveLasker

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,7 @@
 			"args": ["--extensionDevelopmentPath=${workspaceRoot}" ],
 			"stopOnEntry": false,
 			"sourceMaps": true,
-			"outDir": "${workspaceRoot}/out",
+			"outFiles": ["${workspaceRoot}/out"],
 			"preLaunchTask": "npm"
 		}
 	]
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -57,14 +57,14 @@ EXPOSE ${port}
         case '.net core':
 
             return `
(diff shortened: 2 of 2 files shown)


chrisdias:
Thanks for the PR!

Can we separate this out into changes for the dockerfile for .NET and the generation of the PS and bash scripts?  When we first added support for commands we explicitly didn't include these scripts as they are mostly exposed through commands in the extension. Could we instead look at improving  the extension (e.g. settings) to provide any missing support that the scripts provide?

//cc @lostintangent

galvesribeiro:
Guys, did you saw my last comment on #56? We don't need scripts at all.

spboyer:
@galvesribeiro What @chrisdias is saying here is to do it in steps. Let's update the Dockerfile(s) for .NET Core and then update the built in commands for to add the configurations for the launch.json and tasks.json. 

It's a combination of both solutions, but with some simplicity.

galvesribeiro:
AH ok.

> Can we separate this out into changes for the dockerfile for .NET and the generation of the PS and bash scripts?

This question mislead me to think that you guys would still use that PS/bash scripts. Sorry :)

galvesribeiro:
Will it not be merged/fixed?

Issue #59, opened 2017-01-31
Add command to run a system prune


lostintangent:
Docker 1.13 added the `docker system prune` command, which is amazingly useful for development, since it cleans up all unused containers, images, volumes, networks, etc. in one command. We should consider adding a command to this extension for it (e.g. `Docker: System Prune`).

lostintangent:
Closed via #59. Thanks @chrisdias!

PR #64, opened 2017-02-10
add docker system prune #59

chrisdias:


commands/system-prune.ts
```diff
@@ -0,0 +1,9 @@
+import vscode = require('vscode');
+
+export function systemPrune() {
+
+    let terminal = vscode.window.createTerminal("docker system prune");
```
chrisdias: i thought about this too, but then it opens a can of worms... do you start doing this for all commands?  the nice thing is that since it simply shells to the terminal, if you dont have 1.13 then the command will simply fail there.

lostintangent: Yep I totally agree. That's why I didn't feel too strongly and wanted to check my thinking on it with your thoughts. Thanks dude!

PR #65, opened 2017-02-13
check to ensure running linux containers on windows

chrisdias:
i should have bunched up other fixes into a single pull request. this fix got a little bigger so would appreciate a review, thx.

### DIFF
--- a/commands/start-container.ts
+++ b/commands/start-container.ts
@@ -1,5 +1,7 @@
 import vscode = require('vscode');
-import {ImageItem, quickPickImage} from './utils/quick-pick-image';
+import { ImageItem, quickPickImage } from './utils/quick-pick-image';
+import { DockerEngineType, docker } from './utils/docker-endpoint';
+import * as cp from 'child_process';
 
 
 function doStartContainer(interactive: boolean) {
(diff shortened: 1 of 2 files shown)


commands/start-container.ts
```diff
@@ -22,13 +24,33 @@ export function startContainerInteractive() {
...
+                    if (!selected || selected.isCloseAffordance) {
+                        return;
+                    }
+                    return cp.exec('start https://docs.docker.com/docker-for-windows/#/switch-between-windows-and-linux-containers');
+                })
+            };
+        });
+    } else {
+        let option: string = process.platform === 'linux' ? '--net=host' : '';
+        let cmd: string = `docker run ${option} -it --rm azuresdk/azure-cli-python:latest`;
```
lostintangent: I'm not sure whether it's a good idea, but we could potentially mound the `~/.azure` directory into the container, so that your login sessions and configuration settings (e.g. default output type, cloud type) persisted across runs of this command.

chrisdias: I didn't think of that... Wicked good idea. If you know the folder location on any of the platforms let me know otherwise I can figure it out.

chrisdias: I didn't think of that... Wicked good idea. If you know the folder location on any of the platforms let me know otherwise I can figure it out.

chrisdias:
@lostintangent i added mapping for the `.azure` folder and it works great!  can you take one more look?

chrisdias:
ping @lostintangent  :)

PR #66, opened 2017-02-20
Use docker-machine from environment variables

tasosval:
The dockerode library that is used by the extension already has the necessary code to connect to docker-machine instances, as long as there is a valid  DOCKER_HOST environment variable set. 

We can take advantage of this fact and let users connect to other docker daemons. 

Tested with VSCode on windows with:
- a secondary local docker-machine (HyperV)
- remote docker (setup on ubuntu and manually connected as a docker-machine)

### DIFF
--- a/commands/utils/docker-endpoint.ts
+++ b/commands/utils/docker-endpoint.ts
@@ -9,7 +9,10 @@ class DockerClient {
     private endPoint:Docker;
 
     constructor() {
-        if (process.platform === 'win32') {
+        if (process.env.DOCKER_HOST){
+            this.endPoint = new Docker();
+        }
+        else if (process.platform === 'win32') {
             this.endPoint = new Docker({ socketPath: "//./pipe/docker_engine" });
         } else {
             this.endPoint = new Docker({ socketPath: '/var/run/docker.sock' });
--- a/commands/utils/docker-endpoint.ts
+++ b/commands/utils/docker-endpoint.ts
@@ -10,7 +10,7 @@ class DockerClient {
 
     constructor() {
         // Pass no options so that the defaultOpts of docker-modem will be used
-        this.endPoint = new Docker(null);
+        this.endPoint = new Docker();
     }
 
     public getContainerDescriptors(): Thenable<Docker.ContainerDesc[]>{
--- a/commands/utils/docker-endpoint.ts
+++ b/commands/utils/docker-endpoint.ts
@@ -9,14 +9,8 @@ class DockerClient {
     private endPoint:Docker;
 
     constructor() {
-        if (process.env.DOCKER_HOST){
-            this.endPoint = new Docker();
-        }
-        else if (process.platform === 'win32') {
-            this.endPoint = new Docker({ socketPath: "//./pipe/docker_engine" });
-        } else {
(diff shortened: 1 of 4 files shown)


lostintangent:
@tasosval This is awesome, thanks! This is a change that we've been meaning to make, so I appreciate you driving it.

I wonder whether we can simplify this init code a little bit, since `docker-modem` will actually default to setting the right path for Mac/Linux, and give priority for the `DOCKER_HOST` env var if its set: https://github.com/apocas/docker-modem/blob/1387b3e4757254d4324eb187f645bf7fd5ca390f/lib/modem.js#L17.

It seems like the only edge case, which prevents us from being able to simply call `new Docker()`, is the need to support Windows. Long-term, it seems like `docker-modem` should simply check for Windows and set the right path automatically, but we can continue to work around it for a while.

lostintangent:
The docker-modem PR was just merged, so depending on how soon the author releases an update to it and dockerode, we can potentially just update our dockerode dependency to the latest and take advantage of its improved support for various Docker hosts.

msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

tasosval:
It was a really immediate answer from the maintainer of the docker-model project. I guess this development will make this pull request obsolete. I would only keep the changes to the README file so people know this is a feature.

lostintangent:
@tasosval @chrisdias FYI, the change in `docker-modem` was published to NPM, so we should be able to update the `dockerode` dependency and allow it to deal with the environment variables and default host on both Windows and Linux.

tasosval:
Hi. I have made the changes required and will update this pull request. I just want to test with a remote docker-machine just to be on the safe side.

commands/utils/docker-endpoint.ts
```diff
@@ -9,14 +9,8 @@ class DockerClient {
...
-        if (process.env.DOCKER_HOST){
-            this.endPoint = new Docker();
-        }
-        else if (process.platform === 'win32') {
-            this.endPoint = new Docker({ socketPath: "//./pipe/docker_engine" });
-        } else {
-            this.endPoint = new Docker({ socketPath: '/var/run/docker.sock' });
-        }
+        // Pass no options so that the defaultOpts of docker-modem will be used
+        this.endPoint = new Docker(null);
```
tasosval: Yes I did not like it either but I did it to keep the ts compiler from complaining.

The problem seems to be with the embedded typings file that contain the following offending line.
`constructor(options: Docker.DockerOptions);`

The options argument should be optional. Unfortunately I do not know the place where this file was sourced as I have seen other implementations fixing that problem. If we change the typings file we can delete the null argument.

lostintangent: @tasosval Ah good call. Why don't we just update our local copy of the typings file for now? Then we can discuss pushing the improvements back to a shared copy. Ideally, the `dockerode` lib would ship with its d.ts file via NPM.

chrisdias: we crafted the typings file  by hand, so go ahead and make a local change to it and include it in the PR

lostintangent:
@chrisdias I'm comfortable merging this in now, but I'll wait for you to sign off and merge it.

chrisdias:
👍

PR #71, opened 2017-03-11
Adding yaml detection

### LINKED ISSUE #70: Recognize docker-compose.yaml files
Currently, completion support is provided for docker-compose.yml files, but not for docker-compose.yaml files.

According to the docker-compose documentation, both are valid:

> [Tip: You can use either a .yml or .yaml extension for this file. They both work.](https://docs.docker.com/compose/compose-file/#service-configuration-reference)

lostintangent:
This simply resolves #70 by adding support for the `.yaml` extension for `docker-compose` files, in addition to the existing `.yml` support.

### DIFF
--- a/commands/docker-compose.ts
+++ b/commands/docker-compose.ts
@@ -1,6 +1,6 @@
-import vscode = require('vscode');
-import * as path from "path";
-
+import * as vscode from 'vscode';
+import * as path from 'path';
+import { COMPOSE_FILE_GLOB_PATTERN } from '../dockerExtension';
 
 function hasWorkspaceFolder() : boolean {
     return vscode.workspace.rootPath ? true : false;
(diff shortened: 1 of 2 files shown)


rhencke:
@lostintangent That was fast!  Thank you!

Is the following something that might also benefit from recognizing the '.yaml' extension?

https://github.com/Microsoft/vscode-docker/blob/bae341a8396c48bedfb4997016a46633debaa3e1/commands/docker-compose.ts#L13

lostintangent:
@rhencke Yep you're right, that needs to be fixed as well. I'll update this PR to address it, thanks!

lostintangent:
@rhencke Done! Sorry for the delay.

@chrisdias Let me know if this looks good to you, and merge if so. Thanks!

PR #83, opened 2017-05-18
Automatically publish exposed ports when running a container

lostintangent:
This PR updates the `Run` commands, so that they automatically infer the ports the image is exposing, and publishes them on the host machine. This makes the basic `Run` experience simpler, and provides a little more value-add on top of the developer simply executing `docker run` from the integrated terminal.

In addition, I simplified the generated `Dockerfile` for Node.js a bit, by removing the need for the temp folder for NPM installation, and removing some superfluous whitespace. I also updated the default image to use `alpine` instead of `slim`, which is much smaller.


chrisdias:
LGTM

PR #86, opened 2017-05-22
Generate a .dockerignore file as part of workspace configuration

lostintangent:
This PR primarily introduces the creation of a `.dockerignore` file, as part of running the `Docker: Add docker files to workspace` command. This file can make a noticeable difference in build time, as well as helps to keep the built containers as "clean" as possible. As part of helping devs get started with best practices, I think it will be really helpful to provide this file, on top of the existing assets we generate.

As part of making this change, I decided to refactor the code for the workspace configuration code a little bit, since there was a bit of duplication. In the process, I updated some of the `Promise` code to use `async/await`, since we have access to it via TypeScript, and it makes the code so much easier on the eyes :)

The current support for the `.dockerignore` file is entirely optimized for Node.js apps, but I wanted to get this PR out for review/discussion first, since I'm not entirely sure what the best `.dockerignore` files look like for each platform.

### DIFF
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -11,8 +11,8 @@ function genDockerFile(serviceName: string, platform: string, port: string, { cm
             return `FROM node:6-alpine
 ENV NODE_ENV production
 WORKDIR /usr/src/app
-COPY ["package.json", "yarn.lock*", ".yarnclean*", "./"]
-RUN yarn install --production --modules-folder ../node_modules
+COPY ["package.json", "npm-shrinkwrap.json*", "./"]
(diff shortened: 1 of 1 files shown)


lostintangent:
@chrisdias Let me know what you think! I'm not sure if this is the right/best `.dockerignore` file, but in my testing, this can make a noticeable difference on build times for some apps.

chrisdias:
added `docker-compose*` to `.dockerignore` file, i'm sure there is more we can add later (e.g. look at .NET, Go artifacts) but looks good for now.

lostintangent:
@chrisdias I was just pushing a commit to add `docker-compose*` and it got rejected due to yours :) Nice timing!

PR #87, opened 2017-05-22
Run non-interactive containers in the background

lostintangent:
This PR simply adds the `-d` flag when running the `Docker: Run` command. This way, you can use `Docker: Run` to spin up background tasks (and use `Docker: Show Logs`, etc. at any time), and `Docker: Run Interactive` to spin up interactive tasks. Otherwise, the non-interactive command captures your terminal, which might not be the expected behavior.

// CC @chrisdias Let me know what you think of this behavior.

### DIFF
--- a/commands/start-container.ts
+++ b/commands/start-container.ts
@@ -10,12 +10,7 @@ function doStartContainer(interactive: boolean) {
     quickPickImage(false).then(function (selectedItem: ImageItem) {
         if (selectedItem) {
             docker.getExposedPorts(selectedItem.label).then((ports: string[]) => {
-                let options = '--rm'
-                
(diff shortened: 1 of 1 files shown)


lostintangent:
@chrisdias Depending on what you think, we could also rename the `Docker: Run` command to `Docker: Run (Background)`, to contrast it more clearly with `Docker: Run (Interactive)`. Or something similar.

chrisdias:
I like the simplicity of just using Run as opposed to adding "(Background)".

lostintangent:
@chrisdias Yep, agreed. Just wanted to double-check that you were cool with adding this behavior to the existing command.

PR #91, opened 2017-06-06
move to async await

chrisdias:


lostintangent:
LGTM! This is a massively welcome change

PR #98, opened 2017-07-14
Use an SPDX license ID in the package.json file

rcjsuen:
The [npm website](https://docs.npmjs.com/files/package.json#license) suggests that an SPDX license ID be used in the `package.json` file if possible. Since the vscode-docker project is licensed under the MIT license, which does have a corresponding SPDX license ID, we should use that ID in the `package.json` file.


chrisdias:
@rcjsuen thanks for the PR! 

The reason we don't use the SPDX license ID is that the license we have includes the text "Copyright (c) Microsoft Corporation".

Issue #104, opened 2017-07-26
add docker inspect command


chrisdias:
show results in an editor as json

lostintangent:
@chrisdias I assume you mean `docker inspect` command? :)

lostintangent:
I'll send out a PR for this later this evening, and we can iterate on the design a bit.

lostintangent:
Addressed via #105

PR #105, opened 2017-07-29
Introduce image inspect command

lostintangent:
This PR addresses #104 by introducing a new `Docker: Inspect Image` command, which displays a pick-list of images, and when one is selected, will open up a read-only document which contains the JSON returned by calling `docker inspect` under the covers.

screen shot 2017-07-29 at 3 20 37 pm


.vscode/launch.json
```diff
@@ -11,7 +11,7 @@
 			"stopOnEntry": false,
 			"sourceMaps": true,
 			"outFiles": ["${workspaceRoot}/out/**/*.js"],
-			"preLaunchTask": "npm"
+			"preLaunchTask": "npm: compile"
```
lostintangent: The `tasks.json` file is no longer neccessary now that `npm` scripts are auto-detected, so we can just simply specify the `npm` script we want to run directly from here (I love this capability!).

documentContentProviders/dockerInspect.ts
```diff
@@ -0,0 +1,22 @@
+import { TextDocumentContentProvider, Uri, window, workspace } from "vscode";
+import { docker } from "../commands/utils/docker-endpoint";
+
+export const IMAGE_DOMAIN = "image";
+export const SCHEME = "docker-inspect";
+export const URI_EXTENSION = ".json";
```
lostintangent: @chrisdias I'm adding the `.json` extension to the document `Uri` in order to get the right language service experience (e.g. colorization, folding), but I kind of hate it :) Is there a way to open a `Uri`-based document that specifies the language without an extension? I didn't see this capability in the `workspace.openTextDocument` method or any other API.

chrisdias: @jrieken ?

chrisdias: @lostintangent what was the motivation behind removing the `-watch` from `package.json`?

lostintangent: @chrisdias I removed it because the watch (which obviously doesn't exit) was preventing the debugger from starting. That said, maybe I was doing something wrong, so I could revert this change.

chrisdias: I guess its ok, the only thing you lose is the ability to make a change and then reload the window in the running instance to pick up that change, without having to restart the debugger. I typically restart the debugger anyway so it isn't a big deal.

Otherwise, changes look good!

PR #107, opened 2017-08-11
Chrisdias/explorer

chrisdias:
add explorer contribution


commands/push-image.ts
```diff
@@ -1,14 +1,29 @@
 import vscode = require('vscode');
 import { ImageItem, quickPickImage } from './utils/quick-pick-image';
 import { reporter } from '../telemetry/telemetry';
+import { DockerNode } from '../explorer/dockerExplorer';
 const teleCmdId: string = 'vscode-docker.image.push';
 
-export async function pushImage() {
+export async function pushImage(context?: DockerNode) {
+    let imageToPush: Docker.ImageDesc;
```
lostintangent: Is this variable even used? Could we simply track the image name and use it to determine whether we push below?

commands/push-image.ts
```diff
@@ -1,14 +1,29 @@
...
+    let imageName: string = "";
 
-    const selectedItem: ImageItem = await quickPickImage();
-    if (selectedItem) {
-        const terminal = vscode.window.createTerminal(selectedItem.label);
-        terminal.sendText(`docker push ${selectedItem.label}`);
+    // if invokde from the explorer we have the name of the image
```
lostintangent: Typo with "invoke", though we can probably just remove this comment entirely, since every command handler now has the same logic in it.

commands/remove-container.ts
```diff
@@ -0,0 +1,55 @@
...
+export async function removeContainer(context?: DockerNode) {
+
+    let containersToRemove: Docker.ContainerDesc[];
+
+    if (context && context.containerDesc) {
+        containersToRemove = [context.containerDesc];
+    } else {
+        const opts = {
+            "filters": {
+                "status": ["created", "restarting", "running", "paused", "exited", "dead"]
```
lostintangent: It looks like the `options` argument of `quickPickContainer` is only being used to specify a container filter. Would it simpler to have it take a bool that indicates whether to display all containers or just the running ones? Then the commands that display the quick pick don't need to worry about defining this filter object.

explorer/dockerExplorer.ts
```diff
@@ -0,0 +1,163 @@
...
+export class DockerExplorerProvider implements vscode.TreeDataProvider<DockerNode> {
+
+    private _onDidChangeTreeData: vscode.EventEmitter<DockerNode | undefined> = new vscode.EventEmitter<DockerNode | undefined>();
+    readonly onDidChangeTreeData: vscode.Event<DockerNode | undefined> = this._onDidChangeTreeData.event;
+    private _imagesNode: DockerNode;
+    private _containersNode: DockerNode;
+    private _registriesNode: DockerNode;
+    private _debounceTimer: NodeJS.Timer;
+    
+    constructor() {
```
lostintangent: This can be removed.

explorer/dockerExplorer.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    refresh(): void {
+        this.refreshImages()
+        this.refreshContainers()
+        //this.refreshRegistries()
```
lostintangent: Do we want to just strip out the registry code for now? I'm cool either way.

explorer/dockerExplorer.ts
```diff
@@ -0,0 +1,163 @@
...
+                                nodes.push(node);
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (element.label === 'Containers') {
+
+                opts = {
```
lostintangent: Similar comment as above. Could this filter be moved into `getContainerDescriptors`, and allow simply passing a bool to indicate that you want all or only running containers?

package.json
```diff
@@ -66,7 +67,105 @@
...
           "group": "docker"
         }
-      ]
+      ],
+      "view/title": [
+        {
+          "command": "dockerExplorer.refreshExplorer",
+          "when": "view == dockerExplorer",
+          "group": "navigation"
+				},
```
lostintangent: Could we format this? The spacing seems a little off.

package.json
```diff
@@ -110,89 +209,142 @@
...
       }
     },
     "commands": [
       {
         "command": "vscode-docker.configure",
-        "title": "Docker: Add docker files to workspace",
-        "description": "Add Dockerfile, docker-compose.yml files"
+        "title": "Add docker files to workspace",
+        "description": "Add Dockerfile, docker-compose.yml files",
+        "category": "Docker"
```
lostintangent: Does the `category` property append `Docker: ` to the name in the command palette?

commands/push-image.ts
chrisdias: Yes, on line 22.  if the user never selects anything from the quick pick then its undefined.  The alternative is to test `imageName`, but the `imageToPush` pattern is more consistent across the other commands.

commands/push-image.ts
chrisdias: removed

commands/remove-container.ts
chrisdias: I used the options/filter to give us some flexibility with the explorer contribution. we may want to limit what we show in the tree in the future.

explorer/dockerExplorer.ts
chrisdias: `_registriesNode`? I commented it out to match what i did with other "registry" nodes.

explorer/dockerExplorer.ts
chrisdias: i would like to leave it in for now.

explorer/dockerExplorer.ts
chrisdias: as noted above, I'm just leaving some options open for what we show in the tree.

package.json
chrisdias: hmmmm... i'm not seeing this...

explorer/dockerExplorer.ts
lostintangent: Ah apologies, I was referring to the constructor, which didn't seem to be used.

package.json
chrisdias: Yes, it prepends the "Docker" category in the Command Palette. The reason i move do this was so that in the explorer, when you right click on a node, you don't see "Docker:" on every command.

lostintangent: Cool! I wasn't aware of this behavior, so that's nice to know.

chrisdias:
argh... i didn't push my local fixes in response to your comments.

explorer/dockerExplorer.ts
lostintangent: Sounds good!

PR #110, opened 2017-08-13
Chrisdias/remove linter

chrisdias:
removes the linter and cleans up the README, CHANGELOG, etc. for publishing.


rcjsuen:
I actually thought the two would co-exist since users can customize the linting rules with dockerfile_lint and the language server doesn't offer that feature.

In any case, if the decision is to drop dockerfile_lint from the extension, the [documentation](https://github.com/Microsoft/vscode-docs/blob/c274121410792cc1b5bbbba18ab7a7471a3a125e/docs/languages/dockerfile.md) will also need to be updated.

chrisdias:
@rcjsuen we found the "older" linter to be pretty aggressive, so it is actually turned off by default.

Great catch on the docs, I will update there!

chrisdias:
https://github.com/Microsoft/vscode-docs/issues/1137

PR #122, opened 2017-09-13
Fix missing explorer issue

LazarusX:
In 6343e3f651c443142b7b6fafb179b8227efc7a3e, the `docker.showExplorer` configuration property is removed, which breaks the docker sash. This PR fix this issue.


chrisdias:
Nice catch!

PR #149, opened 2017-10-16
Chrisdias/optionalazure

chrisdias:


explorer/deploy/azureAccountWrapper.ts
```diff
@@ -8,6 +8,8 @@ import { ServiceClientCredentials } from 'ms-rest';
 import { AzureEnvironment } from 'ms-rest-azure';
 import { SubscriptionClient, SubscriptionModels } from 'azure-arm-resource';
 import { AzureAccount, AzureSession, AzureLoginStatus } from '../../typings/azure-account.api';
+import { azureAccount } from '../../dockerExtension';
```
chrmarti: If this runs before `azureAccount` in dockerExtension was initialized, `azureAccount` here will always be undefined. Sure this works?

I'd avoid using globals and would instead pass `azureAccount` as an argument to the constructor. Similar in other modules below. Does that seem reasonable?

chrisdias: The azureAccountWrapper is only used during the "Deploy to Azure" process, which is effectively blocked by virtue of the fact that a) we don't show the Azure node if the azure-account extension isn't there, and in the DockerHub scenario I put up a message saying they need to install that extension.  

Admittedly, the way I'm blocking all of this today is a hack.

chrisdias: @chrmarti Per our talk, I now pass `azureAccount` around instead of importing it from the `dockerExtension`.

PR #152, opened 2017-10-20
Add 'Restart' command for containers.

orfevr:
Currently user can stop container and remove stopped container from the context menu. To start existing one (stopped) they need to use console command. I propose implementation of 'Restart' command which allows to restart running containers as well as start stopped.


chrisdias:

bsvipas:
Very good idea, because I was thinking to open an issue, because there's no start/restart.
@orfevr So as far as I understand if container is stopped and I will press restart it will start it? And if it's running and I press restart it will be restarted?

orfevr:
@bsvipas Exactly! I faced with the same thing and decided to make a pull request.

bsvipas:
@orfevr You're my hero! 😄

PR #155, opened 2017-10-21
Only match Dockerfile and *.dockerfile

### LINKED ISSUE #134: DOCKERFILE_GLOB_PATTERN is too permissive
We have a `dockerfile.go` file in at least one repo and it's being detected as belonging to this plugin instead of the Go plugin.

This is likely because the [current DOCKERFILE_GLOB_PATTERN](https://github.com/Microsoft/vscode-docker/blob/master/dockerExtension.ts#L32) is too permissive.

dlech:
The current matches are too permissive and cause false positives.

Fixes #134


chrisdias:

PR #156, opened 2017-10-24
Support workspace folders for the language server's settings

rcjsuen:
VS Code's workspace feature will now allow resources to have settings that are scoped to the given resource's folder hierarchy. We need to add support to the `workspace/configuration` language server protocol request that's currently being proposed so that the extension can send changes to scoped settings to the language server.


chrisdias:

Issue #60, opened 2017-01-31
Allow customization of shell to run 


chrisdias:
We should allow for the user to specify which shell to run (both windows and linux) when executing the command "Attach shell to running container".

PrashanthCorp:
Fixed by #160

PR #160, opened 2017-10-31
Make the shell attach command customizable

FredrikFolkesson:
Allows the user to specify the shell command to be used when attaching to containers by using the settings

`docker.attachShellCommand.LinuxContainer` and `docker.attachShellCommand.windowsContainer`

Defaulting to the current commands `/bin/sh` and `powershell`

This closes, or partly closes, #60.
Depending on if you want the user to be able to specify the shell command to run in "runtime" or only using the settings as this PR provides. 
(the former makes the ease of just right clicking a container and pressing `Attach Shell` a bit worse in my opinion since you would then have to specify the attach command anyways)

### DIFF
--- a/commands/open-shell-container.ts
+++ b/commands/open-shell-container.ts
@@ -5,9 +5,10 @@ import { ContainerNode } from '../explorer/models/containerNode';
 import { reporter } from '../telemetry/telemetry';
 const teleCmdId: string = 'vscode-docker.container.open-shell';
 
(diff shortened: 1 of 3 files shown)


FredrikFolkesson:
@chrisdias Are you still on vacation? :)

FredrikFolkesson:
@chrisdias Any change to look at this? :)

chrisdias:
this looks better than https://github.com/Microsoft/vscode-docker/pull/159. Can we close that one and only take this one?

FredrikFolkesson:
Yeah that makes sense

Issue #167, opened 2017-11-11
False positive error message `Invalid containerPort: ...`


31z4:
Having the following Dockerfile
```dockerfile
FROM openjdk:8-jre-alpine
ENV ZOO_PORT=2181
EXPOSE $ZOO_PORT
```
I get an error message `Invalid containerPort: $ZOO_PORT` which is incorrect, since environment variables [are supported](https://docs.docker.com/engine/reference/builder/#environment-replacement) by the following list of instructions in the `Dockerfile`:
* `ADD`
* `COPY`
* `ENV`
* `EXPOSE`
* `FROM`
* `LABEL`
* `STOPSIGNAL`
* `USER`
* `VOLUME`
* `WORKDIR`
as well as:
* `ONBUILD` (when combined with one of the supported instructions above)

rcjsuen:
Thank you very much for your bug report, @31z4. I have opened rcjsuen/dockerfile-language-server-nodejs#192 to track this problem in the language server.

rcjsuen:
This can be closed since #173 has been merged.

chemidy:
With the last version 0.0.23, it's not resolved.

**[dockerfile-lsp] Invalid containerPort:**

```dockerfile
FROM scratch
ARG GRPC_PORT
EXPOSE $ZOO_PORT
```

rcjsuen:
Hi @chemidy, if you use the Dockerfile that was [originally posted](https://github.com/Microsoft/vscode-docker/issues/167#issue-273119689), you shouldn't get any errors.

In the Dockerfile you have provided, you defined an argument named `GRPC_PORT` but your `EXPOSE` instruction is using `$ZOO_PORT`. Did you make a copy/paste error? Is this the Dockerfile you intended to  write in your comment?
```Dockerfile
FROM scratch
ARG GRPC_PORT
EXPOSE $GRPC_PORT
```

rcjsuen:
In any case, I have opened rcjsuen/dockerfile-language-server-nodejs#199 for the problem that you have found regarding build arguments with no default values.

Thank you for the bug report!

chemidy:
Hi @rcjsuen , yes it was a copy/paste error. Thks for the bug report

rcjsuen:
@chemidy Please update to 0.0.24 for a fix to the problem you reported. Thanks again for the bug report!

Issue #165, opened 2017-11-07
Intellisense should stop "helping" inside comments


ssteinerx:
This only seems to happen with the Docker support.

Inside a comment in a Dockerfile, it's almost impossible to just type 'cause "helpful" suggestions pop up constantly.

`#` at the beginning of a line starts a comment and everything else after that should be left alone for regular typing.

rcjsuen:
Hi @ssteinerx, thank you for the bug report.

Could you attach a screenshot that shows what you are seeing? I want to ensure that the behavior I am seeing matches what you are encountering.

ssteinerx:
Pretty simple...
`#` to start a comment
`<space>` brings up a list of Docker commands and makes typing virtually impossible

![vscode-helpful-docker](https://user-images.githubusercontent.com/43063/32671360-1b905314-c615-11e7-8d40-8a8a4b85e120.gif)

ssteinerx:
@rcjsuen Thanks!  How would one get that updated into their VSCode installation?  At the moment, mine says there are no updates available.  Does it require also pushing an update to this extension?  Thanks again!

rcjsuen:
@ssteinerx The bug is in one of the dependencies of the the vscode-docker extension. It would require both an update to the dependency and an update to the extension (to increment the version of the dependency).

rcjsuen:
I think we can close this since #173 has been merged?

ssteinerx:
@rcjsuen Is this supposed to be in most recent code insiders release?

rcjsuen:
@ssteinerx The fix should be in the 0.0.23 release of this extension that was published on January 6th.

PR #173, opened 2017-11-15
Update to the 0.0.11 release of the language server

rcjsuen:
While there have been no new features added to the 0.0.11 release, this new release includes fixes for both #165 and #167.

Please use the following Dockerfile to verify the two fixes.
```Dockerfile
FROM openjdk:8-jre-alpine
ENV ZOO_PORT=2181
# no suggestions should appear when inserting
# a space in the next commented line
#
EXPOSE $ZOO_PORT
```


chrisdias:
LGTM

Issue #183, opened 2017-11-27
prompt for confirmation on system prune


chrisdias:
per feedback from https://marketplace.visualstudio.com/items?itemName=PeterJausovec.vscode-docker#review-details

PR #213, opened 2018-02-06
Change promptOnSystemPrune default

driskell:
Relates to #183

Clicking System Prune does not produce the necessary warnings that Docker normally products when you run a system prune. The icon is also unlabelled and easy to mis-click so close the to the Refresh and resize grabber. It also insufficiently explains purpose.

This will restore the prompt that was there before as a default but allow those that are fully aware of consequences to disable the prompt, so that any new users are fully informed of consequences. Might be worth checking the messaging explains that volumes are removed too as that was a little unexpected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -423,7 +423,7 @@
         },
         "docker.promptOnSystemPrune": {
           "type": "boolean",
-          "default": false,
+          "default": true,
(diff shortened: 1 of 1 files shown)


chrisdias:
Since there have been multiple issues regarding prune, I think it makes sense to change the default.

PR #219, opened 2018-02-21
Add Python support

brettcannon:
While I tested the Dockerfile template, the other aspects I simply copy and pasted as I do not have extensive Docker experience.

### DIFF
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -52,10 +52,12 @@ ENTRYPOINT dotnet ${serviceName}.dll
         case 'python':
 
             return `
-# Python support extends back to 3.4 and 2.7 and can be specified down
-# to the micro version (e.g. 3.6.3).
+# Python support can be specified down to the minor or micro version
+# (e.g. 3.6 or 3.6.3).
 # OS Support also exists for jessie & stretch (slim and full).
-FROM python:3.6-alpine3.6
+# See https://hub.docker.com/r/library/python/ for all supported Python
+# tags from Docker Hub.
+FROM python:alpine
 
 # If you prefer miniconda:
 #FROM continuumio/miniconda3
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -49,6 +49,37 @@ COPY $source .
 ENTRYPOINT dotnet ${serviceName}.dll
 `;
 
+        case 'python':
+
+            return `
+# Python support extends back to 3.4 and 2.7 and can be specified down
+# to the micro version (e.g. 3.6.3).
+# OS Support also exists for jessie & stretch (slim and full).
+FROM python:3.6-alpine3.6
+
+# If you prefer miniconda:
+#FROM continuumio/miniconda3
+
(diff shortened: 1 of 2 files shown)


configureWorkspace/configure.ts
```diff
@@ -47,6 +47,37 @@ WORKDIR /app
...
 ENTRYPOINT dotnet ${serviceName}.dll
+`;
+
+        case 'python':
+
+            return `
+# Python support extends back to 3.4 and 2.7 and can be specified down
+# to the micro version (e.g. 3.6.3).
+# OS Support also exists for jessie & stretch (slim and full).
+FROM python:3.6-alpine3.6
```
berndverst: Prefer `FROM python:alpine` so we always have the latest stable Python version (right now 3.6.4) by default with the latest stable Alpine.

Note that for some reason this tag currently doesn't use the latest stable Alpine version. That's a separate issue I will take up with the official Docker Python Image maintainers.

Can add a commend you can specify the exact Python and Alpine version like:
`# FROM python:3.6-alpine3.6`

I'll reach out

Probably should make sure that tag is used for newer

berndverst: FYI: Filed the issue with the `alpine` tags here: https://github.com/docker-library/python/issues/265

berndverst: Actually, why not add a comment to refer to `https://hub.docker.com/r/library/python/` for a list of all supported Python version / distribution pairs :)

brettcannon: Done and done!

chrisdias:
This looks good, thanks for the PR! My concern is that I don't have the ability to test that the generated Dockerfile is up to date, continues to work, etc. Of course, I have the same problem with the other platforms as well besides Node :).  Can people here sign up to test and report issues/submit fixes ASAP?  I'm happy to take the fixes and publish an update as soon as i get them, I just need help ensuring that this stays fresh.

brettcannon:
So are you saying all of the languages are going to require a contact point to keep the image up-to-date, otherwise they are going to be removed? Or is Python just a victim of bad timing? 😉 

As for signing up to stay on top of this like a hawk, I don't know if my manager will want that as I did this as a favour for someone from MS. There is talk of using the Dockerfile as the basis of something else within Linux, so it might be possible to have that team take ownership of keeping it updated.

chrisdias:
@brettcannon I'm just asking folks who are interested to submit issues (and PRs if possible) as quickly as they find them so that I can fix them and publish, that's all.

brettcannon:
@chrisdias Ah! Yes, of course. 😄

configureWorkspace/configure.ts
```diff
@@ -94,6 +127,17 @@ services:
     ports:
       - ${port}:${port}`;
 
+        case 'python':
+            return `version: '2.1'
+
+services:
+${serviceName}:
```
mikedld: Improper indentation in YAML here and in the following lines.

brettcannon: PR to fix at https://github.com/Microsoft/vscode-docker/pull/242

Issue #192, opened 2017-12-18
Support for Dockerfile.some-env


maxime1992:
Since Docker 1.8 we can do something like that in order to have multiple Dockerfile in the same folder:

```
docker build -f Dockerfile.db .
docker build -f Dockerfile.web .
```

See https://stackoverflow.com/a/32108873/2398593

With VSC, we loose intellisense within a file `Dockerfile.something` file.

kamcma:
If you add the following to your VSCode settings, you should get it back:
```json
"files.associations": {
    "[Dd]ockerfile*": "dockerfile",
    "docker-compose*": "yaml"
}
```
For the Dockerfile entry, I'm getting IntelliSense and file icons, but the docker-compose entry isn't giving me file icons for some reason.

abbas001:
Thanks its working.

bwateratmsft:
I think this scenario is entirely covered by #102, so I'll resolve this as duplicate.

Issue #193, opened 2017-12-19
Syntax is not working if the name is not "Dockerfile"


riveryc:
Hi, as reading the file https://github.com/Microsoft/vscode-docker/blob/fb97733640fae72d4098c021dbf08198a0a3e13b/dockerExtension.ts#L43-L44

I think if I'm using "Dockerfile-centos" it should also give me the syntax highlighting for it, however it didn't happen to me:
![image](https://user-images.githubusercontent.com/12391265/34140230-6a1290c0-e4cd-11e7-812a-157cb30a579f.png)

Please correct me if I'm wrong, or maybe I got something didn't setup correctly?
The version for this extension I'm using is 0.0.22

Thx
R

chrisdias:
We've gone back and forth on this (see #134) issue unfortunately. Perhaps the right answer is to make the DOCKERFILE_GLOB_PATTERN and COMPOSE_FILE_GLOB_PATTERN configurable so that users can set up how they want to do it.

@riveryc in the meantime, you could add something like this to your workspace or user settings:

``` json
    "files.associations": {
        "Dockerfile-centos": "dockerfile"
    }
```

riveryc:
@chrisdias Thanks so much for this workaround. It's working now as expected. :)

River

PR #248, opened 2018-04-10
Add Support for Dockerfile.<some-env>

### LINKED ISSUE #102: Allow custom file names
[Right now, `COMPOSE_FILE_GLOB_PATTERN` and `DOCKERFILE_GLOB_PATTERN` are hardcoded patterns](https://github.com/Microsoft/vscode-docker/blob/91a56df16058eaa134855eb2438be4216fa2b74d/dockerExtension.ts#L29-L30).

We'd need a way to configure those, either by user or by project.

Thanks!

MattMorgis:
I _think_ this closes #102 and #192 

It should add icon and syntax support for `Dockerfile.<some-env>`

Before:
screen shot 2018-04-10 at 3 53 21 pm

After:
screen shot 2018-04-10 at 3 53 05 pm

### DIFF
--- a/dockerExtension.ts
+++ b/dockerExtension.ts
@@ -41,7 +41,7 @@ import { browseAzurePortal } from './explorer/utils/azureUtils';
 
 export const FROM_DIRECTIVE_PATTERN = /^\s*FROM\s*([\w-\/:]*)(\s*AS\s*[a-z][a-z0-9-_\\.]*)?$/i;
 export const COMPOSE_FILE_GLOB_PATTERN = '**/[dD]ocker-[cC]ompose*.{yaml,yml}';
-export const DOCKERFILE_GLOB_PATTERN = '**/{*.dockerfile,[dD]ocker[fF]ile}';
+export const DOCKERFILE_GLOB_PATTERN = '**/{*.dockerfile,[dD]ocker[fF]ile,[Dd]ockerfile*}';
 
 export var diagnosticCollection: vscode.DiagnosticCollection;
 export var dockerExplorerProvider: DockerExplorerProvider;


chrisdias:
@MattMorgis thanks for the PR. We've gone back and forth on the glob pattern here (see https://github.com/Microsoft/vscode-docker/issues/193#issuecomment-352898885 for example). There's a workaround to use `files.associations` in settings:

``` json
    "files.associations": {
        "Dockerfile-centos": "dockerfile"
    }
```

Another idea was to have a setting that let you customize the glob pattern for your environment that people seemed to like (https://github.com/Microsoft/vscode-docker/issues/102#issuecomment-317503519).  I'd be more inclined to take a PR for that than switching up the glob pattern again.

StephenWeatherford:
@chrisdias, @fiveisprime I don't see the drawback of adding * to the default dockerfile glob pattern, whether or not we add customization of it.  I'd like to get this merged (with minor changes).  Do you disagree?  The compose file glob already has that:


export const FROM_DIRECTIVE_PATTERN = /^\s*FROM\s*([\w-\/:]*)(\s*AS\s*[a-z][a-z0-9-_\\.]*)?$/i;
export const COMPOSE_FILE_GLOB_PATTERN = '**/[dD]ocker-[cC]ompose*.{yaml,yml}';
export const DOCKERFILE_GLOB_PATTERN = '**/{*.dockerfile,[dD]ocker[fF]ile}';

chrisdias:
The pattern used to be more permissive, but then it was too intrusive.  See https://github.com/Microsoft/vscode-docker/issues/134 for the original issue where we changed it to what it is today.  I like the idea of making this a setting so that it can be controlled by the user.

MattMorgis:
I like the idea of making this a setting, too. I will poke around and update this PR.

EricJizbaMSFT:
FYI this PR was created before our recent switch from Travis CI to Azure Dev Ops. Push a commit (merging with master or addressing comments) and the "checks" for this PR will be updated

EricJizbaMSFT:
Closing this PR since this solution is too intrusive as mentioned above and hasn't had any recent activity. We're actively investigating the alternative approach to add settings for these glob patterns

PR #278, opened 2018-06-16
Remove deprecated "go-wrapper" script

korservick:
https://github.com/docker-library/docs/commit/e19ba52efe31aefb75f6753da40d0a2d68fe571f

### DIFF
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -29,8 +29,8 @@ FROM golang:alpine AS builder
 WORKDIR /go/src/app
 COPY . .
 RUN apk add --no-cache git
-RUN go-wrapper download   # "go get -d -v ./..."
-RUN go-wrapper install    # "go install -v ./..."
+RUN go get -d -v ./...
+RUN go install -v ./...
 
 #final stage
 FROM alpine:latest


korservick:
@chrisdias Hi! 
Here small fix for new Dockerfile

StephenWeatherford:
@chrisdias Who needs to approve this?  Thx.

chrisdias:
@ramya-rao-a @joshgav can we get your feedback here?

joshgav:
Thanks for pointing this out @korservick! LGTM. @vladbarosan thoughts?

Related: https://github.com/docker-library/golang/pull/205.

vladbarosan:
Yes we should remove, did we test that it works well on alpine ? 

Also : 
1. Can we add an extra option in the dialog so users can choose Ubuntu based image or alpine based ? ( depending on the use case )
2. Do we want to add some logic taking into consideration that they might be using dep ?

StephenWeatherford:
Who can test this?  I don't have the expertise to manage testing for all the different languages.

@vladbarosan 1) If you provide me with exact files to use for each option, I can do that.  2) Again, I'd need exact files.  Thanks.

StephenWeatherford:
@vladbarosan @joshgav @korservick @ItalyPaleAle I'm planning on merging this soon if there are no more comments. I've verified with https://github.com/cloudfoundry-community/simple-go-web-app/archive/master.zip that the old Dockerfile does not work:

Step 5/12 : RUN go-wrapper download   # "go get -d -v ./..."
 ---> Running in 2e2298220ea6
/bin/sh: go-wrapper: not found
The command '/bin/sh -c go-wrapper download   # "go get -d -v ./..."' returned a non-zero code: 127

The new one does work.

If you'd like to add comments on what image to use for Ubuntu, please let me know (some of the other Dockerfiles have such comments).  Thx.

vladbarosan:
@StephenWeatherford ill do a quick test with the changes to see if everything is fine and let you know by tomorrow

ItalyPaleAle:
@StephenWeatherford Thinking again about this.

For the "destination" image (second step), do we need to use Alpine? Since we're not installing any dependency, would it make sense to use the "scratch" image directly?

See [this short ebook](https://azure.microsoft.com/en-us/resources/how-to-containerize-your-go-code/en-us/) my team sponsored a few months ago.

StephenWeatherford:
@vladbarosan Thanks!  Could you also comment on @ItalyPaleAle's comment above?  Thx.

vladbarosan:
@StephenWeatherford ran a quick test and everything seems fine. I would have added some commented lines in case dep is being used for dependencies, something like 
```
ENV DEP_RELEASE_TAG=v0.5.0
RUN curl -sSL https://raw.githubusercontent.com/golang/dep/master/install.sh | sh \
    && dep ensure -v -vendor-only 
```

Also go modules are around the corner and although those will be part of the tooling we will need if we need to update this.

@ItalyPaleAle we could , but a couple things come in mind:
1. A more complicated compile command so we have static compilation. 
2. Not sure how this would work with cgo 
3. Some packages need to be added ( for example here we want to add the CA certificates). How would they do that be done ? i guess manually mapping them from the local box ? that would be less straightforward.

ItalyPaleAle:
Thanks for the answer:

1. I'm not sure if that's actually an issue. Aren't go binaries already containing all static libraries? I might be wrong, however.
2. As long as the code is statically linked, it should work also if it's c code. If it needs dynamic libraries, then it won't work; but so, neither with the current template.
3. Yes, the CA Certificate bundles could be a real issue. There's an example here: https://github.com/drone/ca-certs/blob/master/Dockerfile

I agree this might not be the best solution given that we need to be as broad as possible. I just thought it would have been an interesting option to consider.

StephenWeatherford:
@vladbarosan @ItalyPaleAle So the consensus is to merge as is, correct?  Thanks!

StephenWeatherford:
I've merged, let me know if anything else needs changing.  Thanks!

PR #281, opened 2018-06-20
Make python dockerfile a bit more resilient

paddie:
The [recommended approach](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) when adding to a dockerfile is to use `COPY` and not `ADD`, since `ADD` will extract `gz` files etc. 

I've also made the `pip install -r requirements.txt` step not fail when the file does not exist.

### DIFF
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -153,7 +153,7 @@ LABEL Name=${serviceName} Version=${version}
 EXPOSE ${port}
 
 WORKDIR /app
-ADD . /app
+COPY . .
 
 # Using pip:
 RUN python3 -m pip install -r requirements.txt
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -155,13 +155,15 @@ EXPOSE ${port}
 WORKDIR /app
 COPY . .
 
+# install dependencies if dependencies file is present
+RUN python3 -m pip install -r requirements.txt || echo ‘No requirements.txt - skipping pip install’
+
 # Using pip:
-RUN python3 -m pip install -r requirements.txt
 CMD ["python3", "-m", "${serviceName}"]
 
 # Using pipenv:
-#RUN python3 -m pip install pipenv
-#RUN pipenv install --ignore-pipfile
+#RUN python3 -m pip install pipenv  || echo ‘No requirements.txt - skipping pip install’
+#RUN pipenv install --ignore-pipfile  || echo ‘No requirements.txt - skipping pip install’
 #CMD ["pipenv", "run", "python3", "-m", "${serviceName}"]
 
 # Using miniconda (make sure to replace 'myenv' w/ your environment name):


qubitron:
@brettcannon can you take a look?

configureWorkspace/configure.ts
```diff
@@ -153,15 +153,17 @@ LABEL Name=${serviceName} Version=${version}
 EXPOSE ${port}
 
 WORKDIR /app
-ADD . /app
+COPY . .
+
+# install dependencies if dependencies file is present
```
qubitron: This comment I think is implied by the echo statement, can you move the RUN command back under the " Using pip" comment?

qubitron:
One thing to consider while we're making changes to the file is to install pip using the --no-cache-dir argument, which removes any cached packages from the docker image and should reduce the size

configureWorkspace/configure.ts
```diff
@@ -153,15 +153,17 @@ LABEL Name=${serviceName} Version=${version}
...
 
 # Using pip:
-RUN python3 -m pip install -r requirements.txt
 CMD ["python3", "-m", "${serviceName}"]
 
 # Using pipenv:
-#RUN python3 -m pip install pipenv
-#RUN pipenv install --ignore-pipfile
+#RUN python3 -m pip install pipenv  || echo ‘No requirements.txt - skipping pip install’
+#RUN pipenv install --ignore-pipfile  || echo ‘No requirements.txt - skipping pip install’
```
brettcannon: Pipenv actually uses a `Pipfile` and not a `requirements.txt` file.

configureWorkspace/configure.ts
```diff
@@ -153,15 +153,17 @@ LABEL Name=${serviceName} Version=${version}
 EXPOSE ${port}
 
 WORKDIR /app
-ADD . /app
+COPY . .
+
+# install dependencies if dependencies file is present
+RUN python3 -m pip install -r requirements.txt || echo ‘No requirements.txt - skipping pip install’
```
brettcannon: I wouldn't use opening and closing quotes and just stick with single quotes (`'`).

configureWorkspace/configure.ts
brettcannon: Pipenv uses a `Pipfile`, not a `requirements.txt` file, and you wouldn't be running pipenv if you didn't already have a `Pipfile`, so I don't think this change should go in.

configureWorkspace/configure.ts
```diff
@@ -153,15 +153,17 @@ LABEL Name=${serviceName} Version=${version}
...
 
 # Using pip:
-RUN python3 -m pip install -r requirements.txt
 CMD ["python3", "-m", "${serviceName}"]
 
 # Using pipenv:
-#RUN python3 -m pip install pipenv
-#RUN pipenv install --ignore-pipfile
+#RUN python3 -m pip install pipenv  || echo ‘No requirements.txt - skipping pip install’
```
brettcannon: This failing has nothing to do with a missing `requirements.txt` file and everything to do with `pip` failing to install `pipenv` which would be fatal for anyone choosing to use `pipenv`, so this change should be reverted.

brettcannon:
The `ADD` usage is quite possibly wrong, so I have no issue with that. And Dan's suggestion about adding `--no-cache-dir` seems reasonable.

But the other changes to allow errors I don't think should go in. If you don't have a `requirements.txt` file then simply remove the file from the `Dockerfile` rather than potentially masking failures not related to not having a `requirements.txt` file. And the other changes actually have nothing to do with `requirements.txt` files.

chrisdias:
@paddie can you take another pass and address the feedback? thx.

StephenWeatherford:
@paddie Are you still planning on updating this PR?  Thanks.

StephenWeatherford:
Closing due to lack of response.  If you'd like to address the requested changes, please feel free to reopen or resubmit.  Thanks!

PR #282, opened 2018-06-20
Allow changing the connected endpoint

tfenster:
We have multiple docker hosts and it is cumbersome to always set the environment variable and restart VS Code in order to make the extension connect to the right host. With the new command "change endpoint" it is possible to just switch endpoints without breaking the env variable way to set the endpoint

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ Of course, you will need to have Docker installed on your computer in order to r
 
 ## Running commands on Linux
 
-By default, Docker runs as the root user, requiring other users to access it with `sudo`. This extension does not assume root access, so you will need to create a Unix group called docker and add users to it. Instructions can be found here: [Create a Docker group](https://docs.docker.com/engine/installation/linux/linux-postinstall/)
+By default, Docker runs as the root user, requiring other users to access it with `sudo`. This extension does not assume root access, so you will need to create a Unix group called docker and add users to it. Instructions can be found here: [Create a Docker group](https://docs.docker.com/install/linux/linux-postinstall/)
 
 ## Connecting to `docker-machine`
 


tfenster:
Does anyone have feedback on this? Doesn't make sense or not good enough or anything? Maybe @chrisdias? I'd be happy to accept it if you don't like that change (and build the extension from my fork in the future) but would appreciate some kind of feedback

Thanks a lot in advance

StephenWeatherford:
@tfenster Would it be better to have this as a VS Code setting?

tfenster:
@StephenWeatherford not sure if I understand you correctly but the use case is to be able to change the connected endpoint on the fly, so I am not sure how a setting would help here?

StephenWeatherford:
BTW, how often do you tend to change this?

tfenster:
@StephenWeatherford Would the extension pick up the changed settings immediately? I could implement it there as well. Not as convenient but better then restarting. Currently I do it only once or twice a day but open multiple cmds because that is quicker. If the extension would allow easier changing, I'd do it there

But again, if I am alone with this and you think this would clutter the extension, I am happy to have my own fork

StephenWeatherford:
@chrisdias @fiveisprime What are you thoughts?  I'm inclined to think adding a VS Code setting it the best idea, but that adding a command is probably overkill for the use case, but don't know how common this is.

tfenster:
maybe making that setting an array and allowing to select one of the values through a command would be a compromise?

commands/utils/create-terminal.ts
```diff
@@ -0,0 +1,9 @@
+import * as vscode from "vscode";
+import { setEnvInTerminal } from '../change-endpoint';
+
+export async function createTerminal(name: string) : Promise<vscode.Terminal> {
```
StephenWeatherford: Why did you make createTerminal async?  Doesn't seem to be required, and changes a bunch of calls.

tfenster:
Understood. I'll close it here and open a new one containing the new setting

PR #283, opened 2018-06-26
Add build, lint, testing, CI support

### LINKED ISSUE #232: Add CI build support
#219 was merged even though it had a compiler error. It was quickly fixed with ec5e219e1c311be0bdfd1cfbda1f72b360d72348 but it would have been nice if the pull request kicked off a CI build to catch this.

Perhaps Travis CI could be used for this?

StephenWeatherford:
Fixes #232


test/test.test.ts
```diff
@@ -0,0 +1,13 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// The module 'assert' provides assertion methods from node
+import * as assert from 'assert';
+
+suite("Tests", () => {
+    test("test", async () => {
+        assert.equal(true, true);
```
rcjsuen: What does this test accomplish exactly?

StephenWeatherford: Just tests that tests are actually running.  Should be removed when we have actual tests in place.

rcjsuen: Makes sense to me. Thanks for the clarification!

StephenWeatherford: Thanks for looking.

.travis.yml
```diff
@@ -0,0 +1,28 @@
+language: node_js
+
+sudo: false
+
+node_js:
+  - 'stable'
```
rcjsuen: Does `'stable'` make sense? VS Code bundles its own Electron/Node.js/v8, correct? Should that be used instead?

Also, is there any value in asking Travis CI to build for OS X? Probably not that much I suppose?

tslint.json
```diff
@@ -0,0 +1,315 @@
...
+        "no-empty-interfaces": false,
+        "no-missing-visibility-modifiers": false,
+        "no-multiple-var-decl": false,
+        "no-switch-case-fall-through": false,
+        "typeof-compare": false
+    },
+    "rulesDirectory": "node_modules/tslint-microsoft-contrib/",
+    "linterOptions": {
+        "exclude": [
+            "src/mongo/grammar/**", // generated files
```
EricJizbaMSFT: nit: no reason to exclude mongo stuff in this repo

.travis.yml
StephenWeatherford: @EricJizbaMSFT You have any thoughts on these?  

@rcjsuen It's what we've been using for our other extensions without issue, I'm going to go ahead and merge for now, but we can always revisit.

rcjsuen:
Why was `package-lock.json` deleted?

StephenWeatherford:
Main reason is that it keeps us from automatically getting security updates (GitHub was showing a warning about this).

PR #287, opened 2018-06-29
Estebanreyl/bugfixes

estebanreyl:
- Fixed HTTP request issue in explorer when viewing images within a selected repository.

- Improved azure explorer speed dramatically (7.5x +) by using a nonblocking approach to api calls and fetch requests. This was also implemented for images themselves

- Alphabetically organized registries listed from azure.


chrisdias:
@estebanreyl do we need to add the `.vs` folder?  i assume this comes from Visual Studio, but you need to use VS Code to develop the extension, so these become unnecessary.

estebanreyl:
I don't think we do, that's my mistake, also I just realized that the nonblocking model I used does not limit outgoing connections, as a result azure could potentially throttle the connections. Ill update these, along with remove the .vs folder and then resubmit the PR.

PR #301, opened 2018-07-06
Some lint clean-up

StephenWeatherford:
Simple rules fixed, none of these should be functionality changes.  Also disabled a lot of less-useful rules.


StephenWeatherford:
ping

tslint.json
```diff
@@ -220,26 +220,26 @@
...
         "import-spacing": true,
         "indent": [
-            false, // TODO: true,
+            true, // TODO: true,
             "spaces"
         ],
         "linebreak-style": false, //changed
         "newline-before-return": false, // changed
-        // TODO: "no-consecutive-blank-lines": true,
+        "no-consecutive-blank-lines": false, // changed
```
EricJizbaMSFT: 'no-consecutive-blank-lines' and 'no-trailing-whitespace' are easily auto-fixed and make code easier to read IMO. What's the downside?

tslint.json
```diff
@@ -220,26 +220,26 @@
             "arguments",
             "statements"
         ],
-        // TODO: "curly": true,
-        // TODO: "eofline": true,
+        "curly": false, // changed
```
EricJizbaMSFT: I've never heard you complain about 'curly' before - was this intentional? Does this one annoy you?

It feels like a lack of curly braces is just asking for bugs when copying/pasting, etc. or as described in the main rule [here](https://palantir.github.io/tslint/rules/curly/)
```
if (foo === bar)
    foo++;
    bar++;
```
> In the code above, the author almost certainly meant for both foo++ and bar++ to be executed only if foo === bar. However, they forgot braces and bar++ will be executed no matter what. This rule could prevent such a mistake.

tslint.json
```diff
@@ -70,7 +70,7 @@
         "no-regex-spaces": true,
         "no-sparse-arrays": true,
         "no-stateless-class": true,
-        // TODO: "no-string-literal": true,
+        "no-string-literal": false, // changed
```
EricJizbaMSFT: I'm not sure I understand why `foo['bar']` would ever be better than `foo.bar`? Don't you lose a bunch of useful things like autocomplete and type safety?

tslint.json
```diff
@@ -81,19 +81,19 @@
...
         "no-with-statement": true,
-        // TODO: "promise-function-async": true,
+        // TODO:"promise-function-async": true,
         // TODO: "promise-must-complete": true,
         "radix": true,
         "react-this-binding-issue": true,
         "react-unused-props-and-state": true,
-        // TODO: "restrict-plus-operands": true,
+        "restrict-plus-operands": false, // changed
         "strict-boolean-expressions": [
```
EricJizbaMSFT: I get annoyed by `strict-boolean-expressions`, too, but this is one that I begrudgingly keep because it's saved me several times. Specifically, when I had something like this:
```
async function foo(): Promise<boolean> {

}

if (foo()) {

}
```

You'll notice I forgot the 'await' when calling 'foo' and tslint will catch that because it doesn't like having a promise in the boolean expression.

tslint.json
```diff
@@ -220,26 +220,26 @@
...
             "statements"
         ],
-        // TODO: "curly": true,
-        // TODO: "eofline": true,
+        "curly": false, // changed
+        "eofline": false, // changed
         "import-spacing": true,
         "indent": [
-            false, // TODO: true,
+            true, // TODO: true,
```
EricJizbaMSFT: nit: remove `// TODO: true`

tslint.json
StephenWeatherford: I learned that rule a long time ago, but I've never actually seen it occur in the wild.

The main reason for disabling though is just the number of exceptions that are already in the codebase, not wanting to cause a bunch of churn, esp. since there are a lot of customer PRs out.  I'll postpone some of these for reconsideration later, though, since you are right about the auto-correct.

tslint.json
StephenWeatherford: Not ready to completely throw in the towel on this one, although I'll have to see if there are some reasonable compromises with the options.

PR #303, opened 2018-07-12
Add tests for 'configure' (Add docker files to workspace)

StephenWeatherford:
1) Added vscode-azureextensionui and support, using it for "configure" code
2) Launch tests from F5
3) NO_TIMEOUTS support for F5 (test/index.ts)
4) Open tests in context of workspace with a single folder (testOutput) by opening test/test.code-workspace
5) Add tests and supporting code

### DIFF
--- a/test/configure.test.ts
+++ b/test/configure.test.ts
@@ -20,7 +20,7 @@ import { getTestRootFolder, constants } from './global.test';
 let testRootFolder: string = getTestRootFolder();
 
 suite("configure (Add Docker files to Workspace)", function (this: Suite): void {
-    this.timeout(10 * 1000);
+    this.timeout(30 * 1000);
 
     const outputChannel: vscode.OutputChannel = vscode.window.createOutputChannel('Docker extension tests');
     ext.outputChannel = outputChannel;


test/assertEx.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export function unorderedArraysEqual<T>(actual: T[], expected: T[], message?: string): void {
+    let result = areUnorderedArraysEqual(actual, expected);
+    assert(result.areEqual, `${message || "Unordered arrays are not equal"}\n${result.message}`);
+}
+
+export function notUnorderedArraysEqual<T>(actual: T[], expected: T[], message?: string): void {
+    let result = areUnorderedArraysEqual(actual, expected);
+    assert(!result.areEqual, `${message}\n${result.message}`);
+}
```
PrashanthCorp: Question: Would it make sense to use have unorderedArraysEqual return a boolean, then use `console.assert(!unorderedArraysEqual)`. It looks like you're defining a second test B to solely test the negation of test A.

helpers/randomness.ts
```diff
@@ -0,0 +1,6 @@
+import * as crypto from "crypto";
+
+export function getRandomHexString(length: number = 10): string {
+    const buffer: Buffer = crypto.randomBytes(Math.ceil(length / 2));
+    return buffer.toString('hex').slice(0, length);
+}
```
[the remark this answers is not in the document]
StephenWeatherford: I've deleted it - ended up not using it.

test/assertEx.ts
StephenWeatherford: > It looks like you're defining a second test B to solely test the negation of test A.

It's not a test, it's an assertion to be used in tests. Having separate functions like this allows both to provide helpful messages for analyzing tests.  That didn't happen correctly here, though, fixed the message:

    assert(!result.areEqual, `${message || "Unordered arrays are equal but were expected not to be"}\n${result.message}`);

configureWorkspace/configure.ts
```diff
@@ -399,8 +398,8 @@ interface PackageJson {
     artifactName: string
 }
 
-async function getPackageJson(folder: vscode.WorkspaceFolder): Promise<vscode.Uri[]> {
-    return vscode.workspace.findFiles(new vscode.RelativePattern(folder, 'package.json'), null, 1, null);
+async function getPackageJson(folderPath: string): Promise<vscode.Uri[]> {
+    return vscode.workspace.findFiles(new vscode.RelativePattern(folderPath, 'package.json'), null, 1, null);
```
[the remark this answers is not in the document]
StephenWeatherford: Using Uri instead of string is a more interesting suggestion. But personally not a fan of using it for paths that you know are file paths, because Uri carries the idea that it might be a URL, so then do you have to assert that it's a file path before calling fsPath?  Because right now the code makes that assumption whenever it calls fsPath, so it might as well be explicit from the start that it's not a URL.

Node.js already handles both '/' and '\' interchangeably and as long as you are careful to use Node.js functions (e.g. path.join, path.basename) instead of using string manipulation, everything works fine.

configureWorkspace/config-utils.ts
```diff
@@ -1,42 +1,70 @@
...
+export async function quickPickOS(): Promise<OS> {
     var opt: vscode.QuickPickOptions = {
         matchOnDescription: true,
         matchOnDetail: true,
         placeHolder: 'Select Operating System'
     }
 
-    const items: string[] = ['Windows', 'Linux'];
+    const OSes: OS[] = ['Windows', 'Linux'];
+    const items = OSes.map(p => <IAzureQuickPickItem<OS>>{ label: p, data: p });
```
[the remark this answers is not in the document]
StephenWeatherford: Not much duplicate code, IMO not worth more refactoring.  I already improved it.

configureWorkspace/config-utils.ts
```diff
@@ -1,42 +1,70 @@
...
-    items.push('Other');
+    const platforms: Platform[] = [
+        'Go',
+        'Java',
+        '.NET Core Console',
+        'ASP.NET Core',
+        'Node.js',
+        'Python',
+        'Other'
+    ];
```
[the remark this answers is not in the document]
StephenWeatherford: Personally no. I like intermediate variables that essentially comment what's going on by their names and types.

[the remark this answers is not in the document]
StephenWeatherford: I mean yes, I think it would be less readable, or maybe about the same.

helpers/async.ts
```diff
@@ -0,0 +1,15 @@
+import * as glob from 'glob';
```
EricJizbaMSFT: nit: So does this repo not use the copyright header consistently?

configureWorkspace/configure.ts
```diff
@@ -481,24 +481,24 @@ async function readPomOrGradle(folder: vscode.WorkspaceFolder): Promise<PackageJ
...
     const opt: vscode.QuickPickOptions = {
         matchOnDescription: true,
         matchOnDetail: true,
         placeHolder: 'Select Project'
     }
 
     const projectFiles: string[] = await new Promise<string[]>((resolve, reject) => {
-
-        glob('**/*.csproj', { cwd: folder.uri.fsPath }, (err, matches: string[]) => {
+        glob('**/*.csproj', { cwd: folderPath }, (err, matches: string[]) => {
```
EricJizbaMSFT: globAsync?

test/assertEx.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    if (!error) {
+        throw new Error(`Expected exception or rejection: ${parseError(expected).message}`);
+    }
+    for (let prop of Object.getOwnPropertyNames(expected)) {
+        assert.equal(error[prop], expected[prop], `Error did not have the expected value for property '${prop}'`);
+    }
+}
+
+suite("assertEx", () => {
```
EricJizbaMSFT: Can you add a little more description to this file? Not sure what 'Ex' stands for (Extensions? Extra?). I honestly thought these were just dummy tests at first and didn't realized these functions were primarily for use elsewhere

And usually the pattern is to have a `*.test.ts` to contain the actual tests. I know this is bit of an edge case, but it honestly confuses me that you're testing 'test' code and I think splitting this into `assertEx.ts` and `assertEx.test.ts` might help

test/configure.test.ts
```diff
@@ -0,0 +1,462 @@
...
+            // Delete everything in the root testing folder
+            assert(path.basename(testRootFolder) === constants.testOutputName, "Trying to delete wrong folder");;
+            await fse.emptyDir(testRootFolder);
+            await func();
+        });
+    }
+
+    // Node.js
+
+    suite("Node.js", () => {
```
EricJizbaMSFT: screen shot 2018-07-17 at 1 32 09 pm

Not a good sign when adding a new file - can we split it up by language maybe?

helpers/async.ts
StephenWeatherford: Oh, it's pretty consistent about *not* using it. :-)

StephenWeatherford: On my list.

PR #311, opened 2018-07-17
Jackson esteban/unified client

estebanreyl:
- Created a unified client in a singleton to facilitate acquiring the azure account, subscriptions, resource groups, registries etc. from a unified simple location.

 Note that I have not refactored existing code to fit the new singleton, this will be done later.

Any feedback would be highly appreciated before refactoring other code and opening a PR on the main branch.


estebanreyl:
Sorry wrong location.

PR #312, opened 2018-07-17
Lint: typedef

StephenWeatherford:


configureWorkspace/configure.ts
```diff
@@ -525,7 +525,10 @@ async function findCSProjFile(folder: vscode.WorkspaceFolder): Promise<string> {
     return projectFiles[0].slice(0, -'.csproj'.length);
```
nturinski: Any concern about projectFIles[0] ever returning undefined?

Issue #73, opened 2017-03-14
Allow restart container (Compose Up and Compose Down) with one command


akalcik:
Often when you develop something new, you make lot of changes to __config.yml_. It will be helpful when I could quick restart container with one command. It will be also cool when the __config.yml_ could be monitored and restart  container automatically on change.

StephenWeatherford:
@akalcik Just to let you know, we shipped a new version of the extension today that contains "Docker compose restart".

PrashanthCorp:
PR shipped, closing this

PR #316, opened 2018-07-18
Add docker compose restart

### LINKED ISSUE #250: suggestion: add "docker compose restart" command
I find as I'm working on a `Dockerfile` I tend to run the commands `docker-compose up` and `docker-compose down` quite a bit in sequence.  I run the container, see that something is wrong, make edits to the `Dockerfile` and then run down and up.  

Maybe a `docker compose: restart` command would be useful.

PrashanthCorp:
Fixes #250. 
Addresses #73.

### DIFF
--- a/commands/docker-compose.ts
+++ b/commands/docker-compose.ts
@@ -1,6 +1,7 @@
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { COMPOSE_FILE_GLOB_PATTERN } from '../dockerExtension';
+import { ext } from '../extensionVariables';
 import { reporter } from '../telemetry/telemetry';
 import { createTerminal } from './utils/create-terminal';
 const teleCmdId: string = 'vscode-docker.compose.'; // we append up or down when reporting telemetry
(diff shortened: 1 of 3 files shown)


PrashanthCorp:
The build failed due to a timeout. Restarting build

commands/docker-compose.ts
```diff
@@ -100,3 +100,8 @@ export function composeUp(dockerComposeFileUri?: vscode.Uri, selectedComposeFile
 export function composeDown(dockerComposeFileUri?: vscode.Uri, selectedComposeFileUris?: vscode.Uri[]) {
     compose('down', 'to take down', dockerComposeFileUri, selectedComposeFileUris);
 }
+
```
StephenWeatherford: Have you tried running this from the command palette?  I think you'll find something undesirable happen.

StephenWeatherford:
Also, I don't see any CTI tests...  Please add them for compose up/down as well and anything related.

commands/docker-compose.ts
PrashanthCorp: Yes. I don't know of any other way to run docker compose Up or Down (the package.json only specifies them in the commands section, not in any context menu). 

The undesirable side-effect is that I'm asked to choose a compose file _once for each of the compose down and up operations_. 
@fiveisprime, is that a critical point of friction?

commands/docker-compose.ts
```diff
@@ -77,34 +77,34 @@ async function compose(command: string, message: string, dockerComposeFileUri?:
...
-            /* __GDPR__
-               "command" : {
-                  "command" : { "classification": "SystemMetaData", "purpose": "FeatureInsight" }
-               }
-             */
-            reporter.sendTelemetryEvent('command', {
-                command: teleCmdId + command
+        for (let command of commands) {
+            selectedItems.forEach((item: Item) => {
+                terminal.sendText(command.toLowerCase() === 'up' ? `docker-compose -f ${item.file} ${command} ${detached} ${build}` : `docker-compose -f ${item.file} ${command}`);
```
StephenWeatherford: Could you move the ? : clauses onto separate lines for better readability?  Thx.

commands/docker-compose.ts
```diff
@@ -34,7 +34,7 @@ function computeItems(folder: vscode.WorkspaceFolder, uris: vscode.Uri[]): vscod
     return items;
 }
 
-async function compose(command: string, message: string, dockerComposeFileUri?: vscode.Uri, selectedComposeFileUris?: vscode.Uri[]): Promise<void> {
+async function compose(commands: string[], message: string, dockerComposeFileUri?: vscode.Uri, selectedComposeFileUris?: vscode.Uri[]): Promise<void> {
```
StephenWeatherford: commands: ('up' | 'down')[]

commands/docker-compose.ts
```diff
@@ -77,34 +77,34 @@ async function compose(command: string, message: string, dockerComposeFileUri?:
...
-        if (reporter) {
-            /* __GDPR__
-               "command" : {
-                  "command" : { "classification": "SystemMetaData", "purpose": "FeatureInsight" }
-               }
-             */
-            reporter.sendTelemetryEvent('command', {
-                command: teleCmdId + command
+        for (let command of commands) {
+            selectedItems.forEach((item: Item) => {
```
StephenWeatherford: nit: I know, not your code, but...  personally I don't like using forEach because you can't just debug through the loop with F10, you have to set a breakpoint in the function. for of is so much cleaner now.

Issue #229, opened 2018-03-06
Unable to connect to Docker on Windows 7 with correctly configured DOCKER_HOST env


thadguidry:
Correctly configured and working Docker on Windows 7 as shown in next 3 screen capture snippets...

![capture4](https://user-images.githubusercontent.com/986438/37055951-a12fadec-2148-11e8-8ab1-baaf256766c2.PNG)

![capture](https://user-images.githubusercontent.com/986438/37055796-48887570-2148-11e8-9568-d9c468131c29.PNG)
![capture2](https://user-images.githubusercontent.com/986438/37055981-af4d3f02-2148-11e8-9405-aef75d6cec1a.PNG)

Yet, when I start VSC as Administrator , it still cannot connect...

![capture3](https://user-images.githubusercontent.com/986438/37055992-b3ccccfa-2148-11e8-9936-f6df601cdfd3.PNG)

chrisdias:
@thadguidry is this when you are expanding the tree in the explorer, or when you run a command from the command palette?

chrisdias:
Unfortunately I don't have a win7 machine readily available (not sure where I can even get one). 

You could clone this repo, open it in VS Code and run it.  You would want to search for `Daemon` and then put a breakpoint on each one of these, and when hit inspect the `error` object.

thadguidry:
@chrisdias yeah, when I expand the Docker explorer tree.. and click images.

Awesome, ok...I'll breakpoint it and see what I find...perhaps contribute :)

chrisdias:
This will be very helpful, thanks.  Here's the two lines to put breakpoints on:

https://github.com/Microsoft/vscode-docker/blob/82e3038e73efe666c7ae8d0e221e2e9c6715efa0/explorer/models/rootNode.ts#L147

https://github.com/Microsoft/vscode-docker/blob/82e3038e73efe666c7ae8d0e221e2e9c6715efa0/explorer/models/rootNode.ts#L246

Then inspect `error` and let me know what you find.

Thanks!!

thadguidry:
@chrisdias Sure thing.  Will let you know tomorrow when I get back to @Ericsson office. (5th largest software company, right behind @Microsoft) :)

thadguidry:
I cloned this repo, opened in VSC, and clicked green Launch Extension.
New VSC opened, I then clicked on Docker panel and this error came up.

[...]

Issue #146, opened 2017-10-16
Using VSCode with Docker Toolbox "sslv3 alert bad certificate"


sergeiwaigant:
Hi there,

is there a workaround for this error?

`write EPROTO 101057795:error:14094412:SSL routines:ssl3_read_bytes:sslv3 alert bad certificate:d:\buildagent\workspace\318698\vendor\node\deps\openssl\openssl\ssl\s3_pkt.c:1493:SSL alert number 42`

I am using Docker Toolbox for Windows (running boot2docker in VBox) and I have configured the DOCKER_HOST environment variable to the IP:Port taken from `docker-machine ls` command.

```
docker-machine ls
NAME      ACTIVE   DRIVER       STATE     URL                         SWARM   DOCKER        ERRORS
default   *        virtualbox   Running   tcp://192.168.99.101:2376           v17.09.0-ce
```

When starting with Docker Toolbox Quickstart Terminal I figured out that the docker-machine command generates certificates and place them in `$HOME\.docker\machine\machines\default`.
For using this certificates in Windows 10 Ubuntu Bash I just copied them into the lxss-home\.docker\machine\machines\default and this works fine after configuring docker-machine.

Where should I copy the certificates so that the vscdoe extension can use them?

Thanks!
-Sergei

sergeiwaigant:
Problem solved...

I just have to set the following environment variables...

```
docker-machine env
$Env:DOCKER_TLS_VERIFY = "1"
$Env:DOCKER_HOST = "tcp://192.168.99.101:2376"
$Env:DOCKER_CERT_PATH = "$HOME\.docker\machine\machines\default"
$Env:DOCKER_MACHINE_NAME = "default"
$Env:COMPOSE_CONVERT_WINDOWS_PATHS = "true"
# Run this command to configure your shell:
# & "C:\Program Files\Docker Toolbox\docker-machine.exe" env | Invoke-Expression
```

PR #324, opened 2018-07-22
Improve docker-machine documentation.

mikepatrick:
Note the need for a DOCKER_CERT_PATH environment variable when using TLS.

The `docker-machine` section of the README calls out the need for a DOCKER_HOST environment variable, but does not mention the DOCKER_CERT_PATH environment variable.

Failure to have DOCKER_CERT_PATH set when using TLS causes strange looking TLS errors, which are only reported to the user as "Could not connect to docker daemon".  

This seems to cause a fair amount of confusion; see [here](https://github.com/Microsoft/vscode-docker/issues/146), [here](https://github.com/Microsoft/vscode-docker/issues/229), and [here](https://stackoverflow.com/questions/51402080/vscode-docker-extension-cannot-connect-to-docker-machine).

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,8 @@ By default, Docker runs as the root user, requiring other users to access it wit
 
 The default connection of the extension is to connect to the local docker daemon. You can connect to a docker-machine instance if you launch Visual Studio Code and have the DOCKER_HOST environment variable set to a valid host or if you set the `docker.host` configuration setting.
 
+If the docker daemon is using TLS, the DOCKER_CERT_PATH environment variable must also be set (e.g. `$HOME\.docker\machine\machines\default`).
+
 ## Contributing
 
 There are a couple of ways you can contribute to this repository:


StephenWeatherford:
Really appreciate the input.  How does the following look (https://github.com/Microsoft/vscode-docker/pull/325)?

## Connecting to `docker-machine`

The default connection of the extension is to connect to the local docker daemon. You can connect to a docker-machine instance if you launch Visual Studio Code and have the DOCKER_HOST environment variable set to a valid host (and DOCKER_CERT_PATH if you're using TLS) or if you set the `docker.host` configuration setting. See [the Docker documentation](https://docs.docker.com/machine/reference/env/) for more information.

README.md
```diff
@@ -124,6 +124,8 @@ By default, Docker runs as the root user, requiring other users to access it wit
 
 The default connection of the extension is to connect to the local docker daemon. You can connect to a docker-machine instance if you launch Visual Studio Code and have the DOCKER_HOST environment variable set to a valid host or if you set the `docker.host` configuration setting.
 
+If the docker daemon is using TLS, the DOCKER_CERT_PATH environment variable must also be set (e.g. `$HOME\.docker\machine\machines\default`).
```
StephenWeatherford: Much thanks for this contribution.

Should you point to (https://docs.docker.com/machine/reference/env/) for more information?

StephenWeatherford: I'll add this after the merge.

StephenWeatherford:
@mikepatrick Thanks!   Really appreciate this contribution.

PR #340, opened 2018-07-30
Mention admin accounts for ACR in readme.

PrashanthCorp:


README.md
```diff
@@ -144,7 +144,8 @@ Before we can accept your pull request you will need to sign a **Contribution Li
 
 Make sure you are signed in to the Docker Hub or Azure container registry from the docker CLI via `docker login` (using your username, not your e-mail address).
 
-If you are using an Azure container registry, you will need to get the username and password from Azure by right-clicking on the Azure container registry in the extension and selecting "Browse in the Azure Portal", then selecting the "Access Keys" tab.
+If you are using an Azure container registry, you will need to ensure admin user is enabled. To do this, right-click on the Azure container registry in the extension and select "Browse in the Azure Portal" - select the "Access Keys" tab in the portal. You will then see the username and password from Azure.
```
StephenWeatherford: If admin user is not enabled, they will not see it in the extension anyway, so I think this is confusing to put here.

estebanreyl: As a note, just submitted PR #341 which addresses this, Admin enabled is no longer necessary for listing container registries at all, only for launching an image into app services and this is only because app services does not support non admin enabled container registries.

EricJizbaMSFT: @PrashanthCorp can this PR be closed based on above comments? Been open for a few months now...

EricJizbaMSFT:
Closing due to lack of response and above comments

PR #332, opened 2018-07-26
estebanreyl/Unified Azure Utilities

estebanreyl:
Adds a singleton client for managing Azure credentials/resources internally. This helps facilitate production of further azure connectivity and extends on many of the existing methods adding optimized async support where possible. The client includes the following features:

- Acquire azureaccount
- Acquire ContainerRegistry/ResourceGroup management clients
- Validate sign in
- Get Azure Container Registries (Optional scoping by Subscription and ResourceGroup 
- Get Azure Subscriptions
- Get Azure Resource Groups (Optional scoping by Subscription)
- Get Session Credentials

PR #341, opened 2018-07-31
Estebanreyl/admin enabled fixes

### LINKED ISSUE #336: Error when listing Azure Registries
When I click on "Registries" -> "Azure" and succesfully log in, the following error pops up, and I cannot view any of my Azure container registries. 

```
The client '<my email>' with object id '6da91d27-c2b4-453c-959c-0d21bfff3999' does not have authorization to perform action 'Microsoft.ContainerRegistry/registries/listCredentials/action' over scope '/subscriptions/e4272367-5645-4c4e-9c67-3b74b59a6982/resourceGroups/contosoitnodeada3/providers/Microsoft.ContainerRegistry/registries/contosoitnode9fee'.
```

### LINKED ISSUE #359: Show non-admin-enabled Azure container registries in tree

estebanreyl:
-Fixes #336, fixes #359 

- Eliminated bugs related to failing to fetch a single container registry for any particular reason.

- Eliminated the need for registries to be admin enabled at all to be listed in Azure Container Registries explorer. 

- Provided a fix for launching to app services after the aforementioned change and an overall fix due to a very obscure bug that prevented launches when an available resource group had a location outside of the locations available for a subscription.

- As a side note rather than a feature, speed for container registries has further increased as a result of admin enabled elimination.


explorer/models/azureRegistryNodes.ts
```diff
@@ -202,23 +196,27 @@ export class AzureRepositoryNode extends NodeBase {
...
-                    node = new AzureImageNode(`${element.label}:${tags[i]}`, 'azureImageNode');
-                    node.azureAccount = element.azureAccount;
-                    node.password = element.password;
-                    node.registry = element.registry;
-                    node.serverUrl = element.repository;
-                    node.subscription = element.subscription;
-                    node.userName = element.userName;
-                    node.created = moment(new Date(JSON.parse(manifest.history[0].v1Compatibility).created)).fromNow();
-                    imageNodes.push(node);
+                    let data: any;
```
northtyphoon: looks data is only used in try-catch block?

explorer/deploy/webAppCreator.ts
```diff
@@ -172,9 +177,16 @@ class ResourceGroupStep extends WebAppCreatorStepBase {
             resourceGroups = results[0];
             locations = results[1];
             resourceGroups.forEach(rg => {
+                let location: string;
+                try {
+                    location = `(${locations.find(l => l.name.toLowerCase() === rg.location.toLowerCase()).displayName})`;
```
northtyphoon: >` [](start = 31, length = 1)

is backtick needed here?

northtyphoon: >location [](start = 20, length = 8)

You might want to change the name to locationDisplayName or locationDescription.

explorer/deploy/webAppCreator.ts
```diff
@@ -408,13 +420,22 @@ class WebsiteStep extends WebAppCreatorStepBase {
...
         super(wizard, 'Create Web App', azureAccount);
 
         this._serverUrl = context.serverUrl;
-        this._serverPassword = context.password;
-        this._serverUserName = context.userName;
+        if (context instanceof DockerHubImageNode) {
+            this._serverPassword = context.password;
+            this._serverUserName = context.userName;
+        }
+        if (context instanceof AzureImageNode) {
```
northtyphoon: >if [](start = 8, length = 2)

`else if` to avoid additional check

northtyphoon: you also might to log if context is neither DockerHubImageNode or AzureImageNode

---
In reply to: [206671989](https://github.com/Microsoft/vscode-docker/pull/341#discussion_r206671989) [](ancestors = 206671989)

explorer/deploy/webAppCreator.ts
```diff
@@ -447,6 +468,12 @@ class WebsiteStep extends WebAppCreatorStepBase {
                 await vscode.window.showWarningMessage(nameAvailability.message);
             }
         }
+        try {
+            await this.loginCredentials();
```
northtyphoon: >loginCredentials [](start = 23, length = 16)

Can you rename to something like acquireRegistryLoginCredentials? It's not clear from the name what credentials are about since it's in WebsiteStep class.

explorer/deploy/webAppCreator.ts
```diff
@@ -543,4 +568,74 @@ class WebsiteStep extends WebAppCreatorStepBase {
...
+                this._serverPassword = creds.passwords[0].value;
+                this._serverUserName = creds.username;
+                return;
+            } catch (error) {
+                vscode.window.showErrorMessage(error.message);
+                throw new Error(('Admin not enabled'));
+            }
+        }
+        await vscode.window.showErrorMessage('Azure App service currently only supports running images from Azure Container Registries with admin enabled');
+        throw new Error(('Admin not enabled'));
```
northtyphoon: >Admin [](start = 26, length = 5)

Azure Container Registry Admin is not enabled.

explorer/deploy/webAppCreator.ts
```diff
@@ -543,4 +568,74 @@ class WebsiteStep extends WebAppCreatorStepBase {
...
+        // Un-Comment when Admin enabled is not required
+        // let convertedCredentials: { password: string, username: string } = await this.getSessionCredentials();
+        // this._serverUserName = convertedCredentials.password;
+        // this._serverPassword = convertedCredentials.username;
```
northtyphoon: Is the support of service principal ready?

explorer/deploy/webAppCreator.ts
estebanreyl: Sorry, that was a leftover from before

explorer/deploy/webAppCreator.ts
estebanreyl: That seems reasonable to me

explorer/deploy/webAppCreator.ts
estebanreyl: It can be done on our side and it will probably be added on a per feature thing as we add functionality such as pull and push but app services itself does not seem to support them at the moment.

estebanreyl:
@sajayAntony

explorer/deploy/webAppCreator.ts
```diff
@@ -3,13 +3,18 @@
...
  *--------------------------------------------------------------------------------------------*/
 
+import { ContainerRegistryManagementClient } from 'azure-arm-containerregistry';
+import { Registry } from 'azure-arm-containerregistry/lib/models';
 import { ResourceManagementClient, ResourceModels, SubscriptionModels } from 'azure-arm-resource';
+import { Subscription } from 'azure-arm-resource/lib/subscription/models';
 import WebSiteManagementClient = require('azure-arm-website');
 import * as fs from 'fs';
 import * as path from 'path';
+import * as request from 'request-promise';
```
ankurkhemani: unused import, remove.

explorer/deploy/webAppCreator.ts
```diff
@@ -3,13 +3,18 @@
...
 import * as vscode from 'vscode';
 import * as WebSiteModels from '../../node_modules/azure-arm-website/lib/models';
 import { reporter } from '../../telemetry/telemetry';
+import { AzureAccount, AzureLoginStatus, AzureSession } from '../../typings/azure-account.api';
```
ankurkhemani: unused import, remove.

explorer/deploy/webAppCreator.ts
```diff
@@ -543,4 +569,26 @@ class WebsiteStep extends WebAppCreatorStepBase {
...
+        const resourceGroup: string = this._registry.id.slice(this._registry.id.search('resourceGroups/') + 'resourceGroups/'.length, this._registry.id.search('/providers/'));
+
+        if (this._registry.adminUserEnabled) {
+            try {
+                let creds = await client.registries.listCredentials(resourceGroup, this._registry.name);
+                this._serverPassword = creds.passwords[0].value;
+                this._serverUserName = creds.username;
+                return;
+            } catch (error) {
+                vscode.window.showErrorMessage(error.message);
```
ankurkhemani: await?

explorer/deploy/webAppCreator.ts
```diff
@@ -543,4 +569,26 @@ class WebsiteStep extends WebAppCreatorStepBase {
...
+    //Implements new Service principal model for ACR container registries while maintaining old admin enabled use
+    private async acquireRegistryLoginCredentials(): Promise<void> {
+        if (this._serverPassword && this._serverUserName) { return; }
+
+        const client = new ContainerRegistryManagementClient(this.azureAccount.getCredentialByTenantId(this._imageSubscription.tenantId), this._imageSubscription.subscriptionId);
+        const resourceGroup: string = this._registry.id.slice(this._registry.id.search('resourceGroups/') + 'resourceGroups/'.length, this._registry.id.search('/providers/'));
+
+        if (this._registry.adminUserEnabled) {
+            try {
+                let creds = await client.registries.listCredentials(resourceGroup, this._registry.name);
```
ankurkhemani: move this to `_util` file inside a new function `listRegistryCredentials `?

explorer/deploy/webAppCreator.ts
estebanreyl: In this case await pauses execution until the user clicks to exit the window which is not a guarantee, additionally it makes no difference as the error is shown when ready and execution ends nonetheless.

explorer/deploy/webAppCreator.ts
estebanreyl: While this makes sense to me conceptually I am honestly having trouble with how to go about it, I could create a function say in azureUtils that takes an azure account, a subscription, and a registry to do this and while this seems reasonable, and maybe I am biased because of PR #332 , it doesn't seem to follow the existing pattern of this file and seems more like something to add to my unified credentials that would look odd anywhere else. It seems most reasonable to me to add it to the Unified credentials branch and leave this like it is, at some point it may be good to refactor this code around that other branch however but that goes for the overall file,

ankurkhemani: Sounds good, you can consider refactoring in the other branch. Also, you are right about the overall file in general.

explorer/deploy/webAppCreator.ts
```diff
@@ -172,9 +173,16 @@ class ResourceGroupStep extends WebAppCreatorStepBase {
             resourceGroups = results[0];
             locations = results[1];
             resourceGroups.forEach(rg => {
+                let locationDisplayName: string;
+                try {
+                    locationDisplayName = locations.find(l => l.name.toLowerCase() === rg.location.toLowerCase()).displayName;
+                } catch (error) {
```
StephenWeatherford: What error is occurring?  Null ref?  I'd rather avoid exceptions than catch and recover.

explorer/deploy/webAppCreator.ts
StephenWeatherford: We shouldn't need to both show a message and throw an error - the error should get displayed by the caller, is that not happening?

One issue is that this code was copied from elsewhere and we need to actually update it to use the shared code in https://github.com/Microsoft/vscode-azuretools, so maybe that needs to be done before integrating these changes.

explorer/models/registryRootNode.ts
```diff
@@ -133,43 +133,43 @@ export class RegistryRootNode extends NodeBase {
...
-                    });
+                    let regs: ContainerModels.Registry[];
+                    try {
+                        regs = await client.registries.list();
+                        subsAndRegistries.push({
+                            'subscription': subs[i],
+                            'registries': regs
+                        });
+                    } catch (error) {
+                        vscode.window.showErrorMessage(error);
```
StephenWeatherford: Let caller show error.

StephenWeatherford: and other places.

Let me know if the caller is not showing the error.  There are further changes around command registration and error handling that I intend to make.

explorer/models/azureRegistryNodes.ts
```diff
@@ -202,23 +196,26 @@ export class AzureRepositoryNode extends NodeBase {
...
+                        node = new AzureImageNode(`${element.label}:${tags[i]}`, 'azureImageNode');
+                        node.azureAccount = element.azureAccount;
+                        node.registry = element.registry;
+                        node.serverUrl = element.repository;
+                        node.subscription = element.subscription;
+                        node.created = moment(new Date(JSON.parse(manifest.history[0].v1Compatibility).created)).fromNow();
+                        imageNodes.push(node);
+
+                    } catch (error) {
+                        vscode.window.showErrorMessage(error);
```
PrashanthCorp: I'm trying to understand where you expect the error to be coming from. Is it from the parse, or the request? I don't see lines 209-213 causing an error. 

If you're surer of where the error is expected to come from, it might be better to wrap just that line in a catch.

explorer/deploy/webAppCreator.ts
estebanreyl: Exactly, I could also store the result in a variable and check if it is available before assigning the value, say : 
const location : Location = locations.find(l => l.name.toLowerCase() === rg.location.toLowerCase());

if(location){
 locationDisplayName = location.displayName;
}else{
 ...
 }

Alternatively I could try changing the location acquisition by subscription strategy and do so by overall locations which would prevent the error in the grand majority of the very rare cases although I suspect this would have problems in cases when there is a private azure location. Also, is there a specific reason as to prefer avoiding rather than catch and recover?

explorer/models/registryRootNode.ts
estebanreyl: Do you mean handle this within asyncpool? I could do that so its a single location for this.

explorer/models/azureRegistryNodes.ts
estebanreyl: I anticipate  it would come from the request , that seems reasonable, I think I just moved it to the try to avoid having an additional check in case it failed but it is more unreadable.

explorer/deploy/webAppCreator.ts
estebanreyl: This does not happen indeed, in fact there is no visible information about an error occurring if I propagate it all the way to the code I didn't write, although I could modify it to act in this way. Additionally, while I agree that we should update the original code for the whole deploy to app services, this PR is essentially an important bugfix which by the nature of the bug broke the existing app services functionality and thus required me to solve a small number of issues on that side to keep existing functionality post the PR. Would it not be best to integrate this functional code with the added bugfixes before attempting the necessary refactoring that appears to me as less pressing?

explorer/deploy/webAppCreator.ts
StephenWeatherford: Yes, that's preferable, thanks.  Exceptions should be for things that are exceptional.  Really you should never see a null ref exception, it's completely avoidable with proper checking.  Having the location not match is a fully normal event that we just need to handle.

(Also, having a bunch of exceptions throw all over the place is darn annoying when you're trying to debug.)

explorer/deploy/webAppCreator.ts
StephenWeatherford: Makes sense.  Added https://github.com/Microsoft/vscode-docker/issues/358 to track that we want telemetry for such failures.

BTW, with my merge of https://github.com/Microsoft/vscode-docker/pull/350, the caller should now be catching and handling (telemetry and error messages) anything you throw from commands.  That should now be the correct way to handle things (let exceptions bubble up), except for cases like there where you need to continue.

explorer/deploy/webAppCreator.ts
```diff
@@ -447,6 +465,12 @@ class WebsiteStep extends WebAppCreatorStepBase {
                 await vscode.window.showWarningMessage(nameAvailability.message);
             }
         }
+        try {
+            await this.acquireRegistryLoginCredentials();
+        } catch (error) {
+            //Admin was not enabled, cannot proceed
+            throw new Error(('Admin not enabled'));
```
StephenWeatherford: nit: double parens

StephenWeatherford: You're making an assumption about why this failed - it could fail for any number of reasons, including network out - this message could mislead users.

StephenWeatherford: You should simply not have the try/catch here - let the exception bubble up (sync up with my PR I mentioned above).

explorer/deploy/webAppCreator.ts
```diff
@@ -543,4 +565,25 @@ class WebsiteStep extends WebAppCreatorStepBase {
...
+                let creds = await client.registries.listCredentials(resourceGroup, this._registry.name);
+                this._serverPassword = creds.passwords[0].value;
+                this._serverUserName = creds.username;
+                return;
+            } catch (error) {
+                vscode.window.showErrorMessage(error.message);
+                throw new Error(('Azure Container Registry Admin is not enabled'));
+            }
+        }
+        vscode.window.showErrorMessage('Azure App service currently only supports running images from Azure Container Registries with admin enabled');
```
StephenWeatherford: Again, you can't make that assumption.  Does the error that you get back in this scenario have a particular code that you can check for in order to throw a more specific error in this case?

StephenWeatherford: Just throw (use the longer message), don't show a message.

estebanreyl: I could check for whether the list failed because of admin was not enabled specifically and then throw accordingly. As for the Azure App service does not support... etc I have to make the assumption as the only way to identify this error is to go into the azure portal and note that your web app fails due to admin not being enabled, no error is actually given to the user from the API as long as credentials are valid.

explorer/deploy/webAppCreator.ts
estebanreyl: Sounds good, I'll try to follow that paradigm.

explorer/deploy/webAppCreator.ts
StephenWeatherford: I'm confused - you're throwing this error specifically when you get an exception.  Can you not check for the error code?

explorer/deploy/webAppCreator.ts
estebanreyl: sweet, Ill update accordingly

estebanreyl:
Ive updated the error handling, nonetheless, bubbling up all the way to the item in command registration does not seem to yield either the telemetry logging (based on console output I've seen) or an error for the user. Note I have rebased my branch with master. I am manually outputting the error to the user however at the highest level possible, let me know if this is better.

StephenWeatherford:
@estebanreyl I'm not seeing the error handling/telemetry problem.  For instance I tried publishing to an app service without admin enabled and got this:
image
and this:

** TELEMETRY("vscode-docker/vscode-docker.createWebApp", 0.1.1-alpha) properties={"isActivationEvent":"false","cancelStep":"","result":"Failed","error":"Error","errorMessage":"Azure App service currently only supports running images from Azure Container Registries with admin enabled"}, measures={"duration":23.597}
createTelemetryReporter.js:26

explorer/models/azureRegistryNodes.ts
```diff
@@ -202,23 +196,28 @@ export class AzureRepositoryNode extends NodeBase {
...
-                    imageNodes.push(node);
+                    let data: string;
+                    try {
+                        data = await request.get('https://' + element.repository + '/v2/' + element.label + `/manifests/${tags[i]}`, {
+                            auth: {
+                                bearer: accessTokenARC
+                            }
+                        });
+                    } catch (error) {
+                        vscode.window.showErrorMessage(error);
```
StephenWeatherford: vscode.window.showErrorMessage(parseError(error).message);

explorer/models/registryRootNode.ts
StephenWeatherford: Except change to show parseError(error).message

StephenWeatherford:
Okay, just a couple minor changes left.  The wizard code has much improved since it was copied into this repo.  Thx.

estebanreyl:
Okay, I think its ready, just fixed the merge conflicts as well.

estebanreyl:
Oh I was confused about that, Ill add it now.

estebanreyl:
@StephenWeatherford  I added the set output channel again.

PR #342, opened 2018-07-31
fix tagname check

chrisdias:


commands/push-image.ts
```diff
@@ -35,7 +35,7 @@ export async function pushImage(context?: ImageNode): Promise<void> {
                 command: teleCmdId
             });
 
-            if (imageName.toLowerCase().indexOf('azurecr.io')) {
+            if (imageName.toLowerCase().includes('azurecr.io')) {
```
StephenWeatherford: I'd suggest you change the name of the event so we can easily distinguish the old, incorrect ones from the new, correct ones.

No need to run the tool, we'll be able to classify them the first time they appear in Nova.

chrisdias:
I kind of like the current event name :) so i'm not sure what to change it to ("`2`"?).  We will know that after a certain date the data is correct and eventually we'll purge the old incorrect data, so i'm leaning towards keeping it the same and dealing with it on the data side. let me know if you think this is a bad idea.

PR #348, opened 2018-08-02
Debug telemetry reporter

StephenWeatherford:


StephenWeatherford:
Depends on https://github.com/Microsoft/vscode-docker/pull/348

telemetry/telemetry.ts
```diff
@@ -1,35 +1,15 @@
...
-    if (extensionPackage) {
-        return {
-            name: extensionPackage.name,
-            version: extensionPackage.version,
-            aiKey: extensionPackage.aiKey
-        };
-    }
-    return;
+export function initializeTelemetryReporter(ctx: vscode.ExtensionContext): void {
+    reporter = createTelemetryReporter(ctx);
```
EricJizbaMSFT: Can we just get rid of this file now and move these two lines over to dockerExtension.ts?

StephenWeatherford: Almost. Lots of places use the reporter from this file, I'll be getting rid of it once I've moved to using the shared registerCommand.

PR #349, opened 2018-08-03
 Change getExposedPorts to return port w/ protocol

### LINKED ISSUE #284: UDP exposed ports not launching correctly
For example in my config if I have:

```
        "ExposedPorts": {
            "20000/udp": {}
        },
```
The launched container is given the following flags:

`$ docker run --rm -d -p 20000:20000 <redacted image>`

It should be:

`$ docker run --rm -d -p 20000:20000/udp <redacted image>`

I think it's just a bug with vscode parsing the config and not knowing to correctly add /udp when it's a udp service.

If I'm doing something wrong myself please do let me know.

PrashanthCorp:
Fixes #284

### DIFF
--- a/commands/start-container.ts
+++ b/commands/start-container.ts
@@ -1,12 +1,12 @@
 import * as cp from 'child_process';
 import * as fs from 'fs';
+import os = require('os');
+import vscode = require('vscode');
 import { ImageNode } from '../explorer/models/imageNode';
 import { reporter } from '../telemetry/telemetry';
 import { createTerminal } from './utils/create-terminal';
 import { docker, DockerEngineType } from './utils/docker-endpoint';
 import { ImageItem, quickPickImage } from './utils/quick-pick-image';
-import os = require('os');
-import vscode = require('vscode');
 
 const teleCmdId: string = 'vscode-docker.container.start';
 
--- a/commands/start-container.ts
+++ b/commands/start-container.ts
@@ -1,12 +1,12 @@
 import * as cp from 'child_process';
 import * as fs from 'fs';
-import os = require('os');
-import vscode = require('vscode');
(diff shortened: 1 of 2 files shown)


commands/start-container.ts
```diff
@@ -29,7 +29,7 @@ export async function startContainer(context?: ImageNode, interactive?: boolean)
         docker.getExposedPorts(imageToStart.Id).then((ports: string[]) => {
             let options = `--rm ${interactive ? '-it' : '-d'}`;
             if (ports.length) {
-                const portMappings = ports.map((port) => `-p ${port}:${port}`);
+                const portMappings = ports.map((port) => `-p ${port.split("/")[0]}:${port}`);
```
StephenWeatherford: Please comment with what you're expecting and the output you're going for.

StephenWeatherford: Are there are testcases?

PrashanthCorp: current output: 
`-p 8080:8080`

output I'm going for: 
`-p 8080:8080/tcp` , `-p 3010:3010/tcp`

Notice the difference in 8080/tcp. Earlier, we stripped the protocol when returning the function (see `getExposedPorts`). 

The earlier command was: 
`docker run --rm -d -p 8080:8080 node-todo:latest`
Now with the change: 
`docker run --rm -d -p 8080:8080/tcp node-todo:latest`

PrashanthCorp: Done.

StephenWeatherford: I wanted code comments. If you need to explain it to me, you need to explain it to code maintainers.

PrashanthCorp: Ah, my bad. Added comment

PR #366, opened 2018-08-08
Add configure command telemetry, plus code improvements

StephenWeatherford:


configureWorkspace/configure.ts
```diff
@@ -459,45 +468,55 @@ function getDefaultPackageJson(): PackageJson {
...
+ */
+async function readPomOrGradle(folderPath: string): Promise<{ foundPath?: string, packageContents: PackageJson }> {
     let pkg: PackageJson = getDefaultPackageJson(); //default
+    let foundPath: string | undefined;
+
+    let pomPath = path.join(folderPath, 'pom.xml');
+    let gradlePath = path.join(folderPath, 'build.gradle');
 
-    if (fs.existsSync(path.join(folderPath, 'pom.xml'))) {
+    if (fs.existsSync(pomPath)) {
```
EricJizbaMSFT: I know - not your code - but pref `await fse.pathExists`

configureWorkspace/configure.ts
```diff
@@ -637,28 +675,35 @@ export async function configure(folderPath?: string): Promise<void> {
...
-        });
-    }
+    // Don't wait
+    vscode.window.showInformationMessage(
+        filesWritten.length ?
+            `The following files were written into the workspace:${EOL}${EOL}${filesWritten.join(', ')}` :
+            "No files were written"
+    );
 
     async function createWorkspaceFileIfNotExists(fileName: string, generatorFunction: GeneratorFunction): Promise<void> {
```
EricJizbaMSFT: Again - not your code - but I'm really not a fan of these nested functions. It makes it difficult to read. Specifically, it makes it seem like filesWritten is always empty when you're showing the info message above (until I realized it was being modified in this function)

configureWorkspace/configure.ts
```diff
@@ -637,28 +675,35 @@ export async function configure(folderPath?: string): Promise<void> {
...
-        const workspacePath = path.join(folderPath, fileName);
+        const workspacePath = path.join(rootFolderPath, fileName);
+        let writeFile = false;
         if (fs.existsSync(workspacePath)) {
-            const item: vscode.MessageItem = await vscode.window.showErrorMessage(`A ${fileName} already exists. Would you like to override it?`, ...YES_OR_NO_PROMPT);
-            if (item.title.toLowerCase() === 'yes') {
-                fs.writeFileSync(workspacePath, generatorFunction(serviceName, platformType, os, port, pkg), { encoding: 'utf8' });
+            const response: vscode.MessageItem = await vscode.window.showErrorMessage(`A ${fileName} already exists.Would you like to overwrite it ? `, ...YES_OR_NO_PROMPTS);
```
EricJizbaMSFT: nit: Seems like the whitespace changes to this message might've been an accident? Specifically the spaces around the punctuation

Also rather than `A ${filename} already exists.` I would prefer `"${filename}" already exists.`

configureWorkspace/configure.ts
```diff
@@ -596,15 +619,17 @@ export async function configure(folderPath?: string): Promise<void> {
...
     }
 
     const platformType: Platform = await quickPickPlatform();
+    properties.platform = platformType;
 
     let os: OS | undefined;
     if (platformType.toLowerCase().includes('.net')) {
         os = await quickPickOS();
     }
+    properties.os = os;
```
EricJizbaMSFT: So this is _not_ the user's os, correct? It's the OS for the docker image? It seems pretty obvious here, but might be less obvious on the telemetry side

PR #378, opened 2018-08-15
Improve build-image and fix some strict typescript issues 

### LINKED ISSUE #377: Don't prompt for Dockerfile if only one in project (command palette->Build Image)

StephenWeatherford:
fixes #377

### DIFF
--- a/commands/build-image.ts
+++ b/commands/build-image.ts
@@ -3,9 +3,6 @@ import * as vscode from "vscode";
 import { DialogResponses, IActionContext, UserCancelledError } from "vscode-azureextensionui";
 import { DOCKERFILE_GLOB_PATTERN } from '../dockerExtension';
 import { ext } from "../extensionVariables";
-import { reporter } from '../telemetry/telemetry';
-
-const teleCmdId: string = 'vscode-docker.image.build';
 
 async function getDockerFileUris(folder: vscode.WorkspaceFolder): Promise<vscode.Uri[]> {
     return await vscode.workspace.findFiles(new vscode.RelativePattern(folder, DOCKERFILE_GLOB_PATTERN), undefined, 1000, undefined);
@@ -117,15 +114,4 @@ export async function buildImage(actionContext: IActionContext, dockerFileUri?: 
     const terminal: vscode.Terminal = ext.terminalProvider.createTerminal('Docker');
     terminal.sendText(`docker build --rm -f "${dockerFileItem.fileName}" -t ${value} ${contextPath}`);
     terminal.show();
-
-    if (reporter) {
-        /* __GDPR__
-           "command" : {
-              "command" : { "classification": "SystemMetaData", "purpose": "FeatureInsight" }
-           }
-         */
-        reporter.sendTelemetryEvent('command', {
-            command: teleCmdId
-        });
-    }
 }
--- a/commands/build-image.ts
+++ b/commands/build-image.ts
@@ -1,5 +1,6 @@
 import * as path from "path";
 import * as vscode from "vscode";
+import { DialogResponses, IActionContext, UserCancelledError } from "vscode-azureextensionui";
 import { DOCKERFILE_GLOB_PATTERN } from '../dockerExtension';
 import { ext } from "../extensionVariables";
 import { reporter } from '../telemetry/telemetry';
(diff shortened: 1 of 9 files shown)


commands/build-image.ts
```diff
@@ -72,27 +74,33 @@ export async function buildImage(dockerFileUri?: vscode.Uri): Promise<void> {
...
-    if (!uri) { return; }
+    while (!dockerFileItem) {
+        let resolvedItem: Item | undefined = await resolveImageItem(workspaceFolder, dockerFileUri);
+        if (resolvedItem) {
+            dockerFileItem = resolvedItem;
+        } else {
+            let msg = "Couldn't find a Dockerfile in your workspace. Would you like to add Docker files to the workspace?";
+            let response = await ext.ui.showWarningMessage(msg, DialogResponses.yes, DialogResponses.no);
+            if (response === DialogResponses.no) {
+                actionContext.properties.cancelStep = msg;
```
EricJizbaMSFT: Users can also cancel the dialog by clicking the 'x' icon on the notification, so your 'cancelStep' telemetry here won't be entirely accurate. You could use a try/catch or you could do something like this to fix that: https://github.com/Microsoft/vscode-azureappservice/blob/e9046407094fedbb77d994c4125b2d2bfa5819b8/src/util.ts#L63

commands/build-image.ts
```diff
@@ -72,27 +74,33 @@ export async function buildImage(dockerFileUri?: vscode.Uri): Promise<void> {
...
 
-    let contextPath: string = uri.path;
+    let contextPath: string = dockerFileItem.folderPath;
     if (defaultContextPath && defaultContextPath !== '') {
         contextPath = defaultContextPath;
     }
 
+    // Get imageName based on name of subfolder containing the Dockerfile, or else workspacefolder
     let imageName: string;
-    if (process.platform === 'win32') {
```
EricJizbaMSFT: omg thank you for fixing this to use `path.basename` - looks so much cleaner 😂

dockerExtension.ts
```diff
@@ -165,26 +165,23 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
-    return client.stop();
+    return await client.stop();
 }
 
 namespace Configuration {
 
     let configurationListener: vscode.Disposable;
 
     export function computeConfiguration(params: ConfigurationParams): vscode.WorkspaceConfiguration[] {
-        if (!params.items) {
```
EricJizbaMSFT: Don't quite understand what this check was doing. It's safe to remove now?

commands/build-image.ts
```diff
@@ -1,69 +1,71 @@
...
 }
 
 function createItem(folder: vscode.WorkspaceFolder, uri: vscode.Uri): Item {
     let filePath = path.join(".", uri.fsPath.substr(folder.uri.fsPath.length));
 
     return <Item>{
-        description: null,
-        file: filePath,
+        description: undefined,
+        fileName: filePath,
```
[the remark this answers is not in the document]
StephenWeatherford: Fixed. filePath

commands/build-image.ts
```diff
@@ -1,69 +1,71 @@
...
     };
 }
 
-function computeItems(folder: vscode.WorkspaceFolder, uris: vscode.Uri[]): vscode.QuickPickItem[] {
-    let items: vscode.QuickPickItem[] = [];
-    // tslint:disable-next-line:prefer-for-of // Grandfathered in
-    for (let i = 0; i < uris.length; i++) {
-        items.push(createItem(folder, uris[i]));
-    }
+function createItems(folder: vscode.WorkspaceFolder, uris: vscode.Uri[]): Item[] {
```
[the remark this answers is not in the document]
StephenWeatherford: Done

commands/build-image.ts
StephenWeatherford: Thanks for catching.  Changed to this for now:

            let msg = "Couldn't find a Dockerfile in your workspace. Would you like to add Docker files to the workspace?";
            actionContext.properties.cancelStep = msg;
            await ext.ui.showWarningMessage(msg, DialogResponses.yes, DialogResponses.cancel);
            actionContext.properties.cancelStep = undefined;
            await vscode.commands.executeCommand('vscode-docker.configure');
            // Try again

commands/build-image.ts
StephenWeatherford: :-)

dockerExtension.ts
StephenWeatherford: Now it returns [] instead of undefined, which is what the signature says. Also signature says params should never be undefined (as pass to us by vscode). AFAIK this is correct.

Issue #334, opened 2018-07-26
In building/pushing an image, make it easier to choose a registry


PrashanthCorp:
Triage decided feature:
- [x] After first push, when registry is specified, ask to save setting (workspace or global)
- [x] Fuse `defaultRegistry` and `defaultRegistryPath` settings into one setting. Note: If a user already has both set, we should either warn the user about it, or change the setting ourselves. 
- [x] Select the prefix in the edit box so that user can easily type over it
- [x] Add menu item to set registry as default
- [x] If they try to push an image with no prefix (username/whatever), we ask if they wanted to provide a prefix - need to figure out wording/etc
- [x] While building an image, we use a flag that lets us know whether or not to prefix registry.
- [x] Discover-ability of the above flag: If the flag is not set, we haven't asked before, and a default registry is set .
- [x] Add prefix for Tag, and for Build if flag set.  Don't apply to push (push might send user to Tag)

StephenWeatherford:
See https://github.com/Microsoft/vscode-docker/issues/253

StephenWeatherford:
1) After first push, ask to save setting (workspace or global)
2) Add menu item to set registry as default 
3) If they try to push an image with no prefix (username/whatever), we ask if they wanted to provide a prefix - need to figure out wording/etc

PrashanthCorp:
> - [ ] After first push, ask to save setting (workspace or global)

By first, do we mean: 
a) first ever push in the extension
b) first push of an image (discern this by ID perhaps?)

@fiveisprime

fiveisprime:
Originally, I was thinking first push to each endpoint, though now I'm thinking first ever push. Just something to indicate that the functionality exists.

PrashanthCorp:
Closing via above PR's

PR #398, opened 2018-08-22
Prompt user on first push (containing a prefix) to store the path; combine defaultRegistry&defaultRegistryPath

### LINKED ISSUE #396: Fuse defaultRegistry and defaultRegistryPath settings into one setting.
Note: If a user already has both set, we should either warn the user about it, or change the setting ourselves.

PrashanthCorp:
Functionally important changes in the title

Chronologically - 
- Refactor RegisterCommand due to the max func length tslint rule - the function was at the limit of the number of lines. This prevents any change from happening there until this was addressed. 
- combine two workspace settings (defaultRegistry&defaultRegistryPath)
- Prompt user on first push (containing a prefix) to store the path

Fix #396 

Addresses #334

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The Docker extension comes with a number of useful configuration settings allowi
 | `docker.attachShellCommand.windowsContainer` | Attach command to use for Windows containers | `powershell`
 | `docker.dockerComposeBuild` | Run docker-compose with the --build argument, defaults to true | `true`
 | `docker.dockerComposeDetached` | Run docker-compose with the --d (detached) argument, defaults to true | `true`
-| `docker.defaultRegistry` | Default registry when tagging an image, empty string will target Docker Hub when pushing. | `""`
+| `docker.defaultRegistry` | (Deprecated) Default registry when tagging an image, empty string will target Docker Hub when pushing. | `""`
 | `docker.defaultRegistryPath` | Path within registry to push to. | `""`
 | `docker.explorerRefreshInterval` | Explorer refresh interval, default is 1000ms. | `1000`
 | `docker.imageBuildContextPath` | Build context PATH to pass to Docker build command. | `""`
--- a/README.md
+++ b/README.md
@@ -92,7 +92,6 @@ The Docker extension comes with a number of useful configuration settings allowi
 | `docker.attachShellCommand.windowsContainer` | Attach command to use for Windows containers | `powershell`
 | `docker.dockerComposeBuild` | Run docker-compose with the --build argument, defaults to true | `true`
 | `docker.dockerComposeDetached` | Run docker-compose with the --d (detached) argument, defaults to true | `true`
-| `docker.defaultRegistry` | (Deprecated) Default registry when tagging an image, empty string will target Docker Hub when pushing. | `""`
 | `docker.defaultRegistryPath` | Path within registry to push to. | `""`
 | `docker.explorerRefreshInterval` | Explorer refresh interval, default is 1000ms. | `1000`
 | `docker.imageBuildContextPath` | Build context PATH to pass to Docker build command. | `""`
--- a/commands/push-image.ts
+++ b/commands/push-image.ts
@@ -42,7 +42,7 @@ export async function pushImage(context?: ImageNode): Promise<void> {
         }
         if (userPrefixPreference === DialogResponses.yes) {
             await configOptions.update('defaultRegistryPath', prefix, vscode.ConfigurationTarget.Workspace);
-            await vscode.window.showInformationMessage('You can change this value at any time via the docker.defaultRegistryPath setting.');
+            vscode.window.showInformationMessage('Default registry path saved. You can change this value at any time via the docker.defaultRegistryPath setting.');
         }
     }
     if (imageToPush) {
--- a/commands/push-image.ts
+++ b/commands/push-image.ts
@@ -14,7 +14,7 @@ const teleAzureId: string = 'vscode-docker.image.push.azureContainerRegistry';
 
 export async function pushImage(context?: ImageNode): Promise<void> {
     const configOptions: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('docker');
-    let askToSavePrefix = configOptions.get('askToSavePrefix', undefined);
+    let askToSaveRegistryPath: boolean = configOptions.get<boolean>('askToSaveRegistryPath');
 
     let prefix = "";
     let imageToPush: Docker.ImageDesc;
@@ -34,15 +34,16 @@ export async function pushImage(context?: ImageNode): Promise<void> {
     if (imageName.includes('/')) {
         prefix = imageName.substring(0, imageName.lastIndexOf('/'));
     }
-    if (prefix && askToSavePrefix !== false) { //account for undefined
-        let userPrefixPreference: vscode.MessageItem = await ext.ui.showWarningMessage("Would you like to save the prefix for autocomplete later?", DialogResponses.yes, DialogResponses.no, DialogResponses.skipForNow);
+    if (prefix && askToSaveRegistryPath !== false) { //account for undefined
+        let userPrefixPreference: vscode.MessageItem = await ext.ui.showWarningMessage(`Would you like to save '${prefix}' as your default registry path?`, DialogResponses.yes, DialogResponses.no, DialogResponses.skipForNow);
         if (userPrefixPreference === DialogResponses.yes || userPrefixPreference === DialogResponses.no) {
-            askToSavePrefix = false;
+            askToSaveRegistryPath = false;
+            await configOptions.update('askToSaveRegistryPath', false, vscode.ConfigurationTarget.Workspace);
         }
         if (userPrefixPreference === DialogResponses.yes) {
             await configOptions.update('defaultRegistryPath', prefix, vscode.ConfigurationTarget.Workspace);
+            await vscode.window.showInformationMessage('You can change this value at any time via the docker.defaultRegistryPath setting.');
         }
-        await configOptions.update('askToSavePrefix', askToSavePrefix, vscode.ConfigurationTarget.Workspace);
     }
     if (imageToPush) {
         const terminal = ext.terminalProvider.createTerminal(imageName);
--- a/commands/push-image.ts
+++ b/commands/push-image.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import vscode = require('vscode');
+import { DialogResponses } from 'vscode-azureextensionui';
 import { ImageNode } from '../explorer/models/imageNode';
 import { ext } from '../extensionVariables';
 import { reporter } from '../telemetry/telemetry';
@@ -12,6 +13,10 @@ const teleCmdId: string = 'vscode-docker.image.push';
 const teleAzureId: string = 'vscode-docker.image.push.azureContainerRegistry';
 
 export async function pushImage(context?: ImageNode): Promise<void> {
+    const configOptions: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('docker');
+    let askToSavePrefix = configOptions.get('askToSavePrefix', undefined);
+
+    let prefix = "";
     let imageToPush: Docker.ImageDesc;
     let imageName: string = "";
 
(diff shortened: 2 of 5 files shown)


dockerExtension.ts
```diff
@@ -166,13 +152,40 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+}
+
+export function deactivate(): Thenable<void> {
     if (!client) {
         return undefined;
     }
     // perform cleanup
     Configuration.dispose();
-    return await client.stop();
+    return client.stop();
```
StephenWeatherford: Why this change?

dockerExtension.ts
```diff
@@ -166,13 +152,40 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+}
+
+function registerComposeCommands(): void {
+    registerCommand('vscode-docker.compose.up', composeUp);
+    registerCommand('vscode-docker.compose.down', composeDown);
+    registerCommand('vscode-docker.compose.restart', composeRestart);
+    registerCommand('vscode-docker.system.prune', systemPrune);
+}
+
+export function deactivate(): Thenable<void> {
```
StephenWeatherford: Why this change?

dockerExtension.ts
```diff
@@ -166,13 +152,40 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
     activateLanguageClient(ctx);
 }
 
-export async function deactivate(): Promise<void> {
+function registerContainerCommands(): void {
```
StephenWeatherford: Personally I'd rather not break them up into separate categories - it just becomes a nuisance.  And what if you add new categories, or a command doesn't fit neatly into one or the other?  I regret doing that in the storage extension.

What I would like to see:
1) Move vscode-docker.createWebApp's code into a function
2) move all of the registerCommands into a single separate function

dockerExtension.ts
PrashanthCorp: The original function isn't async. I got an error regarding this. I'll admit this felt weird.

dockerExtension.ts
PrashanthCorp: Sounds good. I see where that comes from. 

Thanks. Will get on it.

dockerExtension.ts
StephenWeatherford: You mean deactivate?  This is what I see:

export async function deactivate(): Promise<void> {
    if (!client) {
        return undefined;
    }
    // perform cleanup
    Configuration.dispose();
    return await client.stop();
}

package.json
```diff
@@ -242,12 +242,12 @@
         "docker.defaultRegistry": {
           "type": "string",
           "default": "",
-          "description": "Default registry when tagging an image, empty string will target Docker Hub when pushing."
+          "description": "(Deprecated) Default registry when tagging an image, empty string will target Docker Hub when pushing."
```
StephenWeatherford: I'd say we should go ahead and remove it so they get errors in the settings file.  You can still query for the value even if it's removed.

package.json
```diff
@@ -242,12 +242,12 @@
...
         },
         "docker.defaultRegistryPath": {
           "type": "string",
           "default": "",
-          "description": "Path within registry to push to."
+          "description": "Default registry and path when tagging an image, empty string will target Docker Hub when pushing"
```
StephenWeatherford: empty string part no longer applies

README.md
```diff
@@ -92,7 +92,7 @@ The Docker extension comes with a number of useful configuration settings allowi
 | `docker.attachShellCommand.windowsContainer` | Attach command to use for Windows containers | `powershell`
 | `docker.dockerComposeBuild` | Run docker-compose with the --build argument, defaults to true | `true`
 | `docker.dockerComposeDetached` | Run docker-compose with the --d (detached) argument, defaults to true | `true`
-| `docker.defaultRegistry` | Default registry when tagging an image, empty string will target Docker Hub when pushing. | `""`
+| `docker.defaultRegistry` | (Deprecated) Default registry when tagging an image, empty string will target Docker Hub when pushing. | `""`
```
StephenWeatherford: ditto (remove)

package.json
PrashanthCorp: @fiveisprime your thoughts?

PrashanthCorp: I'll go ahead and remove this...

commands/tag-image.ts
```diff
@@ -37,11 +37,6 @@ export async function tagImage(context?: ImageNode): Promise<void> {
             imageName = defaultRegistryPath + '/' + imageName;
         }
 
-        const defaultRegistry = configOptions.get('defaultRegistry', '');
```
StephenWeatherford: Weren't you going to read the old setting, combine with the new one, and resave to the new one?

package.json
PrashanthCorp: The docker command still assumes no prefix as pushing to Dockerhub. 
image

Does it make sense to expose Docker's assumption as part of this statement?

commands/tag-image.ts
PrashanthCorp: Combining happens as part of `consolidateDefaultRegistrySettings`, which I intend to run on extension activation. 

Note: There is a change in that function -having the config setting update be async- that hasn't been pushed. I'm pushing another commit that fixes that, but also implements part 1 of the bug filed (ask to save setting on first push).

package.json
```diff
@@ -239,15 +239,15 @@
...
-          "default": "",
-          "description": "Default registry when tagging an image, empty string will target Docker Hub when pushing."
-        },
         "docker.defaultRegistryPath": {
           "type": "string",
           "default": "",
-          "description": "Path within registry to push to."
+          "description": "Default registry and path when tagging an image, empty string will target Docker Hub when pushing"
+        },
+        "docker.askToSavePrefix": {
```
StephenWeatherford: Why adding another concept of prefix here?  We already have the concept of defaultRegistryPath, so how about askToSaveRegistryPath?

commands/push-image.ts
```diff
@@ -12,6 +13,10 @@ const teleCmdId: string = 'vscode-docker.image.push';
 const teleAzureId: string = 'vscode-docker.image.push.azureContainerRegistry';
 
 export async function pushImage(context?: ImageNode): Promise<void> {
+    const configOptions: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('docker');
+    let askToSavePrefix = configOptions.get('askToSavePrefix', undefined);
```
StephenWeatherford: configOptions.get<boolean>(...

StephenWeatherford: default should be true not undefined

commands/push-image.ts
```diff
@@ -26,6 +31,19 @@ export async function pushImage(context?: ImageNode): Promise<void> {
         }
     }
 
+    if (imageName.includes('/')) {
+        prefix = imageName.substring(0, imageName.lastIndexOf('/'));
+    }
+    if (prefix && askToSavePrefix !== false) { //account for undefined
+        let userPrefixPreference: vscode.MessageItem = await ext.ui.showWarningMessage("Would you like to save the prefix for autocomplete later?", DialogResponses.yes, DialogResponses.no, DialogResponses.skipForNow);
```
StephenWeatherford: `Would you like to save '${prefix}' as your default registry path?`

commands/push-image.ts
```diff
@@ -26,6 +31,19 @@ export async function pushImage(context?: ImageNode): Promise<void> {
...
+        if (userPrefixPreference === DialogResponses.yes || userPrefixPreference === DialogResponses.no) {
+            askToSavePrefix = false;
+        }
+        if (userPrefixPreference === DialogResponses.yes) {
+            await configOptions.update('defaultRegistryPath', prefix, vscode.ConfigurationTarget.Workspace);
+        }
+        await configOptions.update('askToSavePrefix', askToSavePrefix, vscode.ConfigurationTarget.Workspace);
```
StephenWeatherford: nit: I'd just do 

await configOptions.update('askToSavePrefix', false, vscode.ConfigurationTarget.Workspace); 

instead of 'askToSavePrefix = false' above, rather than updating it even when you're not actually changing its value.

commands/push-image.ts
```diff
@@ -26,6 +31,19 @@ export async function pushImage(context?: ImageNode): Promise<void> {
...
+    if (imageName.includes('/')) {
+        prefix = imageName.substring(0, imageName.lastIndexOf('/'));
+    }
+    if (prefix && askToSavePrefix !== false) { //account for undefined
+        let userPrefixPreference: vscode.MessageItem = await ext.ui.showWarningMessage("Would you like to save the prefix for autocomplete later?", DialogResponses.yes, DialogResponses.no, DialogResponses.skipForNow);
+        if (userPrefixPreference === DialogResponses.yes || userPrefixPreference === DialogResponses.no) {
+            askToSavePrefix = false;
+        }
+        if (userPrefixPreference === DialogResponses.yes) {
+            await configOptions.update('defaultRegistryPath', prefix, vscode.ConfigurationTarget.Workspace);
```
StephenWeatherford: After updating:

  showInformationMessage('You can change this value at any time via the docker.defaultRegistryPath setting.');

dockerExtension.ts
```diff
@@ -157,13 +168,23 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
-    activateLanguageClient(ctx);
+    if (defaultRegistry) {
+        if (defaultRegistryPath) { // combine both the settings, then notify the user
+            await configOptions.update('defaultRegistryPath', `${defaultRegistry}/${defaultRegistryPath}`, vscode.ConfigurationTarget.Workspace);
+        }
+        if (!defaultRegistryPath) {// assign defaultRegistry to defaultRegistryPath
+            await configOptions.update('defaultRegistryPath', `${defaultRegistry}`, vscode.ConfigurationTarget.Workspace);
+        }
+        await configOptions.update('defaultRegistry', '', false);
+        vscode.window.showInformationMessage("Moving forward, we shall deprecate the use of the defaultRegistry setting, combining it with defaultRegistryPath.")
```
StephenWeatherford: This implies you're *going* to do this, but you just did it.

`The 'docker.defaultRegistry' setting is now obsolete, and you should just use the 'docker.defaultRegistryPath' setting. Your settings have been updated to reflect this change.`

dockerExtension.ts
```diff
@@ -157,13 +168,23 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
 
-    activateLanguageClient(ctx);
+    if (defaultRegistry) {
+        if (defaultRegistryPath) { // combine both the settings, then notify the user
+            await configOptions.update('defaultRegistryPath', `${defaultRegistry}/${defaultRegistryPath}`, vscode.ConfigurationTarget.Workspace);
+        }
+        if (!defaultRegistryPath) {// assign defaultRegistry to defaultRegistryPath
+            await configOptions.update('defaultRegistryPath', `${defaultRegistry}`, vscode.ConfigurationTarget.Workspace);
+        }
+        await configOptions.update('defaultRegistry', '', false);
```
StephenWeatherford: I *think* you should leave off the false parameter.

package.json
StephenWeatherford: Okay, I guess that makes sense.

commands/push-image.ts
```diff
@@ -34,15 +34,16 @@ export async function pushImage(context?: ImageNode): Promise<void> {
...
+    if (prefix && askToSaveRegistryPath !== false) { //account for undefined
+        let userPrefixPreference: vscode.MessageItem = await ext.ui.showWarningMessage(`Would you like to save '${prefix}' as your default registry path?`, DialogResponses.yes, DialogResponses.no, DialogResponses.skipForNow);
         if (userPrefixPreference === DialogResponses.yes || userPrefixPreference === DialogResponses.no) {
-            askToSavePrefix = false;
+            askToSaveRegistryPath = false;
+            await configOptions.update('askToSaveRegistryPath', false, vscode.ConfigurationTarget.Workspace);
         }
         if (userPrefixPreference === DialogResponses.yes) {
             await configOptions.update('defaultRegistryPath', prefix, vscode.ConfigurationTarget.Workspace);
+            await vscode.window.showInformationMessage('You can change this value at any time via the docker.defaultRegistryPath setting.');
```
StephenWeatherford: Don't await... It will block the user from continuing.

Let's change message to:
'Default registry path saved. You can change this value at any time via the docker.defaultRegistryPath setting.'

commands/push-image.ts
StephenWeatherford: And it's also high priority to avoid writing when not needed now that we know updating a setting could write it into a different place than it currently is.

dockerExtension.ts
StephenWeatherford: Workspace is good for now.

PR #405, opened 2018-08-24
Fix image refresh

### LINKED ISSUE #371: Images list does not refresh after tagging an image
Repros in 0.1.0

StephenWeatherford:
Fixes #371 

The old code only compared images, but we also need to compare tags, and might as well compare everything for better reliability.

### DIFF
--- a/explorer/models/rootNode.ts
+++ b/explorer/models/rootNode.ts
@@ -26,7 +26,7 @@ const containerFilters = {
 };
 
 export class RootNode extends NodeBase {
-    private _imageCache: Docker.ImageDesc[];
+    private _sortedImageCache: Docker.ImageDesc[];
     private _imageDebounceTimer: NodeJS.Timer;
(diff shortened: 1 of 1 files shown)


explorer/models/rootNode.ts
```diff
@@ -67,34 +67,24 @@ export class RootNode extends NodeBase {
...
-                            needToRefresh = true;
-                            break
-                        }
-                    }
-                }
-
-                if (needToRefresh) {
+                let imagesAsJson = JSON.stringify(images);
+                let cacheAsJson = JSON.stringify(this._imageCache);
+                if (imagesAsJson !== cacheAsJson) {
```
PrashanthCorp: Would you also need to sort `this._imageCache` ?

StephenWeatherford: It comes directly from images, which is being sorted here.  Changed it to _sortedImageCache to make that clear.

PR #415, opened 2018-08-27
Add menu item to set registry as default

PrashanthCorp:
Addresses part 4 of #334


package.json
```diff
@@ -592,25 +609,34 @@
         "title": "Connect to a private registry...",
         "category": "Docker"
       },
+      {
+        "command": "vscode-docker.setRegistryAsDefault",
+        "title": "Set registry as default registry path",
```
StephenWeatherford: "Set as default registry path"

commands/registrySettings.ts
```diff
@@ -0,0 +1,36 @@
...
+import { CustomRegistryNode } from "../explorer/models/customRegistryNodes";
+import { DockerHubOrgNode } from "../explorer/models/dockerHubNodes";
+
+export async function setRegistryAsDefault(node: CustomRegistryNode | AzureRegistryNode | DockerHubOrgNode): Promise<void> {
+    let registryName: string;
+    if (node instanceof DockerHubOrgNode) {
+        registryName = node.namespace;
+    } else if (node instanceof AzureRegistryNode) {
+        registryName = node.registry.loginServer || node.label;
+    } else { //CustomREgistryNode
```
StephenWeatherford: Do an actual if here, and assert on the else.  Otherwise you won't get notified if a new case gets added to package.json but not to here.

Same for switch's - you should always assert on default.

commands/registrySettings.ts
```diff
@@ -0,0 +1,36 @@
...
+        registryName = node.registryName;
+    }
+
+    const configOptions: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('docker');
```
StephenWeatherford: constant

commands/registrySettings.ts
```diff
@@ -0,0 +1,36 @@
...
+    await configOptions.update('defaultRegistryPath', registryName, vscode.ConfigurationTarget.Workspace);
```
StephenWeatherford: constant (local to file is fine)

commands/registrySettings.ts
```diff
@@ -0,0 +1,36 @@
...
+    vscode.window.showInformationMessage(`Update defaultRegistryPath to ${registryName}`);
```
StephenWeatherford: Updated

commands/registrySettings.ts
```diff
@@ -0,0 +1,36 @@
...
+    let defaultRegistry = configOptions.get('defaultRegistry', '');
+
+    if (defaultRegistry) {
+        if (defaultRegistryPath) { // combine both the settings, then notify the user
+            await configOptions.update('defaultRegistryPath', `${defaultRegistry}/${defaultRegistryPath}`, vscode.ConfigurationTarget.Workspace);
+        }
+        if (!defaultRegistryPath) {// assign defaultRegistry to defaultRegistryPath
+            await configOptions.update('defaultRegistryPath', `${defaultRegistry}`, vscode.ConfigurationTarget.Workspace);
+        }
+        await configOptions.update('defaultRegistry', undefined, vscode.ConfigurationTarget.Workspace);
```
StephenWeatherford: No longer works since deleted...  Need to add flag instead.

commands/registrySettings.ts
```diff
@@ -0,0 +1,36 @@
...
+    vscode.window.showInformationMessage(`Update defaultRegistryPath to ${registryName}`);
+}
+
+export async function consolidateDefaultRegistrySettings(): Promise<void> {
+    const configOptions: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('docker');
+    let defaultRegistryPath = configOptions.get('defaultRegistryPath', '');
+    let defaultRegistry = configOptions.get('defaultRegistry', '');
+
+    if (defaultRegistry) {
+        if (defaultRegistryPath) { // combine both the settings, then notify the user
```
StephenWeatherford: Use String.join instead of having an if/else.

Remember, duplicate code (or near-duplicate) is bad.

commands/registrySettings.ts
```diff
@@ -1,36 +1,40 @@
...
-        if (defaultRegistryPath) { // combine both the settings, then notify the user
-            await configOptions.update('defaultRegistryPath', `${defaultRegistry}/${defaultRegistryPath}`, vscode.ConfigurationTarget.Workspace);
-        }
-        if (!defaultRegistryPath) {// assign defaultRegistry to defaultRegistryPath
-            await configOptions.update('defaultRegistryPath', `${defaultRegistry}`, vscode.ConfigurationTarget.Workspace);
-        }
-        await configOptions.update('defaultRegistry', undefined, vscode.ConfigurationTarget.Workspace);
+    if (defaultRegistry && combineRegistryPaths) {
+        let updatedPath = defaultRegistryPath ? `${defaultRegistry}/${defaultRegistryPath}` : `${defaultRegistry}`;
+        await configOptions.update('defaultRegistryPath', updatedPath, vscode.ConfigurationTarget.Global);
```
StephenWeatherford: Not using the constants.  BTW, my own guidelines for when to use a constant are essentially: 1) if used multiple times, 2) if a constant would help document it.

commands/registrySettings.ts
```diff
@@ -1,36 +1,40 @@
...
+        await configOptions.update('combineRegistryPathsFlag', false, vscode.ConfigurationTarget.Global);
```
StephenWeatherford: use constant

StephenWeatherford: Move updating the flag to above combining the paths (first line after "then").  Otherwise if for some reason there's an error before updating the flag, the user will get asked every single time (not likely here, but that's what I generally do in scenarios like this).

PrashanthCorp: You raise a good point. Thanks for the catch! 

Did you mean first lint after `let`?

commands/registrySettings.ts
```diff
@@ -1,36 +1,40 @@
...
+    vscode.window.showInformationMessage(`Updated defaultRegistryPath to ${registryName}`);
 }
 
 export async function consolidateDefaultRegistrySettings(): Promise<void> {
     const configOptions: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('docker');
-    let defaultRegistryPath = configOptions.get('defaultRegistryPath', '');
-    let defaultRegistry = configOptions.get('defaultRegistry', '');
+    const combineRegistryPaths: boolean = configOptions.get(combineRegistryPathsKey);
+    let defaultRegistryPath: string = configOptions.get(defaultRegistryPathKey, '');
+    let defaultRegistry: string = configOptions.get(defaultRegistryKey, '');
```
StephenWeatherford: I don't think combineRegistryPathsFlag should be exposed to the user - I'd use ext.context.workspaceState for it.

commands/registrySettings.ts
```diff
@@ -1,36 +1,40 @@
...
     let registryName: string;
     if (node instanceof DockerHubOrgNode) {
         registryName = node.namespace;
     } else if (node instanceof AzureRegistryNode) {
         registryName = node.registry.loginServer || node.label;
-    } else { //CustomREgistryNode
+    } else if (node instanceof CustomRegistryNode) {
         registryName = node.registryName;
+    } else {
+        assert.equal(false, true, "Type of registry node encountered not handled");
```
StephenWeatherford: assert.fail

commands/registrySettings.ts
StephenWeatherford: Then you can remove it from package.json.

commands/registrySettings.ts
```diff
@@ -1,36 +1,40 @@
+import * as assert from 'assert';
 import * as vscode from 'vscode';
 import { AzureRegistryNode } from "../explorer/models/azureRegistryNodes";
 import { CustomRegistryNode } from "../explorer/models/customRegistryNodes";
 import { DockerHubOrgNode } from "../explorer/models/dockerHubNodes";
 
+const defaultRegistryPathKey = "defaultRegistryPath";
+const defaultRegistryKey = "defaultRegistry";
+const combineRegistryPathsKey = "combineRegistryPathsFlag"
```
StephenWeatherford: This name means absolutely nothing.  How about 'hasCheckedRegistryPaths' (and reverse its boolean meaning)?

Issue #382, opened 2018-08-16
Add "Pull" menu option for repositories and tags in a container registry


StephenWeatherford:

estebanreyl:
We have one for pull at least for the tags that we can PR soon, is there such a thing as a local repository?

StephenWeatherford:
Not sure what you mean.

estebanreyl:
Wait, I see how it works, sorry I wasn't quite certain how you'd pull a repository.

PR #417, opened 2018-08-28
AzureCR/Pull Image from Azure

rsamai:
-Added pull-from-azure command that allows user to pull image(s) from Azure to their local images
-Right click on an Image Node and select "Pull image from Azure" or invoke through command palette and select your registry, repository, and image
-Right click on Repository Node and select "Pull image from Azure" to pull all images in repository
-Sends docker login and docker pull to terminal using loginCredentials and works even if user doesn't have Azure CLI 
-Accounts for error saving credentials. Guides user to remove wincred from their Docker config file 
-Addresses Issue #382


commands/azureCommands/pull-from-azure.ts
```diff
@@ -0,0 +1,56 @@
...
+export async function pullFromAzure(context?: AzureImageTagNode | AzureRepositoryNode): Promise<any> {
+    let registryName: string;
+    let registry: Registry;
+    let imageName: string;
+
+    if (context) { // Right Click
+        registryName = context.registry.loginServer;
+        registry = context.registry;
+
+        if (context.contextValue === "azureImageTagNode") { // Right Click on AzureImageNode
```
StephenWeatherford: if (context instanceof AzureImageTagNode) {
…
} else if (context instanceof AzureRepositoryNode) {
…
} else {
assert.fail(`Unexpected node type: ${context.contextValue}`);
}

commands/azureCommands/pull-from-azure.ts
```diff
@@ -0,0 +1,56 @@
...
+            imageName = context.label;
```
StephenWeatherford: Please don't use context.label to get the image name.  That's a problem with the original codebase (it's assuming the label will be the same as the programmatic image name) that I've been slowly correcting.  Please make the following changes/additions to AzureImageTagNode and then use the ImageNameWithTag property:

export class AzureImageTagNode extends NodeBase {
    constructor(
...
    ) {
CHANGED:
        super(AzureImageTagNode.getImageNameWithTag(repositoryName, tag));
    }

ADDED:
    public static getImageNameWithTag(repositoryName: string, tag: string): string {
        return `${repositoryName}:${tag}`
    }

    public get ImageNameWithTag(): string {
        return AzureImageTagNode.getImageNameWithTag(this.repositoryName, this.tag);
    }

commands/azureCommands/pull-from-azure.ts
```diff
@@ -0,0 +1,56 @@
...
+    } else { // Command Palette
+        registry = await quickPicks.quickPickACRRegistry();
+        registryName = registry.loginServer;
+        const repository: Repository = await quickPicks.quickPickACRRepository(registry, 'Select the repository of the image you want to pull');
+        const image: AzureImage = await quickPicks.quickPickACRImage(repository, 'Select the image you want to pull');
+        imageName = `${repository.name}:${image.tag}`;
+    }
+
+    // Using loginCredentials function to get the username and password. This takes care of all users, even if they don't have the Azure CLI
+    const credentials = await acrTools.loginCredentials(registry);
```
StephenWeatherford: rename to getLoginCredentials

commands/azureCommands/pull-from-azure.ts
```diff
@@ -0,0 +1,56 @@
...
+    const username = credentials.username;
```
StephenWeatherford: Why is the username "00000000-0000-0000-0000-000000000000" instead of something unique?

commands/azureCommands/pull-from-azure.ts
```diff
@@ -0,0 +1,56 @@
...
+    // Send commands to terminal
+    const terminal = ext.terminalProvider.createTerminal("Docker");
+    terminal.show();
+    let cont = (err, stdout, stderr) => {
+        ext.outputChannel.append(stdout);
+        ext.outputChannel.append(stderr);
+        terminal.sendText(`docker pull ${registryName}/${imageName}`);
+    }
+
+    exec(`docker login ${registryName} -u ${username} -p ${password}`, cont);
```
StephenWeatherford: I'd prefer to see this inline:

   exec(`docker login ${registryName} -u ${username} -p ${password}`, (err, stdout, stderr) => {
        ext.outputChannel.append(stdout);
        ext.outputChannel.append(stderr);

        terminal.sendText(`docker pull ${registryName}/${imageName}`);
    });

dockerExtension.ts
```diff
@@ -124,6 +125,10 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 
     ctx.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('docker', new DockerDebugConfigProvider()));
 
+    if (azureAccount) {
```
StephenWeatherford: Remove if
Remove setAccount
Move registerCommand to where the other ACR commands are, and use register***Azure***Command

package.json
```diff
@@ -608,6 +613,11 @@
         "command": "vscode-docker.disconnectCustomRegistry",
         "title": "Disconnect from private registry",
         "category": "Docker"
+      },
+      {
+        "command": "vscode-docker.pullFromAzure",
```
StephenWeatherford: vscode-docker.pull-ACR-Image to be consistent with others

commands/azureCommands/pull-from-azure.ts
northtyphoon: Empty GUID is a special user name to indicate the login credential is based on JWT token.

commands/azureCommands/pull-from-azure.ts
rsamai: I can't print context.contextValue in the else statement. tslint says "Property does not exist on type 'never'." It does a flow check and won't hit this statement- in theory due to the parameters of the function. I did  "Unexpected node type: ${context}"

commands/azureCommands/pull-from-azure.ts
StephenWeatherford: Could you add a comment to that effect?  Thx.

commands/azureCommands/pull-from-azure.ts
```diff
@@ -22,11 +25,13 @@ export async function pullFromAzure(context?: AzureImageTagNode | AzureRepositor
...
 
-        if (context.contextValue === "azureImageTagNode") { // Right Click on AzureImageNode
+        if (context instanceof AzureImageTagNode) { // Right Click on AzureImageNode
             imageName = context.label;
-        } else { // Right Click on AzureRepositoryNode
+        } else if (context instanceof AzureRepositoryNode) { // Right Click on AzureRepositoryNode
             console.log(context.repositoryName);
             imageName = `${context.label} -a`; // Pull all images in repository
+        } else {
+            assert.fail(`Unexpected node type: ${context}`);
```
StephenWeatherford: That converts to this string: "Unexpected node type: [object Object]".  Use ${context.contextValue} to give us something useful.

commands/azureCommands/pull-from-azure.ts
```diff
@@ -38,19 +43,46 @@ export async function pullFromAzure(context?: AzureImageTagNode | AzureRepositor
...
     }
+    terminal.sendText(`docker pull ${registryName}/${imageName}`);
+}
 
-    exec(`docker login ${registryName} -u ${username} -p ${password}`, cont);
+function isLoggedIntoDocker(registry: Registry): boolean {
+    let home = process.env.HOMEPATH;
+    let configPath: string = path.join(home, '.docker', 'config.json');
+    let buffer: Buffer;
+    try {
```
StephenWeatherford: Get rid of the try/catch and do it this way, which will give us telemetry if (and only if) it fails (otherwise we won't know if anything wrong is happening on users' machines):
```
        await callWithTelemetryAndErrorHandling('findDockerConfig', async function (this: IActionContext): Promise<void> {
            this.suppressTelemetry = true;

            buffer = fs.readFileSync(configPath); 
});
```

commands/azureCommands/pull-from-azure.ts
```diff
@@ -38,19 +43,46 @@ export async function pullFromAzure(context?: AzureImageTagNode | AzureRepositor
...
+        buffer = fs.readFileSync(configPath);
+    } catch (err) {
+        return false; // If config.json is not found and not in the default location
+    }
 
+    let index = buffer.indexOf(registry.loginServer);
+    console.log("index:");
```
StephenWeatherford: remove console.log()'s

commands/azureCommands/pull-from-azure.ts
```diff
@@ -38,19 +43,46 @@ export async function pullFromAzure(context?: AzureImageTagNode | AzureRepositor
...
+    console.log(index);
+    return index !== -1; // Returns -1 if user is not logged into docker
```
StephenWeatherford: nit: I prefer "return index >= 0;"

commands/azureCommands/pull-from-azure.ts
rsamai: Tslint says "Property not found on type never". It shouldn't get to this else statement since the parameters are either AzureImageTagNode or AzureRepositoryNode. I'll just remove the variable all together

estebanreyl:
@StephenWeatherford  Are we missing anything here? Also what was the final call on following the CLI strategy?

StephenWeatherford:
Sorry, go some other deliverables before I can get back to this.

northtyphoon:
@rosanch does your PR #506 include this change? If yes, we can close it and focus on #506

rosanch:
I don't have this changes, in my current branch. But there is already a pull from azure command. I don't know if this wasn't committed to the dev branch or if they changed it and did a different commit.

However this looks like a more robust function, than the one that it's currently merged. So I wouldn't skip this PR.

StephenWeatherford:
Nice job, @rsamai, thanks!

Made just one change: Request permission before logging in to docker CLI,.

PR #432, opened 2018-08-31
build image and push image: prompt for using or update defaultRegPath based on tag for container.

PrashanthCorp:
Addresses #334 . 
Specifically: 
- Push : If they try to push an image with no prefix (username/whatever), we ask if they wanted to provide a prefix - need to figure out wording/etc
- Tag: Add prefix for Tag, and for Build if flag set. Don't apply to push (push might send user to Tag). 
Both build and Tag now use the same code.  
-  Build: When we notice that the user is building an image with '/' in the path, we prompt the user asking whether to save the registryPath. 
Fetch the user-input prefix, verify that it is not the same as the one in defaultRegistryPath, then prompt


commands/build-image.ts
```diff
@@ -100,15 +102,12 @@ export async function buildImage(actionContext: IActionContext, dockerFileUri?:
...
-        value: imageName + ':latest'
-    };
-
-    const value: string = await ext.ui.showInputBox(opt);
-
+    const imageWithTag: string = await getTagFromUserInput(imageName + ":latest");
+    const defaultPath = configOptions.get('defaultRegistryPath', '');
+    if (!imageWithTag.includes(defaultPath)) { //user has entered a prefix different from stored
+        await askToSavePrefix(imageWithTag);
+    }
```
PrashanthCorp: Needs to be fixed in a separate PR.

commands/push-image.ts
```diff
@@ -9,24 +9,15 @@ import { ImageNode } from '../explorer/models/imageNode';
...
-        imageToPush = context.imageDesc;
-        imageName = context.label;
-    } else {
-        const selectedItem: ImageItem = await quickPickImage();
-        if (selectedItem) {
-            imageToPush = selectedItem.imageDesc;
-            imageName = selectedItem.label;
-        }
-    }
+    let descriptor = await contextToImageDescriptor(context);
```
StephenWeatherford: let [imageToPush, imageName] = await contextToImageDescriptor(context);

commands/push-image.ts
```diff
@@ -35,13 +26,15 @@ export async function pushImage(context?: ImageNode): Promise<void> {
         let askToPushPrefix: boolean = ext.context.workspaceState.get(addPrefixImagePush, true);
         let defaultRegistryPath = vscode.workspace.getConfiguration('docker').get('defaultRegistryPath');
         if (askToPushPrefix && defaultRegistryPath) {
-            let useDefaultRegistry: vscode.MessageItem = { title: "Use default registry path" };
-            let response: vscode.MessageItem = await ext.ui.showWarningMessage(`You are about to push the image to dockerhub. You may not have permissions to push this image. Continue pushing to dockerhub?`, DialogResponses.yes, useDefaultRegistry, DialogResponses.dontWarnAgain);
-            if (response === DialogResponses.dontWarnAgain) {
+            let alwaysPush: vscode.MessageItem = { title: "Always Push" };
+            let tagFirst: vscode.MessageItem = { title: "Tag first" };
+            let options: vscode.MessageItem[] = [DialogResponses.yes, alwaysPush, tagFirst];
+            let response: vscode.MessageItem = await ext.ui.showWarningMessage(`You are about to push the image to dockerhub. You may not have permissions to push this image. Continue pushing to dockerhub?`, ...options);
```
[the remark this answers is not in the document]
PrashanthCorp: Heads up, it looks big: 
image

PR #437, opened 2018-09-01
Docker-build: Prompt with last used image tag for each DockerFile.

PrashanthCorp:


commands/build-image.ts
```diff
@@ -95,20 +93,28 @@ export async function buildImage(actionContext: IActionContext, dockerFileUri?:
...
     }
+
+    const imageWithTag: string = await getTagFromUserInput(imageName);
+    await ext.context.globalState.update(dockerFileKey, imageWithTag);
     const terminal: vscode.Terminal = ext.terminalProvider.createTerminal('Docker');
     terminal.sendText(`docker build --rm -f "${dockerFileItem.relativeFilePath}" -t ${imageWithTag} ${contextPath}`);
     terminal.show();
 }
+
+function buildImageNameFromPath(filePath: string): string {
```
StephenWeatherford: This is not an image name that you're creating

commands/build-image.ts
```diff
@@ -95,20 +93,28 @@ export async function buildImage(actionContext: IActionContext, dockerFileUri?:
     if (defaultContextPath && defaultContextPath !== '') {
         contextPath = defaultContextPath;
     }
+    let absFilePath: string = path.join(vscode.workspace.workspaceFolders[0].uri.fsPath, dockerFileItem.relativeFolderPath, dockerFileItem.relativeFilePath);
```
StephenWeatherford: Paths are relative to rootFolder

StephenWeatherford: It's relativeFile*Path*, not relativeFile*Name* (which wouldn't make sense).  So I expect you don't need relativeFolderPath

commands/build-image.ts
```diff
@@ -95,20 +93,28 @@ export async function buildImage(actionContext: IActionContext, dockerFileUri?:
...
-
-    const imageWithTag: string = await getTagFromUserInput(imageName + ":latest");
-    const defaultPath = configOptions.get(configurationKeys.defaultRegistryPath, '');
-    if (!imageWithTag.includes(defaultPath)) { //user has entered a prefix different from stored
-        await askToSavePrefix(imageWithTag);
+    if (!imageName.match(/:[^\/]+$/)) {
+        imageName += ":latest"
     }
+
+    const imageWithTag: string = await getTagFromUserInput(imageName);
```
StephenWeatherford: We don't want to use getTagFromUserInput, because it's adding the default registry path (which we only want for the first build) and it's highlighting the registry path, which I don't think makes sense for build.

PR #475, opened 2018-09-11
Turn on strict null checking, part 1

StephenWeatherford:


utils/nonNull.ts
```diff
@@ -0,0 +1,48 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Registry } from 'azure-arm-containerregistry/lib/models';
+import { Subscription } from 'azure-arm-resource/lib/subscription/models';
+import { isNullOrUndefined } from 'util';
```
PrashanthCorp: Looks like this function could be housed here. Your thoughts?

commands/utils/docker-endpoint.ts
```diff
@@ -5,14 +5,19 @@
...
     Linux,
     Windows
 }
 
 class DockerClient {
-    private endPoint: Docker;
+    private _endPoint: Docker | undefined;
+
+    private get endPoint(): Docker {
+        return nonNullValue(this._endPoint, 'endPoint');
```
PrashanthCorp: This could throw an internal error, correct? Where is that being handled?

utils/nonNull.ts
StephenWeatherford: Given that this is a Node.js internal function, that would be pretty hard.  :-)

commands/utils/docker-endpoint.ts
StephenWeatherford: It's not.  Assuming it's inside an command handler, vscode will display any unhandled exceptions (and we'll send telementry).

tsconfig.json
```diff
@@ -7,7 +7,16 @@
...
-        "rootDir": "."
+        "rootDir": ".",
+        "alwaysStrict": true,
+        "noFallthroughCasesInSwitch": true,
+        "noImplicitAny": false, // TODO
+        "noImplicitReturns": false, // TODO
+        "noImplicitThis": true,
+        "noUnusedLocals": false, // TODO
+        "noUnusedParameters": false, // TODO
+        "strictNullChecks": false, // TODO
```
EricJizbaMSFT: It feels like there are more settings here than other repos. Are they helpful? Should we add them to other repos?

StephenWeatherford:
> I feel like we lose type safety with nonNullProp since you're accessing the property through a string value (which could easily have typos). I much prefer using nonNullValue in every case

Take a look at the function again.  I added comments to make it clear:
```
/**
 * Retrieves a property by name from an object and checks that it's not null and not undefined.  It is strongly typed
 * for the property and will give a compile error if it's not a property of the source.
 */
```
nonNullValue doesn't have that advantage, although it is strongly typed.

utils/nonNull.ts
StephenWeatherford: To be honest, I had exactly the same thought when reviewing my own code... :-)

PR #479, opened 2018-09-12
Work around vscode/58388 to get debugging tests working again

StephenWeatherford:


rcjsuen:
My opinion carries zero weight but I do not agree with this.

I suggest including a comment that both references the bug and also provides an example of a hardcoded path. However, the JSON content should remain the same.

StephenWeatherford:
Fair enough

Issue #391, opened 2018-08-22
Is Empty PORT EXPOSE By Design


gangachris:
I was generating a Dockerfile with the `> Dockerfile` command, and decided to leave the `PORT` section as an empty string, as I'm generating a base image that doesn't require exposing. This is what is generated 
<img width="457" alt="screen shot 2018-08-22 at 14 37 00" src="https://user-images.githubusercontent.com/3910877/44461281-e0752680-a618-11e8-8e3b-4a5f7a849fdc.png">

I know it's as easy as just deleting the last line, but I'm wondering if this is by design, or if someone doesn't specify a port, EXPOSE should not be generated.

StephenWeatherford:
Ideally we will not add the EXPOSE if port is empty.

StephenWeatherford:
Need to deal with compose files, too.

PrashanthCorp:
Moving forward we won't be generating dockerfiles with empty expose statements. It shouldn't be an issue moving forward.

PR #485, opened 2018-09-17
Support multiple versions of .net core

### LINKED ISSUE #465: Support Dockerfile for .net core 2.2 apps

StephenWeatherford:
Fixes #465 

Sync up with what Visual Studio does for .NET core apps (console and asp.net).  I've taken their templates verbatim to make it easier to sync up with changes.


configureWorkspace/configure.ts
```diff
@@ -415,8 +412,11 @@ async function configureCore(actionContext: IActionContext, options: ConfigureAp
 
         if (writeFile) {
             // Paths in the docker files should be relative to the Dockerfile (which is in the output folder)
-            fs.writeFileSync(filePath, generatorFunction(serviceNameAndPathRelativeToOutput, platformType, os, port, packageInfo), { encoding: 'utf8' });
-            filesWritten.push(fileName);
+            let fileContents = generatorFunction(serviceNameAndPathRelativeToOutput, platformType, os, port, packageInfo);
+            if (fileContents) {
+                fs.writeFileSync(filePath, fileContents, { encoding: 'utf8' });
+                filesWritten.push(fileName);
+            }
```
PrashanthCorp: This looks like a bug that would have affected other usecases too. Nice catch.

configureWorkspace/configure_dotnetcore.ts
```diff
@@ -0,0 +1,224 @@
...
+    // example: COPY Core2.0ConsoleAppWindows/Core2.0ConsoleAppWindows.csproj Core2.0ConsoleAppWindows/
+    let copyProjectCommands = `COPY ["${serviceNameAndRelativePath}.csproj", "${projectDirectory}/"]`
+    let exposeStatements = `EXPOSE ${port}`;
+
+    // Parse version from TargetFramework
+    // Example: netcoreapp1.0
+    const defaultNetCoreVersion = '2.1';
+    let [netCoreAppVersion] = extractRegExGroups(version, /^netcoreapp([0-9.]+)$/, [defaultNetCoreVersion]);
+
+    // semver requires a patch in the version, so add it if only major.minor
```
PrashanthCorp: Comment looks incomplete? did you mean `if version contains only major.minor`?

configureWorkspace/configure_dotnetcore.ts
```diff
@@ -0,0 +1,224 @@
...
+            break;
+        case "ASP.NET Core":
+            template = os === "Linux" ? aspNetCoreLinuxTemplate : aspNetCoreWindowsTemplate;
+            break;
+        default:
+            throw new Error(`Unexpected platform "${platform}"`);
+    }
+
+    let contents = template.replace('$base_image_name$', baseImageName)
+        .replace(/\$expose_statements\$/g, exposeStatements)
```
PrashanthCorp: It it not clear to me whether #391 would be handled. There is no conditional logic in inserting the `$expose_statements$`. It is present b default on both aspnetCore templates,  and absent by default on both dotnetCoreTemplates

commands/utils/docker-endpoint.ts
```diff
@@ -117,12 +118,20 @@ class DockerClient {
...
-                const ports = Object.keys(exposedPorts);
-                resolve(ports);
+    public async getExposedPorts(imageId: string): Promise<string[]> {
+        return await new Promise<string[]>((resolve, reject) => {
+            this.getImage(imageId).inspect((error: {}, data: { Config: { ExposedPorts: {} } }) => {
+                if (error) {
+                    reject(error);
+                } else {
+                    let exposedPorts = data.Config.ExposedPorts;
+                    if (!exposedPorts) {
```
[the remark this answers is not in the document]
StephenWeatherford: True.  Not worth changing now that it's been tested, though.

configureWorkspace/configure.ts
StephenWeatherford: You mean if (fileContents)?
This is actually just a change in behavior - moving special-case code out of the common code.

configureWorkspace/configure_dotnetcore.ts
StephenWeatherford: It's always present in the template now, but exposeStatements will be empty string if no port.  As you see, I handled this more thoroughly for other platforms in https://github.com/Microsoft/vscode-docker/pull/490 but don't plan to merge until after shipping.

StephenWeatherford: (by now, I mean with the PR that came after this)

PrashanthCorp: Agreed. Thanks for the fix there! I was just pointing out a bug we triaged.

PR #490, opened 2018-09-19
 Don't output EXPOSE if empty port specified

### LINKED ISSUE #391: Is Empty PORT EXPOSE By Design

StephenWeatherford:
Fixes #391 

For next milestone.

### DIFF
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -53,11 +53,18 @@ export type ConfigureTelemetryProperties = {
     packageFileSubfolderDepth?: string; // 0 = project/etc file in root folder, 1 = in subfolder, 2 = in subfolder of subfolder, etc.
 };
 
-const generatorsByPlatform = new Map<Platform, {
+export interface IPlatformGeneratorInfo {
     genDockerFile: GeneratorFunction,
     genDockerCompose: GeneratorFunction,
-    genDockerComposeDebug: GeneratorFunction
-}>();
+    genDockerComposeDebug: GeneratorFunction,
+    defaultPort: string
+}
+
+export function getExposeStatements(port: string): string {
+    return port ? `EXPOSE ${port}` : '';
+}
+
+const generatorsByPlatform = new Map<Platform, IPlatformGeneratorInfo>();
 generatorsByPlatform.set('ASP.NET Core', configureAspDotNetCore);
 generatorsByPlatform.set('Go', configureGo);
 generatorsByPlatform.set('Java', configureJava);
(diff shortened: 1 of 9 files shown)


configureWorkspace/configure.ts
```diff
@@ -53,11 +53,18 @@ export type ConfigureTelemetryProperties = {
...
+export interface IPlatformGeneratorInfo {
     genDockerFile: GeneratorFunction,
     genDockerCompose: GeneratorFunction,
-    genDockerComposeDebug: GeneratorFunction
-}>();
+    genDockerComposeDebug: GeneratorFunction,
+    defaultPort: string
+}
+
+export function getExposeStatements(port: string): string {
```
PrashanthCorp: getExposeStatement? I don't see the reasoning behind the plural name.

StephenWeatherford: VS has scenarios where there are more than one (HTTPS and HTTP ports).  I'm trying to keep similar to their logic.

configureWorkspace/configure_go.ts
```diff
@@ -3,15 +3,18 @@
...
+export let configureGo: IPlatformGeneratorInfo = {
   genDockerFile,
   genDockerCompose,
-  genDockerComposeDebug
+  genDockerComposeDebug,
+  defaultPort: '3000'
 };
 
 function genDockerFile(serviceNameAndRelativePath: string, platform: string, os: string | undefined, port: string, { cmd, author, version, artifactName }: Partial<PackageInfo>): string {
+  let exposeStatements = getExposeStatements(port);
```
EricJizbaMSFT: There seems to be an inconsistency where sometimes the `genDockerFile` has a 'port' parameter and sometimes it has an 'exposeStatements' parameter. Can we make that consistent?

StephenWeatherford: done

configureWorkspace/configure.ts
```diff
@@ -70,7 +77,14 @@ function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, o
     let generators = generatorsByPlatform.get(platform);
     assert(generators, `Could not find dockerfile generator functions for "${platform}"`);
     if (generators.genDockerFile) {
-        return generators.genDockerFile(serviceNameAndRelativePath, platform, os, port, { cmd, author, version, artifactName });
+        let contents = generators.genDockerFile(serviceNameAndRelativePath, platform, os, exposeStatements, { cmd, author, version, artifactName });
```
EricJizbaMSFT: Looks like you missed changing this 'exposeStatements' back to 'port'

PR #500, opened 2018-09-20
Add debugging support for .NET Core in Docker containers

philliphoff:
Adds support for debugging .NET Core applications within Docker containers, in a manner similar to that performed by Visual Studio's Docker-related tooling.

The basic feature set is:
 
 - Adds a palette command to scaffold a “Docker .NET Core” debug launch configuration.
   - The intent of this configuration is to minimize input/configuration from/by the user; properties will all be inferred
   - Optional properties can be used to override image tags, names, arguments, variables, etc.
 - Adds a debug configuration provider that:
   - Pulls down the latest .NET Core debugger from the Cloud, as needed
   - Generates a debuggable Docker image for the application (e.g. building just the first stage of the multi-stage Dockerfile)
   - Runs the Docker image in a container, applying shared volumes as necessary to expose debugger/application to the container
   - Dynamically generates a .NET Core debug launch configuration allowing VS Code to initiate and attach the debugger within the container
   - Launches the browser to the running application via the container’s mapped ports
   - Cleans up containers when debugging is done
 
> NOTE: Only debugging of *Linux* containers is supported today (in VS Code running on either Mac or Windows).  We cannot debug *Windows* containers because the .NET Core debugger (`vsdbg`) is not supported on Windows containers, and VS Code cannot use `msvsmon` (used by Visual Studio in that scenario).

The feature is disabled by default, using the new `docker.debugging.enabled` setting.


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.3.1 - xx September 2018
```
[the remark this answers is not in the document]
philliphoff: I was just following what I saw in an earlier version of the file for an as-yet-unreleased feature.  Changed to `0.x.x - Unreleased` if that's better.

dockerExtension.ts
```diff
@@ -281,6 +297,90 @@ namespace Configuration {
     }
 }
 
+function registerDebugConfigurationProvider(ctx: vscode.ExtensionContext): void {
+    const configuration = vscode.workspace.getConfiguration('docker.debugging');
+
+    if (configuration.get('enabled', false)) {
```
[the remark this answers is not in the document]
philliphoff: My understanding is that debugger adaptors have two ways to specify configuration snippets: declaratively in the `package.json` and via the API of the adaptor itself. In my testing, VS Code seems to use the former when no configuration file yet exists and the latter when it does exist.  This means that, to try to cover both scenarios, it's not possible to completely hide the feature behind such a feature flag.  It's admittedly awkward and I debated just removing the declarative snippet, but then that makes the getting started steps more complicated.  I could go either way.

I've updated the setting description to indicate that a restart is necessary.

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,241 @@
...
+            }
+        ];
+    }
+
+    public resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration> {
+        return this.resolveDockerDebugConfiguration(folder, debugConfiguration);
+    }
+
+    private static resolveFolderPath(folderPath: string, folder: WorkspaceFolder): string {
+        return folderPath.replace(/\$\{workspaceFolder\}/g, folder.uri.fsPath);
```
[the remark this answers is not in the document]
philliphoff: Yes and no. VS Code will handle replacement if you let the paths in the configuration flow through back to VS Code (for execution of the debugger).  This adaptor needs to perform work before that configuration flows back to VS Code, and that work requires knowing the real (local) paths, and I couldn't find a VS Code extension API to do so, hence the manual resolution.

It was an attempt to support the same style of configuration used in other debug configurations, though you're correct in that it doesn't support all of them (yet).  We could decide instead to just require all configuration paths to be relative to the workspace folder.  I am also willing to add support for the others (as they make sense), though I would prefer to do that as a separate PR given the size of this PR.

dockerExtension.ts
```diff
@@ -281,6 +297,90 @@ namespace Configuration {
...
+                        fileSystemProvider,
+                        new CommandLineMSBuildClient(processProvider),
+                        new OSTempFileProvider(
+                            osProvider,
+                            processProvider)),
+                    async () => {
+                        // NOTE: Debugging .NET Core in Docker containers requires the C# (i.e. .NET Core debugging) extension.
+                        //       As Docker debugging is experimental, we don't want the extension as a whole to depend on it.
+                        //       Hence, we only check for its existence if/when asked to debug .NET Core in Docker containers.
+                        // TODO: Remove this check once debugging is enabled-by-default (and then add a package.json extension dependency).
```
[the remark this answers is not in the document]
philliphoff: Fair point; removed.

dockerExtension.ts
```diff
@@ -281,6 +297,90 @@ namespace Configuration {
...
+                            }
+                        }
+
+                        return dependenciesSatisfied;
+                    })));
+
+        ctx.subscriptions.push(
+            vscode.debug.onDidChangeActiveDebugSession(
+                session => {
+                    if (session === undefined) {
```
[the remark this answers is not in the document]
philliphoff: The intent is to shutdown any containers used for Docker debugging still running when debugging ends in VS Code, whether from the ending session or a previous one.  I'm not sure it matters which kind of session was ending, in that if there are no containers running it's a no-op and if there are, then we'd still want them shutdown.

(As far as I know, VS Code only supports one debugging session at a time.)

debugging/netcoreapp/fsProvider.ts
```diff
@@ -0,0 +1,150 @@
...
+    makeDir(path: string): Promise<void>;
+    readDir(path: string): Promise<string[]>;
+    readFile(filename: string, encoding?: string): Promise<string>;
+    unlinkFile(filename: string): Promise<boolean>;
+    // tslint:disable-next-line:no-any
+    writeFile(filename: string, data: any): Promise<void>;
+}
+
+export class LocalFileSystemProvider implements FileSystemProvider {
+    public async dirExists(path: string): Promise<boolean> {
```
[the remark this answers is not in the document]
philliphoff: Ah, hadn't realized that was already in use.  Updated to use that instead.  (The use of the provider structure itself is an intentional abstraction of the file system for (future) testing purposes.)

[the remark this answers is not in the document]
EricJizbaMSFT: I understand the desire to plan ahead, but I'm also not a fan of over-complicating code before it's actually needed. How soon do you plan to write the tests? Are you sure the tests need an abstraction and can't just run against the real thing?

debugging/netcoreapp/osProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    get os(): PlatformType {
+        return os.platform() === 'win32' ? 'Windows' : 'Linux';
+    }
+
+    get tmpdir(): string {
+        return os.tmpdir();
+    }
+
+    public pathJoin(pathOS: PlatformType, ...paths: string[]): string {
```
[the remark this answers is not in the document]
philliphoff: Ah, didn't realize `path` offered the OS-specific ones.  Switched to that.

debugging/netcoreapp/appStorage.ts
```diff
@@ -0,0 +1,63 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as path from 'path';
+import { FileSystemProvider } from "./fsProvider";
+
+export interface MementoAsync {
```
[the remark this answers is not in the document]
philliphoff: `AppStorage` represents a state at the "application" or "project" level, which may often be sub-workspace-state.  (You might have several applications within the workspace folder, say, if using VS Code to open a solution created by VS comprised of several individual ASP.NET projects.)  It's intended to be used for things that are specific to the application (e.g. cache settings used to build/run its Docker images/containers) and not dependent on which folder was opened in VS Code (e.g. the solution level or project level folder) to edit that project.  This parallels how VS stores such data.

philliphoff:
> Does that mean anything to you? I admittedly haven't investigated much and plan to look into it further.

I'll have to look into that too.  Does that path exist on your machine?  What folders are shared in your Docker preferences?  (Mine does not explicitly state that `/usr/...` is shared, but it somehow still works.)

dockerExtension.ts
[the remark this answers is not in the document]
EricJizbaMSFT: I'm just concerned because this code will be run for _every_ user that has the docker extension installed - even if they never touch the 'docker-netcoreapp' experience (well - every user once we get rid of the feature flag). No matter how perfect the code is today (which could change in the future) I would prefer to have some checking here to prevent the docker extension from affecting users that are debugging entirely different things.

Can you use `onDidTerminateDebugSession` and check `session.type === 'docker-netcoreapp'`?

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Sounds good. If I were you, I wouldn't spend time adding support for other variables - I would just file an issue on VS Code.

dockerExtension.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Who is the intended audience for this preview feature? If it's just for internal people or MVPs, then I think "more complicated" is okay. If you want the average user to try this as well (based solely on instructions from the README) then I would prefer the less complicated scenario.

If the current implementation is only meant to hide the snippet from `provideDebugConfigurations`, I would prefer if we read the setting directly in that method. That way the debug type is still defined for people who find the snippet the other way.

I'm also open to removing the feature flag completely if we make sure the entry points are labeled with 'Preview' somehow.

CHANGELOG.md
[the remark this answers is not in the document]
EricJizbaMSFT: Ah okay - I wasn't sure if you wanted to include this in our next release train or if you wanted a specific release for this feature right after this PR was merged.

If this is just the "unreleased" case, I don't have a strong preference.

debugging/netcoreapp/appStorage.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Okay I'll have to try out the scenario you described. I just want to make sure we're following VS Code conventions here, which may be different than VS. In general, VS Code has a much bigger reliance on the fact that the project is at the root of the workspace. They also have a concept of 'multi-root' workspaces, which might be how they handle the multi-project scenario you're describing (which is handled appropriately by workspaceState)

philliphoff:
@EricJizbaMSFT Did you happen to recently install OS X Mojave (10.14)?  I'm now seeing that same Docker mount error after doing so.  I had to manually add `/usr/local/share/dotnet` as a shared folder in the Docker preferences; after that, it worked.  I'll look into why that's happening and what the extension might be able to do about it.

EricJizbaMSFT:
Uh nope not on Mojave
screen shot 2018-09-28 at 11 55 48 am

philliphoff:
It looks like we ran into this issue with the VS for Mac Docker Tools and submitted a [request](https://github.com/docker/for-mac/issues/2214) to allow programmatic additions to the shared folder list.  So strange that I hadn't run into this issue until now on my machine.  I'll add instructions to the user to add the shared folder in the `README.md`.

philliphoff:
I've added a "prerequisite check" that the shared folder exists in the user's Docker preferences.

dockerExtension.ts
```diff
@@ -281,6 +298,75 @@ namespace Configuration {
     }
 }
 
+function registerDebugConfigurationProvider(ctx: vscode.ExtensionContext): void {
```
EricJizbaMSFT: Can you move this function into it's own file in the 'debugging' folder? We try to keep the main extension.ts file as clean as possible otherwise it gets out of hand quickly

dockerExtension.ts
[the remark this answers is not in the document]
philliphoff: I ran into issues with `onDidTerminateDebugSession` when the VS Live Share extension was installed, as it may "wrap" other debug adaptors, even when not actively used.  In that case, the returned `type` is the Live Share adaptor.  It exposes the underlying adaptor type via a custom request, but when a session is terminated you cannot make custom requests.

Instead, I add a "session manager" that's informed by my debug adaptor when debugging is about to start, at which time it starts listening to the termination event.  Then, on termination, it performs the necessary cleanup and stops listening for future events.  This should limit it the cleanup to running only when the user has attempted to debug within a Docker container, but no more than once per attempt.

dockerExtension.ts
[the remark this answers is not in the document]
philliphoff: The feature flag was just paranoia on my part, to ensure that the experimental feature could not impact other production features.  To keep things simple, that is, to ensure that users can get to the feature from both scaffolding entry points and without requiring a reload/restart, I've just removed the feature flag and labeled the scaffolded items/commands as `(Preview)`.

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,242 @@
...
+        private readonly osProvider: OSProvider,
+        private readonly netCoreProjectProvider: NetCoreProjectProvider,
+        private readonly prerequisite: Prerequisite) {
+    }
+
+    public provideDebugConfigurations(folder: WorkspaceFolder | undefined, token?: CancellationToken): ProviderResult<DebugConfiguration[]> {
+        return [
+            {
+                name: 'Docker: Launch .NET Core',
+                type: 'docker-netcoreapp',
```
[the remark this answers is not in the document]
philliphoff: I think I originally chose `netcoreapp` because it's the name of the target platform for .NET Core (e.g. in the project file).  I agree that `docker-coreclr` is more consistent with the naming of the debug adaptor for .NET Core in VS Code.  Changed.

debugging/netcoreapp/osProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+        return pathOS === 'Windows' ? path.win32.join(...paths) : path.posix.join(...paths);
+    }
+
+    public pathNormalize(pathOS: PlatformType, rawPath: string): string {
+        return rawPath.replace(
+            pathOS === 'Windows' ? /\//g : /\\/g,
```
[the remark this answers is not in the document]
philliphoff: I've found that `path.win32.normalize()` will convert `/` to `\` as desired, but `path.posix.normalize()` doesn't seem to do the equivalent; `\` will *not* be converted to `/`.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.x.x - Unreleased
+
+### Added
+* Adds preview support for debugging .NET Core web applications running in Linux Docker containers.
```
StephenWeatherford: Is it just web apps or also console apps

README.md
```diff
@@ -27,6 +28,38 @@ Rich IntelliSense (completions) for `Dockerfile` and `docker-compose.yml` files:
 
 ![IntelliSense for DockerFiles](images/intelliSense.gif)
 
+## Debugging (Preview)
+
+To debug a .NET Core web application running in a Linux Docker container, add a Docker .NET Core launch configuration:
```
StephenWeatherford: Would it be better to say .NET Core ASP.NET app?  That's what they're called when adding dockerfiles to the workspace.

README.md
```diff
@@ -27,6 +28,38 @@ Rich IntelliSense (completions) for `Dockerfile` and `docker-compose.yml` files:
...
+
+1. Switch to the debugging tab.
+1. Select `Add configuration...`
+1. Select `Docker: Launch .NET Core (Preview)`
+1. Set a breakpoint.
```
StephenWeatherford: It would be good to mention the prerequisites that will be required.

README.md
```diff
@@ -27,6 +28,38 @@ Rich IntelliSense (completions) for `Dockerfile` and `docker-compose.yml` files:
...
+1. Start debugging.
+
+> Mac OS X users: before debugging, add `/usr/local/share/dotnet/sdk/NuGetFallbackFolder` as a shared folder in your Docker preferences.
```
StephenWeatherford: A picture is worth a thousands words

README.md
```diff
@@ -27,6 +28,38 @@ Rich IntelliSense (completions) for `Dockerfile` and `docker-compose.yml` files:
...
+
+Upon debugging, a Docker image will be built and a container run based on that image.  The container will have a volumes mapped to the locally-built application and the .NET Core debugger.  After the debugger is attached, the browser will be launched and navigated to the application's initial page.
```
StephenWeatherford: a volumes mapped?

package.json
```diff
@@ -261,6 +262,75 @@
...
+        "label": "Docker: Launch .NET Core (Preview)",
+        "configurationSnippets": [
+          {
+            "label": "Docker: Launch .NET Core (Preview)",
+            "description": "Docker: Launch .NET Core (Preview)",
+            "body": {
+              "name": "Docker: Launch .NET Core (Preview)",
+              "type": "docker-netcoreapp",
+              "request": "launch",
+              "preLaunchTask": "build"
```
StephenWeatherford: This should have an empty dockerOptions in it.

README.md
StephenWeatherford: suggest: container will be run

README.md
```diff
@@ -27,6 +28,38 @@ Rich IntelliSense (completions) for `Dockerfile` and `docker-compose.yml` files:
...
+
+Most properties of the configuration are optional will be inferred from the project. If not, or if there are additional customizations to the Docker image build or container run process to be made, those can be added under the `dockerOptions` property of the configuration.
```
StephenWeatherford: and will be

StephenWeatherford: suggest: If not, or if there are additional customizations to be made to the Docker image build or container run process,

README.md
```diff
@@ -27,6 +28,38 @@ Rich IntelliSense (completions) for `Dockerfile` and `docker-compose.yml` files:
...
+```json
+{
+    "configurations": [
+        {
+            "name": "Docker: Launch .NET Core",
+            "type": "docker-netcoreapp",
+            "request": "launch",
+            "preLaunchTask": "build",
+            "dockerOptions": {
+                // Add customizations here.
```
StephenWeatherford: What are the options available?

debugging/netcoreapp/appStorage.ts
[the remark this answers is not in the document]
StephenWeatherford: I think it would be good to have a meeting to discuss how we expect users to manage these types of projects/solutions in vscode.  Do we have actual customer usage data?

debugging/netcoreapp/debuggerClient.ts
```diff
@@ -0,0 +1,25 @@
...
+import { VsDbgClient } from "./vsdbgClient";
+
+export interface DebuggerClient {
+    getDebugger(os: PlatformType): Promise<string>;
+}
+
+export class DefaultDebuggerClient {
+    private static debuggerVersion: string = 'vs2017u5';
+    private static debuggerLinuxRuntime: string = 'debian.8-x64';
+    private static debuggerWindowsRuntime: string = 'win7-x64';
```
StephenWeatherford: I thought Windows wasn't supported...?

debugging/netcoreapp/debuggerClient.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { PlatformType } from "./osProvider";
+import { VsDbgClient } from "./vsdbgClient";
+
+export interface DebuggerClient {
+    getDebugger(os: PlatformType): Promise<string>;
+}
+
+export class DefaultDebuggerClient {
+    private static debuggerVersion: string = 'vs2017u5';
```
StephenWeatherford: Do we need settings to override these?

debugging/netcoreapp/osProvider.ts
```diff
@@ -0,0 +1,42 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as os from 'os';
+import * as path from 'path';
+
+export type PlatformType = 'Windows' | 'Linux';
```
StephenWeatherford: There's already an "OS" in config-utils.ts.  It should be moved out of there and made common.  Maybe rename "PlatformOS"?

debugging/netcoreapp/prereqManager.ts
```diff
@@ -0,0 +1,93 @@
...
+    constructor(
+        private readonly browserClient: BrowserClient,
+        // tslint:disable-next-line:no-any
+        private readonly getExtension: (extensionId: string) => vscode.Extension<any>,
+        private readonly showErrorMessage: showErrorMessageFunction) {
+    }
+
+    public async checkPrerequisite(): Promise<boolean> {
+        // NOTE: Debugging .NET Core in Docker containers requires the C# (i.e. .NET Core debugging) extension.
+        //       As Docker debugging is experimental, we don't want the extension as a whole to depend on it.
```
StephenWeatherford: We wouldn't want this even if it weren't experimental.  This is a generic docker extension, not just Azure.

debugging/netcoreapp/prereqManager.ts
```diff
@@ -0,0 +1,93 @@
...
+    checkPrerequisite(): Promise<boolean>;
+}
+
+type showErrorMessageFunction = (message: string, ...items: vscode.MessageItem[]) => Thenable<vscode.MessageItem>;
+
+export class DotNetExtensionInstalledPrerequisite implements Prerequisite {
+    constructor(
+        private readonly browserClient: BrowserClient,
+        // tslint:disable-next-line:no-any
+        private readonly getExtension: (extensionId: string) => vscode.Extension<any>,
```
StephenWeatherford: any -> unknown

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+            {
+                name: 'Docker: Launch .NET Core (Preview)',
+                type: 'docker-netcoreapp',
+                request: 'launch',
+                preLaunchTask: 'build'
+            }
+        ];
+    }
+
+    public resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration> {
```
StephenWeatherford: Where is resolveDebugConfiguration used?

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+        return this.resolveDockerDebugConfiguration(folder, debugConfiguration);
+    }
+
+    private static resolveFolderPath(folderPath: string, folder: WorkspaceFolder): string {
+        return folderPath.replace(/\$\{workspaceFolder\}/g, folder.uri.fsPath);
+    }
+
+    private async resolveDockerDebugConfiguration(folder: WorkspaceFolder, debugConfiguration: DockerDebugConfiguration): Promise<DebugConfiguration> {
```
StephenWeatherford: Promise<DebugConfiguration|undefined>.

It would be great if you could temporarily turn on strictNullChecks in tsconfig.json and resolve all errors in your code before merging since we're trying to get the codebase to that point.

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+
+        const resolvedAppProject = DockerDebugConfigurationProvider.resolveFolderPath(appProject, folder);
+
+        const context = this.inferContext(folder, resolvedAppFolder, debugConfiguration);
+
+        const resolvedContext = DockerDebugConfigurationProvider.resolveFolderPath(context, folder);
+
+        let dockerfile = debugConfiguration.dockerOptions && debugConfiguration.dockerOptions.build && debugConfiguration.dockerOptions.build.dockerfile
+            ? DockerDebugConfigurationProvider.resolveFolderPath(debugConfiguration.dockerOptions.build.dockerfile, folder)
+            : path.join(appFolder, 'Dockerfile'); // TODO: Omit dockerfile argument if not specified or possibly infer from context.
```
StephenWeatherford: Are these TODOs required before shipping?  If not, I'd use "CONSIDER"

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,249 @@
...
+
+        const files = await this.fsProvider.readDir(resolvedAppFolder);
+
+        const projectFile = files.find(file => path.extname(file) === '.csproj');
+
+        if (projectFile) {
+            return path.join(resolvedAppFolder, projectFile);
+        }
+
+        throw new Error('Unable to infer the application project file. It must be explicitly set in the Docker debug configuration.');
```
StephenWeatherford: You don't say how to do that or why it couldn't automatically be inferred.  My test project hits this.

debugging/netcoreapp/outputManager.ts
```diff
@@ -0,0 +1,82 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as vscode from 'vscode';
```
StephenWeatherford: At a high level, what does this do?

package.json
```diff
@@ -261,6 +262,75 @@
...
+            }
+          }
+        ],
+        "configurationAttributes": {
+          "launch": {
+            "properties": {
+              "dockerOptions": {
```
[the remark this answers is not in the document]
philliphoff: When transforming the `docker-coreclr` debug configuration into the debug configuration for `coreclr`, some of the configuration is merged (e.g. the `launchBrowser` property) while some is overwritten (e.g. the `program` and `cwd` properties).  Because of the merge, and because the debug configurations don't seem to be rigorously defined, I was concerned about collisions with new Docker-related properties, so made the properties hang off of a single, Docker-specific property.

I'll admit to being overly paranoid (for the first PR in a new codebase).  I've updated the configuration such that the properties now hang off of the root, and updated the `build` and `run` properties to `dockerBuild` and `dockerRun`, respectively, to avoid any confusion about the local .NET Core build/run.

dockerExtension.ts
philliphoff: Sure! Moved.

package.json
```diff
@@ -261,6 +262,75 @@
...
+            "body": {
+              "name": "Docker: Launch .NET Core",
+              "type": "docker-netcoreapp",
+              "request": "launch",
+              "preLaunchTask": "build"
+            }
+          }
+        ],
+        "configurationAttributes": {
+          "launch": {
```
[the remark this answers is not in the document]
philliphoff: Yes, there are standard ways of passing build arguments during image creation and environment variables during container run; I had intended to flesh out these options (in the `dockerBuild` and `dockerRun` properties of the debug configuration) as a follow on PR.

debugging/netcoreapp/prereqManager.ts
```diff
@@ -0,0 +1,93 @@
...
+            return true;
+        }
+
+        const settingsPath = path.join(this.osProvider.homedir, 'Library/Group Containers/group.com.docker/settings.json');
+
+        if (await this.fileSystemProvider.fileExists(settingsPath)) {
+            const settingsContent = await this.fileSystemProvider.readFile(settingsPath);
+            const settings = <DockerSettings>JSON.parse(settingsContent);
+
+            if (settings.filesharingDirectories && settings.filesharingDirectories.find(directory => directory === MacNuGetPackageFallbackFolderPath) !== undefined) {
```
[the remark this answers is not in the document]
philliphoff: Did you press the "Apply & Restart" button after the addition?  I had occasion to flatten and rebuild my machine yesterday which gave me the opportunity to test this scenario again; I saw the expected error before adding the folder as well as success after adding the folder (and after  Apply & Restart).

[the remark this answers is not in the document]
EricJizbaMSFT: I still see the error today, so definitely not just a `Apply & Restart` problem. This is what `settingsContent` looks like for me:
```json
{
    "analyticsEnabled": true,
    "proxyHttpMode": "system",
    "diskPath": "/Users/ericjizba/Library/Containers/com.docker.docker/Data/com.docker.driver.amd64-linux/Docker.qcow2",
    "diskSizeMiB": 65536,
    "cpus": 4,
    "defaultMachineMigrationStatus": 4,
    "memoryMiB": 2048,
    "displayedWelcomeWhale": true,
    "buildNumber": "19611",
    "autoStart": true,
    "channelID": "stable",
    "checkForUpdates": true,
    "settingsVersion": 1,
    "version": "17.09.0-ce-mac35",
    "displayedWelcomeMessage": true,
    "experimentalFeaturesState": {
        "discoverEEV1": true,
        "releaseNotesV1": true,
        "passthroughHTTPProxy": true
    },
    "linuxDaemonConfigCreationDate": "2017-10-31 00:52:23 +0000",
    "dockerAppLaunchPath": "/Applications/Docker.app"
}
```

No `filesharingDirectories` in sight.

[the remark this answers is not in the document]
philliphoff: It looks like you're running an earlier version of Docker for Mac than me (`18.06.1-ce-mac73`).  Perhaps yours stores that data in a different location.  I don't suppose you could do a quick search and see if that added folder shows up in another JSON file somewhere?

CHANGELOG.md
philliphoff: Web apps are the initial target (and the only thing I'm testing at the moment).  Console apps *should* likely be accommodated with perhaps minor tweaks.

README.md
philliphoff: Since the "preview" targets only ASP.NET, makes sense.  Updated.

README.md
philliphoff: Added note about dependency on C# VS Code extension.

README.md
philliphoff: Updated.

philliphoff: Upated.

package.json
philliphoff: Added.

README.md
philliphoff: Updated.

philliphoff: Sure; updated.

README.md
philliphoff: Added tables with descriptions of all of the options (to date).

debugging/netcoreapp/debuggerClient.ts
philliphoff: It's not...*quite*.  There *is* a Windows version of `vsdbg` (hence the constant) but isn't currently supported on the base Windows .NET Core images; the maintainers state that `msvsmon` is still the PoR for Windows container debugging as it "supports more debugging scenarios", but for which VS Code does not have an adaptor (AFAIK).

debugging/netcoreapp/debuggerClient.ts
philliphoff: To date, we haven't seen the need to make these configurable in VS.

debugging/netcoreapp/prereqManager.ts
philliphoff: Fair point; updated comment.

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
philliphoff: Updated.

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
philliphoff: Added explicit instruction to set either the `appFolder` or `appProject` configuration property.

debugging/netcoreapp/outputManager.ts
philliphoff: It wraps output going to a VS Code `OutputChannel`, allowing it to be grouped into operations, i.e. with a starting message, ending message, error message, where the operation content is prefixed.  It allows a simple visualization of a hierarchy of operations.  This is often needed when performing Docker operations, where it's desired to display the raw output of the operation but ensure it's distinct from the output of the extension itself.

For example:

```
Starting Docker build...
> Sending build context to Docker daemon   2.47MB
> Step 1/16 : FROM microsoft/dotnet:2.1-aspnetcore-runtime AS base
> ---> 2967770cac05
> Step 2/16 : WORKDIR /app
> ---> Using cache
> ---> 7c9558751bf3
> ...
Docker build complete.
```

debugging/netcoreapp/prereqManager.ts
philliphoff: Isn't `unknown` only available in TypeScript 3.0?

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
philliphoff: Made updates to build cleanly with `strictNullChecks` enabled.

debugging/netcoreapp/osProvider.ts
philliphoff: Moved out, renamed, and used from debugging implementation.

dockerExtension.ts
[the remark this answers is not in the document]
StephenWeatherford: sounds good

README.md
philliphoff: Added a thousand words (i.e. one picture).

README.md
StephenWeatherford: Doesn't it also require the az CLI?

README.md
```diff
@@ -10,6 +10,7 @@ The Docker extension makes it easy to build, manage and deploy containerized app
 * Command Palette (`F1`) integration for the most common Docker commands (for example `docker build`, `docker push`, etc.)
 * Explorer integration for managing Images, running Containers, and Docker Hub registries
 * Deploy images from Docker Hub and Azure Container Registries directly to Azure App Service
+* Debug .NET Core applications running in Linux Docker containers
 * [Working with docker](https://code.visualstudio.com/docs/azure/docker) will walk you through many of the features of this extension
```
StephenWeatherford: Seems like the ASP.NET debugging should appear lower in the README after the more general functionality.

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
StephenWeatherford: Thanks!  Much appreciated.

debugging/netcoreapp/prereqManager.ts
StephenWeatherford: You can merge this in: https://github.com/Microsoft/vscode-docker/pull/528

debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,250 @@
...
+
+        const files = await this.fsProvider.readDir(resolvedAppFolder);
+
+        const projectFile = files.find(file => path.extname(file) === '.csproj');
+
+        if (projectFile) {
+            return path.join(resolvedAppFolder, projectFile);
+        }
+
+        throw new Error('Unable to infer the application project file. Set either the `appFolder` or appProject` property in the Docker debug configuration.');
```
StephenWeatherford: Missing tick on appProject

README.md
```diff
@@ -27,6 +28,78 @@ Rich IntelliSense (completions) for `Dockerfile` and `docker-compose.yml` files:
 
 ![IntelliSense for DockerFiles](images/intelliSense.gif)
 
+## Debugging .NET Core ASP.NET (Preview)
+
+To debug a .NET Core ASP.NET application running in a Linux Docker container, add a Docker .NET Core launch configuration:
```
StephenWeatherford: Be more explicit about saying Windows is not currently supported (preferably verify from code).

debugging/coreclr/appStorage.ts
```diff
@@ -0,0 +1,63 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as path from 'path';
```
StephenWeatherford: We need to discuss this.

debugging/coreclr/dockerClient.ts
```diff
@@ -0,0 +1,194 @@
...
+        const buildProgress =
+            (content: string) => {
+                if (progress) {
+                    progress(content);
+                }
+
+                // Last line of expected output is: 'Successfully built 7cc5654ca3b6'
+                const buildSuccessPrefix = 'Successfully built ';
+
+                if (content.startsWith(buildSuccessPrefix)) {
```
StephenWeatherford: TODO

debugging/coreclr/dockerClient.ts
```diff
@@ -0,0 +1,194 @@
...
+        await this.processProvider.exec(command, { progress: buildProgress });
+
+        if (!imageId) {
+            throw new Error('The Docker image was built successfully but the image ID could not be retrieved.');
+        }
+
+        return imageId;
+    }
+
+    public async inspectObject(nameOrId: string, options?: DockerInspectObjectOptions): Promise<string | undefined> {
```
StephenWeatherford: TODO

debugging/coreclr/dockerClient.ts
```diff
@@ -0,0 +1,194 @@
...
+
+        command += ' ' + imageTagOrId;
+
+        if (options && options.command) {
+            command += ' ' + options.command;
+        }
+
+        const result = await this.processProvider.exec(command, {});
+
+        const containerId = result.stdout.substr(0, result.stdout.length - 1 /* Exclude trailing <CR>. */);
```
StephenWeatherford: And if this is Windows?  CRLF?

debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,250 @@
...
+            }
+        ];
+    }
+
+    public resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration | undefined> {
+        return this.resolveDockerDebugConfiguration(folder, debugConfiguration);
+    }
+
+    private static resolveFolderPath(folderPath: string, folder: WorkspaceFolder): string {
+        return folderPath.replace(/\$\{workspaceFolder\}/g, folder.uri.fsPath);
```
StephenWeatherford: should this be /gi for case-insensitive?

debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    private async resolveDockerDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration): Promise<DebugConfiguration | undefined> {
+        if (!folder) {
+            return undefined;
+        }
+
+        const prerequisiteSatisfied = await this.prerequisite.checkPrerequisite();
+
+        if (!prerequisiteSatisfied) {
+            return undefined;
```
StephenWeatherford: should this throw?

debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,250 @@
...
+    constructor(
+        private readonly debugSessionManager: DebugSessionManager,
+        private readonly dockerManager: DockerManager,
+        private readonly fsProvider: FileSystemProvider,
+        private readonly osProvider: OSProvider,
+        private readonly netCoreProjectProvider: NetCoreProjectProvider,
+        private readonly prerequisite: Prerequisite) {
+    }
+
+    public provideDebugConfigurations(folder: WorkspaceFolder | undefined, token?: CancellationToken): ProviderResult<DebugConfiguration[]> {
```
StephenWeatherford: You should wrap stuff that gets called directly from vscode with callWithTelemetryAndErrorHandling.  E.g.:

        await callWithTelemetryAndErrorHandling('getCustomRegistryUsernamePwd', async function (this: IActionContext): Promise<void> {
              this.rethrowError = true;
              …

StephenWeatherford: Not needed for commands because our registerCommand handles it automatically.

debugging/coreclr/dockerManager.ts
```diff
@@ -0,0 +1,330 @@
...
+            ? '-t localhost'
+            : '-f /dev/null';
+
+        const entrypoint = options.os === 'Windows'
+            ? 'ping'
+            : 'tail';
+
+        const volumes = this.getVolumes(debuggerFolder, options);
+
+        const containerId = await this.dockerOutputManager.performOperation(
```
StephenWeatherford: asdf

debugging/coreclr/outputManager.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    public appendLine(content: string): void {
+        if (this.level) {
+            this.outputChannel.append(this.generatePrefix());
+        }
+
+        this.outputChannel.appendLine(content);
+    }
+
+    public async performOperation<T>(startContent: string, operation: (outputManager: OutputManager) => Promise<T>, endContent?: string | outputCallback<T>, errorContent?: string | outputCallback<Error>): Promise<T> {
```
StephenWeatherford: This might be the right place to put in the callWithTelemetryAndErrorHandling call...?

debugging/coreclr/outputManager.ts
```diff
@@ -0,0 +1,82 @@
...
+export class DefaultOutputManager implements OutputManager {
+    private skipFirstLine: boolean = false;
+    private isShown: boolean = false;
+
+    constructor(private readonly outputChannel: vscode.OutputChannel, private readonly level: number = 0) {
+    }
+
+    public append(content: string): void {
+        if (this.level) {
+            const split = content.split(/[\n\r]/g);
```
StephenWeatherford: On Windows this will probably result in empty entries in the split array.

README.md
philliphoff: Moved to below the private registries (preview) section.

debugging/coreclr/dockerDebugConfigurationProvider.ts
philliphoff: Fixed.

debugging/coreclr/dockerClient.ts
philliphoff: Fair point; updated to use `trim()` instead of making assumptions about exact number(s) of whitespace.  While Docker explicitly states the value in the output is the container ID, it doesn't appear to document their container ID format, so a regex here would still be relying on an assumption.

StephenWeatherford: trim() works

debugging/coreclr/dockerDebugConfigurationProvider.ts
philliphoff: Reasonable; updated.

debugging/coreclr/outputManager.ts
philliphoff: Yes, it was a naive implementation.  I've replaced it with something more robust and has unit tests.

debugging/coreclr/dockerClient.ts
```diff
@@ -0,0 +1,194 @@
...
+
+        if (!containerId) {
+            throw new Error('The Docker container was run successfully but the container ID could not be retrieved.')
+        }
+
+        return containerId;
+    }
+
+    public trimId(id: string): string {
+        return id.substring(0, 12);
```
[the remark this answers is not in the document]
philliphoff: Added some additional error handling.  A trimmed ID is just the first 12 characters of the full Docker ID (whatever format it happens to be using for that).

README.md
philliphoff: Added not to `README.MD` and prerequisite check for a Linux Docker host.

debugging/coreclr/dockerClient.ts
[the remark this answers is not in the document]
philliphoff: Noted and I think that can *probably* be accommodated.  For VS we don't explicitly support debugging via non-local Docker hosts, though it will likely work assuming you've setup the Docker CLI context to default to the remote host.  Since debugging requires passing the Core CLR debugging an actual Docker command line, we'd need to ensure that the Docker host used by `docker-endpoint` matches up with that command line used for debugging.

In any case, I would prefer to make such a swap as a separate PR to simplify testing.  I've added a `CONSIDER` comment to `dockerClient`.

debugging/coreclr/dockerDebugConfigurationProvider.ts
philliphoff: In my opinion no, as the prerequisite checks themselves already displaying their corresponding actionable messages.  If we throw here, then VS Code would display another error message which would be less actionable as well as tend to dilute the prerequisite check errors.

debugging/coreclr/dockerDebugConfigurationProvider.ts
philliphoff: Wrapped call with `callWithTelemetryAndErrorHandling()`.

debugging/netcoreapp/dockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,242 @@
...
+                name: 'Docker: Launch .NET Core',
+                type: 'docker-netcoreapp',
+                request: 'launch',
+                preLaunchTask: 'build'
+            }
+        ];
+    }
+
+    public resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration> {
+        return this.resolveDockerDebugConfiguration(folder, debugConfiguration);
```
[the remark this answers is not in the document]
philliphoff: Added telemetry around the resolution of the debug configuration (that indicates debugging being invoked).

debugging/netcoreapp/outputManager.ts
```diff
@@ -0,0 +1,82 @@
...
+        try {
+            const result = await operation(new DefaultOutputManager(this.outputChannel, this.level + 1));
+
+            if (endContent) {
+                this.appendLine(typeof endContent === 'string' ? endContent : endContent(result));
+            }
+
+            return result;
+        } catch (error) {
+            if (errorContent) {
```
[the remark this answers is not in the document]
philliphoff: `performOperation()` only handles errors to the extent that of ensuring symmetry of logical operations as shown in the Docker output window.  It wasn't intended to interact with the user or generate VS Code error items.  (At this point it's mostly used to "nest" Docker output between logical operation start/end messages.

README.md
philliphoff: Yes, true.  Added a prerequisite check for the `dotnet` CLI.

debugging/coreclr/appStorage.ts
philliphoff: As discussed in a comment with @EricJizbaMSFT, while VS Code has a memento service for data associated with the user as well as with a given workspace, it doesn't have something (that I know of) that is associated with an arbitrary folder (which may be one of many subfolders within a workspace and, in the case of debugging, corresponds to a given application being debugged).  Consider, for example, opening VS Code to the solution-level folder that has several applications, each of which could be debugged individually; we want to cache some data for each individual application.

I could likely come up with a key-naming scheme to stuff this data in the workspace memento service, but it just feels like the data should belong to each application itself (as well as is consistent with where this data is stored in VS).

philliphoff:
> Also still need CTI testcases 
Requested access to the OneNote in order to add test cases.

> Should do some basic testing on Linux.
Ran through the basic scenario on Ubuntu (and added prerequisite check for user inclusion in `docker` group).

philliphoff:
> Also still need CTI testcases 

Added a CTI test page including a main end-to-end scenario test:

https://microsoft.sharepoint.com/teams/azcode/_layouts/OneNote.aspx?id=%2Fteams%2Fazcode%2FSiteAssets%2Fazcode%20Notebook&wd=target%28Testing.one%7CD12CBB51-9002-4371-B958-B88191DF5CF5%2F.NET%20Core%20Debugging%20%28Not%20yet%20checked%20in%5C%29%7C818BF24C-34CD-8940-A165-3407A13A5F9E%2F%29
onenote:https://microsoft.sharepoint.com/teams/azcode/SiteAssets/azcode%20Notebook/Testing.one#.NET%20Core%20Debugging%20(Not%20yet%20checked%20in)&section-id={D12CBB51-9002-4371-B958-B88191DF5CF5}&page-id={818BF24C-34CD-8940-A165-3407A13A5F9E}&end

philliphoff:
Ok, so all comments should now be addressed one way or another.

@StephenWeatherford Let me know if there's anything else I can do.

debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -71,7 +71,9 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
     }
 
     public resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration | undefined> {
-        return this.resolveDockerDebugConfiguration(folder, debugConfiguration);
+        return callWithTelemetryAndErrorHandling(
+            'debugCoreClr',
+            async () => await this.resolveDockerDebugConfiguration(folder, debugConfiguration));
```
StephenWeatherford: nit: debugCoreClr doesn't seem to say what it's really doing. How about "resolveDebugConfiguration"?

philliphoff: I chose `debugCoreClr` as that was *effectively* what was being done (i.e. an attempt to initiate debugging of .NET Core in Docker) even though `resolveDebugConfiguration` is what VS Code happens to name the interface function (and PMs looking at the telemetry data typically care more about the intent than the implementation specifics).  I'm amenable to other IDs or schemes; are there any conventions/restrictions in the naming (as the codebase uses a couple different styles)?

StephenWeatherford:
Other than the suggested rename of telemetry, ship it!  Thanks!

debugging/coreclr/dockerDebugConfigurationProvider.ts
StephenWeatherford: If that's what it's effectively doing, it's fine, thanks.

philliphoff:
Many thanks to @EricJizbaMSFT and @StephenWeatherford for their very thorough review!

PR #503, opened 2018-09-22
Add buttons for Azure CLI and configure commands

### LINKED ISSUE #473: Make "Azure CLI" command more discoverable

StephenWeatherford:
Fixes #473 
TODO:
- [ ] Only show CLI button if Azure extension installed

Add docker files:
image
Azure CLI:
image

image
image
image


package.json
```diff
@@ -136,18 +136,33 @@
...
+          "command": "vscode-docker.configure",
+          "group": "docker"
         }
       ],
       "view/title": [
         {
           "command": "vscode-docker.explorer.refresh",
           "when": "view == dockerExplorer",
-          "group": "navigation"
+          "group": "navigation@1"
```
EricJizbaMSFT: Every other extension has refresh on the farthest right. The file explorer also has it on the right of the 'create' actions

package.json
```diff
@@ -489,9 +508,13 @@
       },
       {
         "command": "vscode-docker.container.start.azurecli",
-        "title": "Azure CLI",
+        "title": "Run Azure CLI",
```
EricJizbaMSFT: I'm kind of confused by this command. It feels like generic Azure tooling that _happens_ to leverage a container, rather than container-specific tooling. Is it really worth putting in the title bar?

Also - doesn't this overlap with the cloud shell commands? In both cases you don't need the Azure cli installed. Maybe we just need to make _those_ commands more obvious somehow
screen shot 2018-09-24 at 1 32 02 pm

Maybe another button next to the filter button?
screen shot 2018-09-24 at 1 45 27 pm

images/FileGroup_16x.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><style>.icon-canvas-transparent{opacity:0;fill:#f6f6f6}.icon-vs-bg{fill:#424242}.icon-vs-fg{fill:#f0eff1}.icon-vs-out{fill:#f6f6f6}</style><path class="icon-canvas-transparent" d="M0 0h16v16H0z" id="canvas"/><path class="icon-vs-out" d="M16 3.379L12.621 0H5v3H0v13h11v-3h5z" id="outline"/><path class="icon-vs-bg" d="M12.207 1H6v2h1V2h4v3h3v6h-3v1h4V3.793L12.207 1zM12 4V2.207L13.793 4H12zM1 4v11h9V6.793L7.207 4H1zm8 10H2V5h4v3h3v6zM7 7V5.207L8.793 7H7z" id="iconBg"/><path class="icon-vs-fg" d="M7 5.207L8.793 7H7V5.207zM6 8V5H2v9h7V8H6zm6-5.793V4h1.793L12 2.207zM11 5V2H7v1h.621L11 6.379V11h3V5h-3z" id="iconFg"/></svg>
```
EricJizbaMSFT: You've got 6 new svgs here, but looks like only 4 are used

package.json
StephenWeatherford: Good question.  @chrisdias @fiveisprime Is there a reason to have both "Run Azure CLI" and "Open Bash in Cloud Shell"?

fiveisprime: The CLI container will mount in the file system, right? That (along with not provisioning time) makes it much more usable than the cloud shell.

EricJizbaMSFT: @fiveisprime so should this be in the azure account extension so that more people benefit? Again - this doesn't feel like docker-specific tooling. It just _happens_ to use docker

StephenWeatherford: Correct.  It looks like (from quick docs perusal) that cloud shell is limited to drives in Azure storage.

EricJizbaMSFT:
Closing this PR due to lack of activity

PR #525, opened 2018-10-03
Update to latest telemetry module

### LINKED ISSUE #522: Please use the latest version of vscode-extension-telemetry dependency
This extension uses version 0.0.18 of `vscode-extension-telemtetry`  module.

The latest version of `vscode-extension-telemtetry` module (0.0.22) uses the latest version of `applicationinsights` module (1.0.5) which has the fix to prevent unnecessary patching of the require and other 3rd party modules. 

For more details, please see https://github.com/Microsoft/vscode-extension-telemetry/issues/16

StephenWeatherford:
Fixes #522

### DIFF
--- a/package.json
+++ b/package.json
@@ -693,7 +693,7 @@
     "request-promise-native": "^1.0.5",
     "semver": "^5.5.1",
     "vscode-azureextensionui": "^0.17.0",
-    "vscode-extension-telemetry": "0.0.18",
+    "vscode-extension-telemetry": "^0.0.22",
     "vscode-languageclient": "^4.4.0"
   }
 }


package.json
```diff
@@ -693,7 +693,7 @@
     "request-promise-native": "^1.0.5",
     "semver": "^5.5.1",
     "vscode-azureextensionui": "^0.17.0",
-    "vscode-extension-telemetry": "0.0.18",
+    "vscode-extension-telemetry": "^0.0.22",
```
EricJizbaMSFT: Just want to make sure you know that `^0.0.22` is functionally the same as `0.0.22` and thus we still won't get any updates automatically. The [caret](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) "Allows changes that do not modify the left-most non-zero digit"

I already filed an issue to ask that they change their versioning strategy: https://github.com/Microsoft/vscode-extension-telemetry/issues/22

PR #526, opened 2018-10-03
Remove files written notification

### LINKED ISSUE #515: notification overload
when i "add dockerfiles to the workspace", i get a prompt that tells me files were added to my workspace.  This seems unnecessary:

* the command i executed indicated that i wanted to add files to my workspace
* i can see that new files were added because they show up in my explorer
* i can see that files were added because i am using source code control

For the past few years i don't remember anyone complaining that this was an issue, but now i have to click again to dismiss the notification.

StephenWeatherford:
Fixes #515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.3.1 - 25 September 2018
+
+### Fixed
+
+* Error while generating Dockerfile for 'other' [#504](https://github.com/Microsoft/vscode-docker/issues/504)
+
 ## 0.3.0 - 21 September 2018
 
 ### Added
--- a/commands/azureCommands/pull-from-azure.ts
+++ b/commands/azureCommands/pull-from-azure.ts
@@ -1,0 +1,106 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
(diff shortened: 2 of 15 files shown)


chrisdias:
i dont see any reason to open it... again, i'm probably under SCC so i'll see the new files added to the workspace in the explorer pretty quickly.

EricJizbaMSFT:
I was only suggesting we open the Dockerfile _if it's useful_ (not purely as a notification). I'm coming from experience with the functions extension where I think it's pretty fair to assume users will edit their new 'index.js' file or set a breakpoint. If users never interact with their Dockerfile, then I agree it's more noise than it's worth.

Back to the notification, though - I do want to make sure we're "designing for the user" and not "designing for the demo". When I'm dogfooding VS Code I hardly ever dismiss info messages (that don't have a button) because they disappear in 10 seconds anyways. I literally don't care about them. However, I can imagine being annoyed during a demo where "Add docker files" was the main attraction and I wanted it to be spotless.

I can't vouch for Docker, but we _have_ received feedback in the functions extension that people want  feedback once files have been added. The entry point for 'Create new project' and 'Create function' is in the Azure view, so users have to explicitly switch to the SCC or file view to notice if things were added. New users have said they end up clicking the button multiple times before they thought to check the other views.

StephenWeatherford:
Personally I like the idea of opening the Dockerfile.  I would expect most users to at least check it over if not make modifications.  And if the next step is to build, that's easier with the file open (right-click -> Build Image).

chrisdias:
i can see why you would want to open the `Dockerfile`, so I would be ok with that.  i just have not seen any customer feedback that suggests its a problem (as you have with functions).

PR #528, opened 2018-10-05
Update mtslint contrib

StephenWeatherford:


StephenWeatherford:
That's weird, apparently I already updated it, must have been looking at an older branch? will update title.

PR #530, opened 2018-10-06
Fix truncation of multi-level image IDs

### LINKED ISSUE #527: UNUSABLE OUTPUT: Truncate long Image and Container registry paths in the Explorer
SETTING:

```json
  // Truncate long Image and Container registry paths in the Explorer
  "docker.truncateLongRegistryPaths": false,
```

OUTPUT:

```
container-registry.domain.com:5000/username/image-name:tag (3 days ago)
```

`true` OUTPUT:

```
con...5000/username (3 days ago)
```

When `true` is selected, the output is unusable. The most significant value in a image's path is the image name, followed by the tag, then user and registry path. I would expect the output to only truncate the registry path, not the image name and tag.

EXPECTED OUTPUT:

```
con...5000/username/image-name:tag (3 days ago)
```

Version: 0.3.1
OS: darwin

StephenWeatherford:
Fixes #527

### DIFF
--- a/explorer/models/containerNode.ts
+++ b/explorer/models/containerNode.ts
@@ -23,7 +23,7 @@ export class ContainerNode extends NodeBase {
 
     public getTreeItem(): vscode.TreeItem {
         let config = vscode.workspace.getConfiguration('docker');
-        let displayName: string = getImageOrContainerDisplayName(this.label, config.get('truncateLongRegistryPaths'), config.get('truncateMaxLength', 10));
+        let displayName: string = getImageOrContainerDisplayName(this.label, config.get('truncateLongRegistryPaths'), config.get('truncateMaxLength'));
 
         return {
             label: `${displayName}`,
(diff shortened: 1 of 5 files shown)


explorer/models/containerNode.ts
```diff
@@ -21,14 +22,8 @@ export class ContainerNode extends NodeBase {
...
-        let displayName: string = this.label;
-
-        if (vscode.workspace.getConfiguration('docker').get('truncateLongRegistryPaths', false)) {
-            if (/\//.test(displayName)) {
-                let parts: string[] = this.label.split(/\//);
-                displayName = trimWithElipsis(parts[0], vscode.workspace.getConfiguration('docker').get('truncateMaxLength', 10)) + '/' + parts[1];
-            }
-        }
+        let config = vscode.workspace.getConfiguration('docker');
+        let displayName: string = getImageOrContainerDisplayName(this.label, config.get('truncateLongRegistryPaths'), config.get('truncateMaxLength', 10));
```
PrashanthCorp: `config.get('truncateLongRegistryPaths', false)` was the previous usage. I see package.json properly handles the default in this case. 
Looking at package.json again, the 10 in `config.get('truncateMaxLength', 10)` should not be required in any of these calls.

StephenWeatherford:
ping

PR #531, opened 2018-10-08
Add syntax highlighting for Dockerfile.*

### LINKED ISSUE #511: Add syntax highlighting for Dockerfile(.*)
Source issue https://github.com/Microsoft/vscode/issues/38797

frozzare:
Fixes #511

### DIFF
--- a/package.json
+++ b/package.json
@@ -271,7 +271,8 @@
         ],
         "filenamePatterns": [
           "*.dockerfile",
-          "Dockerfile"
+          "Dockerfile",
+          "Dockerfile.*"
         ]
       }
     ],


StephenWeatherford:
@frozzare Thanks for contributing.  Unfortunately, it's not quite that simple - see https://github.com/Microsoft/vscode-docker/pull/248.  What we really need is a way to customize this per configuration, but vscode doesn't currently support that.  I'm planning on filing an issue soon but need to take time to investigate more thoroughly what is not supported and what's needed.

frozzare:
@StephenWeatherford I see how that can be a problem after reading #134, e.g `Dockerfile.go` or something like that. User configuration can be one way to solve it and only support those filenames Docker supports by default (I assume that's only `Dockerfile`?) and/or having support some variants (e.g `Dockerfile.windows`) within this extension.

EricJizbaMSFT:
FYI this PR was created before our recent switch from Travis CI to Azure Dev Ops. Push a commit (merging with master or addressing comments) and the "checks" for this PR will be updated

StephenWeatherford:
#511
I'll be looking into this group of extension issues soon.

slxiao:
When could this PR be merged? I am looking forward to such function

StephenWeatherford:
Sorry for taking so long.  There just doesn't seem to be a good way to do this right now in vscode (there's no way to use user-configured settings in telling vscode what filename patterns to use for determining the file type).  I do have an idea I'd like to try, though, I'll try to get to it soon.

EricJizbaMSFT:
Closing this PR since this solution is too intrusive as mentioned above. We're actively investigating the alternative approach to add settings for these glob patterns. Tracked here: https://github.com/Microsoft/vscode-docker/issues/102

Issue #532, opened 2018-10-08
sqlserver


211003514:
&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: vscode-docker.configure
Error type: Error
Error Message: Docker files can only be generated if VS Code is opened on a folder.

Version: 0.3.1
OS: darwin

StephenWeatherford:
@211003514 Please follow the error message - you need to actually open your project or solution folder inside VS before using 'Add Docker Files to Workspace' so that we can find related files needed to figure out the current context.  Thanks.

StephenWeatherford:
I've improved the error message a bit.  Thanks for reporting, let us know if you're still having issues.

PR #533, opened 2018-10-08
Improve message that opening a folder is required for some commands

StephenWeatherford:
Improves message for #532.
Also combines the code and correctly throws rather than just showing messages.

### DIFF
--- a/commands/build-image.ts
+++ b/commands/build-image.ts
@@ -10,6 +10,7 @@ import { DOCKERFILE_GLOB_PATTERN } from '../dockerExtension';
 import { delay } from "../explorer/utils/utils";
 import { ext } from "../extensionVariables";
 import { addImageTaggingTelemetry, getTagFromUserInput } from "./tag-image";
+import { quickPickWorkspaceFolder } from "./utils/quickPickWorkspaceFolder";
 
 async function getDockerFileUris(folder: vscode.WorkspaceFolder): Promise<vscode.Uri[]> {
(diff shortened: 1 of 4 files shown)


commands/build-image.ts
```diff
@@ -56,25 +57,7 @@ export async function buildImage(actionContext: IActionContext, dockerFileUri: v
...
-
-    if (!rootFolder) {
-        if (!vscode.workspace.workspaceFolders) {
-            vscode.window.showErrorMessage('Docker files can only be built if VS Code is opened on a folder.');
-        } else {
-            vscode.window.showErrorMessage('Docker files can only be built if a workspace folder is picked in VS Code.');
-        }
-        return;
-    }
+    let rootFolder: vscode.WorkspaceFolder = await quickPickWorkspaceFolder('To build Docker files you must first open a folder or workspace in VS Code.');
```
PrashanthCorp: I see `you must first open a folder or workspace in VS Code.` being repeated in all 3 calls. Possibly add that to the function itself?

StephenWeatherford: I'd rather not. Doing string concatenation of parts of sentences causes problems if we want to localize later.

PrashanthCorp: Interesting. I thought the loc could happen in the function, at the point we throw the error. 
I don't feel strongly about this, I was just wondering if there was a way to force consistency in the messages in the future. These errors would stand out to the user more than to they would to the dev.

StephenWeatherford: I assumed you mean something like:
   `To ${message} you need to first open...`

That tends to be problematic.  I agree this is non-optimal, but I haven't found a better way, except to restrict the message to be more generic. I think this is good enough, though.  Thx.

PR #573, opened 2018-10-19
Support additional Docker build and run customization

philliphoff:
Allows users to further customize Docker images and containers used during debugging.  

The following build customizations were added:

 - **`args`:** Set or override build arguments (i.e. `ARG`) declared in the `Dockerfile`
 - **`labels`:** Dynamically apply labels to the image

The following run customizations were added:

 - **`env`:** Set or override environment variables applied to the container
 - **`envFiles`:** Apply sets of environment variables, defined in arbitrary files, to the container
 - **`labels`:** Dynamically apply labels to the container

This brings the Docker debugging customizations closer to parity with VS.

> NOTE: This PR adds a dependency on [deep-equal](https://github.com/substack/node-deep-equal)

### DIFF
--- a/debugging/coreclr/commandLineBuilder.ts
+++ b/debugging/coreclr/commandLineBuilder.ts
@@ -6,9 +6,7 @@ export class CommandLineBuilder {
     public static create(...args: (undefined | string | CommandLineArgFactory)[]): CommandLineBuilder {
         let builder = new CommandLineBuilder();
 
-        for (let i = 0; i < args.length; i++) {
-            const arg = args[i];
-
+        for (let arg of args) {
             if (arg) {
                 if (typeof arg === 'string') {
                     builder = builder.withArg(arg);
@@ -29,7 +27,7 @@ export class CommandLineBuilder {
         return this.withArgFactory(() => arg);
     }
 
-    public withArrayArgs<T>(name: string, values: T[] | undefined, formatter?: (T) => string): CommandLineBuilder {
+    public withArrayArgs<T>(name: string, values: T[] | undefined, formatter?: (value: T) => string): CommandLineBuilder {
         formatter = formatter || ((value: T) => value.toString());
 
         return this.withArgFactory(() => values ? values.map(value => `${name} "${formatter(value)}"`).join(' ') : undefined);
@@ -43,7 +41,7 @@ export class CommandLineBuilder {
         return this;
     }
 
-    public withFlagArg(name: string, value: boolean | undefined) {
+    public withFlagArg(name: string, value: boolean | undefined): CommandLineBuilder {
         return this.withArgFactory(() => value ? name : undefined);
     }
 
@@ -65,7 +63,7 @@ export class CommandLineBuilder {
         return this.withArgFactory(() => value ? `${name} "${value}"` : undefined);
     }
 
-    public withQuotedArg(value: string | undefined) {
+    public withQuotedArg(value: string | undefined): CommandLineBuilder {
         return this.withArgFactory(() => value ? `"${value}"` : undefined);
     }
 }
--- a/debugging/coreclr/commandLineBuilder.ts
+++ b/debugging/coreclr/commandLineBuilder.ts
@@ -1,0 +1,71 @@
+export type CommandLineArgFactory = () => (string | undefined);
+
+export class CommandLineBuilder {
+    private readonly args: CommandLineArgFactory[] = [];
+
+    public static create(...args: (undefined | string | CommandLineArgFactory)[]): CommandLineBuilder {
+        let builder = new CommandLineBuilder();
+
+        for (let i = 0; i < args.length; i++) {
+            const arg = args[i];
+
+            if (arg) {
+                if (typeof arg === 'string') {
+                    builder = builder.withArg(arg);
+                } else {
+                    builder = builder.withArgFactory(arg);
+                }
+            }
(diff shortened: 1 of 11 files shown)


debugging/coreclr/dockerClient.ts
```diff
@@ -64,6 +69,14 @@ export class CliDockerClient implements DockerClient {
             command += ` -f ${options.dockerfile}`;
         }
 
+        if (options && options.args) {
+            command += Object.keys(options.args).map(arg => ` --build-arg "${arg}=${options.args[arg]}"`).join('');
```
EricJizbaMSFT: There's gotta be a cleaner way to add all of these arguments. Maybe instead of appending them to the command as you go, you do something like this:

```typescript
let command = 'docker build --rm';
let commandArgs: string[] = [];
options = options || {}; // so that you don't have to check `if (options)` every time

commandArgs.push(...formatKeyValArg('label', options.labels));

command += commandArgs.join(' ');

function formatKeyValArg(argName: string, values?: { [key: string]: string } | undefined): string[] {
    if (values) {
        return Object.keys(values).map(key => `--${argName} "${key}=${values[key]}"`);
    }
}
```

I'm mainly suggesting this to avoid a possible bug due to a missing/misplaced quote or space.

debugging/coreclr/dockerManager.ts
```diff
@@ -283,6 +276,61 @@ export class DefaultDockerManager implements DockerManager {
...
     private static readonly ProgramFilesEnvironmentVariable: string = 'ProgramFiles';
 
+    private static compareProperty<T, U>(obj1: T | undefined, obj2: T | undefined, getter: (obj: T) => (U | undefined)): boolean {
+        const prop1 = obj1 ? getter(obj1) : undefined;
+        const prop2 = obj2 ? getter(obj2) : undefined;
+
+        return prop1 === prop2;
+    }
+
+    private static compareDictionary<T>(obj1: T | undefined, obj2: T | undefined, getter: (obj: T) => ({ [key: string]: string } | undefined)): boolean {
```
EricJizbaMSFT: We should be able to find an existing way in nodejs to compare objects rather than implementing your own. For example [deep-equal](https://www.npmjs.com/package/deep-equal) is probably a good option

StephenWeatherford:
@philliphoff  Have you considered updating the testcases?

debugging/coreclr/dockerManager.ts
philliphoff: I looked at `deep-equal`, but it didn't quite have the semantics I was looking for (less absolute equality as "effective" equality in the context of the Docker commands.  For example, when specifying an array of custom tag names for an image, for the purposes of deciding whether to reuse an image for debugging, the order or those tags is irrelevant.

debugging/coreclr/dockerManager.ts
```diff
@@ -283,6 +276,61 @@ export class DefaultDockerManager implements DockerManager {
...
+        if (keys1.length !== keys2.length) {
+            return false;
+        }
+
+        for (let i = 0; i < keys1.length; i++) {
+            if (keys1[i] !== keys2[i]) {
+                return false;
+            }
+
+            if (dict1[i] !== dict2[i]) {
```
EricJizbaMSFT: I don't understand how this check is working. How are you using an index on a dictionary instead of the key? Even if the index worked - wouldn't it be out of order since you sorted the keys but I don't think you sorted the dict?

debugging/coreclr/dockerManager.ts
EricJizbaMSFT: Fair enough about array equality, but I think you should still use `deep-equal` to replace `compareDictionary` since dictionaries have no concept of order (right?). Aka I'm fine if you don't call it on the options itself, but you could at least call it on 'args' and 'labels'

debugging/coreclr/dockerManager.ts
EricJizbaMSFT: If this is actually a bug then I'm totally using it as more reason to use deep-equal :)

debugging/coreclr/commandLineBuilder.ts
```diff
@@ -0,0 +1,71 @@
+export type CommandLineArgFactory = () => (string | undefined);
```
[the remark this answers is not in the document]
philliphoff: Added.

debugging/coreclr/dockerManager.ts
philliphoff: Ach, you're right!  That's what I get for not writing tests for this comparison of options (which I've now done).

debugging/coreclr/dockerManager.ts
philliphoff: Happy to switch to `deep-equal` for this purpose; what's the ceremony around pulling in additional 3rd-party dependencies?

StephenWeatherford:
> what's the typical granularity for individual tests? Would it be better to try to work the various customizations into the main scenario tests, create an additional test that adds the various customizations, or create fine-grained tests for each customization?

I'd suggest optimizing on making it easiest for the testers to understand and perform. Many of our tests have a "setup" section that other tests reference.  But you could put them into the main scenario if you think that makes sense.

As for how many tests - we attempt to optimize bang for the buck.  Also, mark a test as P2 if it's long and only needs to be done infrequently or before a release. Thx.

PR #578, opened 2018-10-22
Update opn to 5.4.0

tsalmela:


EricJizbaMSFT:
@tsalmela could you provide a little more detail about why this is necessary? Was there a change in 5.4 that you're specifically after? 5.4 should automatically be picked up by `^5.2.0` so we don't normally update each package version unless there's some specific reason

tsalmela:
> @tsalmela could you provide a little more detail about why this is necessary? Was there a change in 5.4 that you're specifically after? 5.4 should automatically be picked up by `^5.2.0` so we don't normally update each package version unless there's some specific reason

Not a real reason. Just went through old packages to see if there are newer versions available. Mainly just aiming for the t-shirt with a minimum effort.

PR #591, opened 2018-10-24
Fix lint task warning

StephenWeatherford:
Fix this error from vscode when opening project:

Error: the description can't be converted into a problem matcher:
{
    "base": "$tslint5",
    "fileLocation": "absolute"
}


StephenWeatherford:
Yes, I have it installed. Not sure exactly why this error, but auto-complete does show simply "$tslint5" as a possibility and stops the error.

Issue #606, opened 2018-11-01
Enable Azure Pipeline tests on Mac/Windows


EricJizbaMSFT:
- [x] Linux enabled
- [ ] Mac enabled
- [x] Windows enabled

Tests are passing on Linux, but not mac or windows so I disabled them. Mac is probably failing since I don't think the host agents have "docker" installed by default. Not sure about the Windows failures.

PR #607, opened 2018-11-01
Switch from travis ci to azure pipelines

EricJizbaMSFT:
Mostly copied from [functions](https://github.com/Microsoft/vscode-azurefunctions) repo. Unfortunately tests weren't working on Mac or Windows agents by default so I disabled them and filed an issue for someone (maybe me if I get bored? Maybe Stephen?) to address separately: https://github.com/Microsoft/vscode-docker/issues/606


grhm:
@EricJizbaMSFT Does the azure pipeline work now?  Only it looks like all PR builds are failing with permissions issues?  (Looking at the last 3 builds [here](https://dev.azure.com/ms-azuretools/AzCode/_build?definitionId=8))

EricJizbaMSFT:
@grhm yeah I'll look into it. Builds are working from this repo - just not from other forks

grhm:
Did you want me to raise an issue for it, or is my comment here enough?  Not sure of the etiquette here...

EricJizbaMSFT:
In terms of etiquette on issues vs comments: it depends. If you want something fixed then an issue is generally better because it's harder for us to accidentally lose/forget. In this case an issue isn't necessary because I already have a PR out: https://github.com/Microsoft/vscode-docker/pull/610

PR #609, opened 2018-11-05
Different icons for unhealthy containers

grhm:
Added the ability to have the icons change from green to yellow if a container is running but is in an "unhealthy" state.

### DIFF
--- a/explorer/models/rootNode.ts
+++ b/explorer/models/rootNode.ts
@@ -187,7 +187,8 @@ export class RootNode extends NodeBase {
                             // can't do a full object compare because "Status" keeps changing for running containers
                             if (ctr.Id === cont.Id &&
                                 ctr.Image === cont.Image &&
-                                ctr.State === cont.State) {
+                                ctr.State === cont.State &&
+                                ctr.Status.includes('(unhealthy)') === cont.Status.includes('(unhealthy)')) {
                                 found = true;
                                 break;
                             }
--- a/explorer/models/rootNode.ts
+++ b/explorer/models/rootNode.ts
@@ -152,6 +152,10 @@ export class RootNode extends NodeBase {
         return imageNodes;
     }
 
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT:
@grhm I merged my fix for PRs from forks. I tried just doing a "rebuild", but that didn't pick up my changes. Can you merge with master and push again? Hopefully that will trigger the right build

grhm:
Just merged and pushed and it looks like the build is now working.  Thanks.

StephenWeatherford:
@grhm Thanks for adding this!  Do you know how I can get a container into an unhealthy state so I can test?  Thx.

grhm:
@StephenWeatherford I've been using the following Dockerfile

```
FROM nginx:stable  
HEALTHCHECK --interval=3s CMD [ "/bin/false" ]
```

Any base image will do, I just happened to have nginx to hand.
The container will be reported as unhealthy after 3 failed healthchecks.  I've added the 3 second interval as the default is 30s, so this should go unhealthy after 9 seconds instead on 90.

grhm:
OK.  I've had a dig and found that dockerode is using the Docker API (from https://www.npmjs.com/package/dockerode#usage) and that said API *doesn't* present the health in its return (see https://docs.docker.com/engine/api/v1.37/#operation/ContainerInspect).

So, I'm going to add a string comparison check into the autoRefreshContainers() function

grhm:
Rebased my changes to fix merge conflicts as another PR was merged first.

grhm:
Fixes #615

StephenWeatherford:
Looking good, thanks! I only have one request:  Could you please pull out a function rather than repeating the code in multiple places?  

function isContainerUnhealthy(container: ContainerDesc): boolean {
    return container.Status.includes('(unhealthy)');
}

StephenWeatherford:
FYI, we're releasing in a few days, so if you can get this fixed, it will be in the next release.

PR #618, opened 2018-11-13
make the launch.json generation leaner

### LINKED ISSUE #612: Simplify generated launch.json
Hey,

VSCode dev here. This milestone I am looking into simplifing generated `launch.json` for various extensions https://github.com/Microsoft/vscode/issues/62851

The `launch.json` that docker generates is the following:
```
{
	// Use IntelliSense to learn about possible attributes.
	// Hover to view descriptions of existing attributes.
	// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
	"version": "0.2.0",
	"configurations": [
		{
			"name": "Docker: Attach to Node",
			"type": "node",
			"request": "attach",
			"port": 9229,
			"address": "localhost",
			"localRoot": "${workspaceFolder}",
			"remoteRoot": "/usr/src/app",
			"protocol": "inspector"
		}
	]
}
```

Here are my suggestions on how to simplify this:
- [ ] Remove the protocol field, by default it is auto which should work fine
- [ ] Remove the `address` and `localRoot`. Also contribute a [`DebugConifgurationProvider`](https://github.com/Microsoft/vscode/blob/e839e4560c19ce53c717d6f3078fd01e35c737a2/src/vs/vscode.d.ts#L8190) which will just resovle every launch configuration by appending the localRoot and address if they are not already present
- [ ] Change the default for localRoot to be `${workspaceFolder}`, not `null` as it is now (auto complete it in json)

By removing these unnecesry fields we will make it look simpler for the user, making him see only what is important to him.
Also what is the field that the user has to configure here the most? I would think it is the port and the remoteRoot. Would you profit from some quick pick which would ask the user which port he is using? Similar to your create a `dockerfile` experience?

All of this should be pretty straightforward. Please let me know if you need help.

isidorn:
fixes #612

This PR simplifies the generated `launch.json` in the following ways:
* The snippets do not have the default fields any more, making them more concise. The missing fields should be nicely added by node
* When the initial configuraiton is provided we ask the user for the remoteRoot, making it clearer what is the attribute that users change the most

What still needs to be done imho:
* We should read the dockerfile, such that we see the value of the dockerfile WORKDIR and take that as the default, not the hardcoded `/usr/src/app`
* Read the dockerfile and see if the userconfigured a remote port, if he did than insert the port in the configuration. Currently we do not set any because the node debug will corectly pick the port based on node version
* Tune the wording of the input box, not sure if `Please enter your Docker remote root` is correct here, maybe WORKDIR instead of remote root.
* Simplify the docker .net launch configuraiton in a similar way

@StephenWeatherford let me know what you think. Thanks a lot
fyi @chrisdias

### DIFF
--- a/configureWorkspace/configDebugProvider.ts
+++ b/configureWorkspace/configDebugProvider.ts
@@ -9,20 +9,14 @@ import * as vscode from 'vscode';
 export class DockerDebugConfigProvider implements vscode.DebugConfigurationProvider {
 
     public provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): vscode.ProviderResult<vscode.DebugConfiguration[]> {
-
-        const config: vscode.DebugConfiguration = {
-            name: 'Docker: Attach to Node',
-            type: 'node',
-            request: 'attach',
-            port: 9229,
-            address: 'localhost',
-            localRoot: '\${workspaceFolder}',
-            remoteRoot: '/usr/src/app',
-            protocol: 'inspector'
-        };
-
-        return [config];
-
+        return vscode.window.showInputBox({ value: '/usr/src/app', prompt: 'Please enter your Docker remote root'}).then(remoteRoot => {
+            return [{
+                name: 'Docker: Attach to Node',
+                type: 'node',
+                request: 'attach',
+                remoteRoot
+            }];
+        });
     }
 
 }
--- a/configureWorkspace/configDebugProvider.ts
+++ b/configureWorkspace/configDebugProvider.ts
@@ -6,17 +6,17 @@
 'use strict';
 
 import * as vscode from 'vscode';
+import { ext } from '../extensionVariables';
+
 export class DockerDebugConfigProvider implements vscode.DebugConfigurationProvider {
 
-    public provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): vscode.ProviderResult<vscode.DebugConfiguration[]> {
-        return vscode.window.showInputBox({ value: '/usr/src/app', prompt: 'Please enter your Docker remote root'}).then(remoteRoot => {
-            return [{
-                name: 'Docker: Attach to Node',
-                type: 'node',
-                request: 'attach',
-                remoteRoot
-            }];
-        });
+    public async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
+        const remoteRoot = await ext.ui.showInputBox({ value: '/usr/src/app', prompt: 'Please enter your Docker remote root' });
+        return [{
+            name: 'Docker: Attach to Node',
+            type: 'node',
+            request: 'attach',
+            remoteRoot
+        }];
     }
-
 }
--- a/package.json
+++ b/package.json
@@ -287,11 +287,7 @@
               "type": "node",
               "request": "attach",
               "name": "Docker: Attach to Node",
-              "port": 9229,
-              "address": "localhost",
-              "localRoot": "^\"\\${workspaceFolder}\"",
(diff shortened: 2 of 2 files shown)


philliphoff:
The scaffolded launch configuration for Docker .NET Core debugging is fairly lean; it adds only the optional `dockerBuild` and `dockerRun` properties, which were an explicit ask when adding the feature.  All other properties are inferred by its `DebugConfigurationProvider` (DCP).

> We should read the dockerfile, such that we see the value of the dockerfile WORKDIR and take that as the default, not the hardcoded /usr/src/app

That's not a bad idea, though multi-stage Dockerfiles (such as scaffolded by .NET Core tooling) could complicate that (as you'd want the last `WORKDIR` specified by the most-descendent stage).  An alternative may be to use a DCP to identify the ultimate `WorkingDir` of the built image/container (via `docker inspect`) and use that.  The Docker .NET Core DCP does this to identify the exposed HTTP port (if any), in order to know where to point the browser on debug.

That said, the `/app` convention is fairly embedded within the .NET Core tooling. The VS Code DCP was patterned after the VS equivalent, which also makes that assumption.

configureWorkspace/configDebugProvider.ts
```diff
@@ -9,20 +9,14 @@ import * as vscode from 'vscode';
...
-        return [config];
-
+        return vscode.window.showInputBox({ value: '/usr/src/app', prompt: 'Please enter your Docker remote root'}).then(remoteRoot => {
+            return [{
+                name: 'Docker: Attach to Node',
+                type: 'node',
+                request: 'attach',
+                remoteRoot
+            }];
+        });
```
PrashanthCorp: For any new code added, we tend to avoid thenable in favor of async-await. 

Debugging this code has seemed easier when the code is written as : 
```
let remoteRoot = await vscode.window.showInputBox({ value: '/usr/src/app', prompt: 'Please enter your Docker remote root'}); 
return [{
name: 'Docker: Attach to Node',
type: 'node',
request: 'attach',
remoteRoot
}]; 
```

You would need to make the function async however, so I'll defer to @StephenWeatherford  whether requiring this standard is worth it here. :)

StephenWeatherford: Yes, please do convert.  Also, please use ext.ui.showInputBox instead of vscode.window.showInputBox. Note that this will throw on user cancel, which I assume makes sense here?

isidorn: Thanks for the review and suggestions I have pushed a commit which takes care of this.

isidorn:
@philliphoff thanks for your reply.
Reading fields from the docker file when possible makes good sense to me. And if you get some prototpye of this to work let me know and I will be glad to try it out and give feedback. Note that we do not have to be right 100% of the time, my suggestion is just to present a good default to the user in QuickPick and he can always change it.

In the end I leave making this leaner and more simply to you since I do not know what are the biggest pain points that users face when setting up docker debugging. Let me know if there is something on the vscode side we can do to improve this.

Just for reference what is the scaffolded launch configuration for Docker .NET?

isidorn:
Is there something else that needs to be done on my end here?
Thanks

philliphoff:
@isidorn For reference, this is the scaffolded Docker .NET Core launch configuration:

```json
{
     "name": "Docker: Launch .NET Core (Preview)",
     "type": "docker-coreclr",
     "request": "launch",
     "preLaunchTask": "build",
     "dockerBuild": {},
     "dockerRun": {}
}
```

StephenWeatherford:
@philliphoff I'm assuming you're signing off on this PR...  Thx.

isidorn:
Great thanks. Should I click the merge button or will you, not sure what is the procedure here. Thanks!
Also thanks for the scaffolded configuration.

philliphoff:
@StephenWeatherford Are you ok with the changes?

StephenWeatherford:
@isidorn, @philliphoff Thanks!

PR #622, opened 2018-11-13
Change default bash command for attaching the shell in Linux container

### LINKED ISSUE #518: Fail to 'Attach Shell' for 'registry' with error
Repro Steps:
1. Run 'docker run registry'.
2. Navigate to 'Containers' node -> Right click the 'registry' container ->Select 'Attach Shell'.
3. Check the result.

Expect:
Succeed to execute 'Attach Shell' fro container 'registry'.

Actual:
Fail to Fail to 'Attach Shell' for 'registry' with error.
compare

PrashanthCorp:
Fixes #518

Powershell: 
image

CMD: 
image

 Bash on Mac: 
image

### DIFF
--- a/commands/open-shell-container.ts
+++ b/commands/open-shell-container.ts
@@ -16,9 +16,9 @@ function getEngineTypeShellCommands(engineType: DockerEngineType): string {
     const configOptions: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('docker');
     switch (engineType) {
         case DockerEngineType.Linux:
-            return configOptions.get('attachShellCommand.linuxContainer', '/bin/sh -c "[ -e /bin/bash ] && /bin/bash || /bin/sh"');
+            return configOptions.get('attachShellCommand.linuxContainer');
         case DockerEngineType.Windows:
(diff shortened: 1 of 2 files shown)


commands/open-shell-container.ts
```diff
@@ -17,7 +16,7 @@ function getEngineTypeShellCommands(engineType: DockerEngineType): string {
     const configOptions: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration('docker');
     switch (engineType) {
         case DockerEngineType.Linux:
-            return configOptions.get('attachShellCommand.linuxContainer', '/bin/sh');
+            return configOptions.get('attachShellCommand.linuxContainer', '/bin/sh -c "[ -e /bin/bash ] && /bin/bash || /bin/sh"');
```
StephenWeatherford: This is totally unnecessary, could you remove the default here?  Thanks.

PrashanthCorp: I just realized that changing the setting to blank would still not pick up this default. Thanks for the heads up.

PrashanthCorp: Should I remove the `powershell`  default too below?

PR #627, opened 2018-11-16
Add support for certificates

### LINKED ISSUE #483: Unable to connect do a https repository with enterprise CA certificate.
I'm trying to connect to a private repository that uses a https certificate signed by our enterprise CA. My docker does to login command correctly with this repository and and CA certificate is in windows certificate store. One more informations this repository works fine and it's not in the insecure registry settings in my docker.

Action: vscode-docker.connectCustomRegistry
Error type: RequestError
Error Message: Error: unable to verify the first certificate

Version: 0.2.0
OS: win32

StephenWeatherford:
Fixes #483


utils/getTrustedCertificates.ts
```diff
@@ -0,0 +1,107 @@
...
+    // tslint:disable-next-line:no-function-expression
+    return callWithTelemetryAndErrorHandling('docker.certificates', async function (this: IActionContext): Promise<(string | Buffer)[]> {
+        this.suppressTelemetry = true;
+
+        let systemCerts = getCertificatesFromSystem();
+
+        let certificatePaths = vscode.workspace.getConfiguration('docker').get<string[] | undefined>('certificates');
+        if (certificatePaths === undefined) {
+            if (isLinux()) {
+                certificatePaths = ['/etc/ssl/certs/ca-certificates', '/etc/openssl/certs', '/etc/pki/tls/certs', '/usr/local/share/certs'];
```
EricJizbaMSFT: How'd you pick this list? Maybe add a comment with link to docs

utils/getTrustedCertificates.ts
```diff
@@ -0,0 +1,107 @@
...
+let _systemCertificates: (string | Buffer)[] | undefined;
+
+export async function getTrustedCertificates(): Promise<(string | Buffer)[]> {
+    // tslint:disable-next-line:no-function-expression
+    return callWithTelemetryAndErrorHandling('docker.certificates', async function (this: IActionContext): Promise<(string | Buffer)[]> {
+        this.suppressTelemetry = true;
+
+        let systemCerts = getCertificatesFromSystem();
+
+        let certificatePaths = vscode.workspace.getConfiguration('docker').get<string[] | undefined>('certificates');
```
EricJizbaMSFT: If this is undefined - can't we default to the behavior before your PR? Why do we need to specify any ca stuff?

utils/getTrustedCertificates.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (certificatePaths === undefined) {
+            if (isLinux()) {
+                certificatePaths = ['/etc/ssl/certs/ca-certificates', '/etc/openssl/certs', '/etc/pki/tls/certs', '/usr/local/share/certs'];
+            } else {
+                certificatePaths = [];
+            }
+        }
+        let folderCerts = await getCertificatesFromPaths(certificatePaths);
+
+        this.properties.fromSystem = String(systemCerts.length);
```
EricJizbaMSFT: nit: `fromSystemCount`?

utils/getTrustedCertificates.ts
StephenWeatherford: I'd like to but there is no official list, which is why I added a configuration option.

utils/getTrustedCertificates.ts
StephenWeatherford: That would mean Linux wouldn't work out of the box.  This way it should work most of the time.

utils/getTrustedCertificates.ts
StephenWeatherford: Seems unnecessary but done.

utils/getTrustedCertificates.ts
StephenWeatherford: Not sure why we would want that.

utils/getTrustedCertificates.ts
StephenWeatherford: Added this:

// These are some of the most common paths on Linux, but there is no official standard

utils/getTrustedCertificates.ts
EricJizbaMSFT: Never mind - I thought it was already reading from system store before your PR

utils/getTrustedCertificates.ts
EricJizbaMSFT: Mainly concerned that it looked like we were sending cert info in telemetry. Count makes it seem obvious that it's just a non-important number

utils/getTrustedCertificates.ts
EricJizbaMSFT: Talked offline. I'm fine with this since they're analagous to the "system cert store" on linux

gulpfile.js
```diff
@@ -15,7 +15,7 @@ gulp.task('test', ['install-azure-account'], (cb) => {
     const env = process.env;
     env.DEBUGTELEMETRY = 1;
     env.CODE_TESTS_WORKSPACE = './test/test.code-workspace';
-    env.MOCHA_reporter = 'mocha-junit-reporter';
+    //env.MOCHA_reporter = 'mocha-junit-reporter';
```
EricJizbaMSFT: If you're going to comment this out then you have to set `MOCHA_reporter` somewhere else otherwise the CI builds will never fail for tests

PR #628, opened 2018-11-19
Support multiple reporters

StephenWeatherford:
Thoughts?  This makes it possible to have multiple reporters (currently we don't get any feedback from tests when run locally, and you have to know to click on 'Tests' to see feedback on Azure Pipelines (which I didn't at first).  Also allows adding JSON to MOCHA_ vars (since an object is required to configure the multiple reporters).

The simpler option would be to simply define the reporter as a variable in the pipeline.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -15,7 +15,6 @@ gulp.task('test', ['install-azure-account'], (cb) => {
     const env = process.env;
     env.DEBUGTELEMETRY = 1;
     env.CODE_TESTS_WORKSPACE = './test/test.code-workspace';
-    env.MOCHA_FILE = path.join(__dirname, 'test-results.xml');
     const cmd = cp.spawn('node', ['./node_modules/vscode/bin/test'], { stdio: 'inherit', env });
     cmd.on('close', (code) => {
         cb(code);
--- a/test/index.ts
+++ b/test/index.ts
@@ -3,6 +3,8 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as path from 'path';
+
 //
 // PLEASE DO NOT MODIFY / DELETE UNLESS YOU KNOW WHAT YOU ARE DOING
 //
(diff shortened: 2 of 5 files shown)


test/index.ts
```diff
@@ -43,12 +43,23 @@ for (let envVar of Object.keys(environmentVariables)) {
...
+            if (typeof value === 'string' && !isNaN(parseInt(value, undefined))) {
+                value = parseInt(value, undefined);
+            }
         }
         options[option] = value;
     }
 }
+
+options.reporterOptions = {
+    "reporterEnabled": "mocha-junit-reporter, spec"
```
EricJizbaMSFT: Did you add this here just for the sake of debugging? Doesn't it overwrite what's in the env?

But actually I like this better than trying to read it from the env. I think it makes sense to read things like `MOCHA_grep` or `MOCHA_enableTimeouts` from the env since those change somewhat often, but I don't think the reporter stuff will change since you're automatically writing results to both places now. Seems simpler to just set it one place (here).

StephenWeatherford: Had it there originally.  I'm fine with that.

test/index.ts
```diff
@@ -43,12 +43,23 @@ for (let envVar of Object.keys(environmentVariables)) {
     if (match) {
         let [, option] = match;
         let value: string | number = environmentVariables[envVar];
-        if (typeof value === 'string' && !isNaN(parseInt(value, undefined))) {
-            value = parseInt(value, undefined);
+
+        try {
+            value = JSON.parse(value);
+        } catch
```
PrashanthCorp: It took me a bit of time to realize this catch block does not intend to use anything from the thrown error. Is this syntax common?

StephenWeatherford: It's a little unusual having the extra code in the catch. There's no tryParse so I just have to try it.

It's going away, though.

gulpfile.js
```diff
@@ -15,7 +15,6 @@ gulp.task('test', ['install-azure-account'], (cb) => {
     const env = process.env;
     env.DEBUGTELEMETRY = 1;
     env.CODE_TESTS_WORKSPACE = './test/test.code-workspace';
-    //env.MOCHA_reporter = 'mocha-junit-reporter';
     env.MOCHA_FILE = path.join(__dirname, 'test-results.xml');
```
EricJizbaMSFT: `MOCHA_FILE` is directly related to `mocha-junit-reporter` so might make sense to move that over to index.ts as well

StephenWeatherford: done

PR #631, opened 2018-11-19
Make sure Node.js uses internal certificates if neither certificates …

### LINKED ISSUE #629: Opening Azure registries fails if new docker.useCertificateStore setting not specified

StephenWeatherford:
…option specified

Fixes #629 

Apparently if we pass in an empty array to agentOptions.ca, Node.js will not use it hard-coded internal list of certificates, which breaks exploring ACR.

### DIFF
--- a/utils/getTrustedCertificates.ts
+++ b/utils/getTrustedCertificates.ts
@@ -16,16 +16,27 @@ let _systemCertificates: (string | Buffer)[] | undefined;
 
 export async function getTrustedCertificates(): Promise<(string | Buffer)[]> {
     // tslint:disable-next-line:no-function-expression
-    return callWithTelemetryAndErrorHandling('docker.certificates', async function (this: IActionContext): Promise<(string | Buffer)[]> {
+    return callWithTelemetryAndErrorHandling('docker.certificates', async function (this: IActionContext): Promise<(string | Buffer)[] | undefined> {
         this.suppressTelemetry = true;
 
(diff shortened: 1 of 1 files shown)


utils/getTrustedCertificates.ts
```diff
@@ -16,16 +16,27 @@ let _systemCertificates: (string | Buffer)[] | undefined;
```
EricJizbaMSFT: So if I specify one file in `docker.certificatePaths` does that mean I only get that one file and that's it? I would expect it to be _added_ to the default "hard-coded internal list of certificates"

utils/getTrustedCertificates.ts
```diff
@@ -16,16 +16,27 @@ let _systemCertificates: (string | Buffer)[] | undefined;
 
 export async function getTrustedCertificates(): Promise<(string | Buffer)[]> {
     // tslint:disable-next-line:no-function-expression
```
EricJizbaMSFT: Would it be accurate to say that you expect most Windows/Mac users to use `docker.useCertificateStore` and only Linux users to use `docker.certificatePaths`? Just trying to understand the use-case for each and how common each case is

utils/getTrustedCertificates.ts
EricJizbaMSFT: I think we can mitigate this by improving the default "it just works" behavior of the settings. It doesn't feel super related to this PR, so I filed a separate issue to discuss: https://github.com/Microsoft/vscode-docker/issues/633

PR #638, opened 2018-11-20
Saw/import certificates

### LINKED ISSUE #633: Docker certificates settings should allow me to specify "true" and "just work"
The settings "useCertificateStore" and "certificatePaths" feel pretty confusing to me. One only works on mac/windows and the other only works on Linux. Plus if I specify a single path on Linux I don't get any "default" CAs.

I think it would be great if we have a single setting that the user can turn off or on and get almost all default "good" behavior. Here is my suggested schema for that setting (hacked together - not meant to be blindly copy/pasted haha):
```
"docker.certificates": {
  "oneOf": [
    {
      "type": "boolean"
    },
    {
      "type": "object",
      "properties": {
        "certificatePaths": {
          "type": "array",
          "items": {
            "type": "string"
          },
          "default": [
            "/etc/ssl/certs/ca-certificates",
            "/etc/openssl/certs",
            "/etc/pki/tls/certs",
            "/usr/local/share/certs"
          ],
          "description": "Paths to files or folders containing certificates to import. For Linux, the correct path to pick up system-wide certificates will depend on the distribution."
        },
        "useCertificateStore": {
          "type": "boolean",
          "default": true,
          "description": "On Mac and Windows, indicates whether to automatically import certificates from the system certificate store. Ignored on Linux."
        }
      }
    }
  ],
  "default": false,
  "description": "Something something about set to 'true' to get default cert behavior or an object to have more fine-tuned control"
},
```

StephenWeatherford:
Fixes #633

### DIFF
--- a/utils/getTrustedCertificates.ts
+++ b/utils/getTrustedCertificates.ts
@@ -14,29 +14,49 @@ import { isLinux, isMac, isWindows } from '../helpers/osVersion';
 
 let _systemCertificates: (string | Buffer)[] | undefined;
 
+type ImportCertificatesSetting = boolean | {
+    useCertificateStore: boolean,
+    certificatePaths: string[]
+};
+
+const defaultCertificatePaths: string[] = [
+    "/etc/ssl/certs/ca-certificates",
+    "/etc/openssl/certs",
+    "/etc/pki/tls/certs",
+    "/usr/local/share/certs"
+];
+
 export async function getTrustedCertificates(): Promise<(string | Buffer)[]> {
     // tslint:disable-next-line:no-function-expression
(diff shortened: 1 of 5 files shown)


utils/getTrustedCertificates.ts
```diff
@@ -14,29 +14,49 @@ import { isLinux, isMac, isWindows } from '../helpers/osVersion';
...
+    useCertificateStore: boolean,
+    certificatePaths: string[]
+};
+
+const defaultCertificatePaths: string[] = [
+    "/etc/ssl/certs/ca-certificates",
+    "/etc/openssl/certs",
+    "/etc/pki/tls/certs",
+    "/usr/local/share/certs"
+];
```
PrashanthCorp: Repeat of the defaults from package.json?

Issue #570, opened 2018-10-18
Add template for C++


robotdad:
It'd be nice if there was a template for creating a C++ container. I've got a fork with that available now. I added a test based on Ruby but it isn't clear to me how to run that.
https://github.com/robotdad/vscode-docker

You can test it by cloning this and adding the cpp dockerfile to it.
https://github.com/robotdad/helloworld

StephenWeatherford:
Fixed by #644

PR #644, opened 2018-11-22
Adding C++ template

robotdad:
Proposed change for #570


configureWorkspace/config-utils.ts
```diff
@@ -9,6 +9,18 @@ import { IAzureQuickPickItem, IAzureUserInput } from 'vscode-azureextensionui';
 import { ext } from "../extensionVariables";
 import { Platform, PlatformOS } from '../utils/platform';
 
+export type OS = 'Windows' | 'Linux';
```
StephenWeatherford: Please remove these and instead use the ones from platform.ts (where they were moved to)

robotdad: Thanks, done. The PlatformOS type from platform.ts was already being used so that shouldn't have still been there.

StephenWeatherford: This applies to the extra Platform enum as well...  Thx.

StephenWeatherford:
Regarding the lint error:
>  ERROR: (ordered-imports) D:/a/1/s/configureWorkspace/configure.ts[20, 1]: Import sources within a group must be alphabetized.

Looks alphabetized to me...?  But you can try this:

  npm run lint-fix

If it's still giving you problems, let me know and I'll take a look.

configureWorkspace/config-utils.ts
robotdad: D'oh! Fixed.

robotdad:
I threw a hello world in this repo to save you some typing on getting a C++ app to try this with.
https://github.com/robotdad/helloworld

StephenWeatherford:
Thanks!

StephenWeatherford:
Looks great, thanks!  Next release should be out very soon.

PR #653, opened 2018-11-27
Fix null ref show connection error during prune

### LINKED ISSUE #652: Null ref exception showing connection error in prune
&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: vscode-docker.system.prune
Error type: TypeError
Error Message: Cannot set property 'suppressErrorDisplay' of undefined

Version: 0.4.0
OS: win32

StephenWeatherford:
Fixes #652

### DIFF
--- a/commands/system-prune.ts
+++ b/commands/system-prune.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as assert from 'assert';
 import * as semver from 'semver';
 import vscode = require('vscode');
 import { IActionContext } from 'vscode-azureextensionui';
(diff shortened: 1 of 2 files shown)


commands/system-prune.ts
```diff
@@ -11,6 +12,8 @@ import { ext } from '../extensionVariables';
 import { docker } from './utils/docker-endpoint';
 
 export async function systemPrune(actionContext: IActionContext): Promise<void> {
```
EricJizbaMSFT: I prefer `registerCommand('vscode-docker.system.prune', systemPrune);` to keep dockerExtension.ts as simple as possible.

Which means you would only change this file:
```typescript
export async function systemPrune(this: IActionContext): Promise<void> {
    const actionContext = this; // <-- if you don't like referencing `this` directly
```

But no strong preference.

StephenWeatherford: I like the idea but don't want to deal with changing it now, but will keep it in mind in the future. Thx.

PR #654, opened 2018-11-27
Re-enable Windows/Mac tests

### LINKED ISSUE #606: Enable Azure Pipeline tests on Mac/Windows

StephenWeatherford:
Fixes #606

### DIFF
--- a/.azure-pipelines/main.yml
+++ b/.azure-pipelines/main.yml
@@ -5,8 +5,7 @@ jobs:
   steps:
   - template: common/build.yml
   - template: common/lint.yml
-  # https://github.com/Microsoft/vscode-docker/issues/606
-  # - template: common/test.yml
+  - template: common/test.yml
 
 - job: Linux
   pool:
--- a/.azure-pipelines/main.yml
+++ b/.azure-pipelines/main.yml
@@ -20,7 +20,7 @@ jobs:
   pool:
     vmImage: macOS 10.13
   steps:
+    env:
   - template: common/build.yml
   - template: common/lint.yml
-  # https://github.com/Microsoft/vscode-docker/issues/606
-  # - template: common/test.yml
+  - template: macOs/test.yml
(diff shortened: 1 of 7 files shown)


test/customRegistries.test.ts
```diff
@@ -3,18 +3,25 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+debugger;
```
EricJizbaMSFT: ?

StephenWeatherford: Fixed

StephenWeatherford:
I'll fix Mac in another PR.

.azure-pipelines/main.yml
```diff
@@ -4,9 +4,9 @@ jobs:
     vmImage: VS2017-Win2016
   steps:
   - template: common/build.yml
+  - template: common/webpack-prod.yml # This is required since we're not running publish
```
EricJizbaMSFT: I would prefer if 'webpack-prod' were a part of `common/build.yml` for two reasons:
1. Then it's run for every OS and you don't need separate windows/mac steps
1. I think it works better as a separate step than as a part of `vscode:prepublish` (as currently specified in package.json). I've had issues where commands run as a part of `prepublish` don't output errors

StephenWeatherford: Been wanting to talk to you in person about this.  There's kind of a conflict between what's needed for command-line to work and the Azure pipelines, since they don't share dependency management.  E.g. I think "vsce package" on command-line for a dev should automatically run build and webpack (and vscode:prepublish is the standard mechanism for that), but then if I also have a separate step for it in Azure, it will get run twice, increasing the build times.

StephenWeatherford: At the moment I've been more interested in getting it working, but I do want it optimized.

PR #675, opened 2018-12-05
Upgrade to gulp 4.0.0

StephenWeatherford:


gulpfile.js
```diff
@@ -14,31 +14,21 @@ const glob = require('glob');
...
-    });
+    return await spawn(path.join(__dirname, './node_modules/.bin/webpack'), ['--mode', 'production'], { stdio: 'inherit', env });
 });
 
 /**
  * Installs the azure account extension before running tests (otherwise our extension would fail to activate)
  * NOTE: The version isn't super important since we don't actually use the account extension in tests
  */
-gulp.task('install-azure-account', () => {
+gulp.task('install-azure-account', async () => {
```
PrashanthCorp: Why the async?

gulpfile.js
```diff
@@ -57,7 +47,14 @@ gulp.task('install-azure-account', () => {
...
 
-function spawn(command, args, options, cb) {
+gulp.task('test', gulp.series('install-azure-account', async () => {
+    env.DEBUGTELEMETRY = 1;
+    env.CODE_TESTS_WORKSPACE = path.join(__dirname, 'test/test.code-workspace');
+    env.CODE_TESTS_PATH = path.join(__dirname, 'dist/test');
+    return spawn('node', ['./node_modules/vscode/bin/test'], { stdio: 'inherit', env });
+}));
+
+function spawn(command, args, options) {
```
PrashanthCorp: Is the return type a promise? I presume that given the use of async in the webpack tasks. No a strong preference - annotating the return type might help here

gulpfile.js
```diff
@@ -14,31 +14,21 @@ const glob = require('glob');
 
 const env = process.env;
 
-gulp.task('webpack-dev', (cb) => {
+gulp.task('webpack-dev', async () => {
     preWebpack();
-    spawn(path.join(__dirname, './node_modules/.bin/webpack'), ['--mode', 'development'], { stdio: 'inherit', env }, cb);
+    return await spawn(path.join(__dirname, './node_modules/.bin/webpack'), ['--mode', 'development'], { stdio: 'inherit', env });
```
EricJizbaMSFT: As an alternative you can do this all in javascript:
```
const webpack = require('webpack');
const webpackConfig = require('./webpack.config');

webpackConfig.mode = mode;
return new Promise((resolve, reject) => webpack(webpackConfig, (err, stats) => {
    if (err) {
        reject(err);
    } else {
        console.log(stats.toString());
        resolve();
    }
}));
```

But I'm not sure how best to set `max-old-space-size` in that case

gulpfile.js
StephenWeatherford: Removed the await, thanks.
Unfortunately this is JS, not TS, so no types (at least not in the friendly TS way).

gulpfile.js
StephenWeatherford: Interesting, might consider that later.

gulpfile.js
StephenWeatherford: I think I'd prefer to keep all tasks async or all non-async.  With async, I don't have to worry about returning Promise.resolve(), and I can use await.

EricJizbaMSFT: Actually I think @PrashanthCorp was right to question this. You _can_ return a Promise to signal completion, but [the value is ignored](https://github.com/gulpjs/gulp/blob/master/docs/getting-started/4-async-completion.md#returning-a-promise). Thus when we returned a `Stream` or `ChildProcess` it was ignored. I used [this branch](https://github.com/Microsoft/vscode-docker/commits/ej/test2) to test the theory (The commit `Test` broke the tests, but passed CI. The commit `Remove async` addressed the async thing and properly reported a failure)
    screen shot 2018-12-06 at 11 49 41 am

Anyways I'll fix this in an upcoming PR

PrashanthCorp:
I remember seeing output being ignored, but thought it was just me. Thanks for looking into this. 😅
________________________________

PR #685, opened 2018-12-06
Clean up gulp/pipelines and re-enable mac tests

EricJizbaMSFT:
I learned a lot about gulp 4 while working on [this PR](https://github.com/Microsoft/vscode-azurefunctions/pull/853) and wanted to fix some things here.

Actual fixes:
1. Always use Node 8. I think this is the key thing that fixed webpack on macOS and allowed me to pretty easily re-enable the tests (since Stephen already did the "DOCKER_UNAVAILABLE" stuff) Fixes https://github.com/Microsoft/vscode-docker/issues/606
1. Don't use `async function` because it meant test failures weren't being reported. See https://github.com/Microsoft/vscode-docker/pull/675#discussion_r239591977

Other stuff that was just nice to have:
1. Consolidate into one `common\test.yml`
1. Use typescript for gulpfile (enabled with `ts-node`)
1. Use functions instead of `gulp.task` which is deprecated


gulpfile.ts
```diff
@@ -61,13 +63,11 @@ function spawn(command, args, options) {
...
     }
 
     return cp.spawn(command, args, options);
 }
 
-function preWebpack() {
-    // without this, webpack can run out of memory in some environments
-    env.NODE_OPTIONS = '--max-old-space-size=8192';
-}
+exports['webpack-dev'] = () => webpack('development');
```
StephenWeatherford: How about using export function test() etc?

EricJizbaMSFT: I'm mostly following the pattern in [the docs](https://github.com/gulpjs/gulp/blob/master/docs/getting-started/3-creating-tasks.md). I think they do this because tasks are often the combination of multiple functions using `gulp.series` and `gulp.parallel` rather than just a single function. Also you can't name a function with a hyphen and I don't feel like refactoring the `webpack-*` task names.

I'm not opposed, but not gonna do it right now.

PR #690, opened 2018-12-10
Allow more customization of docker run command

ismael-soriano:
## SUMMARY
Extend Docker Run configuration to add more customisations on the containers.

Now the `DockerRun` config on `launch.json` file can accept the following elements:

```json
"dockerRun": {
    "containerName": "my-container",
    "env": {
        "var1": "value1",
        "var2": "value2"
    },
    "envFiles": [
        "${workspaceFolder}/staging.env"
    ],
    "labels": {
        "label1": "value1",
        "label2": "value2"
    },
    "ports": [
        "80:80",
        "443:443"
    ],
    "externalHosts": [
        "some-hostname:some-ip",
        "some-other-hostname:some-other-ip"
    ],
    "volumes": [
        {
            "localPath": "path-on-host-machine",
            "containerPath": "path-inside-container",
            "permissions": "ro|rw"
        }
    ]
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,22 @@ Example run customization:
                 "labels": {
                     "label1": "value1",
                     "label2": "value2"
-                }
+                },
+                "ports": [
+                    "80:80",
+                    "443:443"
+                ],
+                "externalHosts": [
+                    "some-hostname:some-ip",
+                    "some-other-hostname:some-other-ip"
+                ],
+                "volumes": [
+                    {
+                        "localPath": "path-on-host-machine",
+                        "containerPath": "path-inside-container",
+                        "permissions": "ro|rw"
+                    }
+                ]
             }
         }
     ]
--- a/README.md
+++ b/README.md
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
 | `env` | Environment variables applied to the container. | None |
 | `envFiles` | Files of environment variables read in and applied to the container. Environment variables are specified one per line, in `<name>=<value>` format. | None |
 | `labels` | The set of labels added to the container. | `com.microsoft.created-by` = `visual-studio-code` |
+| `ports` | Ports that are going to be mapped on the host. | All ports exposed in the Dockerfile will get binded to a random port on the host machine |
+| `extraHosts` | Hosts to be added on the container `hosts` file for dns resolution. | None |
+| `volumes` | Volumes that are going to be mapped to the container. | None |
+
+# ports
+| Property | Description | Required |
+| --- | --- | --- |
+| `hostPort` | Port number to be binded on the host. | No |
+| `containerPort` | Port number of the container to be binded. | Yes |
+| `protocol` | Specific protocol for the binding (`tcp | udp`). If no protocol is specified it will bind both. | No |
+
+# volumes
+| Property | Description | Default |
+| --- | --- | --- |
+| `localPath` | Path on local machine that would be mapped. If the folder does not exist it will be created.  | None |
+| `containerPath` | Path where the volume will be mapped on the container. If the folder does not exist it will be created. | None |
+| `permissions` | Permissions for the container for the volume mapped (`rw | ro`). | `rw` |
 
 Example run customization:
 
@@ -253,12 +270,36 @@ Example run customization:
                     "label2": "value2"
                 },
                 "ports": [
-                    "80:80",
-                    "443:443"
+                    {
+                        "hostPort": 80,
+                        "containerPort": 80
+                    },
+                    {
+                        "containerPort": 443
+                    },
+                    {
+                        "containerPort": 6029,
+                        "protocol": "udp"
+                    },
+                    {
+                        "containerPort": 6029,
+                        "protocol": "tcp"
+                    },
+                    {
+                        "hostPort": 4562,
+                        "containerPort": 5837,
+                        "protocol": "tcp"
+                    }
                 ],
-                "externalHosts": [
-                    "some-hostname:some-ip",
-                    "some-other-hostname:some-other-ip"
+                "extraHosts": [
+                    {
+                        "hostname": "some-hostname",
+                        "ip": "some-ip"
+                    },
+                    {
+                        "hostname": "some-other-hostname",
+                        "ip": "some-other-ip"
+                    }
                 ],
                 "volumes": [
                     {
--- a/README.md
+++ b/README.md
@@ -228,23 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
 | `env` | Environment variables applied to the container. | None |
 | `envFiles` | Files of environment variables read in and applied to the container. Environment variables are specified one per line, in `<name>=<value>` format. | None |
 | `labels` | The set of labels added to the container. | `com.microsoft.created-by` = `visual-studio-code` |
-| `ports` | Ports that are going to be mapped on the host. | All ports exposed in the Dockerfile will get binded to a random port on the host machine |
-| `extraHosts` | Hosts to be added on the container `hosts` file for dns resolution. | None |
+| `ports` | Ports that are going to be mapped on the host. | All ports exposed by the Dockerfile will be bound to a random port on the host machine |
+| `extraHosts` | Hosts to be added to the container's `hosts` file for DNS resolution. | None |
 | `volumes` | Volumes that are going to be mapped to the container. | None |
 
 # ports
-| Property | Description | Required |
-| --- | --- | --- |
-| `hostPort` | Port number to be binded on the host. | No |
-| `containerPort` | Port number of the container to be binded. | Yes |
-| `protocol` | Specific protocol for the binding (`tcp | udp`). If no protocol is specified it will bind both. | No |
+| Property | Description | Required | Default |
+| --- | --- | --- | --- |
+| `hostPort` | Port number to be bound on the host. | No | None |
+| `containerPort` | Port number of the container to be bound. | Yes | None |
+| `protocol` | Specific protocol for the binding (`tcp | udp`). If no protocol is specified it will bind both. | No | None |
 
 # volumes
-| Property | Description | Default |
-| --- | --- | --- |
-| `localPath` | Path on local machine that would be mapped. If the folder does not exist it will be created.  | None |
-| `containerPath` | Path where the volume will be mapped on the container. If the folder does not exist it will be created. | None |
-| `permissions` | Permissions for the container for the volume mapped (`rw | ro`). | `rw` |
+| Property | Description | Required | Default |
+| --- | --- | --- | --- |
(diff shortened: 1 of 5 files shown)


StephenWeatherford:
+ @philliphoff

README.md
```diff
@@ -251,7 +251,22 @@ Example run customization:
                 "labels": {
                     "label1": "value1",
                     "label2": "value2"
-                }
+                },
+                "ports": [
```
philliphoff: The new properties should be documented in the table above.  They should also be added to the [JSON Schema for the configuration](https://github.com/Microsoft/vscode-docker/blob/34b406052e3224817421317e2de34b7e275b4145/package.json#L383) in the `package.json`.

README.md
```diff
@@ -251,7 +251,22 @@ Example run customization:
...
+                    "80:80",
```
philliphoff: I wonder if it'd be better to split this port mapping string (as well as the external hosts string below) into a separate object, like is done for volumes.  The string format is tied to the Docker CLI (which could change, though unlikely) and may not be the most straightforward to a new-to-Docker user.  It's also harder to (statically) validate and provide IntelliSense for.

Maybe something like this:

```json
"ports": [
  {
    "containerPort": 80,
    "localPort": 80
  },
  {
    "containerPort": 443,
    "localPort": 443
  }
]
```

It's admittedly more verbose, but the user's unlikely edit these configurations often.  It seems better to aim for clarity than terseness.

debugging/coreclr/dockerClient.ts
```diff
@@ -193,6 +195,8 @@ export class CliDockerClient implements DockerClient {
             .withArrayArgs('--env-file', options.envFiles)
             .withKeyValueArgs('--label', options.labels)
             .withArrayArgs('-v', options.volumes, volume => `${volume.localPath}:${volume.containerPath}${volume.permissions ? ':' + volume.permissions : ''}`)
+            .withArrayArgs('-p', options.ports)
```
philliphoff: While it may happen that adding an explicit mapping via `-p` will override an earlier implicit mapping via `-P`, it would probably be better to instead just switch between using `-p` and `-P` depending on whether ports have been explicitly mapped, or not.

debugging/coreclr/dockerManager.ts
```diff
@@ -217,7 +217,7 @@ export class DefaultDockerManager implements DockerManager {
                         env: options.env,
                         envFiles: options.envFiles,
                         labels: options.labels,
-                        volumes
+                        volumes: [...volumes, ...options.volumes]
```
philliphoff: Shouldn't the `options.externalHosts` and `options.ports` be passed as well?

README.md
ismael-soriano: I see your point here and I think the object way it's more accurate so it can provide intelliSense for it and abstract from the docker related form, so the user does not need to know hoy to write the exact string.

For the ports I'll provide an object like the following:

```json
"ports": [
  {
    "hostPort": 80,
    "containerPort": 80
  },
  {
    "containerPort": 443
  },
  {
    "containerPort": 6029,
    "protocol": "udp"
  },
  {
    "containerPort": 6029,
    "protocol": "tcp"
  }
]
```
Alongside this object I'll check if ports have been configured so it does not execute with the `-P` flag, and only ports indicated will be mapped.

With this, you can provide if you want to allocate an exact port for the container port associated or just want it to get randomized, also you can just expose a port from an specified protocol `(tcp|udp)`.

For the external hosts:

I'll change the name to `extraHosts` to keep consinstence with the docker compose label as here we are adding an "extra host" under the `--add-host` instruction.

The final object would look like this:

```json
"extraHosts": [
  {
    "hostname": "some-hostname",
    "ip": "some-ip"
  },
  {
    "hostname": "some-other-hostname",
    "ip": "some-other-ip"
  }
]
```

debugging/coreclr/dockerClient.ts
ismael-soriano: As i've said in a comment earlier, I'll check for the existence of port mapping configuration so the command only binds those.

debugging/coreclr/dockerManager.ts
ismael-soriano: Yes, I'll fix this point too.

package.json
```diff
@@ -407,6 +407,50 @@
                     "additionalProperties": {
                       "type": "string"
                     }
+                  },
+                  "ports": {
+                    "type": "array",
+                    "description": "Ports that are going to be mapped on the host.",
+                    "hostPort": {
```
philliphoff: I'm not sure this is [JSON Schema format](https://json-schema.org/).  I thought that it would look something like:

```json
{
  "ports": {
    "type": "array",
    "description": "...",
    "items": {
      "type": "object",
      "properties": {
        "hostPort": { ... },
        "containerPort": { ... },
        "protocol": { ... }
      },
      "required": [ "containerPort" ]
    }
  }
}
```

README.md
```diff
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
 | `env` | Environment variables applied to the container. | None |
 | `envFiles` | Files of environment variables read in and applied to the container. Environment variables are specified one per line, in `<name>=<value>` format. | None |
 | `labels` | The set of labels added to the container. | `com.microsoft.created-by` = `visual-studio-code` |
+| `ports` | Ports that are going to be mapped on the host. | All ports exposed in the Dockerfile will get binded to a random port on the host machine |
+| `extraHosts` | Hosts to be added on the container `hosts` file for dns resolution. | None |
+| `volumes` | Volumes that are going to be mapped to the container. | None |
+
+# ports
+| Property | Description | Required |
+| --- | --- | --- |
```
philliphoff: Any reason to use a `Required` column here for ports but a `Default` column for volumes? Could we use the same column in both tables?

README.md
```diff
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
 | `env` | Environment variables applied to the container. | None |
 | `envFiles` | Files of environment variables read in and applied to the container. Environment variables are specified one per line, in `<name>=<value>` format. | None |
 | `labels` | The set of labels added to the container. | `com.microsoft.created-by` = `visual-studio-code` |
+| `ports` | Ports that are going to be mapped on the host. | All ports exposed in the Dockerfile will get binded to a random port on the host machine |
```
philliphoff:
```suggestion
| `ports` | Ports that are going to be mapped on the host. | All ports exposed by the Dockerfile will be bound to a random port on the host machine |
```

README.md
```diff
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
...
+| `extraHosts` | Hosts to be added on the container `hosts` file for dns resolution. | None |
```
philliphoff:
```suggestion
| `extraHosts` | Hosts to be added to the container's `hosts` file for DNS resolution. | None |
```

README.md
```diff
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
...
+| `volumes` | Volumes that are going to be mapped to the container. | None |
+
+# ports
+| Property | Description | Required |
+| --- | --- | --- |
+| `hostPort` | Port number to be binded on the host. | No |
```
philliphoff:
```suggestion
| `hostPort` | Port number to be bound on the host. | No |
```

README.md
```diff
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
...
+| `containerPort` | Port number of the container to be binded. | Yes |
```
philliphoff:
```suggestion
| `containerPort` | Port number of the container to be bound. | Yes |
```

README.md
```diff
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
...
+| `protocol` | Specific protocol for the binding (`tcp | udp`). If no protocol is specified it will bind both. | No |
+
+# volumes
+| Property | Description | Default |
+| --- | --- | --- |
+| `localPath` | Path on local machine that would be mapped. If the folder does not exist it will be created.  | None |
+| `containerPath` | Path where the volume will be mapped on the container. If the folder does not exist it will be created. | None |
```
philliphoff:
```suggestion
| `containerPath` | Path where the volume will be mapped within the container. The folder will be created if it does not exist. | None |
```

README.md
```diff
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
...
+| Property | Description | Required |
+| --- | --- | --- |
+| `hostPort` | Port number to be binded on the host. | No |
+| `containerPort` | Port number of the container to be binded. | Yes |
+| `protocol` | Specific protocol for the binding (`tcp | udp`). If no protocol is specified it will bind both. | No |
+
+# volumes
+| Property | Description | Default |
+| --- | --- | --- |
+| `localPath` | Path on local machine that would be mapped. If the folder does not exist it will be created.  | None |
```
philliphoff:
```suggestion
| `localPath` | Path on local machine that will be mapped. The folder will be created if it does not exist.  | None |
```

README.md
```diff
@@ -228,6 +228,23 @@ Customize the Docker container run process by adding properties under the `docke
...
+| `containerPath` | Path where the volume will be mapped on the container. If the folder does not exist it will be created. | None |
+| `permissions` | Permissions for the container for the volume mapped (`rw | ro`). | `rw` |
```
philliphoff:
```suggestion
| `permissions` | Permissions for the container for the mapped volume, `rw` for read-write or `ro` for read-only. | `rw` |
```

debugging/coreclr/dockerClient.ts
```diff
@@ -187,12 +200,14 @@ export class CliDockerClient implements DockerClient {
         options = options || {};
 
         const command = CommandLineBuilder
-            .create('docker', 'run', '-dt', '-P')
+            .create('docker', 'run', '-dt', options.ports ? '-P' : '')
```
philliphoff: `options.ports` will be true even if empty; you'll need to test both for validity as well as length > 0.

Ideally this would be replaced with a new function in `CommandLineBuilder` that represents flags, something like:

`.withFlagArg('-P', options.ports === undefined || options.ports.length < 1)`

ismael-soriano: Thanks for pointing out the `.withFlagsArg()` method, I'll use it and remove the `-P` flag from the `.create()`.

README.md
ismael-soriano: Not a good reason at all, to make it consistent I'll apply the same colums for both tables like the following:

| Property | Description | Required | Default |
| --- | --- | --- | --- |

package.json
ismael-soriano: Sorry, the objects were malformed, I'll also apply enums for the protocol and permissions properties so it shows the user by intelliSense which elements to insert on those fields.

PR #731, opened 2019-01-08
Prepare for hotfix 0.5.1

StephenWeatherford:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vscode-docker",
-  "version": "0.5.0",
+  "version": "0.5.1",
```
EricJizbaMSFT: package-lock should be updated, too.

I usually just run an `npm install` to do it for me

StephenWeatherford: Thanks. Did manually so I know no dependencies changed for the hotfix.

Issue #698, opened 2018-12-11
Fail to inspect the created Docker container with error 'No such object...'


v-ruizh:
Extension Build: [20181209.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=1407)

Repro Steps:
1. Create a new .NET Core ASP.NET application.
2. Open application folder in VS Code.
3. Wait for and respond 'yes' when prompted to add required assets to build and debug.
4. Add Docker file to workspace. (Docker: Add Docker files to workspace->ASP.NET Core->Linux->80)
5. Add .NET Core debug launch task. (Debugging tab->Add configuration...->Docker: Launch .NET Core(Preview) )
6. Open the 'launch.json' file and edit .NET Core debugging configuration as below shown.
![image](https://user-images.githubusercontent.com/45614674/49786824-debf9380-fcd9-11e8-9f28-4535b1bd2612.png)
7. Open Dockerfile file and edit it as below shown.
![image](https://user-images.githubusercontent.com/45614674/49786875-07e02400-fcda-11e8-830f-db48056e871e.png)
8. Create a file 'Dockerfile.env' in the root directory and edit it.
![image](https://user-images.githubusercontent.com/45614674/49786946-34943b80-fcda-11e8-856b-25a31a0e670c.png)
9. Debug this application and copy the Docker container ID from Docker output pane.
10. Inspect the created Docker container with below command.
$ docker inspect containerID

Expect:
Inspect the Docker container successfully without any error.

Actual:
Fail to inspect the Docker container with error 'No such object...'.
![image](https://user-images.githubusercontent.com/45614674/49787143-cf8d1580-fcda-11e8-83c5-a2c2031e9187.png)

More Info:
Docker image can be inspected successfully.

StephenWeatherford:
@philliphoff Given that the error is coming from running the docker CLI itself, I'm doubting this is an actual issue.

philliphoff:
@v-ruizh Could you provide a bit more context about step 9?  What was shown in the Docker output window.  (I'd like to verify which IDs were shown.)

Also, note that the Docker container used for debugging is torn down when debugging ends (as it's just an empty shell of a container once debugging ends).  That may be why you can't inspect it after the fact.  It might be good to ensure that we log the killing of the container so the user knows it's happened.

v-ruizh:
@philliphoff Thanks for your guidance. The container can be inspected successfully after container started. 
[...]

PR #737, opened 2019-01-09
Log removal of containers.

philliphoff:


philliphoff:
Resolves #698

philliphoff:
>Question: Do these containers show up on the explorer pane too? (I assume not)

@PrashanthCorp: They do, in fact, show up in the explorer pane (automatically) and are removed (automatically) by whatever Docker event notification means used by the pane.

>Also, runContainer (line 203) forcefully removes a container (and another one is created). In the event the creation throws an error, do the users need to be notified about the container removal?

@PrashanthCorp: That *should* be an edge case.  That is, assuming that container removal generally succeeds after debugging ends, then there generally shouldn't be a container that needs to be removed prior to the start of debugging.

To me, it seems intuitive that anything that needs to be torn down or replaced prior to debugging is *actually* done prior to debugging.  That is, I wouldn't expect a container used for a prior debugging session to exist when I next debug the same application.  In that case, I don't think it's critical that we log everything.  It's less intuitive (as CTI confirmed) that the container would be torn down immediately after debugging.  (E.g. in the case of VS, we sometimes leave the container running between debugging sessions for performance reasons that don't necessarily apply to VS Code.)

PR #745, opened 2019-01-16
Fix user cancel showing messages

### LINKED ISSUE #718: Operation cancelled error canceling "container remove" when used on Containers node
&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: vscode-docker.container.remove
Error type: Error
Error Message: Operation cancelled.

Version: 0.5.0
OS: win32

StephenWeatherford:
Fixes #718

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -377,9 +377,9 @@
       }
     },
     "ajv-errors": {
-      "version": "1.0.0",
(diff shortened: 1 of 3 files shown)


StephenWeatherford:
Another option would be to set a flag inside cancel errors.  I doubt this will make much of a difference in performance, though.

Issue #733, opened 2019-01-08
"A dynamic link library (DLL) initialization routine failed" in insiders 1.31


StephenWeatherford:
Seeing these errors in telemetry, e.g.:

"A dynamic link library (DLL) initialization routine failed. \\?\C:\Users\<user>\.vscode-insiders\extensions\peterjausovec.vscode-docker-0.5.0\dist\node_modules\win-ca\lib\crypt32-x64.node" (win32)

Cannot find module './crypt32-x64' (win32)

Unknown Error (linux)

PR #761, opened 2019-01-29
Use win-ca fallback logic since nAPI isn't currently compatible with Electron

StephenWeatherford:
Not positive if this is the correct fix (it might be), or just a work-around, but it does fix the issue, and we need a fix before vscode ships next week.

#733 

Note that mac-ca works differently (spawns a mac exe) so does not have this same problem.

### DIFF
--- a/utils/getTrustedCertificates.ts
+++ b/utils/getTrustedCertificates.ts
@@ -119,6 +119,7 @@ function getCertificatesFromSystem(): (string | Buffer)[] {
 
         try {
             if (isWindows()) {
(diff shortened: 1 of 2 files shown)


StephenWeatherford:
> So is this a temporary workaround while waiting for a fix in VS Code or is this more "permanent"?

Turns out this is the permanent fix (at least for this version of win-ca). Updated comment.

StephenWeatherford:
https://www.npmjs.com/package/win-ca#availability

Issue #765, opened 2019-01-30
[.NET Core Debugging] Support Alpine-based images


philliphoff:
Alpine-based images require a variant of the .NET Core debugger and logic to determine which variant to use for a given image.

We should add such logic to enable debugging of Alpine-based images.

philliphoff:
Closed with #771.

PR #771, opened 2019-01-31
[.NET Core Debugging] Add support for Alpine images

philliphoff:
Adds support for debugging Linux images based on the "Alpine" set of .NET Core base images.

Alpine images require a different variant of the .NET Core debugger (`vsdbg`) than used for the "standard" images currently supported.  Determining which variant is necessary must be done *after* the container is running, by executing a script on the container which returns the image type.

Currently, the debugging provider volume mounts only the folder of a specific variant of debugger, and has the debugging engine use that variant.  This change updates the debugging provider such that containers volume mount the *root* folder for all cached debugging variants (as that must happen during container creation), then starts the container, then runs the script to determine the needed variant, and *then* has the debugging engine use that variant.  Because the volume mount is the root of the debugger cache, the container can use any of the cached variants as needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ The exact folder to use for certificatePaths on Linux will depend on the distrib
 
 ## Debugging .NET Core (Preview)
 
-> Note that Windows containers are **not** currently supported, only Linux containers.
+> Note that Windows containers are **not** currently supported, only Linux containers. However, both standard and Alpine .NET Core runtime base images are supported.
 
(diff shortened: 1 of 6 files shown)


philliphoff:
Resolves #765.

philliphoff:
@StephenWeatherford Thanks for the reminder; I've added a variant of the main scenario but with Alpine-based images.

StephenWeatherford:
Thx.

PR #791, opened 2019-02-12
Resolve local volume paths

### LINKED ISSUE #785: Support for ${workspaceFolder} in dockerRun/Volumes localPath and containerPath
**Issue**: Currently, there seems to be no support for ${workspaceFolder} when configuring volume mappings in localPath or containerPath.  The error is 

`docker: Error response from daemon: create ${workspaceFolder}/.devcerts: "${workspaceFolder}/.devcerts" includes invalid characters for a local volume name`

**Example**:
~~~~json
"dockerRun": {
                "volumes": [
                    {
                        "localPath": "${workspaceFolder}/.devcerts",
                        "containerPath": "/.devcerts",
                        "permissions": "rw"
                    }
                ]
            }
~~~~

**Use case**: when mapping a drive into an ASP.NET Core container containing certificates for development in order to enable SSL.  These certs could be stored in the local workspace and mapped this way, without having to hardcode the full path.  (The full path might vary depending on who checks out the source.)

**Plugin version**: plugin was built from master at commit `a45a9a6601cf7b17cc3f2cc2ad16db339ba13a49`

philliphoff:
Resolves the `localPath` debug configuration of volumes to allow the use of the `${workspaceFolder}` variable.  This allows users to specify workspace-relative volumes rather than absolute volumes, which helps users share launch configurations.

Resolves #785.

### DIFF
--- a/debugging/coreclr/dockerDebugConfigurationProvider.ts
+++ b/debugging/coreclr/dockerDebugConfigurationProvider.ts
@@ -186,8 +186,11 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
 
         const network = debugConfiguration && debugConfiguration.dockerRun && debugConfiguration.dockerRun.network;
         const ports = debugConfiguration && debugConfiguration.dockerRun && debugConfiguration.dockerRun.ports;
-        const volumes = debugConfiguration && debugConfiguration.dockerRun && debugConfiguration.dockerRun.volumes;
         const extraHosts = debugConfiguration && debugConfiguration.dockerRun && debugConfiguration.dockerRun.extraHosts;
+
+        const volumes = debugConfiguration && debugConfiguration.dockerRun && debugConfiguration.dockerRun.volumes
+            ? debugConfiguration.dockerRun.volumes.map(volume => ({ ...volume, localPath: DockerDebugConfigurationProvider.resolveFolderPath(volume.localPath, folder) }))
+            : [];
 
         return {
             containerName,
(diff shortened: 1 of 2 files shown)


debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -202,6 +202,12 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
         };
     }
 
+    private static inferVolumes(folder: WorkspaceFolder, debugConfiguration: DockerDebugConfiguration): DockerContainerVolume[] {
+        return debugConfiguration && debugConfiguration.dockerRun && debugConfiguration.dockerRun.volumes
+            ? debugConfiguration.dockerRun.volumes.map(volume => ({ ...volume, localPath: DockerDebugConfigurationProvider.resolveFolderPath(volume.localPath, folder) }))
```
PrashanthCorp: Just curious: Did you find the spread operator more readable than Javascript's `Object.assign(volume, {localPath: resolveFoderpath})`?

StephenWeatherford: Personally I find spread much easier to read.

philliphoff: Yes, `Object.assign()` has always felt clunky to me.  We used it extensively on other TS projects before the spread operator was available; I'm happy it finally made it into the language.

PrashanthCorp: Sounds good. We've used Object.assign() before, so I was wondering if we should use this convention moving forward.

PR #793, opened 2019-02-13
Fix nuget fallback volume mount for dotnet debugging

hacst:
Previously the NuGetFallbackFolder mounted into the container used for
debugging was hardcoded to /usr/local/share/dotnet/sdk/NuGetFallbackFolder
for all non-windows platforms. However this path is only the valid
default for OSX. On Linux /usr/share/dotnet/sdk/NuGetFallbackFolder is
the default installation location. This patch corrects that behavior.

### DIFF
--- a/debugging/coreclr/dockerManager.ts
+++ b/debugging/coreclr/dockerManager.ts
@@ -73,6 +73,7 @@ export interface DockerManager {
 }
 
 export const MacNuGetPackageFallbackFolderPath = '/usr/local/share/dotnet/sdk/NuGetFallbackFolder';
+const LinuxNuGetPackageFallbackFolderPath = '/usr/share/dotnet/sdk/NuGetFallbackFolder';
 
 function compareProperty<T, U>(obj1: T | undefined, obj2: T | undefined, getter: (obj: T) => (U | undefined)): boolean {
     const prop1 = obj1 ? getter(obj1) : undefined;
@@ -352,7 +353,8 @@ export class DefaultDockerManager implements DockerManager {
         }
 
         const nugetFallbackVolume: DockerContainerVolume = {
-            localPath: this.osProvider.os === 'Windows' ? path.join(programFilesEnvironmentVariable, 'dotnet', 'sdk', 'NuGetFallbackFolder') : MacNuGetPackageFallbackFolderPath,
+            localPath: this.osProvider.os === 'Windows' ? path.join(programFilesEnvironmentVariable, 'dotnet', 'sdk', 'NuGetFallbackFolder') :
+                (this.osProvider.isMac ? MacNuGetPackageFallbackFolderPath : LinuxNuGetPackageFallbackFolderPath),
             containerPath: options.os === 'Windows' ? 'C:\\.nuget\\fallbackpackages' : '/root/.nuget/fallbackpackages',
             permissions: 'ro'
         };


hacst:
Note that I was only able to verify this fixed my issues on Linux and did not seem to break anything on Windows. As I do not have access to a Mac I couldn't test the change there. 

My understanding of the default paths comes from https://github.com/dotnet/core-setup/blob/1d95ece515edbbad0ef8f8980424d3e3bce49cb2/src/corehost/common/pal.unix.cpp#L203 which seems to be a part of what ```dotnet --list-sdks``` etc. uses under the hood.

StephenWeatherford:
@philliphoff Can you review?  Thx.

philliphoff:
@hacst Thanks for the contribution!  Will try to get to this in the next couple of days.  I've got all 3 platforms running on which I can test the change.

hacst:
Awesome. Thanks for the heads-up.

debugging/coreclr/dockerManager.ts
```diff
@@ -341,18 +341,26 @@ export class DefaultDockerManager implements DockerManager {
             permissions: 'ro'
         };
 
-        let programFilesEnvironmentVariable: string | undefined;
-
-        if (this.osProvider.os === 'Windows') {
-            programFilesEnvironmentVariable = this.processProvider.env[DefaultDockerManager.ProgramFilesEnvironmentVariable];
+        let nugetFallbackFolder: string;
+        if (this.osProvider.isMac) {
+            nugetFallbackFolder = MacNuGetPackageFallbackFolderPath;
```
philliphoff: Perhaps `MacNuGetPackageFallbackFolderPath` should just be the .NET installation folder, so that we can build the `nugetFallbackFolder` path in the same way across all the platforms.  Alternatively, just define the constant for Linux (like Mac) and only actually build a path in the case of Windows.

It just feels a little inconsistent that Mac is treated differently than Linux, when both are static paths.

hacst: Thanks for taking a look. Both ways are possible. I decided not to make ```MacNuGetPackageFallbackFolderPath``` relative because it is exported and used somewhere else. Changing its meaning would've meant having to duplicate the knowledge about where to find the fallback folder relative to the sdk in that location too. It would make things much more uniform though.

I also considered just adding a Linux version of the variable but didn't see a reason for it to be exported or in that scope.

It seemed like things would likely develop away from static paths more towards whatever the dotnet binary does for dynamic sdk detection on the different platforms so the patch goes into that direction.

I think the smallest change to just achieve the fix would be to just use an explicit string for linux and the existing behavior for the rest. If that's preferrable I can adjust the PR to that.

philliphoff: Yeah, I see the Mac path used in a pre-requisite check as well.  Ok, I see a couple of options of varying complexity:

1. As you suggest, just make the complete Linux path a constant, in the same way as the Mac path, defined next to the Mac path (whether exported or not).  This has the advantage of making the differences between those platforms more obvious, being able to see the paths next to each other.  Obviously the simplest change.

1. Being a .NET-specific property, one could argue for a new property on the `DotNetClient` interface (e.g. `nuGetPackageFallbackFolderPath: string`) which builds the path based on the local platform OS.  That logic can then be easily shared between the `DockerManager` and the pre-requisite checks.  A bigger change, but feels like the "better" one to me.

It'd be great to do one of these; I'll let you decide which, though.

hacst: Went for option 1 .

hacst:
Sorry for the delay. Updated the patch according to the review recommendations. Please take another look.

PR #795, opened 2019-02-16
Add debugging and dockerfile creation for fsharp dotnet core projects.

gdziadkiewicz:
**Tested only on Windows 10 machine with Docker for Windows** using template generated fsharp app and generated dockerfile.

~~After refreshing my branch with master the debugging feature stopped working for my setup and the changes in commit e6cc443 are workarounds for those problems:~~
1. ~~Before changing the  single quotes to double quotes in platform checking docker exec call it resulted in `Syntax error: Unterminated quoted string`  error.~~
2. ~~The call to ` this.debuggerClient.getDebugger` was returning path with backslashes. This also caused error. Please confirm that this is the right place to do the normalization if you decide to merge this.~~

### DIFF
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -364,17 +364,19 @@ async function configureCore(actionContext: IActionContext, options: ConfigureAp
     }
 
     let targetFramework: string;
+    let projFile: string;
     let serviceNameAndPathRelativeToOutput: string;
     {
         // Scope serviceNameAndPathRelativeToRoot only to this block of code
         let serviceNameAndPathRelativeToRoot: string;
         if (platformType.toLowerCase().includes('.net')) {
             let projFilePath = await findCSProjOrFSProjFile(rootFolderPath);
-            serviceNameAndPathRelativeToRoot = projFilePath;
+            serviceNameAndPathRelativeToRoot = projFilePath.slice(0, -(path.extname(projFilePath).length));
             let projFileContents = (await fse.readFile(path.join(rootFolderPath, projFilePath))).toString();
 
             // Extract TargetFramework for version
             [targetFramework] = extractRegExGroups(projFileContents, /<TargetFramework>(.+)<\/TargetFramework/, ['']);
+            projFile = projFilePath;
 
             properties.packageFileType = projFilePath.endsWith('.csproj') ? '.csproj' : '.fsproj';
             properties.packageFileSubfolderDepth = getSubfolderDepth(serviceNameAndPathRelativeToRoot);
@@ -406,6 +408,7 @@ async function configureCore(actionContext: IActionContext, options: ConfigureAp
 
     if (targetFramework) {
         packageInfo.version = targetFramework;
+        packageInfo.artifactName = projFile;
     }
 
     let filesWritten: string[] = [];
--- a/configureWorkspace/configure.ts
+++ b/configureWorkspace/configure.ts
@@ -51,7 +51,7 @@ interface PomXmlContents {
 export type ConfigureTelemetryProperties = {
     configurePlatform?: Platform;
     configureOs?: PlatformOS;
-    packageFileType?: string; // 'build.gradle', 'pom.xml', 'package.json', '.csproj'
+    packageFileType?: string; // 'build.gradle', 'pom.xml', 'package.json', '.csproj', '.fsproj'
     packageFileSubfolderDepth?: string; // 0 = project/etc file in root folder, 1 = in subfolder, 2 = in subfolder of subfolder, etc.
 };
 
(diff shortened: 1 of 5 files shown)


debugging/coreclr/dockerManager.ts
```diff
@@ -236,7 +236,7 @@ export class DefaultDockerManager implements DockerManager {
 
         await this.addToDebugContainers(containerId);
 
-        const debuggerPath = await this.debuggerClient.getDebugger(options.run.os, containerId);
+        const debuggerPath = this.osProvider.pathNormalize(options.run.os, await this.debuggerClient.getDebugger(options.run.os, containerId));
```
gdziadkiewicz: Is this the right place to do that?

debugging/coreclr/debuggerClient.ts
```diff
@@ -19,7 +19,7 @@ export class DefaultDebuggerClient {
 
     // This script determines the "type" of Linux release (e.g. 'alpine', 'debian', etc.).
     // NOTE: The result may contain line endings.
-    private static debuggerLinuxReleaseIdScript: string = '/bin/sh -c \'ID=default; if [ -e /etc/os-release ]; then . /etc/os-release; fi; echo $ID\'';
+    private static debuggerLinuxReleaseIdScript: string = '/bin/sh -c "ID=default; if [ -e /etc/os-release ]; then . /etc/os-release; fi; echo $ID"';
```
gdziadkiewicz: Not checked on Linux or OS X machines.

StephenWeatherford:
@philliphoff

debugging/coreclr/debuggerClient.ts
philliphoff: Unfortunately, double-quotes only works on Windows; Mac and Linux require single quotes (else the interpretation is performed by the host machine rather than the container).  I've created a separate PR (#798) with just this fix (to disambiguate it from your F# changes) and updated it to work across the platforms.

debugging/coreclr/dockerManager.ts
philliphoff: I feel like the `DebuggerClient` should be responsible for returning a relative path to the debugger *in the container platform's format*, so I've moved it there in PR #798, which is based on your workaround.

configureWorkspace/configure.ts
```diff
@@ -369,14 +369,14 @@ async function configureCore(actionContext: IActionContext, options: ConfigureAp
         // Scope serviceNameAndPathRelativeToRoot only to this block of code
         let serviceNameAndPathRelativeToRoot: string;
         if (platformType.toLowerCase().includes('.net')) {
-            let csProjFilePath = await findCSProjFile(rootFolderPath);
-            serviceNameAndPathRelativeToRoot = csProjFilePath.slice(0, -'.csproj'.length);
-            let csProjFileContents = (await fse.readFile(path.join(rootFolderPath, csProjFilePath))).toString();
+            let projFilePath = await findCSProjOrFSProjFile(rootFolderPath);
+            serviceNameAndPathRelativeToRoot = projFilePath;
```
philliphoff: What's the impact of *not* trimming the file extension as the previous version did?

philliphoff: Ah, ok, I see, the file extension gets used later in the .NET Core Dockerfile `genDockerFile()` generator function.  Since the generator signature is supposed to be generic across platforms, I'm a little leery of making `serviceNameAndPathRelativeToRoot` semantically different *only* for .NET Core.

I wonder if it'd be better to use the `PackageInfo` object to pass along something like the `packageType`, which is already being collected.  That can then be used by the generator function and keep the existing semantics of the `serviceNameAndPathRelativeToRoot`.  (It looks like the .NET Core generator is already using the `PackageInfo.version` to hold the target .NET Framework, so it'd be a similar idea.  Maybe the `artifactName` would be a good fit?)

philliphoff:
You might also look over the `README.md` and see if there is anything that needs to be documented (e.g. machine prerequisites, extension prerequisites, special configuration steps, etc.) in order for a user to be successful with debugging F# in Docker containers.

debugging/coreclr/dockerManager.ts
gdziadkiewicz: Ok, I will remove this from my branch.

debugging/coreclr/debuggerClient.ts
gdziadkiewicz: Great! I will remove this commit from the branch.

configureWorkspace/configure.ts
gdziadkiewicz: I went with the `artifactName`.

gdziadkiewicz:
Review issues solved - ready for review.

philliphoff:
@gdziadkiewicz Sorry, I've been out of the office the past couple of days; will try to get to this in the next couple.

StephenWeatherford:
@philliphoff How close is this to being ready?  Need it in by tomorrow if for this release, thx.

philliphoff:
@StephenWeatherford The changes themselves now look ok to me; let me see if I can do some testing this afternoon.

philliphoff:
Tested successfully across Windows, Mac, and Linux (Ubuntu). :shipit:

PR #798, opened 2019-02-19
Resolve .NET Core debugging on Windows

philliphoff:
gdziadkiewicz had discovered that I broke debugging .NET Core in Docker on Windows with the previous Alpine support commit, and was also kind enough to suggest a workaround in PR #795 (a larger change to support F# project debugging).

For simplicity, I've split that workaround into a separate PR and updated/verified it to work across Mac/Ubuntu/Windows.  The changes are:

 - Use a double-quoted script string to obtain the container debugger type when VS Code is running on Windows.
 - Normalize the relative (host) path returned by the debugging client for the target (i.e. container's) platform.

### DIFF
--- a/debugging/coreclr/debuggerClient.ts
+++ b/debugging/coreclr/debuggerClient.ts
@@ -4,6 +4,7 @@
 
 import { PlatformOS } from '../../utils/platform';
 import { DockerClient } from './dockerClient';
+import { OSProvider } from './osProvider';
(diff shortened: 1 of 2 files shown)


philliphoff:
@StephenWeatherford @EricJizbaMSFT Could either of you take a quick look at this when you get a chance (as a community PR is dependent upon it)?  Thanks!

PR #800, opened 2019-02-20
Add docker compose start, stop

### LINKED ISSUE #623: Add Compose Stop and Compose Start to Command Palette
`Compose Down` and `Compose Up` are supported but they destroy and rebuild the containers. Please add support for the non-destructive commands `Compose Stop` (`docker-compose stop`) and `Compose Start` (`docker-compose start`).

PrashanthCorp:
Fixes #623

### DIFF
--- a/commands/docker-compose.ts
+++ b/commands/docker-compose.ts
@@ -38,7 +38,7 @@ function computeItems(folder: vscode.WorkspaceFolder, uris: vscode.Uri[]): vscod
     return items;
 }
 
-async function compose(commands: ('up' | 'down')[], message: string, dockerComposeFileUri?: vscode.Uri, selectedComposeFileUris?: vscode.Uri[]): Promise<void> {
+async function compose(commands: composeOperation[], message: string, dockerComposeFileUri?: vscode.Uri, selectedComposeFileUris?: vscode.Uri[]): Promise<void> {
     let folder: vscode.WorkspaceFolder = await quickPickWorkspaceFolder('To run Docker compose you must first open a folder or workspace in VS Code.');
 
     let commandParameterFileUris: vscode.Uri[];
@@ -59,7 +59,7 @@ async function compose(commands: ('up' | 'down')[], message: string, dockerCompo
         }
 
         const items: vscode.QuickPickItem[] = computeItems(folder, uris);
-        selectedItems = [<Item>await ext.ui.showQuickPick(items, { placeHolder: `Choose Docker Compose file ${message}` })];
+        selectedItems = [<Item>await ext.ui.showQuickPick(items, { placeHolder: `Choose Docker Compose file to ${message} containers.` })];
     }
 
     const terminal: vscode.Terminal = ext.terminalProvider.createTerminal('Docker Compose');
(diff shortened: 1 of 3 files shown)


commands/docker-compose.ts
```diff
@@ -38,7 +38,7 @@ function computeItems(folder: vscode.WorkspaceFolder, uris: vscode.Uri[]): vscod
     return items;
```
StephenWeatherford: new testcases?

package.json
```diff
@@ -806,10 +813,22 @@
...
+      {
+        "command": "vscode-docker.compose.start",
+        "title": "Compose Restart",
+        "description": "Builds and starts a composition of containers",
+        "category": "Docker"
+      },
+      {
+        "command": "vscode-docker.compose.stop",
+        "title": "Compose Restart",
+        "description": "Stops a composition of containers (non-destructive)",
```
StephenWeatherford: Stops a composition of containers without removing them.

(see https://docs.docker.com/compose/reference/stop/)

package.json
```diff
@@ -806,10 +813,22 @@
         "description": "Restarts a composition of containers",
         "category": "Docker"
       },
+      {
+        "command": "vscode-docker.compose.start",
+        "title": "Compose Restart",
+        "description": "Builds and starts a composition of containers",
```
StephenWeatherford: Don't think it builds, see https://docs.docker.com/compose/reference/start/

package.json
```diff
@@ -806,10 +813,22 @@
...
+        "command": "vscode-docker.compose.stop",
+        "title": "Compose Restart",
+        "description": "Stops a composition of containers (non-destructive)",
+        "category": "Docker"
+      },
       {
         "command": "vscode-docker.compose.up",
         "title": "Compose Up",
-        "description": "Starts a composition of containers",
+        "description": "Builds a composition of containers",
```
StephenWeatherford: Builds, (re)creates, starts, and attaches to containers for a service.

https://docs.docker.com/compose/reference/up/

package.json
```diff
@@ -797,7 +804,7 @@
       {
         "command": "vscode-docker.compose.down",
         "title": "Compose Down",
-        "description": "Stops a composition of containers",
+        "description": "Stops and removes a composition of containers",
```
StephenWeatherford: Stops containers and removes containers and networks created by up.

https://docs.docker.com/compose/reference/down/

commands/docker-compose.ts
```diff
@@ -38,7 +38,7 @@ function computeItems(folder: vscode.WorkspaceFolder, uris: vscode.Uri[]): vscod
...
 }
```
StephenWeatherford: image

PrashanthCorp: @fiveisprime What do you think about the number of context menu items?

package.json
StephenWeatherford: There was a reason I modified the text from the docs to "Stops containers and removes containers and non-external networks created by Compose Up." If you read the docs, you'll see that using the flags we use does not remove all of those.  Don't want to worry users.

commands/docker-compose.ts
StephenWeatherford: My point was that you have 3 restarts.

Personally I would group up and down together in a group, and start/stop/restart in another group.

fiveisprime: We talked a little about this via email. I was also confused by the three restarts 🤣 My ¢2 on the order:

- Compose Up
- Compose Down
- Compose Restart
- Compose Stop
- Compose Start

Will `exec` come into this as well or are we less worried about that?

StephenWeatherford: Don't you think stop/stop should be in a separate group?  That would make it easier to find the one you want and also break up the related functionality (since start/stop are non-destructive and don't build).

Do you think we should consider keyboard bindings for some of these things?  Or best left to the user?

StephenWeatherford: Actually, I can't remember if restart does up/down or start/stop.

PrashanthCorp: It used to do up/down, but it should be doing stop/start (which I've fixed).

commands/docker-compose.ts
StephenWeatherford: ??

StephenWeatherford: You broke compose up

commands/docker-compose.ts
StephenWeatherford: > Don't you think stop/stop should be in a separate group?  That would make it easier to find the one you want and also break up the related functionality (since start/stop are non-destructive and don't build).
> Do you think we should consider keyboard bindings for some of these things?  Or best left to the user?

@fiveisprime I still think we should make groups.
image

EricJizbaMSFT:
Closing this PR due to lack of activity. Issue still tracked by https://github.com/Microsoft/vscode-docker/issues/623

PR #825, opened 2019-03-08
Create only one output channel

### LINKED ISSUE #715: Two "Docker" tabs in output view
image

StephenWeatherford:
Fixes #715


extension.ts
```diff
@@ -91,7 +91,9 @@ function initializeExtensionVariables(ctx: vscode.ExtensionContext): void {
     ext.ui = new AzureUserInput(ctx.globalState);
   }
   ext.context = ctx;
-  ext.outputChannel = util.getOutputChannel();
+
+  ext.outputChannel = vscode.window.createOutputChannel("Docker");
```
philliphoff: `createOutputChannel()` returns a "disposable"; is that (should that) be added to the disposable list for when the extension is unloaded?

StephenWeatherford: Yep

PR #834, opened 2019-03-11
Add more state icons

StephenWeatherford:
… for next release


explorer/models/rootNode.ts
```diff
@@ -333,24 +341,50 @@ export class RootNode extends NodeBase {
...
+                                dark: path.join(imagesPath, 'dark', 'StatusStop_16x.svg'),
+                            };
+                            break;
+                        case "paused":
+                            iconPath = {
+                                light: path.join(imagesPath, 'light', 'StatusPause_16x.svg'),
+                                dark: path.join(imagesPath, 'dark', 'StatusPause_16x.svg'),
+                            };
+                            break;
+                        case "created":
```
EricJizbaMSFT: What's the reasoning for a "created" container using the "restart" icon? I would think stopped makes sense, and your comment above seems to agree
> // A container that ran and completed("stopped" in other contexts, although a created container is technically also "stopped")

StephenWeatherford: I thought created meant it was in the process of transitioning to starting, but apparently that's not the case.  Will change, thx.

Issue #900, opened 2019-04-16
Extension does not load in older vscode versions - WORKAROUND AVAILABLE: Please upgrade vscode to 1.31.1 or higher


StephenWeatherford:
WORKAROUND: Please update to vscode 1.31.1 or higher (current is 1.33) or else install the previous version of the extension.  A fix will be available  in the next week or two.

Use Help -> Check for Updates...
or download from here: https://code.visualstudio.com/updates/v1_31

StephenWeatherford:
Tracked in https://github.com/Microsoft/vscode-docker/issues/881

StephenWeatherford:
Fixed in 0.6.2

PR #903, opened 2019-04-19
Use executeCommand(vscode.open) instead of opn

PrashanthCorp:
The handler function is still called openExternal. I plan to use openExternal ASAP.


package-lock.json
```diff
@@ -11331,7 +11312,7 @@
     "yazl": {
       "version": "2.5.1",
       "resolved": "https://registry.npmjs.org/yazl/-/yazl-2.5.1.tgz",
-      "integrity": "sha512-phENi2PLiHnHb6QBVot+dJnaAZ0xosj7p3fWl+znIjBDlnMI2PsZCJZ306BPTFOaHf5qdDEI8x5qFrSOBN5vrw==",
+      "integrity": "sha1-o9ZdPdZZpbCTeFDoYJ8i//orXDU=",
```
EricJizbaMSFT: Hmm can we double check why all of these are changing?

package.json
```diff
@@ -1021,7 +1021,7 @@
     }
   },
   "engines": {
-    "vscode": "^1.26.0"
+    "vscode": "^1.31.0"
```
EricJizbaMSFT: We can't change this due to https://github.com/Microsoft/vscode-docker/issues/900

explorer/utils/dockerConnectionError.ts
```diff
@@ -37,8 +37,8 @@ export function showDockerConnectionError(actionContext: IActionContext, error:
     actionContext.suppressErrorDisplay = true;
     vscode.window.showErrorMessage(parseError(wrappedError).message, ...items).then(response => {
         if (response) {
-            // tslint:disable-next-line:no-unsafe-any
-            opn(response.url);
+            // tslint:disable-next-line:no-floating-promises
+            openExternal(response.url);
```
EricJizbaMSFT: How come some places you added an `await` and some places you just disable the lint error? I think in most cases it should be safe to await openExternal. (this was not true of `opn` because awaiting `opn` would never return as far as I could tell)

explorer/models/taskNode.ts
```diff
@@ -55,12 +55,11 @@ export class TaskRootNode extends NodeBase {
             tasks = await client.tasks.list(resourceGroup, element.registry.name);
             if (tasks.length === 0) {
                 const learnHow: vscode.MessageItem = { title: "Learn How to Create Build Tasks" };
-                vscode.window.showInformationMessage(`You do not have any Tasks in the registry '${element.registry.name}'.`, learnHow).then(val => {
-                    if (val === learnHow) {
-                        // tslint:disable-next-line:no-unsafe-any
-                        opn('https://aka.ms/acr/task');
-                    }
-                })
+                let response = await vscode.window.showInformationMessage(`You do not have any Tasks in the registry '${element.registry.name}'.`, learnHow)
```
EricJizbaMSFT: Hmm I think this should still use a `then` (and maybe add a comment like `// don't wait` to make it more clear that this was intentional). Awaiting `showInformationMessage` means it won't return until the user dismisses the notification which might not happen for a long time if the user ignores the notification. Specifically - I wonder if this results in a loading "spinner" in the tree while we await

commands/utils/quick-pick-azure.ts
```diff
@@ -79,12 +79,11 @@ export async function quickPickSubscription(): Promise<Subscription> {
     const subscriptions = await AzureUtilityManager.getInstance().getFilteredSubscriptionList();
     if (subscriptions.length === 0) {
         let openPortal = 'Open Portal';
-        vscode.window.showErrorMessage("You are not signed in to Azure, or you do not have any subscriptions. To sign in, select 'Azure: Sign In' from the command palette. Subscriptions can be created in the Azure portal", openPortal).then(val => {
-            if (val === openPortal) {
-                // tslint:disable-next-line:no-unsafe-any
-                opn('https://portal.azure.com/');
-            }
-        });
+        let response = await vscode.window.showErrorMessage("You are not signed in to Azure, or you do not have any subscriptions. To sign in, select 'Azure: Sign In' from the command palette. Subscriptions can be created in the Azure portal", openPortal);
```
EricJizbaMSFT: Same thing here regarding other comment about `.then`

package-lock.json
PrashanthCorp: Was on an older version of npm. I updated it to 6.9.0, and the integrity keys have reverted AFAICT.

explorer/utils/dockerConnectionError.ts
EricJizbaMSFT: Fair enough - I think awaiting `openExternal` is technically cleaner, but realistically I don't think it matters

PrashanthCorp:
Done.

PR #906, opened 2019-04-22
Ensure debugger directory exists

### LINKED ISSUE #897: '/home/$USER/.vsdbg/getvsdbg.sh' is missing after fresh install
TLDR:
Error: EACCES: permission denied, open '/home/$USER/.vsdbg/getvsdbg.sh' but '/home/$USER/.vsdbg/getvsdbg.sh' is missing.
Fix install plox :D

Repro steps:

1. Fresh install of pop_os
2. Install vs code
3. dotnet new mcv
4. standard dockerfile
5. standard docker-coreclr buildtask
6. Build goes thru but run does not.

Error Message:
Acquiring the latest .NET Core debugger...
Unable to acquire the .NET Core debugger.
Error: EACCES: permission denied, open '/home/$USER/.vsdbg/getvsdbg.sh'

Action: debugCoreClr
Error type: EACCES
Error Message: EACCES: permission denied, open '/home/tehho/.vsdbg/getvsdbg.sh'

Version: 0.6.1
OS: linux
VS Code Version: 1.33.1

philliphoff:
Ensure the debugger directory exists prior to starting the container (or, rather, before any possible use of the directory), to prevent it potentially being created with permissions different than those used by VS Code (to actually acquire the debugger).

This resolves #897.

### DIFF
--- a/debugging/coreclr/vsdbgClient.ts
+++ b/debugging/coreclr/vsdbgClient.ts
@@ -98,11 +98,7 @@ export class RemoteVsDbgClient implements VsDbgClient {
     }
 
     private async ensureVsDbgFolderExists(): Promise<void> {
-        const directoryExists = await this.fileSystemProvider.dirExists(this.vsdbgPath);
-
-        if (!directoryExists) {
-            await this.fileSystemProvider.makeDir(this.vsdbgPath);
-        }
+        await this.fileSystemProvider.ensureDir(this.vsdbgPath);
     }
 
     private async getVsDbgAcquisitionScript(): Promise<void> {
--- a/debugging/coreclr/vsdbgClient.ts
+++ b/debugging/coreclr/vsdbgClient.ts
@@ -59,6 +59,11 @@ export class RemoteVsDbgClient implements VsDbgClient {
     }
 
     public async getVsDbgFolder(): Promise<string> {
+        // NOTE: If non-existant, other processes (e.g. Docker when volume mounting) may attempt to create this folder.
+        //       This can lead to "access denied" if those processes run elevated compared to VS Code.
+        //       Therefore, ensure it's created before they're ever given the folder name.
+        await this.ensureVsDbgFolderExists();
+
         return this.vsdbgPath;
     }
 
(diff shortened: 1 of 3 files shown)


debugging/coreclr/vsdbgClient.ts
```diff
@@ -92,6 +97,14 @@ export class RemoteVsDbgClient implements VsDbgClient {
             'Unable to acquire the .NET Core debugger.');
     }
 
+    private async ensureVsDbgFolderExists(): Promise<void> {
+        const directoryExists = await this.fileSystemProvider.dirExists(this.vsdbgPath);
+
+        if (!directoryExists) {
+            await this.fileSystemProvider.makeDir(this.vsdbgPath);
```
PrashanthCorp: You can use `fse.EnsureDir` for this.

PrashanthCorp:
I have larger questions about the use of a fileSystemHandler - all the functionality I could see in the codebase could be done with fs-extra's functionality. I don't think that should block this PR, but I'd like some context on that soon.

debugging/coreclr/vsdbgClient.ts
philliphoff: Sure can!  Updated.

philliphoff:
@PrashanthCorp 
> I have larger questions about the use of a fileSystemHandler - all the functionality I could see in the codebase could be done with fs-extra's functionality. I don't think that should block this PR, but I'd like some context on that soon.

The `FileSystemProvider` serves several purposes, as a shield from the external library actually used (i.e. `fs-extra`) which allows it to be easily swapped with another if needed in the future, as an adaptor for when that library doesn't exactly provide the desired API (e.g. separate file/dir exist functions), as a convenient unit of dependency injection, and as a point for mocking in unit tests.

Could `fs-extra` have been used directly, yes.  But consider the (currently) pending PR (#903) to replace `opn`, where a number of places in the code had to be touched due to the change.  Had the `openExternal()` function existed as an abstraction layer for `opn` in the beginning, the change likely would have been much smaller.

PR #907, opened 2019-04-26
Update the scope of commands that require file system access

RMacfarlane:


commands/utils/TerminalProvider.ts
```diff
@@ -11,15 +11,32 @@ export interface ITerminalProvider {
...
+  private _cwd: vscode.Uri | undefined;
+
+  constructor(extensionExecutionContext: vscode.ExtensionExecutionContext) {
+    // When the extension is executing locally but there is a remote present,
+    // creating a terminal defaults to the remote. Passing a file URI overrides
+    // this behavior so the terminal is created locally.
+    this._cwd = extensionExecutionContext === vscode.ExtensionExecutionContext.Remote
+      ? undefined
+      : process.platform === 'win32'
+        ? vscode.Uri.file(process.env.USERPROFILE || 'c:\\')
```
EricJizbaMSFT: I assume this would work:
```
import * as os from 'os';

os.homedir()
```

commands/utils/TerminalProvider.ts
```diff
@@ -11,15 +11,32 @@ export interface ITerminalProvider {
 }
 
 export class DefaultTerminalProvider {
+  private _cwd: vscode.Uri | undefined;
+
+  constructor(extensionExecutionContext: vscode.ExtensionExecutionContext) {
+    // When the extension is executing locally but there is a remote present,
```
EricJizbaMSFT: Is there any documentation for this behavior? I want to make sure this is actually an intended design choice for how terminals work in VS Code - rather than a random "bug" (aka "feature") that we should be relying on

extension.ts
```diff
@@ -103,10 +103,27 @@ function initializeExtensionVariables(ctx: vscode.ExtensionContext): void {
   registerUIExtensionVariables(ext);
 }
 
+async function setContextKeys(ctx: vscode.ExtensionContext): Promise<void> {
+  const isRemote = ctx.executionContext === vscode.ExtensionExecutionContext.Remote;
+  vscode.commands.executeCommand('setContext', 'docker:isRemote', isRemote);
+
+  const hasRemote = vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders[0].uri.authority;
+  const hasShownInfoMessage = ctx.globalState.get('docker.hasShownRemoteNotification');
+  if (!hasShownInfoMessage && !isRemote && hasRemote) {
```
EricJizbaMSFT: So we're worried it'll be confusing if commands are just hidden and I'm not sure how effective a single global notification would be. Instead, what if we have a per-command warning like this:
```
This command requires access to the remote workspace, but the Docker extension is running on your local machine.
```
In this case, I think it also makes sense to have a button that would switch Docker from a ui extension to a workspace extension, since that seems difficult to figure out.

package.json
```diff
@@ -27,6 +27,8 @@
     "type": "git",
     "url": "https://github.com/microsoft/vscode-docker"
   },
+  "extensionKind": "ui",
```
EricJizbaMSFT: Do I have to do anything special to debug an extension when the kind is set to "ui"? [These instructions](https://code.visualstudio.com/api/advanced-topics/remote-extensions#debugging-your-extension-in-a-remote-environment) work for me when the kind is "workspace", but I don't see the docker explorer when it's set to "ui"

typings/vscode.proposed.d.ts
```diff
@@ -0,0 +1,36 @@
...
+ * These API are NOT stable and subject to change. They are only available in the Insiders
+ * distribution and CANNOT be used in published extensions.
+ *
+ * To test these API in local environment:
+ * - Use Insiders release of VS Code.
+ * - Add `"enableProposedApi": true` to your package.json.
+ * - Copy this file to your project.
+ */
+
+declare module 'vscode' {
```
EricJizbaMSFT: What's the status of this proposed api? Is it out of proposed yet?

EricJizbaMSFT:
Closing this due to lack of activity and in favor of https://github.com/microsoft/vscode-docker/issues/954

PR #920, opened 2019-05-03
Don't use webpack for local development

EricJizbaMSFT:


.vscode/launch.json
```diff
@@ -11,12 +11,31 @@
...
 			],
 			"env": {
+				"AZCODE_DOCKER_IGNORE_BUNDLE": "1",
 				"DEBUGTELEMETRY": "1"
 			},
 			"stopOnEntry": false,
 			"sourceMaps": true,
 			"outFiles": [
-				"${workspaceFolder}/../dist/**/*.js"
+				"${workspaceFolder}/**/*.js"
```
StephenWeatherford: Do breakpoint work?  I had trouble with them until I set this to ${workspaceFolder}/out/**/*.js

EricJizbaMSFT: Works for me, but I can add that

PR #921, opened 2019-05-04
Remove usage of clipboardy

PrashanthCorp:
The pacakge-lock.json has a lot of https -> http conversions. I've update npm to the latest version and cleaned/rebuilt node_modules, but the issues still persist. Any ideas?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1025,7 +1025,7 @@
     }
   },
   "engines": {
-    "vscode": "^1.26.0"
+    "vscode": "^1.33.0"
   },
   "scripts": {
     "vscode:prepublish": "npm run webpack-prod",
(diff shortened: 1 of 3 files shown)


StephenWeatherford:
Try npm cache clean

package.json
```diff
@@ -1025,7 +1025,7 @@
     }
   },
   "engines": {
-    "vscode": "^1.26.0"
+    "vscode": "^1.33.0"
```
StephenWeatherford: Is this the minimum we can go and still get this functionality?

EricJizbaMSFT: Hmm Stephen had a comment related to this and then deleted it...but I still think this should be 1.31.0 (at least for this PR) - the minimum needed for `vscode.env.clipboard`

Issue #848, opened 2019-03-14
C++ template doesn't have EXPOSE statements


StephenWeatherford:

StephenWeatherford:
function genDockerFile(serviceNameAndRelativePath: string, platform: string, os: string | undefined, port: string, { cmd, author, version, artifactName }: Partial<PackageInfo>): string {
  let exposeStatements = getExposeStatements(port);  <<<< NOT USED

PR #926, opened 2019-05-06
Turn on noUnusedLocals

EricJizbaMSFT:
A few notes:
1. `verifyTelemetryProperties` wasn't used in the configure tests, but seems somewhat useful so I added it and fixed the failing tests
1. the cpp generation was ignoring `exposeStatements`, but I assume that's fine since it says `// We don't open a port for Cpp` a few lines above. Updated the cpp test to reflect this

### DIFF
--- a/commands/azureCommands/pull-from-azure.ts
+++ b/commands/azureCommands/pull-from-azure.ts
@@ -6,11 +6,7 @@
 import * as assert from 'assert';
 import { Registry } from "azure-arm-containerregistry/lib/models";
 import { exec } from 'child_process';
-import * as fse from 'fs-extra';
-import os = require('os');
-import * as path from "path";
 import vscode = require('vscode');
-import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
 import { AzureImageTagNode, AzureRepositoryNode } from '../../explorer/models/azureRegistryNodes';
 import { ext } from '../../extensionVariables';
 import * as acrTools from '../../utils/Azure/acrTools';
@@ -90,19 +86,4 @@ async function pullImage(loginServer: string, imageRequest: string, username: st
     terminal.show();
 
     terminal.sendText(`docker pull ${loginServer}/${imageRequest}`);
-}
-
-async function isLoggedIntoDocker(loginServer: string): Promise<{ configPath: string, loggedIn: boolean }> {
-    const home = os.homedir();
-    let configPath: string = path.join(home, '.docker', 'config.json');
-    let buffer: Buffer;
-
-    await callWithTelemetryAndErrorHandling('findDockerConfig', async function (this: IActionContext): Promise<void> {
-        this.suppressTelemetry = true;
-        buffer = fse.readFileSync(configPath);
-    });
-
-    let index = buffer.indexOf(loginServer);
-    let loggedIn = index >= 0; // Returns -1 if user is not logged into Docker
-    return { configPath, loggedIn }; // Returns object with configuration path and boolean indicating if user was logged in or not
 }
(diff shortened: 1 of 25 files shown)


commands/azureCommands/pull-from-azure.ts
```diff
@@ -91,18 +87,3 @@ async function pullImage(loginServer: string, imageRequest: string, username: st
 
     terminal.sendText(`docker pull ${loginServer}/${imageRequest}`);
 }
-
-async function isLoggedIntoDocker(loginServer: string): Promise<{ configPath: string, loggedIn: boolean }> {
```
PrashanthCorp: Is this an unused function?

debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -100,7 +100,7 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
 
         const { appFolder, resolvedAppFolder } = await this.inferAppFolder(folder, debugConfiguration);
 
-        const { appProject, resolvedAppProject } = await this.inferAppProject(folder, debugConfiguration, resolvedAppFolder);
+        const { resolvedAppProject } = await this.inferAppProject(folder, debugConfiguration, resolvedAppFolder);
```
PrashanthCorp: I don't see changes to the `inferAppProject` function to reflect this change.

commands/azureCommands/pull-from-azure.ts
EricJizbaMSFT: Yes

debugging/coreclr/dockerDebugConfigurationProvider.ts
EricJizbaMSFT: `inferAppProject` returns an object - removing `appProject` is essentially like removing a property name from an interface, which doesn't require any sort of change on the original object.

EricJizbaMSFT:
Related issue: https://github.com/microsoft/vscode-docker/issues/848

EricJizbaMSFT:
ping @nturinski @PrashanthCorp @StephenWeatherford

debugging/coreclr/dockerManager.ts
```diff
@@ -60,11 +60,6 @@ type LastImageBuildMetadata = {
     options: DockerBuildImageOptions;
 };
 
-type LastContainerRunMetadata = {
```
StephenWeatherford: fyi @philliphoff

debugging/coreclr/netCoreProjectProvider.ts
```diff
@@ -2,13 +2,12 @@
...
  *--------------------------------------------------------*/
 
-import * as path from 'path';
 import { DotNetClient } from "./dotNetClient";
 import { FileSystemProvider } from "./fsProvider";
 import { TempFileProvider } from './tempFileProvider';
 
 const getTargetPathProjectFileContent =
-`<Project>
+    `<Project>
```
StephenWeatherford: Why this change?  Personally I prefer them all to be left-indented if any part of the multi-line string is.

test/configure.test.ts
```diff
@@ -143,6 +143,8 @@ async function testConfigureDocker(platform: Platform, expectedTelemetryProperti
             }
         }
     }
+
+    verifyTelemetryProperties(actionContext, expectedTelemetryProperties);
```
StephenWeatherford: Thx.

debugging/coreclr/netCoreProjectProvider.ts
EricJizbaMSFT: My only change was to remove the `path` import above. This must have been auto-formatted

EricJizbaMSFT: Yeah this is getting changed on-save. I like it better left-indented, too, but don't feel like fighting the auto format

PR #904, opened 2019-04-22
Allow custom values for COMPOSE_FILE_GLOB_PATTERN and DOCKERFILE_GLOB_PATTERN (issue #102)

x13machine:
* add setting docker.dockerfileGlobPattern for specifying COMPOSE_FILE_GLOB_PATTERN
* add setting docker.composeFileGlobPattern for specifying COMPOSE_FILE_GLOB_PATTERN 
* add information about the settings in README.md 

Fixes #102

PR #928, opened 2019-05-08
Respect `file.associations` for Dockerfile commands

EricJizbaMSFT:
We've got a number of issues/PRs out for people wanting support for Dockerfile.\<something\>. The most common workaround is to do something like this:
```
"files.associations": {
    "[Dd]ockerfile*": "dockerfile",
    "docker-compose*": "yaml"
}
```

That works great for icons/intellisense, but not command palette commands like "Build Image" (as mentioned [here](https://github.com/microsoft/vscode-docker/issues/102#issuecomment-462759346) by @crstamps2). This PR fixes that.

Note: I'm intentionally take a different approach than PRs like https://github.com/microsoft/vscode-docker/pull/904 which add custom settings. I'm also not dealing with docker-compose files at all. I wanted to unblock a good chunk of users as fast as possible with hopefully an uncontroversial fix.

### DIFF
--- a/src/utils/quick-pick-file.ts
+++ b/src/utils/quick-pick-file.ts
@@ -29,12 +29,17 @@ function createFileItem(rootFolder: vscode.WorkspaceFolder, uri: vscode.Uri): It
     };
 }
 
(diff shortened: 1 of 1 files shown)


StephenWeatherford:
I missed this, will take a look on Monday.

Issue #959, opened 2019-05-23
Excessive Docker polling


jeff-h:
Issue Type: <b>Performance Issue</b>

VS Code version: Code - Insiders 1.35.0-insider (955afd60a5e4177d1fdf1c3a4a02dc4e0ca7d781, 2019-05-22T18:03:07.494Z)
OS version: Darwin x64 18.6.0
- Remote Extension/Connection Type:Docker

Running macOS, with Docker for Mac CE 18.06.

When clicking around inside the Docker sidebar, VSCode appears to start polling Docker for the data it is displaying — I noticed in Console.app that (I think) this is polling `/containers/json` and `images/json`.

Unfortunately each of these is polled once per second, for each kind of information, for every window you might have used the Docker sidebar in, resulting in multiple polls per second if you view enough data in the Docker sidebar. 

Even worse, the polling continues even when you exit the Docker sidebar. In my case this often results in Docker using 50% CPU even with no containers running.

Is this unique to my setup perhaps? Happy to provide further info if I can.

EricJizbaMSFT:
@jeff-h as a workaround you can adjust the VS Code setting `docker.explorerRefreshInterval`, which defaults to 1 second.

I'll leave this issue open, though, because it feels like we might be able to stop polling if the Docker sidebar isn't open.

jeff-h:
I'll give that a try, thanks! However, stopping the polling when the sidebar isn't open would be much better.

Issue #948, opened 2019-05-21
Please add the confirmation before Remove container!!!


grapewheel:
Coz of the menu of the **Restart container** is **BELOW** the **Remove container**, and I used to be clicked the wrong way to remove container, and there was **NOT** any **CONFIMATION** to remove container !!

![深度截图_20190521140905](https://user-images.githubusercontent.com/490426/58072387-1fe59900-7bd2-11e9-87b8-3c8fab075fd3.jpg)

Pleaz fix the BUG! Rerun a mysql container is suffered for anybody!

Issue #937, opened 2019-05-14
There opens a new tab when executing the action 'View Task Logs' every time 


v-xuanzh:
**OS**: Windows
**Extension Build:** [20190513.1
](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=5945)
**Steps to reproduce**:
1. Create a build task then execute 'Run Task' action.
2. Right click on a task node then select 'View Task Logs'.
3. Make sure the task log opens well, then right click the same task node -> Select 'View Task Logs'.

**Expect Experience:**

There only shows one opened tab.

**Actual Experience:**

There opens a new tab even though the task log has been opened.
![image](https://user-images.githubusercontent.com/34729022/57682796-63933e00-75e7-11e9-993c-b6ae7a0477c1.png)

Issue #935, opened 2019-05-13
We should better handle expired credentials for Docker Hub


Dhwanil20:
Repro steps:

1.
2.

Action: getChildren
Error type: Error
Error Message: Docker: Please log out of Docker Hub and then log in again.

Version: 0.6.2
OS: win32
Product: Visual Studio Code - Insiders
Product Version: 1.34.20-insider
Language: en

<details>
<summary>Call Stack</summary>

```
Object.module.exports.t.getUser extension.bundle.js:1:644951
process._tickCallback next_tick.js:68:7
```

</details>

StephenWeatherford:
@Dhwanil20 Thanks for reporting. It would be nice if we handled this better.  Can you verify that logging out (via right-clicking on the Docker Hub node) and logging back in does work?  Thx.

Dhwanil20:
@StephenWeatherford Hello, it's Click On Docker Hub After Getting Error And Also Today i have not start docker and i have open Vs Code after same Error Occur Please check That's

StephenWeatherford:
We're saving credentials that have expired, did you try this?

> Can you verify that logging out (via right-clicking on the Docker Hub node) and logging back in does work? Thx.
![image](https://user-images.githubusercontent.com/6913354/58203261-8f20c300-7c8e-11e9-8dee-fe0f9064605a.png)

PR #966, opened 2019-06-04
Docker Explorer 2.0

EricJizbaMSFT:
# Summary

I set out with the goal to move the Docker explorer over to the [shared tree implementation](https://github.com/microsoft/vscode-azuretools/tree/master/ui#azure-extension-tree-data-provider) used by our other extensions. It was a bit more of a ...massive... undertaking than I expected because I basically had to tear everything down and rebuild it. In the end, I think I fixed a lot of bugs and improved a lot of things, but most importantly this codebase will be much easier for us to maintain because it's now based on our shared building blocks.

We are going to need to have people (on our team and hopefully outside of our team) bash against this for a little while before releasing. With that in mind, any time spent reviewing this is much appreciated!

## **Help Wanted!** Please follow [these instructions](https://aka.ms/azcodeinsiders) to install [this build](https://aka.ms/AA58wrg) and let us know what you think!

# THE NEW EXPLORER

It features separate views for containers/images/registries - which means you can hide/resize/reorder each view on its own.

Screen Shot 2019-06-04 at 9 53 35 AM

It also has in-line error messages when failing to connect, which IMO is much better than that big scary error message we used to have.

Screen Shot 2019-06-04 at 9 54 29 AM

# THE NEW COMMAND PALETTE

Just like the above, the commands are all grouped based on images/containers/registries. It will also be a one-to-one mapping of the tree. So if you have the tree grouped/sorted/etc., you will see that when running from the command palette.

Screen Shot 2019-06-04 at 9 55 54 AM

Since Docker previously had the option for "All Containers", I wanted to maintain and improve on that functionality. Now you will see a multi-select quick pick:

Screen Shot 2019-06-04 at 9 56 33 AM

# THE NEW ACR TASK VIEW

The old way of viewing tasks was through a web view that honestly wasn't accessible, wasn't consistent with other VS Code experiences, and was going to be annoying to maintain going forward. I moved all tasks into the tree. You can still view properties, copy image digest, and view logs - but you can't write a custom filter for viewing tasks. I still think this is a net-positive experience and we can theoretically add back support for filtering in this new world.

Screen Shot 2019-06-04 at 10 52 41 AM (2)

# More detailed list of fixes

### Leverage tree from shared package
- Subscription filter button: Fixes https://github.com/microsoft/vscode-docker/issues/512
- Add "Deploy to Azure" in command palette: Fixes https://github.com/microsoft/vscode-docker/issues/551
- Support "Load more...": Fixes https://github.com/microsoft/vscode-docker/issues/799
- Don't use `node.label`: Fixes https://github.com/microsoft/vscode-docker/issues/842
- Fix excessive docker polling: Fixes https://github.com/microsoft/vscode-docker/issues/959
- Use multi-select quick pick for commands like start/stop/remove: Related to https://github.com/microsoft/vscode-docker/issues/331
- Show task logs in the tree instead of a webview. Fixes https://github.com/microsoft/vscode-docker/issues/661, Fixes https://github.com/microsoft/vscode-docker/issues/937, Fixes https://github.com/microsoft/vscode-docker/issues/853, and Fixes https://github.com/microsoft/vscode-docker/issues/641

### Leverage wizard from shared package
- Incorrect wizard step count: Fixes https://github.com/microsoft/vscode-docker/issues/442
- Consistently use ignoreFocusOut: Fixes https://github.com/microsoft/vscode-docker/issues/810
- Use async validation of web app name: Fixes https://github.com/microsoft/vscode-docker/issues/809
- Use ext.ui.showQuickPick everywhere: Fixes https://github.com/microsoft/vscode-docker/issues/355

### Leverage consistent error handling/telemetry from shared package
- dockerHubUtils isn't properly handling errors: Fixes https://github.com/microsoft/vscode-docker/issues/885
- Wasn't showing errors when failed to log in to Docker Hub: Fixes https://github.com/microsoft/vscode-docker/issues/360
- Confusing error when pulling images: Fixes https://github.com/microsoft/vscode-docker/issues/549

### Other:
- Better handle expired credentials for Docker Hub: Fixes https://github.com/microsoft/vscode-docker/issues/935 and Fixes https://github.com/microsoft/vscode-docker/issues/882
    - We were saving the token, username, and password in keytar, but never using the username/password again. Rather than saving/re-using a token that is likely to expire, we should actually leverage the username/password from keytar and get a new token.
- Show organizations in Docker Hub tree: Fixes https://github.com/microsoft/vscode-docker/issues/466
- Add "Start Container...": Fixes https://github.com/microsoft/vscode-docker/issues/279
- Add prune command for just images or just containers: Fixes https://github.com/microsoft/vscode-docker/issues/789
- Add confirmation message before all destructive actions: Fixes https://github.com/microsoft/vscode-docker/issues/948
- Add pull option for repositories and tags in ACR: Fixes https://github.com/microsoft/vscode-docker/issues/382
- Make naming of tag/image more consistent: Fixes https://github.com/microsoft/vscode-docker/issues/837, Fixes https://github.com/microsoft/vscode-docker/issues/775, and Fixes https://github.com/microsoft/vscode-docker/issues/491
- Move to shared library: Fixes https://github.com/microsoft/vscode-docker/issues/372


1138-4EB:
I'm trying 0.6.3-alpha from https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=6664 on Win10 with VSC:

```
Version: 1.35.1 (user setup)
Commit: c7d83e57cd18f18026a8162d042843bda1bcf21f
Date: 2019-06-12T14:30:02.622Z
Electron: 3.1.8
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Windows_NT x64 10.0.17134
```

> THE NEW EXPLORER

This is great. I was wondering if there was anything wrong with my setup before I saw this '2.0' issue.

> It also has in-line error messages when failing to connect

This works ok. I restarted docker-for-win, to test it. Both showing the messages and automatically reconnecting again.

> Just like the above, the commands are all grouped based on images/containers/registries.

This is great too.

> Since Docker previously had the option for "All Containers", I wanted to maintain and improve on that functionality. Now you will see a multi-select quick pick:

I don't know if this is the desired behaviour, but I found myself trying it multiple times until I guessed it:

- F1
- Docke Cont Start
- Enter
- ArrowDown
- Enter

Nothing happens. Due to the checkbox, the last 'Enter' has no effect, although I 'selected' a container with the arrow key. For it to work, I need to press 'Space' before the last 'Enter'. This will select the container.

I'd say that requiring to press 'space' is unnecessary when the users want to start/stop/whatever a single container.

Note that this does not apply if you do some of the steps with the mouse, because clicking on a container will select it only.

> THE NEW ACR TASK VIEW

I am not a user of ACR.

EricJizbaMSFT:
@1138-4EB thanks for the feedback!

Regarding the space bar vs. enter - that confused me as well when I first tried it. Created a new issue here: https://github.com/microsoft/vscode-docker/issues/1005

PR #983, opened 2019-06-11
make order of container / image name configurable

tfenster:
If you have multiple containers from the same repo / image, the container list can be difficult to use as you need to make it very wide to be able to distinguish the containers 

image

As you can see, the first and third container are not distinguishable and for the second the eye also has to travel some way to the right to find the difference. In that case it would be good to have the config setting proposed in my PR which shows the container name first

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -892,6 +892,11 @@
                     "type": "boolean",
                     "default": true,
                     "description": "Run docker-compose with the --d (detached) argument, defaults to true"
+                },
+                "docker.imageFirst": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "When displaying containers, show the image first if true, otherwise show the container name first; defaults to true"
                 }
             }
         },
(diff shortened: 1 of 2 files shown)


tfenster:
The alternative would be to allow to configure some kind of "template" like {name} - {status} - {image} but that seems too complicated from my point of view. If you would like that more, I would still look into implementing it

EricJizbaMSFT:
@tfenster we've had a lot of asks for many different configurations, so I've actually been working on generic setting support for all views (containers, images, and volumes). It's a work in progress, but you're welcome to take a look:

https://github.com/microsoft/vscode-docker/tree/ej/configExplorer

That should be in PR in the next few days, so I think we can close this (assuming that covers your scenario).

A few screenshots of the proposed experience:
Screen Shot 2019-06-11 at 2 48 24 PM
Screen Shot 2019-06-11 at 2 55 00 PM
Screen Shot 2019-06-11 at 2 55 19 PM
Screen Shot 2019-06-11 at 2 56 29 PM

tfenster:
That looks awesome, thanks!

Issue #962, opened 2019-05-29
Setting "docker.host" description is misleading


joaomoreno:
Testing https://github.com/microsoft/vscode-remote-release/issues/474

The `docker.host` setting has the following description:

> Full host address to connect to (with protocol). Equivalent to setting the DOCKER_HOST environment variable

This is misleading, as reported by @benad in https://github.com/microsoft/vscode-docker/issues/580#issuecomment-494921010.

It just doesn't work:

![image](https://user-images.githubusercontent.com/22350/58546355-73c13500-8205-11e9-9c44-54bb012e7698.png)

It works once I remove the `tcp://` prefix.

---

Either fix the description or, better yet, handle the prefix correctly.

wizcas:
I doesn't works for me even if I removed the `tcp://` prefix

EricJizbaMSFT:
@joaomoreno @wizcas this should now work with or without "tcp://". Feel free to try the latest build from master with these instructions: https://aka.ms/azCodeInsiders

Issue #914, opened 2019-05-01
How to setup configs with docker-machine.


trmaphi:
Currently, I'm on mac os and installed docker-toolbox using brew with version info
```bash
~ brew cask info docker-toolbox
docker-toolbox: 18.06.1-ce
https://www.docker.com/products/docker-toolbox
/usr/local/Caskroom/docker-toolbox/18.06.1-ce (205.3MB)
From: https://github.com/Homebrew/homebrew-cask/blob/master/Casks/docker-toolbox.rb
==> Name
Docker Toolbox
==> Artifacts
DockerToolbox-18.06.1-ce.pkg (Pkg)

```
I read [docs]( https://github.com/Microsoft/vscode-docker#connecting-to-docker-machine) on setup with docker-machine. I got the value of docker host from
```bash
~ docker-machine env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.100:2376"
export DOCKER_CERT_PATH="/Users/trmaphi/.docker/machine/machines/default"
export DOCKER_MACHINE_NAME="default"
# Run this command to configure your shell: 
# eval $(docker-machine env)
```
Then, I copied DOCKER_HOST value inside global `settings.json`, but I got message noti from vscode said `Unable to connect to Docker. Please make sure you have installed Docker and that it is running. Details: connect ECONNREFUSED 127.0.0.1:80`

How do I config my global `settings.json` to connect this extension with docker-machine

atomi:
I think using `DOCKER_HOST` is a much better way to do remote builds.    
Preferably, a per project setting for `DOCKER_HOST`, `DOCKER_TLS_VERIFY` and `DOCKER_CERT_PATH`.

trmaphi:
@atomi could you show me `.vscode/settings.json` in your workspace?

atomi:
@trmaphi I don't use a .vscode/settings.json file.    
By the way VSCode also passes environment variables set in your shell.    
So you can set them there also.

PrashanthCorp:
Is this an issue only with the tree view, or are commands sent to shell affected too? 


We rely on  dockerode for populating the tree view, and we've seen issues ensuring docker_host is honored. We shall look into this.

EricJizbaMSFT:
Hi all - just merged #998 to support the following VS Code settings:

`docker.host` - equivalent to `DOCKER_HOST`
`docker.certPath` - equivalent to `DOCKER_CERT_PATH`
[...]

Issue #649, opened 2018-11-26
Wrong DOCKER_HOST config when using docker.host configuration


ntcong:
vscode-docker tell me to set docker.host in the format \<ip\>:\<port\>. When using Docker build image command, it'll set DOCKER_HOST = docker.host config.

DOCKER_HOST config should be tcp://\<ip\>:\<port\>. Using \<ip\>:\<port\> will result in an error:
`parse <ip>:<port>: first path segment in URL cannot contain colon`

ntcong:
changing https://github.com/Microsoft/vscode-docker/blob/4d7b773fabd8c210ff7392bcc1daffc66b862004/commands/utils/TerminalProvider.ts#L20 from `DOCKER_HOST: value` to `DOCKER_HOST: "tcp://" + value` works. But i'm not sure we should use the correct DOCKER_HOST format (tcp://host:port) for docker.host configuration, or add `tcp://` to the DOCKER_HOST environment

DefineFC:
same here, thanks for the workaround!

PrashanthCorp:
Hey @ntcong, you are correct. Looking at how the host prefix is used by the upstream library (docker-modem), there seem to be multiple allowed protocols. I don't think it's appropriate for us to make assumptions about wanting a tcp connection. We'll look into making this more visible.

zxfishhack:
setting docker.host to tcp://host:port does not work for me.

source from docker-modem 
```javascript
    var parsed = url.parse(self.host);
    address = url.format({
      'protocol': parsed.protocol || self.protocol,
      'hostname': parsed.hostname || self.host,
      'port': self.port
    });
    address = url.resolve(address, options.path);
```
when host is `tcp://192.168.1.1`, options.path is `/containers`, self.port is 8080, the final address will be 'tcp:/containers'.

[DockerClient::refreshEndpoint](https://github.com/Microsoft/vscode-docker/blob/master/commands/utils/docker-endpoint.ts#L27), the errorMessage defind is 'The docker.host configuration setting must be entered as <host>:<port>, e.g. dockerhost:2375'

IMO, `vscode-docker` accept <host>:<port> only and export DOCKER_HOST=tcp://<host>:<port> to terminal is better.

Issue #580, opened 2018-10-23
Unable to connect to Docker Machine


eduardonunesp:
Can't connect to `docker` even settings the `DOCKER_HOST` env var and on config file with `DOCKER_HOST` (`tcp://192.168.99.100:2376`)

Does that extension ever worked on `macOS` and `Windows`? 

macOS: `10.14`
VSCode:  `Version 1.28.2`

StephenWeatherford:
@eduardonunesp Sorry you're having troubles.  So you're trying to connect to Docker Machine?  Have you followed the instructions at https://github.com/microsoft/vscode-docker#connecting-to-docker-machine?

eduardonunesp:
@StephenWeatherford NP, sorry if I sounded rude, actually I have set the `environment variable` also the `docker.host` option on config file without success. I have tried also an year ago and it didn't work also

hawm:
May slove if you docker machine  enable TLS
https://stackoverflow.com/questions/51402080/vscode-docker-extension-cannot-connect-to-docker-machine

mrkz55:
Same issue on MacOS 10.14 running VSCode 1.30.2 (1.30.2). Is there a ballpark timeframe for the PR on this one?

dr3x:
Same issue in Windows 10:

docker-machine start
& "C:\ProgramData\chocolatey\lib\docker-machine\bin\docker-machine.exe" env | Invoke-Expression
docker-machine info
docker build --rm -f "Dockerfile" -t dr3x/dotnet-docker-vscode:latest .
error during connect: Post http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.39/build?
...
docker-vscode%3Alatest&target=&ulimits=null&version=1: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.

akshaybabloo:
I tried this setting:

```json
"docker.host": "tcp://192.168.99.100:2376",
    "docker.importCertificates": {
        "useCertificateStore": true,
        "certificatePaths": [
            "/etc/ssl/certs/ca-certificates",
            "/etc/openssl/certs",
            "/etc/pki/tls/certs",
            "/usr/local/share/certs",
            "C:\\Users\\aksha\\.docker\\machine\\machines\\default"
        ]
    }
```

I get an error as 

```
[...]

PR #998, opened 2019-06-17
Support all `DOCKER_*` environment variables as VS Code settings

EricJizbaMSFT:
Previously we only supported "docker.host" (aka "DOCKER_HOST") as a VS Code setting. I added support for the 4 main environment variables used in [docker machine](https://docs.docker.com/machine/get-started/). Users want these as VS Code settings so that they can have per-workspace setups and so that they can modify a setting without reloading VS Code or changing their environment variables.

Also we were parsing "docker.host" ourselves in the past, which led to a lot of problems because our parsing logic was _very_ simplistic. Instead, I will let docker_modem (a dependency of Dockerode) do all the parsing for us. See [here](https://github.com/apocas/docker-modem/blob/master/lib/modem.js#L26) for more info.

Fixes https://github.com/microsoft/vscode-docker/issues/962
Fixes https://github.com/microsoft/vscode-docker/issues/914
Fixes https://github.com/microsoft/vscode-docker/issues/649
Fixes https://github.com/microsoft/vscode-docker/issues/580
Fixes https://github.com/microsoft/vscode-docker/issues/216

Related to https://github.com/microsoft/vscode-docker/issues/646

### DIFF
--- a/src/tree/LocalRootTreeItemBase.ts
+++ b/src/tree/LocalRootTreeItemBase.ts
@@ -6,6 +6,7 @@
 import { ConfigurationChangeEvent, ConfigurationTarget, TreeView, TreeViewVisibilityChangeEvent, workspace, WorkspaceConfiguration } from "vscode";
 import { AzExtParentTreeItem, AzExtTreeItem, AzureWizard, GenericTreeItem, IActionContext, InvalidTreeItem, registerEvent } from "vscode-azureextensionui";
 import { configPrefix } from "../constants";
+import { ext } from "../extensionVariables";
 import { nonNullProp } from "../utils/nonNull";
 import { isLinux } from "../utils/osUtils";
 import { getThemedIconPath } from "./IconPath";
@@ -293,8 +294,12 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
     }
 
     private async getSortedItems(): Promise<TItem[]> {
-        const items: TItem[] = await this.getItems() || [];
-        return items.sort((a, b) => a.treeId.localeCompare(b.treeId));
+        if (ext.dockerodeError === undefined) {
+            const items: TItem[] = await this.getItems() || [];
+            return items.sort((a, b) => a.treeId.localeCompare(b.treeId));
+        } else {
+            throw ext.dockerodeError;
+        }
     }
 
     private async hasChanged(): Promise<boolean> {
--- a/src/tree/LocalRootTreeItemBase.ts
+++ b/src/tree/LocalRootTreeItemBase.ts
@@ -294,11 +294,11 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
     }
 
     private async getSortedItems(): Promise<TItem[]> {
-        if (ext.dockerodeError === undefined) {
+        if (ext.dockerodeInitError === undefined) {
             const items: TItem[] = await this.getItems() || [];
             return items.sort((a, b) => a.treeId.localeCompare(b.treeId));
         } else {
-            throw ext.dockerodeError;
+            throw ext.dockerodeInitError;
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/tree/LocalRootTreeItemBase.ts
```diff
@@ -293,8 +294,12 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
     }
 
     private async getSortedItems(): Promise<TItem[]> {
-        const items: TItem[] = await this.getItems() || [];
-        return items.sort((a, b) => a.treeId.localeCompare(b.treeId));
+        if (ext.dockerodeError === undefined) {
```
nturinski: Why have this attached to the `extensionVariables` rather than just bubbling the error up from `refreshDockerode`?

src/utils/addDockerSettingsToEnv.ts
```diff
@@ -0,0 +1,28 @@
...
+import { ext } from '../extensionVariables';
+
+export function addDockerSettingsToEnv(env: {}): void {
+    addDockerSettingToEnv("host", 'DOCKER_HOST', env);
+    addDockerSettingToEnv("certPath", 'DOCKER_CERT_PATH', env);
+    addDockerSettingToEnv("tlsVerify", 'DOCKER_TLS_VERIFY', env);
+    addDockerSettingToEnv("machineName", 'DOCKER_MACHINE_NAME', env);
+}
+
+function addDockerSettingToEnv(settingKey: string, envVar: string, env: {}): void {
```
nturinski: So VS Code settings are respected above env settings that may currently exist?  For instance, `DOCKER_HOST` would get overridden if I had the `host` setting in my project?

EricJizbaMSFT: Yeah, that's what users requested. They might have the environment variable set on their machine, but they want VS Code settings to override that on a per-workspace level

nturinski: It definitely makes sense, but I was wondering if we should alert users in the output that their env variable is getting overwritten.  I could totally see myself forgetting I had a setting and wondering why my environment variables weren't being used 😅

src/tree/LocalRootTreeItemBase.ts
EricJizbaMSFT: Because refreshDockerode gets called in two places:
1. On extension activation
1. Every time you update a docker setting

I feel like showing an error in either of those cases is too aggressive and would get annoying. I would rather delay the error and show it in a place that's actually relevant (aka the tree).

nturinski: I just don't like that it has to be manually reset to `undefined`.  I could see a case where the error is stale or not attached to the right tree node.  What happens if you update a bunch of docker settings at once??

src/utils/addDockerSettingsToEnv.ts
EricJizbaMSFT: Added a warning

src/tree/LocalRootTreeItemBase.ts
EricJizbaMSFT: This is the current experience
Screen Shot 2019-06-17 at 3 47 46 PM

This is what you might get if I showed the error in refreshDockerode:
Screen Shot 2019-06-17 at 3 49 35 PM

I'm open to suggestions for refactoring the code, but I definitely don't want to change the user experience.

>  I could see a case where the error is stale or not attached to the right tree node.

Worst case the user has to hit refresh and it won't be stale any more. I don't think it can be attached to the wrong tree node. This logic is unique to `LocalRootTreeItemBase`, which is the root node for all views (containers, images, volumes, etc.)

> What happens if you update a bunch of docker settings at once??


It works? I'm not entirely sure what your concern is.

nturinski: I was wondering if you could have errors from one treeItem and have it affect another valid treeItem since it's only checking of the `ext.dockerodeError` is equal to undefined.  But I'm not familiar with the code flow for this, so if it doesn't seem like a realistic scenario, I'm totally fine with dropping it.  Just wanted to get the reason since it's not a pattern I've seen you use before.

Issue #1008, opened 2019-06-18
command 'vscode-docker.images.selectGroupBy' already exists


tvilot:
Repro steps:

1.
2.

Action: docker.activate
Error type: Error
Error Message: command 'vscode-docker.images.selectGroupBy' already exists

Version: 0.6.3
OS: linux
Product: Visual Studio Code
Product Version: 1.35.1
Language: en

<details>
<summary>Call Stack</summary>

```
d.registerCommand extensionHostProcess.js:497:447
Object.registerCommand extensionHostProcess.js:716:422
Object.module.exports.t.registerCommand extension.bundle.js:211:110531
ue extension.bundle.js:157:609314
Object.<anonymous> extension.bundle.js:157:611927
Object.<anonymous> extension.bundle.js:1:394616
module.exports.n extension.bundle.js:1:392954
module.exports.n extension.bundle.js:1:392731
Object.module.exports.t.callWithTelemetryAndErrorHandling extension.bundle.js:1:394533
Object.module.exports.t.activateInternal extension.bundle.js:157:610822
```

</details>

EricJizbaMSFT:
@tvilot we just changed publisher from "PeterJausovec" to "ms-azuretools". It's possible your machine still has an extension under the old publisher. Can you try reloading VS Code?

If that doesn't work - you could also try manually removing the folder like "PeterJausovec.vscode-docker" from your extensions folder:
* Windows `%USERPROFILE%\.vscode\extensions`
* macOS `~/.vscode/extensions`
* Linux `~/.vscode/extensions`

rrmistry:
@EricJizbaMSFT , a clean uninstall -> reload VS Code -> install of `ms-azuretools.vscode-docker` seems to have resolved the issue. Thank you!

arush0:
I followed @rrmistry steps but I still have issues with the extension. vscode recommends the extension even though I have it installed.
<img width="437" alt="image" src="https://user-images.githubusercontent.com/32102766/59767582-9aa4e100-9270-11e9-8d32-d919469d60d4.png">

rrmistry:
Hello @arush0 , I was not precise in my comment. I had uninstalled the docker extension and not VS Code. As far as I understand, extensions do not get uninstalled when you uninstall VS code.

My apologies for the unintentional misdirection.

arush0:
[...]

PR #1067, opened 2019-06-19
Workaround for multiple publishers installed

EricJizbaMSFT:
Potential temporary workaround for https://github.com/microsoft/vscode-docker/issues/1008

Still testing, but wanted to get eyes on it ASAP. I basically check if both extensions are installed and defer to the old publisher for all functionality.

### DIFF
--- a/extension.ts
+++ b/extension.ts
@@ -112,6 +112,15 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats: 
     this.properties.isActivationEvent = 'true';
     this.measurements.mainFileLoad = (perfStats.loadEndTime - perfStats.loadStartTime) / 1000;
 
+    const extension = vscode.extensions.getExtension('PeterJausovec.vscode-docker');
+    if (extension) {
+      // temporary workaround for https://github.com/microsoft/vscode-docker/issues/1008
+      // We will essentially defer all functionality to the old publisher's extension if we detect both installed
+      vscode.commands.executeCommand('setContext', 'isOldPublisherInstalled', true);
+      this.properties.isOldPublisherInstalled = 'true';
+      return;
+    }
+
     ctx.subscriptions.push(
       vscode.languages.registerCompletionItemProvider(
         DOCUMENT_SELECTOR,
--- a/extension.ts
+++ b/extension.ts
@@ -116,8 +116,8 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats: 
     if (extension) {
       // temporary workaround for https://github.com/microsoft/vscode-docker/issues/1008
       // We will essentially defer all functionality to the old publisher's extension if we detect both installed
-      vscode.commands.executeCommand('setContext', 'isOldPublisherInstalled', true);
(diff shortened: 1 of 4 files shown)


extension.ts
```diff
@@ -112,6 +112,15 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
...
     this.measurements.mainFileLoad = (perfStats.loadEndTime - perfStats.loadStartTime) / 1000;
 
+    const extension = vscode.extensions.getExtension('PeterJausovec.vscode-docker');
+    if (extension) {
+      // temporary workaround for https://github.com/microsoft/vscode-docker/issues/1008
+      // We will essentially defer all functionality to the old publisher's extension if we detect both installed
+      vscode.commands.executeCommand('setContext', 'isOldPublisherInstalled', true);
+      this.properties.isOldPublisherInstalled = 'true';
+      return;
+    }
```
PrashanthCorp: With this PR, the extension would hide all commands when it recognizes the old extension. I don't see code at the registerCommand sites to ensure the command is not run. 
What happens when a dependent extension calls say `docker.deployToWebApp` before the extension is activated? Does : 
1. The command hit both extensions? Then this code will make sure the commands are hidden, but the command might still run, while also running the older extension copy
2. It pings this extension, which prevents the other extension from running the command.

EricJizbaMSFT: The key part is the `return` line. This extension will never even register the commands if we detect the old publisher.

extension.ts
```diff
@@ -112,6 +112,15 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
     this.properties.isActivationEvent = 'true';
     this.measurements.mainFileLoad = (perfStats.loadEndTime - perfStats.loadStartTime) / 1000;
 
+    const extension = vscode.extensions.getExtension('PeterJausovec.vscode-docker');
+    if (extension) {
+      // temporary workaround for https://github.com/microsoft/vscode-docker/issues/1008
+      // We will essentially defer all functionality to the old publisher's extension if we detect both installed
+      vscode.commands.executeCommand('setContext', 'isOldPublisherInstalled', true);
```
StephenWeatherford: The context is global, right?  Key should be specific to the extension.

PrashanthCorp:
LGTM

extension.ts
StephenWeatherford: Should we alert them and ask to uninstall the old one?  Otherwise how will they ever see new functionality?

EricJizbaMSFT: The problem is that if they uninstall the old one, they will immediately be recommended to install it again if they open a Dockerfile. So yes - we need to do that, but not until VS Code stable ships with the updated recommendation id

I think it makes sense to include that logic in our next actual release. Currently this release and the old release have the exact same functionality. I want to be able to say "uninstall the old one _so that you can get this new functionality_"

extension.ts
EricJizbaMSFT: Changed to `isOldDockerPublisherInstalled`

PR #876, opened 2019-04-02
Implement webhooks, version 2

PrashanthCorp:
Identical to #862. I've redone it so that the diff is cleaner.

PR #862, opened 2019-03-20
Create webhook on deploying to App service

PrashanthCorp:

PR #1085, opened 2019-06-21
 Create webhook on app service creation - based on the newer explorer

PrashanthCorp:
Ref; #876, #862


src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -55,11 +60,64 @@ export async function deployImageToAzure(context: IActionContext, node?: RemoteT
...
     const site = nonNullProp(wizardContext, 'site');
     const createdNewWebApp: string = `Successfully created web app "${site.name}": https://${site.defaultHostName}`;
+    if (node instanceof AzureTagTreeItem) {
+        let siteClient = new SiteClient(site, node.parent.parent.parent.root);
+        let appUri: string = (await siteClient.getWebAppPublishCredential()).scmUri;
+        await createWebhookForApp(node, wizardContext, appUri);
+    } else if (node instanceof DockerHubTagTreeItem) {
+        // point to dockerhub to create a webhook
+        // http://cloud.docker.com/repository/docker/<registryName>/<repoName>/webHooks
+        const dockerhubPrompt: string = "Copy web app endpoint and browse to dockerhub";
```
EricJizbaMSFT: Can you show a screenshot? This seems like a big button...

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -55,11 +60,64 @@ export async function deployImageToAzure(context: IActionContext, node?: RemoteT
...
+        let response: string = await vscode.window.showInformationMessage("Please browse to your dockerhub account to set up a CI/CD webhook", dockerhubPrompt);
+        if (response) {
+            let appUri = `https://${site.name}.scm.azurewebsites.net/docker/hook`;
```
EricJizbaMSFT: Use the url from SiteClient instead of constructing it yourself. I don't think this will be sovereign-friendly as-is

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -55,11 +60,64 @@ export async function deployImageToAzure(context: IActionContext, node?: RemoteT
...
+            // tslint:disable-next-line:no-unsafe-any
```
EricJizbaMSFT: What are these tslint disables for? Are they necessary?

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -55,11 +60,64 @@ export async function deployImageToAzure(context: IActionContext, node?: RemoteT
...
+            await vscode.env.clipboard.writeText(appUri);
+            // account.name, repository.name
+            // tslint:disable-next-line:no-unsafe-any
+            vscode.env.openExternal(vscode.Uri.parse(`https://cloud.docker.com/repository/docker/${node.parent.parent.parent.username}/${node.parent.repoName}/webHooks`));
```
EricJizbaMSFT: Use this: https://github.com/microsoft/vscode-docker/blob/master/src/utils/openExternal.ts

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -55,11 +60,64 @@ export async function deployImageToAzure(context: IActionContext, node?: RemoteT
...
+    const crmClient = createAzureClient(node.parent.parent.parent.root, ContainerRegistryManagementClient);
+    let registryResourceGroupName: string = node.parent.parent.resourceGroup;
+    let registryName = node.parent.parent.registryName;
+
+    const registryList = await crmClient.registries.list();
+    const registryHandle = registryList.find((value) => value.name === registryName);
+    if (!registryHandle) {
+        return;
+    }
+    let webhookLocation: string = registryHandle.location;
```
EricJizbaMSFT: Looks like `location` already exists on `AzureRegistryTreeItem._registry`, so you can use that. (It's a private variable, but you can expose it similar to `registryName`).

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -55,11 +60,64 @@ export async function deployImageToAzure(context: IActionContext, node?: RemoteT
...
+    let webhookCreateParameters: WebhookCreateParameters = {
+        location: webhookLocation,
+        serviceUri: appUri,
+        scope: `${node.parent.repoName}:${node.tag}`,
+        actions: ["push"],
+        status: 'enabled'
+    };
+
+    const webhook = await crmClient.webhooks.create(registryResourceGroupName, registryName, webhookName, webhookCreateParameters);
+    ext.outputChannel.appendLine(`Created webhook ${webhook.name} with scope ${webhook.scope}, id: ${webhook.id} and location: ${webhook.location}`);
```
EricJizbaMSFT: Again - can you post a screenshot so that I can see what this message looks like?

Also generally wrap any dynamically inserted things (like `${webhook.name}`) in quotes

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -55,11 +60,64 @@ export async function deployImageToAzure(context: IActionContext, node?: RemoteT
...
+
     ext.outputChannel.appendLine(createdNewWebApp);
     // don't wait
     window.showInformationMessage(createdNewWebApp);
 }
 
+async function createWebhookForApp(node: AzureTagTreeItem, wizardContext: IActionContext & Partial<IAppServiceWizardContext>, appUri: string): Promise<void> {
+    // fields derived from the app service wizard
+    let siteName: string = wizardContext.site.name;
+    let webhookName: string = `webapp${siteName}`;
```
EricJizbaMSFT: how did you choose this name?

PrashanthCorp: I'm mirroring what the portal-created webhook is called.

src/commands/registries/azure/deployImageToAzure.ts
PrashanthCorp: outputwebhook

src/commands/registries/azure/deployImageToAzure.ts
PrashanthCorp: Yes 

image

Any suggestions on the naming?

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -141,3 +148,68 @@ class DockerSiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardConte
         return !context.site;
     }
 }
+
+class DockerWebhookCreateStep extends AzureWizardExecuteStep<IAppServiceWizardContext> {
```
nturinski: I feel like there's enough code here to just be contained in its own file.

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -141,3 +148,68 @@ class DockerSiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardConte
...
+            if (response) {
+                await vscode.env.clipboard.writeText(appUri);
+                await openExternal(`https://cloud.docker.com/repository/docker/${this._treeItem.parent.parent.parent.username}/${this._treeItem.parent.repoName}/webHooks`);
+            }
+        }
+
+    }
+
+    public shouldExecute(context: IAppServiceWizardContext): boolean {
+        return !!context.site;
```
nturinski: Shouldn't this only execute if it's an AzureTagTreeItem or DockerHubTagTreeItem?  It seems like if the treeItem isn't either of these types, it'll display the output message without actually executing anything.

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -141,3 +148,68 @@ class DockerSiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardConte
...
+        if (this._treeItem instanceof AzureTagTreeItem) {
+            await this.createWebhookForApp(this._treeItem, context, appUri);
+        } else if (this._treeItem instanceof DockerHubTagTreeItem) {
+            // point to dockerhub to create a webhook
+            // http://cloud.docker.com/repository/docker/<registryName>/<repoName>/webHooks
+            const dockerhubPrompt: string = "Copy web app endpoint and browse to dockerhub";
+            let response: string = await vscode.window.showInformationMessage("Please browse to your dockerhub account to set up a CI/CD webhook", dockerhubPrompt);
+            if (response) {
+                await vscode.env.clipboard.writeText(appUri);
+                await openExternal(`https://cloud.docker.com/repository/docker/${this._treeItem.parent.parent.parent.username}/${this._treeItem.parent.repoName}/webHooks`);
```
nturinski: I don't think you should await this since "Creating webhook for webapp "${context.newSiteName}"... won't disappear until the user clicks something from the information window.  I wouldn't expect that behavior since I assume I have to actually create the webhook in cloud.docker.com?

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -141,3 +148,68 @@ class DockerSiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardConte
...
+    public async execute(context: IAppServiceWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+        const creatingNewApp: string = `Creating webhook for webapp "${context.newSiteName}"...`;
+        ext.outputChannel.appendLine(creatingNewApp);
+        progress.report({ message: creatingNewApp });
+
+        const site: Site = nonNullProp(context, 'site');
+        let siteClient = new SiteClient(site, context);
+        let appUri: string = (await siteClient.getWebAppPublishCredential()).scmUri;
+        if (this._treeItem instanceof AzureTagTreeItem) {
+            await this.createWebhookForApp(this._treeItem, context, appUri);
```
nturinski: nit: I'd put a notification once it's created here rather than putting it in the function itself.  That's just personal preference, but I like being able to control the notification where the ui is and keep it separate it from the implementation code.

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -141,3 +148,68 @@ class DockerSiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardConte
...
+        let webhookName: string = `webapp${siteName}`;
+
+        // variables derived from the container registry
+        const crmClient = createAzureClient(node.parent.parent.parent.root, ContainerRegistryManagementClient);
+        let registryResourceGroupName: string = node.parent.parent.resourceGroup;
+        let registryName = node.parent.parent.registryName;
+
+        const existingWebhooks = await crmClient.webhooks.list(registryResourceGroupName, registryName);
+        if (existingWebhooks.find((hook) => hook.name === webhookName)) {
+            return;
```
nturinski: The way it is now, the UI will say "Creating webhook..." but never alert the user that the webhook already exists.  I think it'd be confusing for users since they would expect the UI to indicate that the webhook creation was resolved whether by already existing or being created.

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -141,3 +148,68 @@ class DockerSiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardConte
...
+        }
+
+        let webhookCreateParameters: WebhookCreateParameters = {
+            location: node.parent.parent.registryLocation,
```
nturinski: Can you get the location from the context?  Should the location match the site's location?

src/commands/registries/azure/deployImageToAzure.ts
nturinski: I'd follow this pattern: 
image

PrashanthCorp:
Ping

src/commands/registries/azure/deployImageToAzure.ts
PrashanthCorp: :D I had the same confusion, and it cost me a few days rectifying it. Since the webhook is created under the container registry (which polls the container for changes), the location needs to match the location of the container registry. Not doing so would result in an error saying the container registry is not found (at the location of the site).

src/commands/registries/azure/DockerWebhookCreateStep.ts
```diff
@@ -0,0 +1,82 @@
...
+        if (this._treeItem.parent instanceof AzureRepositoryTreeItem) {
+            const creatingNewApp: string = `Creating webhook for webapp "${context.newSiteName}"...`;
+            ext.outputChannel.appendLine(creatingNewApp);
+            progress.report({ message: creatingNewApp });
+            const webhook = await this.createWebhookForApp(this._treeItem, context, appUri);
+            ext.outputChannel.appendLine(`Created webhook '${webhook.name}' with scope '${webhook.scope}', id: '${webhook.id}' and location: '${webhook.location}'`);
+        } else if (this._treeItem.parent instanceof DockerHubRepositoryTreeItem) {
+            // point to dockerhub to create a webhook
+            // http://cloud.docker.com/repository/docker/<registryName>/<repoName>/webHooks
+            const dockerhubPrompt: string = "Copy web app endpoint and browse to dockerhub";
```
nturinski: Still gotta change this button prompt.  I think how the Azure Account extension does it would be a good model.

src/commands/registries/azure/DockerWebhookCreateStep.ts
```diff
@@ -0,0 +1,82 @@
...
+        // fields derived from the app service wizard
+        let siteName: string = wizardContext.site.name;
+        let baseName = `webapp${siteName}`;
+        let webhookName: string = baseName;
+        // variables derived from the container registry
+        const registryTreeItem: AzureRegistryTreeItem = (<AzureRepositoryTreeItem>node.parent).parent;
+        const crmClient = createAzureClient(registryTreeItem.parent.root, ContainerRegistryManagementClient);
+        const existingWebhooks = await crmClient.webhooks.list(registryTreeItem.resourceGroup, registryTreeItem.registryName);
+        let dedupeCount: number = 0;
+        while (existingWebhooks.find((hook) => hook.name === webhookName)) {
```
nturinski: The way it is now, wouldn't create a new webhook every time you deploy to Azure?  You could probably check the scope of the webhook and verify it doesn't match the scope you are creating.

Also, I'd have a dedupeCount max otherwise there's a potential (albeit low) chance that you'll hit an infinite while loop.

src/commands/registries/azure/DockerWebhookCreateStep.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    public async execute(context: IAppServiceWizardContext, progress: Progress<{
+        message?: string;
+        increment?: number;
+    }>): Promise<void> {
+        const site: Site = nonNullProp(context, 'site');
+        let siteClient = new SiteClient(site, context);
+        let appUri: string = (await siteClient.getWebAppPublishCredential()).scmUri;
+        if (this._treeItem.parent instanceof AzureRepositoryTreeItem) {
+            const creatingNewApp: string = `Creating webhook for webapp "${context.newSiteName}"...`;
```
nturinski: nit: `const creatingNewWebHook` Also web app is usually two words.

src/commands/registries/azure/DockerWebhookCreateStep.ts
PrashanthCorp: Thanks for pointing out. I didn't see the original comment since it was part of another thread.

src/commands/registries/azure/DockerWebhookCreateStep.ts
PrashanthCorp: I feel that the naming needs to be updated. Deploy to Azure doesn't merely deploy, but it also creates a web app. 

The ideal flow is that you deploy only once. Anytime you make a change to your code and build your container, pushing to ACR should update the web app (deploy is done separately)

nturinski: Okay, but the webhooks are all listed by registry rather than by web app then?

src/commands/registries/azure/DockerWebhookCreateStep.ts
```diff
@@ -0,0 +1,83 @@
...
+    }>): Promise<void> {
+        const site: Site = nonNullProp(context, 'site');
+        let siteClient = new SiteClient(site, context);
+        let appUri: string = (await siteClient.getWebAppPublishCredential()).scmUri;
+        if (this._treeItem.parent instanceof AzureRepositoryTreeItem) {
+            const creatingNewWebhook: string = `Creating webhook for web app "${context.newSiteName}"...`;
+            ext.outputChannel.appendLine(creatingNewWebhook);
+            progress.report({ message: creatingNewWebhook });
+            const webhook = await this.createWebhookForApp(this._treeItem, context, appUri);
+            ext.outputChannel.appendLine(`Created webhook '${webhook.name}' with scope '${webhook.scope}', id: '${webhook.id}' and location: '${webhook.location}'`);
```
nturinski: Use double quotes (") rather than single quotes(')

src/commands/registries/azure/DockerWebhookCreateStep.ts
```diff
@@ -0,0 +1,83 @@
...
+        } else if (this._treeItem.parent instanceof DockerHubRepositoryTreeItem) {
+            // point to dockerhub to create a webhook
+            // http://cloud.docker.com/repository/docker/<registryName>/<repoName>/webHooks
+            const dockerhubPrompt: string = "Copy & open";
+            const dockerhubUri: string = `https://cloud.docker.com/repository/docker/${this._treeItem.parent.parent.parent.username}/${this._treeItem.parent.repoName}/webHooks`;
```
nturinski: Could you update the screenshot with the new popup?

src/commands/registries/azure/DockerWebhookCreateStep.ts
PrashanthCorp: Yes. The registry owns the webhooks, and the tier of the registry decides how many webhooks you can have, etc.

src/commands/registries/azure/DockerWebhookCreateStep.ts
PrashanthCorp: image

src/commands/registries/azure/deployImageToAzure.ts
PrashanthCorp: Done

src/commands/registries/azure/DockerWebhookCreateStep.ts
```diff
@@ -0,0 +1,83 @@
...
+        if (this._treeItem.parent instanceof AzureRepositoryTreeItem) {
+            const creatingNewWebhook: string = `Creating webhook for web app "${context.newSiteName}"...`;
+            ext.outputChannel.appendLine(creatingNewWebhook);
+            progress.report({ message: creatingNewWebhook });
+            const webhook = await this.createWebhookForApp(this._treeItem, context, appUri);
+            ext.outputChannel.appendLine(`Created webhook "${webhook.name}" with scope "${webhook.scope}", id: "${webhook.id}" and location: "${webhook.location}"`);
+        } else if (this._treeItem.parent instanceof DockerHubRepositoryTreeItem) {
+            // point to dockerhub to create a webhook
+            // http://cloud.docker.com/repository/docker/<registryName>/<repoName>/webHooks
+            const dockerhubPrompt: string = "Copy & open";
```
nturinski: nit: Copy & Open (just looks more aesthetically pleasing 😃)

src/commands/registries/azure/DockerWebhookCreateStep.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    public shouldExecute(context: IAppServiceWizardContext): boolean {
+        return !!context.site && (this._treeItem.parent instanceof AzureRepositoryTreeItem || this._treeItem.parent instanceof DockerHubRepositoryTreeItem);
+    }
+
+    private async createWebhookForApp(node: RemoteTagTreeItem, wizardContext: IActionContext & Partial<IAppServiceWizardContext>, appUri: string): Promise<Webhook | undefined> {
+        // fields derived from the app service wizard
+        let siteName: string = wizardContext.site.name;
```
nturinski: Rather than passing in wizardContext, I would just pass in the site in the execute method above.  If you are going to use wizardContext, don't you have nonNull check wizardContext.site like you did in line 33?

PrashanthCorp: Did you mean pass in just the site _from_ the execute method? I misread what you wrote initially.

nturinski: Yeah, instead of passing in the whole context, you could just pass in the site into createWebhookForApp.  You don't use the context for anything other than the site and you already did a nonNull check in the execute method.

Issue #1092, opened 2019-06-24
The contents of the opened .json file is empty when executing 'Inspect' action for one volume


v-xuanzh:
**Extension Build:** [20190623.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=7099)
**OS**: Windows

**Steps to reproduce:**
1. Run the command to create a volume
`docker volume create my-vol`
2. Open VOLUMES -> Right click on the volume -> Inspect
3. Observe the contents of opened .json file.

**Expect Experience:**

There shows the information like below.
![image](https://user-images.githubusercontent.com/34729022/60010499-7e5ad700-96aa-11e9-8d12-9ef908707db2.png)

**Actual Experience:**

The contents of the opened .json file is empty.
![image](https://user-images.githubusercontent.com/34729022/60010538-92063d80-96aa-11e9-9bdb-f06b720d77d3.png)

PR #1095, opened 2019-06-26
Fix inspect volume

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-docker/issues/1092


EricJizbaMSFT:
I think the only tslint rule that might catch this would be `typedef` and that would be a massive undertaking haha...

(Since `openReadOnlyJson` takes an object and a Promise is an object)

PR #1097, opened 2019-06-28
Generalize registry support

EricJizbaMSFT:
I'm skeptical that we'll ever have time to support all of the providers listed in https://github.com/microsoft/vscode-docker/issues/869, so my goal with this PR was to make it as easy as possible for the community to add support for their provider. I used GitLab as a test subject, and I'll send out a separate PR so that you can see what that looks like.

The main things that are now generalized:
1. A log-in wizard that allows for url/username/password
1. Caching url/username/password across sessions
1. Pull repos/images
1. Set registry as default

Since the list of providers will theoretically grow and make the registries view cluttered, the default view will now be this:
Screen Shot 2019-06-28 at 10 20 49 AM

Once you select to connect one, you will see this:
Screen Shot 2019-06-28 at 10 13 53 AM

That means you will _only_ see the providers you have connected. This is a breaking change from the previous experience and it means all users will essentially have to re-connect to their registry provider when we ship this update. Here's an example of a few registries that are connected:
Screen Shot 2019-06-28 at 10 14 20 AM
(You may notice one side-benefit of this work is that you can log in to multiple docker hub accounts 🙂)

Fixes https://github.com/microsoft/vscode-docker/issues/1091

PR #1098, opened 2019-06-28
Add GitLab support

EricJizbaMSFT:
I used GitLab as a test subject for https://github.com/microsoft/vscode-docker/pull/1097 (see that PR first). This demonstrates the minimum amount of code the community would have to contribute to add support for a provider.

This adds support for viewing repos/tags, pulling repos/images, and setting a registry as default.

Screen Shot 2019-06-28 at 10 45 16 AM


src/tree/registries/gitLab/gitLabRegistryProvider.ts
```diff
@@ -0,0 +1,21 @@
...
+import { RegistryApi } from "../all/RegistryApi";
+import { IRegistryProvider } from "../IRegistryProvider";
+import { GitLabAccountTreeItem } from "./GitLabAccountTreeItem";
+
+export const gitLabRegistryProvider: IRegistryProvider = {
+    label: "GitLab",
+    id: 'gitLab',
+    api: RegistryApi.GitLabV4,
+    logInOptions: {
+        wizardTitle: 'Log In to GitLab',
```
nturinski: nit: I think in wouldn't be capitalized in this case.  But what do you think about using "Sign in" to be consistent with Azure Accounts?

EricJizbaMSFT: This was modeled after the existing Docker Hub support which uses "Log In". I don't have a strong preference

nturinski: Yeah same, just thought I'd mention it

PR #1119, opened 2019-07-09
Remove "images" folder

EricJizbaMSFT:
All of the images that are still used were moved here: https://github.com/microsoft/vscode-docker/tree/master/resources/readme

(I had to wait to delete this until after we published 0.7.0 since the marketplace page references our master branch.)


EricJizbaMSFT:
Can you see READMEs for previous versions? I thought the marketplace always showed the latest

StephenWeatherford:
Fairly sure it shows the one packaged in the vsix, although I could be wrong.

EricJizbaMSFT:
> Fairly sure it shows the one packaged in the vsix, although I could be wrong.

Tested with 0.6.4 installed and it has the latest README (in the screenshot you can tell because the "visit the wiki" line is new).

Screen Shot 2019-07-17 at 10 01 30 AM

We should probably restructure this at some point anyways (before GA-ing?) because you can pass the base url for images when creating a vsix. We should have it point to the `v0.6.4` tag instead of the master branch.

Docs for vsce package:
```
vsce package --help


  Usage: package [options]

  Packages an extension

  Options:

    -h, --help              output usage information
    -o, --out [path]        Output .vsix extension file to [path] location
    --baseContentUrl [url]  Prepend all relative links in README.md with this url.
    --baseImagesUrl [url]   Prepend all relative image links in README.md with this url.
    --yarn                  Use yarn instead of npm
```

StephenWeatherford:
That's cool, didn't know about that.

PR #1124, opened 2019-07-10
Adding Win10 19H1 (aka 1903) support

bwateratmsft:


bwateratmsft:
@StephenWeatherford Presumably yes. We could make a manifest that would simplify updating but there's no predicting what Windows version numbers are going to be.

Issue #1144, opened 2019-07-19
Support HTTPS in ASP.NET Core apps


bwateratmsft:
Currently "dotnet new" will enable HTTPS by default for ASP.NET Core apps, but this extension does not support it. Exposing 443 has no effect, and setting the ASPNETCORE_URLS to include HTTPS results in app startup failure, because no certificate is found.

Visual Studio automatically configures the ASP.NET Core HTTPS development certificate and maps in a volume containing it, and an additional volume with dotnet user-secrets for the certificate password.

The ask is to do that here too.

bwateratmsft:
Things I'd like to do:

1. Change the singular port prompt to one accepting multiple (i.e. a comma-separated list)
2. Automatically export the 'dotnet dev-certs' certificate with a random password
3. Set that password as a secret with 'dotnet user-secrets'
4. Automatically volume map the cert and user-secrets folders
5. In launch.json add environment variables for ASPNETCORE_URLS and ASPNETCORE_ENVIRONMENT
6. When choosing port for browser launch, first look for anything mapped to 443, then to 80

bwateratmsft:
I'm having a lot of trouble with cert trust on Mac. I can't figure out a way to launch `dotnet dev-certs https --trust` outside of VSCode (i.e. in its own window--which we need, because it's prompting for a password). Launching it within VSCode in a new terminal instance _technically_ works, but in practice won't actually work--if it's done during F5 the terminal instance almost immediately gets unfocused and "lost". In Windows there's a separate popup done by the OS so it's not an issue there.

pratiksanglikar:
@bwateratmsft About the issue with Mac, can we execute any shell script from the code? If that's possible we can think of applescript or even a simple shell script to do that. Applescript for sure has the capabilities to open a new terminal window and run a script inside it.

bwateratmsft:
@pratiksanglikar AppleScript was the closest I was able to get to something working. I ended up with something like `osascript -e 'tell app "Terminal" to do script "dotnet dev-certs https --trust"' -e 'tell app "Terminal" to activate'`. There may be a shorter way to do it but I'm not really familiar with AppleScript. Technically it works, but it leaves the resultant terminal window open when done, which isn't ideal.

EricJizbaMSFT:
@bwateratmsft maybe take a look at this code: 
https://github.com/microsoft/vscode-docker/blob/master/src/commands/registries/logInToDockerCli.ts#L33

[...]

PR #1150, opened 2019-07-23
Add SSL support for ASP.NET Core web apps

bwateratmsft:
For https://github.com/microsoft/vscode-docker/issues/1144


src/debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -163,6 +171,7 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
         };
     }
 
+    // tslint:disable-next-line:cyclomatic-complexity
```
EricJizbaMSFT: Would prefer if you fixed the cyclomatic-complexity rather than disable. Looks like there's too many duplicate checks of `debugConfiguration` and `debugConfiguration.dockerRun`

The VS Code api says `debugConfiguration` will always be defined, so you can trust that and remove all instances of `debugConfiguration &&`

Also you can remove all instances of `debugConfiguration.dockerRun && ` by adding this line to the top of the method:
```
debugConfiguration.dockerRun = debugConfiguration.dockerRun || {};
```

As an aside, we have an open work item to turn on `strictNullChecks` which will enforce `undefined` and `null` typing. You can turn it on in `tsconfig.json` if you want to wade through a bunch of errors to temporarily check just your code.

src/debugging/coreclr/dockerManager.ts
```diff
@@ -357,13 +386,65 @@ export class DefaultDockerManager implements DockerManager {
...
+                path.join(this.osProvider.homedir, '.aspnet', 'https'),
+            hostUserSecretsPath: this.osProvider.os === 'Windows' ?
+                path.join(appDataEnvironmentVariable, 'Microsoft', 'UserSecrets') :
+                path.join(this.osProvider.homedir, '.microsoft', 'usersecrets'),
+        };
+    }
+
+    private getContainerSecretsPaths(os: PlatformOS): { containerCertificateExportPath: string, containerUserSecretsPath: string } {
+        return {
+            containerCertificateExportPath: os === 'Windows' ?
```
EricJizbaMSFT: These paths never change?

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +60,75 @@ export class CommandLineDotNetClient implements DotNetClient {
             return undefined;
         }
     }
+
+    public async trustAndExportSslCertificate(projectFile: string, hostExportPath: string, containerExportPath: string): Promise<void> {
+        if (CommandLineDotNetClient.KnownConfiguredProjects.has(projectFile)) {
+            return;
+        }
+
+        await this.addUserSecretsIfNecessary(projectFile);
```
sgreenmsft: >addUserSecretsIfNecessary [](start = 19, length = 25)

Shouldn't we prompt the user before futzing with the contents of the csproj file?

src/debugging/coreclr/dockerManager.ts
```diff
@@ -357,13 +386,65 @@ export class DefaultDockerManager implements DockerManager {
             permissions: 'ro'
         };
 
-        const volumes: DockerContainerVolume[] = [
+        let volumes: DockerContainerVolume[] = [
```
sgreenmsft: >volumes [](start = 12, length = 7)

On macos, does the default volume sharing configuration include the new directories that we're now using?  I know that one pain point in VS for Mac is the necessity to configuration volume sharing to include the dotnet sdk directory...

src/debugging/coreclr/dockerManager.ts
```diff
@@ -307,19 +329,26 @@ export class DefaultDockerManager implements DockerManager {
...
-            return `http://localhost:${webPort}`;
+        let portMappingsString = await this.dockerClient.inspectObject(containerNameOrId, { format: '{{json .NetworkSettings.Ports}}' });
+        let portMappings = <IPortMappings>JSON.parse(portMappingsString);
+
+        if (portMappings) {
+            let httpsPort = portMappings["443/tcp"] && portMappings["443/tcp"][0] && portMappings["443/tcp"][0].HostPort || null;
+            let httpPort = portMappings["80/tcp"] && portMappings["80/tcp"][0] && portMappings["80/tcp"][0].HostPort || null;
+
+            if (httpsPort) {
+                return `https://localhost:${httpsPort}`;
```
sgreenmsft: >return [](start = 16, length = 6)

It seems like this ought to be in the launch.json file.

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -78,7 +79,13 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
                 dockerBuild: {
                 },
                 dockerRun: {
-                }
+                    env: {
```
sgreenmsft: >env [](start = 20, length = 3)

Does this need to be limited to aspnetcore only?

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +60,75 @@ export class CommandLineDotNetClient implements DotNetClient {
...
+
+    private async addUserSecretsIfNecessary(projectFile: string): Promise<void> {
+        const contents = await this.fsProvider.readFile(projectFile);
+
+        if (contents.indexOf('UserSecretsId') >= 0) {
+            return;
+        }
+
+        const dotNetVer = await this.getVersion();
+        if (semver.gte(dotNetVer, '3.0.0')) {
```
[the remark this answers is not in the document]
bwateratmsft: The "dotnet user-secrets init" command is added to the 3.0 CLI. Modifying the project file directly via xml2js absolutely shredded it, so users with < 3.0 CLI are going to have to manually add the UserSecretsId property.

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +60,75 @@ export class CommandLineDotNetClient implements DotNetClient {
...
+            return;
+        }
+
+        await this.addUserSecretsIfNecessary(projectFile);
+        await this.promptAndTrustCertificateIfNecessary();
+
+        const password = uuidv4();
+
+        // Export the certificate
+        const exportCommand = `dotnet dev-certs https -ep "${hostExportPath}" -p "${password}"`;
```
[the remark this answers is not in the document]
bwateratmsft: FWIW, we do this in VS. Generally probably not necessary--but, we can't verify if the password has been set (without another commandline), nor can we verify that the cert matches the password.

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +60,75 @@ export class CommandLineDotNetClient implements DotNetClient {
...
+        const userSecretsPasswordCommand = `dotnet user-secrets --project "${projectFile}" set Kestrel:Certificates:Development:Password "${password}"`;
+        await this.processProvider.exec(userSecretsPasswordCommand, {});
+
+        // This is not honored due to https://github.com/aspnet/AspNetCore.Docs/issues/6199#issuecomment-418194220
+        // Consequently, the certificate name must be equal to <binaryName>.pfx, i.e. MyWebApp.dll => MyWebApp.pfx
+        //const userSecretsPathCommand = `dotnet user-secrets --project "${projectFile}" set Kestrel:Certificates:Development:Path "${containerExportPath}"`;
+        //await this.processProvider.exec(userSecretsPathCommand, {});
+
+        // Cache the project so we don't do this all over again every F5
+        CommandLineDotNetClient.KnownConfiguredProjects.add(projectFile);
```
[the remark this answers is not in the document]
bwateratmsft: (See above--we do this every session in VS to ensure things aren't broken inbetween sessions. The total cost is fractions of a second on the first F5 of the session.)

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +60,75 @@ export class CommandLineDotNetClient implements DotNetClient {
...
+        } catch {
+            if (this.osProvider.os === 'Windows') {
+                const selection = await ext.ui.showWarningMessage(
+                    "The ASP.NET Core HTTPS development certificate is not trusted. Would you like to trust the certificate? A prompt may be shown.",
+                    { modal: true },
+                    DialogResponses.yes, DialogResponses.no);
+
+                if (selection === DialogResponses.yes) {
+                    const trustCommand = `dotnet dev-certs https --trust`;
+                    await this.processProvider.exec(trustCommand, {});
```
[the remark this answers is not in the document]
bwateratmsft: So on Windows, instead of a shell + password prompt, it's simply a dialog box--and it actually is visible, unlike Mac.

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
bwateratmsft: TBH I don't know what this method is for, because it's not actually the source of the debug configuration (which comes from package.json)

src/debugging/coreclr/dockerManager.ts
bwateratmsft: More or less. There's a few other paths permitted by ASP.NET Core but these are the ones we use in VS/VSMac.

src/debugging/coreclr/CommandLineDotNetClient.ts
bwateratmsft: We don't in VS; additionally SSL will not work without this.

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +60,75 @@ export class CommandLineDotNetClient implements DotNetClient {
...
+        const exportCommand = `dotnet dev-certs https -ep "${hostExportPath}" -p "${password}"`;
+        await this.processProvider.exec(exportCommand, {});
+
+        // Set the password to dotnet user-secrets
+        const userSecretsPasswordCommand = `dotnet user-secrets --project "${projectFile}" set Kestrel:Certificates:Development:Password "${password}"`;
+        await this.processProvider.exec(userSecretsPasswordCommand, {});
+
+        // This is not honored due to https://github.com/aspnet/AspNetCore.Docs/issues/6199#issuecomment-418194220
+        // Consequently, the certificate name must be equal to <binaryName>.pfx, i.e. MyWebApp.dll => MyWebApp.pfx
+        //const userSecretsPathCommand = `dotnet user-secrets --project "${projectFile}" set Kestrel:Certificates:Development:Path "${containerExportPath}"`;
```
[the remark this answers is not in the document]
bwateratmsft: I wrestled with that. We've had endless problems with paths when they contain spaces, unicode characters, etc. in VS. I would have strongly preferred explicitly setting the path to use so it doesn't depend on the shaky logic of "MyWebApp.dll" => "MyWebApp.pfx", especially considering the logic that we use to determine the DLL name is itself rather shaky.

I'll remove it, but if they ever fix that bug this would be a point of improvement for us.

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +60,75 @@ export class CommandLineDotNetClient implements DotNetClient {
...
+
+        try {
+            const checkCommand = `dotnet dev-certs https --check --trust`;
+            await this.processProvider.exec(checkCommand, {});
+        } catch {
+            if (this.osProvider.os === 'Windows') {
+                const selection = await ext.ui.showWarningMessage(
+                    "The ASP.NET Core HTTPS development certificate is not trusted. Would you like to trust the certificate? A prompt may be shown.",
+                    { modal: true },
+                    DialogResponses.yes, DialogResponses.no);
```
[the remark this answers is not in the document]
bwateratmsft: So "Trust" is not in the azureextensionui module, would we add it there and reference, or just add it here?

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +60,75 @@ export class CommandLineDotNetClient implements DotNetClient {
...
+
+                if (selection === DialogResponses.yes) {
+                    const trustCommand = `dotnet dev-certs https --trust`;
+                    await this.processProvider.exec(trustCommand, {});
+                }
+            } else if (this.osProvider.isMac) {
+                await ext.ui.showWarningMessage(
```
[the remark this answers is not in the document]
bwateratmsft: Ah, thought there'd be a OK button by default. I'll add it.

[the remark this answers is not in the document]
bwateratmsft: Actually this gets back to the above question; OK is not available in vscode-azureextensionui, should I add it there, or here?

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
bwateratmsft: Most all of the values turn into simple assignments, so I'm gonna condense them down into the return below.

src/configureWorkspace/configure.ts
```diff
@@ -59,11 +59,15 @@ export interface IPlatformGeneratorInfo {
...
     genDockerCompose: GeneratorFunction,
     genDockerComposeDebug: GeneratorFunction,
-    defaultPort: string | undefined // '' = defaults to empty but still asks user if they want a port, undefined = don't ask at all
+    defaultPorts: Number[] | undefined // [] = defaults to empty but still asks user if they want a port, undefined = don't ask at all
 }
 
-export function getExposeStatements(port: string): string {
-    return port ? `EXPOSE ${port}` : '';
+export function getExposeStatements(ports: Number[]): string {
+    return ports ? ports.map(port => `EXPOSE ${port}`).join('\n') : '';
```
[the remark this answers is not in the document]
bwateratmsft: So, this resulted in a unit test failure that exposes an interesting problem. The failure was at configure.test.ts#1231, the ports section was created with `\r\n` (since I ran it on Windows), but the rest of the compose file uses only `\n`, presumably because our templates are all lacking in `\r`. Essentially, we've never put a `\r` into a compose file or Dockerfile before...why start now?

src/debugging/coreclr/dockerManager.ts
bwateratmsft: The default includes `/Users` (all of which these are under) so at least we're not adding a problem there. Users would still need to add the SDK folder though.

src/debugging/coreclr/dockerManager.ts
bwateratmsft: Yes, it probably should. I'm going to file an issue and address that in a separate PR.

src/debugging/coreclr/CommandLineDotNetClient.ts
sgreenmsft: In VS we add UserSecretsId during initial scaffolding of the Dockerfile.  Here we're doing it during F5.  And we're going to modify the user's csproj even if they say "no" to the SSL prompt.

bwateratmsft: If they don't want it done that's what the "configureSsl" flag in launch.json is for.

src/debugging/coreclr/CommandLineDotNetClient.ts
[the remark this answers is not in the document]
EricJizbaMSFT: But why would the password or cert be incorrect after the initial time we set it? If it's changed - doesn't that imply the user changed it? That sounds like the intentional action of an advanced user (aka hard to do "accidentally") and I would prefer to defer to them rather than assume we know better.

Even though you say it doesn't cost much (and I believe you) - I'm also just very hesitant to run things every F5. It's a high occurrence scenario that you want to be as fast and reliable as possible.

src/debugging/coreclr/CommandLineDotNetClient.ts
EricJizbaMSFT: I tend to agree with Steven on this. What if you check everything up front, including:
1. Check if UserSecretsId is in csproj
1. Check if password has been set and that the cert matches the password
1. Check if cert is trusted

And only have one generic prompt to essentially "do it all" if you detect anything wrong.

src/debugging/coreclr/CommandLineDotNetClient.ts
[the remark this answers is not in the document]
bwateratmsft: Yes, monkeying with it would mess things up. Also, it's not every F5, it's first F5 of the session.

src/debugging/coreclr/CommandLineDotNetClient.ts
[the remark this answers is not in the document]
bwateratmsft: Yes, it's a minority with 3.0 right now. If they don't have UserSecretsId, SSL will not work, period. Without it, they'll get an error message that says something like "Unable to find global property UserSecretsId in <projectFile>.csproj. Please ensure this property is present." User has to manually add it. There is no way around that; trying to alter the project file with xml2js ripped it to shreds.

src/debugging/coreclr/CommandLineDotNetClient.ts
bwateratmsft: Checking the cert and password match would require a crypto library, and I highly doubt it'd be faster than cleaning the slate on the first F5 of the session.

src/debugging/coreclr/CommandLineDotNetClient.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Can you post a screenshot of the error message? How and when does it show up? How many users will or won't have UserSecretsId by default in their project?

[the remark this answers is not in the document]
bwateratmsft: Yes, here's a shot:
image

It'll show up during F5. Unfortunately, most users will not have UserSecretsId at this point. Fortunately though, this entire code path is gated on configureSslCertificate = true, so the error won't randomly start showing up in projects already in use--not until they re-add the Docker launch configuration.

As more and more adopt 3.0+ we can expect the incident rate to drop, but unfortunately there's nothing else we can do right now.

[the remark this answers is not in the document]
EricJizbaMSFT: I'm just advocating for better warning messages, documentation, etc. You have a separate prompt for trusting the cert with a "Learn More" button, but not this. If anything this one feels more confusing than the cert thing (which is a simple one-line command) and more deserving of a "Learn More". For example, you pass in a random hex string in the `dotnet user-secrets init` command - does that mean the user has to generate a random id when editing their csproj?

As an aside, for these learn more links you can just add a simple page to our wiki here: https://github.com/microsoft/vscode-docker/wiki

src/debugging/coreclr/CommandLineDotNetClient.ts
bwateratmsft: @EricJizbaMSFT Here's an alternative that keeps us away from the mess of checking certificate existence / password match / etc.

Instead of doing this on F5, do it via a command. This would make using the VSCode terminal for cert trust viable on Mac. The only thing I'm not sure of is this--can we use a command to manipulate the Docker launch config, in order to add the necessary environment variables? What should we do if no such launch config exists yet and they run the command?

Additionally, it would need to be done per-project, per-machine; i.e. if you sync'd source that was already set up for SSL, you would have to do it again on your machine in order to make it work (And maybe we somehow detect that on F5 and ask? What if they say no on this machine?)

EricJizbaMSFT: [Written just before you commented - not sure I quite understand what you mean by "do it via a command"]

> Checking the cert and password match would require a crypto library, and I highly doubt it'd be faster than cleaning the slate on the first F5 of the session.

Yeah I wouldn't want to use a separate crypto library. I thought there might be a dotnet command that could verify at least part of this for you. Never mind if there isn't a command.

> Other than that, checking if UserSecretsId is in the project and checking trust is exactly what we're doing.

You're never prompting about UserSecretsId, just trusting the cert. @sgreenmsft's original comment said you should prompt before messing with their csproj. I'm agreeing with that and saying perhaps you should combine it with the "cert trust" prompt so that you don't have two separate prompts.

bwateratmsft: As in, one of the Ctrl + Shift + P commands.

src/debugging/coreclr/CommandLineDotNetClient.ts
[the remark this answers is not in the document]
EricJizbaMSFT: > Yes, monkeying with it would mess things up.

Right I'm saying if the user messes things up, I'm fine with that. I generally don't trust us to get it right for every single combination of OS, .NET, etc. In other words, I'm saying some users might have successfully configured SSL for ASP.NET and we actually end up breaking them by running this every session. If we decide "users can turn it off in their launch config" is the answer, I guess that's fine. It just feels unnecessary to run every session

> Also, it's not every F5, it's first F5 of the session.

True. Although if you're code isn't perfect (I know - just a hypothetical) and throws an error before `CommandLineDotNetClient._KnownConfiguredProjects.add(projectFile)` it will end up being "every F5"

src/debugging/coreclr/CommandLineDotNetClient.ts
bwateratmsft: Doing it as a command, we avoid several problems--
1. Screwing up already-working configuration without permission
2. Wasting time doing things more frequently than actually needed
3. Mac trust nightmare

EricJizbaMSFT: When does the command get run?

bwateratmsft: Whenever the user does it.

EricJizbaMSFT: I think making it a command that users can run at any time is a good idea regardless. But I still don't get how a command alone would be discovered

Issue #1155, opened 2019-07-25
Remove Generic Docker Registry failed


degus2019:
Repro steps:

1. Add an empty URL for "Generic Docker Registry"
2. Try to Remove them

Action: vscode-docker.registries.disconnectRegistry
Error type: ERR_INVALID_URL
Error Message: Invalid URL: 

Version: 0.7.0
OS: win32
Product: Visual Studio Code
Product Version: 1.36.1
Language: en

<details>
<summary>Call Stack</summary>

```
onParseError url.js:237:17
parse url.js:246:3
new URL url.js:321:5
GenericDockerV2RegistryTreeItem.get host [as host] extension.bundle.js:31:159472
GenericDockerV2RegistryTreeItem.get label [as label] extension.bundle.js:157:487343
GenericDockerV2RegistryTreeItem.get effectiveLabel [as effectiveLabel] extension.bundle.js:1:464668
ConnectedRegistriesTreeItem.compareChildrenImpl extension.bundle.js:31:100285
module.exports.loadMoreChildrenInternal.n._cachedChildren._cachedChildren.concat.sort extension.bundle.js:31:102539
ConnectedRegistriesTreeItem.<anonymous> extension.bundle.js:31:102522
o extension.bundle.js:31:98683
```

</details>

EricJizbaMSFT:
@degus2019 can you give me an example "incorrect URL" that I can try?

degus2019:
As example our internal registry: 
https://nexus3.xx.xxxxx.de:18445
alternative:
https://10.11.11.2:18445

After i added the IP URL, there is no additional entry in "Connected Registries" Tab. 
When I try to add one of the two urls, I am told that the urls already exist

it is not possible to disconnect any registry entry .. 
i tried to remove the credentials in windows store... without result
where are the connect data stored?

EricJizbaMSFT:
@degus2019 Very helpful - thanks!

Unfortunately, I couldn't find an easy workaround with the existing extension. However, I've created a build which should allow you to disconnect the registry with the empty url. If you're interested, install [this build](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=8120) using these instructions: https://aka.ms/azCodeInsiders

EricJizbaMSFT:
FYI @quangkieu see above comment for a workaround

degus2019:
i used the build, but there is no option for disconnect,
you only can reset all connections, because you cant see the entrys..
[...]

PR #1160, opened 2019-07-25
Make disconnect registry more robust to errors

EricJizbaMSFT:
A user was able to connect a registry with an empty URL and get in a situation where they couldn't disconnect it because `ext.registriesTree.showTreeItemPicker` would throw an error. Instead, I will circumvent the tree and use the registry cache more directly to show the quick pick.

Related to https://github.com/microsoft/vscode-docker/issues/1155 (emphasis on _related to_. There's still other aspects of that bug that I want to address, like preventing users from entering empty urls, etc.)

Example screenshot:
Screen Shot 2019-07-25 at 11 48 07 AM

### DIFF
--- a/src/tree/registries/RegistriesTreeItem.ts
+++ b/src/tree/registries/RegistriesTreeItem.ts
@@ -140,13 +140,35 @@ export class RegistriesTreeItem extends AzExtParentTreeItem {
         await this.saveCachedProviders();
     }
 
-    public async disconnectRegistry(context: IActionContext, node: IRegistryProviderTreeItem): Promise<void> {
-        context.telemetry.properties.providerId = node.cachedProvider.id;
-        context.telemetry.properties.providerApi = node.cachedProvider.api;
+    public async disconnectRegistry(context: IActionContext, cachedProvider: ICachedRegistryProvider | undefined): Promise<void> {
+        if (!cachedProvider) {
+            const picks = this._cachedProviders.map(crp => {
+                const provider = getRegistryProviders().find(rp => rp.id === crp.id);
+                let label: string = (provider && provider.label) || crp.id;
+                let descriptions: string[] = [];
+                if (crp.username) {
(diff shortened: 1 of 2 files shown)


src/tree/registries/RegistriesTreeItem.ts
```diff
@@ -140,13 +140,35 @@ export class RegistriesTreeItem extends AzExtParentTreeItem {
...
+        if (!cachedProvider) {
+            const picks = this._cachedProviders.map(crp => {
+                const provider = getRegistryProviders().find(rp => rp.id === crp.id);
+                let label: string = (provider && provider.label) || crp.id;
+                let descriptions: string[] = [];
+                if (crp.username) {
+                    descriptions.push(`Username: "${crp.username}"`)
+                }
+                if (crp.url) {
+                    descriptions.push(`URL: "${crp.url}"`)
```
PrashanthCorp: Why are you pushing these strings to an array and then invoking them by index, instead of just assigning these strings to a `description` and a `detail` variable? It also seems buggy - if `crp` doesn't have a url or username field, an error would be thrown when assigning a value to `returnValue.detail`.

EricJizbaMSFT: > It also seems buggy - if crp doesn't have a url or username field, an error would be thrown when assigning a value to returnValue.detail.

I don't think arrays are as strict as you think they are. You can do `descriptions[9999]` with an empty array and it'll return `undefined` with no error.

> Why are you pushing these strings to an array and then invoking them by index, instead of just assigning these strings to a description and a detail variable?

It's not a simple assignment because `url` and/or `username` could be undefined and I never want to be in a case where `detail` is defined, but `description` isn't (since `description` takes priority). I probably could've more like what you're asking, but it feels like one and the same.

EricJizbaMSFT:
ping @PrashanthCorp

PR #1162, opened 2019-07-26
Bump shared packages

EricJizbaMSFT:
Moved `Test*` classes to dev package and other minor improvements to test code

Also had to change default version of node for Azure Dev Ops. I was running into build failures on Linux only, and narrowed it down to an old version of npm causing issues. Node 10 is LTS anyways

### DIFF
--- a/test/buildAndRun.test.ts
+++ b/test/buildAndRun.test.ts
@@ -11,11 +11,11 @@ import { Context, Suite } from 'mocha';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { commands, Uri } from 'vscode';
-import { IActionContext, TestUserInput } from 'vscode-azureextensionui';
+import { IActionContext } from 'vscode-azureextensionui';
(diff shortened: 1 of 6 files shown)


test/buildAndRun.test.ts
```diff
@@ -82,14 +81,15 @@ suite("Build Image", function (this: Suite): void {
             errorHandling: {}
         };
 
-        await configure(context, testRootFolder);
-        assert.equal(configureInputs.length, 0, 'Not all inputs were used for configure docker files');
+        await testUserInput.runWithInputs(configureInputs, async () => {
+            await configure(context, testRootFolder);
```
nturinski: Just to make sure I understand, but you got rid of these asserts since checking whether or not all inputs were used should be handled in testUserInput anyway, right?

EricJizbaMSFT: Yes it's handled in the shared code now

PR #1167, opened 2019-07-29
Updating netcore Dockerfile to use separate directory for publish

sgreenmsft:
The Dockerfile we scaffold calls build and publish with the same target directory. In .NET Core 3 this causes a problem because some build artifacts are left over in the publish output. See: https://github.com/aspnet/AspNetCore/issues/11609

### DIFF
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -80,7 +80,7 @@ $copy_project_commands$
 RUN dotnet restore "$container_project_directory$/$project_file_name$"
 COPY . .
 WORKDIR "/src/$container_project_directory$"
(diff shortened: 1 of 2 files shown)


src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -83,11 +83,11 @@ WORKDIR "/src/$container_project_directory$"
 RUN dotnet build "$project_file_name$" -c Release -o /app
```
EricJizbaMSFT: If publish gets its own `/app/publish` folder should build get its own `/app/build` folder for consistency?

sgreenmsft: That's quite reasonable.

PR #1173, opened 2019-07-31
Supporting SSL in ASP.NET Core web apps

bwateratmsft:
Part of #1144

### DIFF
--- a/package.json
+++ b/package.json
@@ -659,6 +659,11 @@
                                         }
                                     }
                                 }
+                            },
+                            "configureSslCertificate": {
+                                "type": "boolean",
+                                "default": true,
+                                "description": "Whether to configure an SSL certificate for ASP.NET Core web services."
                             }
                         }
                     }
@@ -1108,6 +1113,11 @@
                     "type": "boolean",
                     "default": true,
                     "description": "Run docker-compose with the --d (detached) argument, defaults to true"
+                },
+                "docker.promptToTrustAspNetCoreCertificate": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Prompt to trust the ASP.NET Core HTTPS development certificate, if necessary"
                 }
             }
         },
--- a/package.json
+++ b/package.json
@@ -1113,11 +1113,6 @@
                     "type": "boolean",
                     "default": true,
                     "description": "Run docker-compose with the --d (detached) argument, defaults to true"
-                },
-                "docker.promptToTrustAspNetCoreCertificate": {
-                    "type": "boolean",
-                    "default": true,
-                    "description": "Prompt to trust the ASP.NET Core HTTPS development certificate, if necessary"
                 }
             }
         },
--- a/package.json
+++ b/package.json
@@ -660,10 +660,10 @@
                                     }
                                 }
                             },
-                            "configureSslCertificate": {
+                            "configureAspNetCoreSsl": {
                                 "type": "boolean",
                                 "default": true,
-                                "description": "Whether to configure an SSL certificate for ASP.NET Core web services."
+                                "description": "Whether to configure certificate and other settings to enable SSL on ASP.NET Core web services."
                             }
                         }
                     }
--- a/src/debugging/coreclr/CommandLineDotNetClient.ts
+++ b/src/debugging/coreclr/CommandLineDotNetClient.ts
@@ -2,7 +2,12 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
+import * as crypto from 'crypto';
+import * as semver from 'semver';
+import { parseError } from 'vscode-azureextensionui';
 import { ProcessProvider } from "./ChildProcessProvider";
+import { FileSystemProvider } from "./fsProvider";
+import { OSProvider } from "./LocalOSProvider";
 
 export type MSBuildExecOptions = {
     target?: string;
@@ -12,10 +17,15 @@ export type MSBuildExecOptions = {
 export interface DotNetClient {
     execTarget(projectFile: string, options?: MSBuildExecOptions): Promise<void>;
     getVersion(): Promise<string | undefined>;
+    isCertificateTrusted(): Promise<boolean>;
+    exportCertificate(projectFile: string, certificateExportPath: string): Promise<void>;
 }
 
 export class CommandLineDotNetClient implements DotNetClient {
-    constructor(private readonly processProvider: ProcessProvider) {
+    constructor(
+        private readonly processProvider: ProcessProvider,
+        private readonly fsProvider: FileSystemProvider,
+        private readonly osProvider: OSProvider) {
     }
 
     public async execTarget(projectFile: string, options?: MSBuildExecOptions): Promise<void> {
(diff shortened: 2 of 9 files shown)


src/debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -303,6 +300,34 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
...
+            //tslint:disable:no-unsafe-any no-any
+            if (launchSettings && launchSettings.profiles) {
+                const projectProfile = Object.values<any>(launchSettings.profiles).find(p => p.commandName === 'Project');
+
+                if (projectProfile && projectProfile.applicationUrl && /https:\/\//i.test(projectProfile.applicationUrl)) {
+                    return true;
+                }
+            }
+            //tslint:enable:no-unsafe-any no-any
+        }
```
bwateratmsft: I can't decide if I want to automatically infer from launchSettings.json at this time. If we decide not to right now, we don't need a feature flag to guard this functionality, since activating it would require them to put the `configureSslCertificate` flag on their launch config.

src/debugging/coreclr/LocalAspNetCoreSslManager.ts
```diff
@@ -0,0 +1,157 @@
...
+import { ext } from '../../extensionVariables';
+import { PlatformOS } from '../../utils/platform';
+import { quickPickProjectFileItem } from '../../utils/quick-pick-file';
+import { quickPickWorkspaceFolder } from '../../utils/quickPickWorkspaceFolder';
+import { ProcessProvider } from './ChildProcessProvider';
+import { DotNetClient } from './CommandLineDotNetClient';
+import { OSProvider } from './LocalOSProvider';
+import { NetCoreProjectProvider } from './netCoreProjectProvider';
+
+export type HostSecretsFolders = {
```
bwateratmsft: Do we need these to be separate?

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
bwateratmsft: Additionally/alternatively, should we add a command to do it now?

src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -48,6 +58,60 @@ export class CommandLineDotNetClient implements DotNetClient {
             return undefined;
         }
     }
+
+    public async isCertificateTrusted(): Promise<boolean | undefined> {
```
PrashanthCorp: We usually stay away from returning results which can also be undefined. It requires a further check at the calling function. It can also be coerced into a false.

Using an enum could help. I'll defer to Eric on this though.

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -75,10 +76,8 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
                 type: 'docker-coreclr',
                 request: 'launch',
                 preLaunchTask: 'build',
-                dockerBuild: {
-                },
-                dockerRun: {
-                }
+                dockerBuild: {},
+                dockerRun: {}
```
PrashanthCorp: If this is on autosave, look into what settings you have here. Changes like these should not be happening.

src/debugging/coreclr/LocalAspNetCoreSslManager.ts
PrashanthCorp: I don't see the advantage. TS does type-guarding by structure, so passing around variables of types `HostSecretsFolders` and `ContainerSecretsFodlers` interchangeably will not throw any warnings (they are structurally the same, names don't matter). If you want separate types for your own clarifications, it makes sense. But if you are planning to do different operations on varaibles of either type, TS won't prevent the next dev from mixing them up.

package.json
```diff
@@ -659,6 +659,11 @@
                                         }
                                     }
                                 }
+                            },
+                            "configureSslCertificate": {
```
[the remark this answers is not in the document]
bwateratmsft: I'm not sure what you mean by nested?

src/debugging/coreclr/CommandLineDotNetClient.ts
bwateratmsft: I'll do an enum here, it's easy enough.

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
bwateratmsft: This was intentionally changed; to match what we have in package.json.

src/debugging/coreclr/LocalAspNetCoreSslManager.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+export interface AspNetCoreSslManager {
+    trustCertificateIfNecessary(): Promise<void>;
+    exportCertificateIfNecessary(projectFile: string | undefined, certificateExportPath: string | undefined): Promise<void>;
+    getHostSecretsFolders(): SecretsFolders;
+    getContainerSecretsFolders(platform: PlatformOS): SecretsFolders;
+}
+
+export class LocalAspNetCoreSslManager implements AspNetCoreSslManager {
```
EricJizbaMSFT: Before you get comfortable with this pattern (Provider/Manager classes being passed around) - I want to quote [what I said](https://github.com/microsoft/vscode-docker/pull/500#pullrequestreview-163939080) to Phil when he first introduced it to this repo:

> I'm still not a huge fan of the Provider pattern for the following reasons:
> 1. I think it makes the code unnecessarily complicated
> 1. I think tests are more effective if they're running against the 'real thing'
> 1. I think you might be using patterns that were helpful testing against VS but are less necessary when testing VS Code (which is a lot easier from my experience)

It feels like the debugging code is off on it's own little world and not really integrated with the rest of the codebase. Regardless of what pattern we choose, I think the repo should be consistent. You obviously don't have to fix it for this PR, but I just want to make sure you're aware of this. If this was one-off work I might not care as much, but seems like y'all will be contributing more regularly now.

src/debugging/coreclr/LocalAspNetCoreSslManager.ts
```diff
@@ -0,0 +1,139 @@
...
+        const trusted = await this.dotNetClient.isCertificateTrusted();
+
+        if (trusted === TrustState.Trusted || trusted === TrustState.NotApplicable) {
+            LocalAspNetCoreSslManager._CertificateTrustedOrSkipped = true;
+            return;
+        }
+
+        if (this.osProvider.os === 'Windows') {
+            const trust: MessageItem = { title: 'Trust' };
+            const message = 'The ASP.NET Core HTTPS development certificate is not trusted. Your browser may show warnings about the site when opened. To trust the certificate manually, run \`dotnet dev-certs https --trust\`, or click "Trust" to run this command automatically. You may be prompted to trust the certificate.';
```
EricJizbaMSFT: Pretty long message - can we get a screenshot? And try to shorten? I feel like the purpose of a "Learn More" button is so that we don't have to explain every detail in the message haha

bwateratmsft: Agreed, it is long. I wasn't super in love with it. Here it is on Windows (Mac loses the sentence about clicking "Trust"):

image

I think we could get rid of the part about browser warning?

src/debugging/coreclr/LocalAspNetCoreSslManager.ts
bwateratmsft: I'd tend to agree. You end up with a very unwieldy factory in registerDockerDebug...ts, to construct a million and one objects, many of which don't actually have state and could be static methods.

src/debugging/coreclr/LocalAspNetCoreSslManager.ts
EricJizbaMSFT: Looks a little gross, but not as bad as I thought. Up to you

bwateratmsft: I worked with the PM to shorten it,
image

PR #1176, opened 2019-08-02
Always Add host and port to `compose.debug.yml`

burkeholland:
It appears that we are looking for an `npm` script which directly references `node` in order to generate the proper debug configuration (--inspect=0.0.0.0:9229). But directly referencing `node` isn't required for an `npm start` script. For instance, the following is valid...

```
"scripts": {
  "start": "./server.js"
}
```

I had this configuration and it took me quite a while to figure out why the debugger wouldn't attach.

It seems to me that we should always generate the proper host:port configuration for the `debug.yml`. What is the use case for not doing that if we don't find "node" as a command in the `package.json` scripts section? If the user has already selected "node", then node is explicit.

### DIFF
--- a/test/configure.test.ts
+++ b/test/configure.test.ts
@@ -462,13 +462,13 @@ suite("Configure (Add Docker files to Workspace)", function (this: Suite): void 
             assertFileContains('docker-compose.debug.yml', '9229:9229');
             assertFileContains('docker-compose.debug.yml', 'image: testoutput');
             assertFileContains('docker-compose.debug.yml', 'NODE_ENV: development');
-            assertFileContains('docker-compose.debug.yml', 'command: node --inspect index.js');
+            assertFileContains('docker-compose.debug.yml', 'command: node --inspect:0.0.0.0:9229 index.js');
 
             assertFileContains('docker-compose.yml', '1234:1234');
             assertNotFileContains('docker-compose.yml', '9229:9229');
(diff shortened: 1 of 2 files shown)


PrashanthCorp:
It looks like the tests are failing because the docker-compose files being tested weren't updated with this PR?

test/configure.test.ts
```diff
@@ -462,13 +462,13 @@ suite("Configure (Add Docker files to Workspace)", function (this: Suite): void
             assertFileContains('docker-compose.debug.yml', '9229:9229');
             assertFileContains('docker-compose.debug.yml', 'image: testoutput');
             assertFileContains('docker-compose.debug.yml', 'NODE_ENV: development');
-            assertFileContains('docker-compose.debug.yml', 'command: node --inspect index.js');
+            assertFileContains('docker-compose.debug.yml', 'command: node --inspect:0.0.0.0:9229 index.js');
```
PrashanthCorp: Did the syntax of the node command change, or is this merely reacting to the other changes you've made?

burkeholland: The syntax should change in the debug.yml. This is updating the test to account for the new correct value.

The question we need to answer is "What is the use case for NOT passing the correct host and port when "node" isn't found as a command in the `package.json` file? I can't think of one, but I'm sure it was originally done like that for a reason.

StephenWeatherford:
Reviewing...

StephenWeatherford:
LGTM

PR #1180, opened 2019-08-02
Improvements to ASP.NET Core SSL support

bwateratmsft:
* Add <UserSecretsId> property when < 3.0 dotnet CLI is present
* Pass in ASPNETCORE_HTTPS_PORT env var to enable automatic redirection to SSL page
* Automatically infer if the app is SSL-enabled

#1144

### DIFF
--- a/src/debugging/coreclr/CommandLineDotNetClient.ts
+++ b/src/debugging/coreclr/CommandLineDotNetClient.ts
@@ -97,8 +97,24 @@ export class CommandLineDotNetClient implements DotNetClient {
 
         const dotNetVer = await this.getVersion();
         if (semver.gte(dotNetVer, '3.0.0')) {
+            // The dotnet 3.0 CLI has `dotnet user-secrets init`, let's use that if possible
             const userSecretsInitCommand = `dotnet user-secrets init --project "${projectFile}" --id ${this.getRandomHexString(32)}`;
             await this.processProvider.exec(userSecretsInitCommand, {});
+        } else {
+            // Otherwise try to manually edit the project file by adding a property group immediately after the <Project> tag
+            // Allowing for leading and trailing whitespace, as well as XML attributes, this regex matches the <Project> tag as long as it is alone on its line
+            const projectTagRegex = /^[ \t]*<Project.*>[ \t]*$/im;
+
+            const matches = contents.match(projectTagRegex);
+            if (matches && matches[0]) {
+                // If found, add the new property group immediately after
+                const propertyGroup = `
+  <PropertyGroup>
+    <UserSecretsId>${this.getRandomHexString(32)}</UserSecretsId>
+  </PropertyGroup>`;
+                const newContents = contents.replace(matches[0], matches[0] + propertyGroup);
+                await this.fsProvider.writeFile(projectFile, newContents);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/debugging/coreclr/CommandLineDotNetClient.ts
```diff
@@ -97,8 +97,24 @@ export class CommandLineDotNetClient implements DotNetClient {
 
         const dotNetVer = await this.getVersion();
         if (semver.gte(dotNetVer, '3.0.0')) {
+            // The dotnet 3.0 CLI has `dotnet user-secrets init`, let's use that if possible
             const userSecretsInitCommand = `dotnet user-secrets init --project "${projectFile}" --id ${this.getRandomHexString(32)}`;
             await this.processProvider.exec(userSecretsInitCommand, {});
+        } else {
+            // Otherwise try to manually edit the project file by adding a property group immediately after the <Project> tag
+            // Allowing for leading and trailing whitespace, as well as XML attributes, this regex matches the <Project> tag as long as it is alone on its line
+            const projectTagRegex = /^[ \t]*<Project.*>[ \t]*$/im;
```
EricJizbaMSFT: `[ \t]` can be simplified to `\s`

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -305,6 +305,28 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
             return debugConfiguration.configureAspNetCoreSsl;
         }
 
+        const launchSettingsPath = path.join(path.dirname(projectFile), 'Properties', 'launchSettings.json');
+
+        if (await this.fsProvider.fileExists(launchSettingsPath)) {
+            let launchSettingsString = await this.fsProvider.readFile(launchSettingsPath);
+            if (launchSettingsString.charCodeAt(0) === 0xFEFF) {
```
EricJizbaMSFT: Not really sure why this is necessary.

Also rather than read _then_ parse, we typically use `fse.readJSON` from 'fs-extra' which does them in the same step. It's simpler and I _think_ more robust, but not entirely sure

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -305,6 +305,28 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
...
+                // Remove byte order mark if needed
+                launchSettingsString = launchSettingsString.slice(1);
+            }
+            const launchSettings = JSON.parse(launchSettingsString);
```
EricJizbaMSFT: If this fails to parse, I feel like we should just return `false` (aka assume it's not ssl) rather than display an error to the user. I assume they would get a better error down the line from some other source

This also applies if it succeeds to parse, but the object doesn't match the structure you expect below

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
bwateratmsft: I'll try it. I had to read/parse separately with JSON.parse because it couldn't comprehend what a byte order mark was.

bwateratmsft: Looks like fse.readJson is smart enough to deal with BOM so I'll definitely use that.

PR #1187, opened 2019-08-05
Update .dockerignore to match VS

bwateratmsft:
Updating of the .dockerignore file to match what is created by VS. There are also corresponding changes in VS.

### DIFF
--- a/src/configureWorkspace/configure.ts
+++ b/src/configureWorkspace/configure.ts
@@ -128,13 +128,13 @@ function genDockerIgnoreFile(service: string, platformType: string, os: string, 
 **/charts
 **/docker-compose*
 **/Dockerfile*
-**/LICENSE
 **/node_modules
 **/npm-debug.log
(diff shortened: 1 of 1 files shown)


src/configureWorkspace/configure.ts
```diff
@@ -113,19 +113,29 @@ function genDockerComposeDebug(serviceNameAndRelativePath: string, platform: Pla
...
-.git
-.gitignore
-.env
-*/bin
-*/obj
-README.md
-LICENSE
-.vscode`;
+    return `**/node_modules
+**/npm-debug.log
```
EricJizbaMSFT: Can we alphabetize this list? I use the [sort lines](https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines) extension in VS Code

bwateratmsft: Certainly can!

PR #1192, opened 2019-08-07
Update icons to match new VS Code style

misolori:
VS Code is updating its icons to [use an outline style](https://github.com/microsoft/vscode/issues/8017#issuecomment-509349133). With this change coming to stable soon (should be released tomorrow), I've gone through and updated all of the icons used in the Docker extension to match this new style. 

Below is a summary of the changes and a sample, I didn't touch any of the registry/azure icons as those respect those brand's icons:

image

image

cc @chrisdias @fiveisprime


EricJizbaMSFT:
Looks good! A few comments:

1. Feels like the generic registry icon should be updated to outline style somehow. The easiest way to reproduce is actually to change the "Group By" in the images view to "Registry".
Screen Shot 2019-08-08 at 10 01 26 AM
Screen Shot 2019-08-08 at 10 02 42 AM
1. Should the clipboard icon for tasks remove its color? Seems like color is only used for things like status and in brand icons. Also note there's actually two separate clipboard icons:
Screen Shot 2019-08-08 at 10 08 48 AM
1. Is the tag icon standard or used elsewhere? If not - can we get rid of the "string" going through the tag? My biggest complaint about the outline style is it makes complex icons more difficult to see. The original icon shape felt simpler:
Screen Shot 2019-08-08 at 10 18 52 AM
1. The "play" and "stop" icons got rid of the enclosing circle (which I think I prefer). Should the "checkmark", "error" and "pause" icons also get rid of the enclosing circle?
Screen Shot 2019-08-08 at 10 19 31 AM

misolori:
@EricJizbaMSFT thanks for the feedback, I went ahead and updated the icons based on your feedback, let me know what you think. I left the Error/Warning in their circle/triangle to be consistent with our icons.

image

EricJizbaMSFT:
Much better thanks! Although a few of the new ones seem to have a white background problem on light theme...
Screen Shot 2019-08-09 at 10 29 36 AMScreen Shot 2019-08-09 at 10 29 41 AMScreen Shot 2019-08-09 at 10 30 00 AM

misolori:
Sorry about, I removed the white background on all of them.

image

PR #1196, opened 2019-08-08
Update docker.svg icon to use line icon

snnsnn:
VS Code version has updated product icons to have uniform style, color and size. This svg icon follows the same specifications with new vs code icons.

Docker icon preview

This PR is for docker.svg only. There is already a pull request for other icons:

https://github.com/microsoft/vscode-docker/pull/1192

### DIFF
--- a/resources/docker.svg
+++ b/resources/docker.svg
@@ -1,10 +1,103 @@
-<svg width="28" height="28" viewBox="0 0 28 28" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-<g id="Canvas" fill="none">
-<g id="Union">
-<g id="Union_2">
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
@snnsnn the Docker icon is a little more complicated because of their marketing guidelines, etc. I know we've discussed this, but I don't remember all of the conclusions. @misolori @fiveisprime @chrisdias for confirmation

misolori:
We've been in contact with the Docker team about options for their logo in the extension. They don't support modifying their current logo and will let us know if they come up with an alternative, so I would close this PR until we get an official one from their team.

snnsnn:
Ok. Thanks.

dbeff:
I saw this PR just after created another. 
https://github.com/microsoft/vscode-docker/pull/1215 
Sorry about that.

PR #1198, opened 2019-08-09
Automatically enable dotnet user secrets if needed

bwateratmsft:


src/debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -190,7 +195,8 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
             os,
             ports: debugConfiguration.dockerRun.ports,
             volumes: DockerDebugConfigurationProvider.inferVolumes(folder, debugConfiguration),
-            configureAspNetCoreSsl: ssl
+            configureAspNetCoreSsl: ssl,
+            configureDotNetUserSecrets: userSecrets,
```
PrashanthCorp: `userSecrets` doesn't look like a boolean based on the definition, while configureUserSecrets looks like one. Is there an implicit type-cast I'm missing here?

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
```diff
@@ -328,6 +334,17 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
         return false;
     }
 
+    private async inferUserSecrets(ssl: boolean, projectFile: string): Promise<boolean> {
+        if (ssl) {
+            return true;
+        }
+
+        if (this.fsProvider.fileExists(projectFile)) {
```
PrashanthCorp: `inferUserSecrets` is async. Does the `fileExists` check have an async version too, that we can use here?

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
bwateratmsft: I could rename it to configureDotNetUserSecrets, but userSecrets is just a yes/no inferred from the project file.

src/debugging/coreclr/dockerDebugConfigurationProvider.ts
bwateratmsft: fileExists in fact is async, and this if statement is broken. It would always think it's true, since the promise would always be defined (either true or false). Good catch!

bwateratmsft:
> Please provide testcases, either in code or in the azcode teams notebook, thx.

bwateratmsft:
> Please provide testcases, either in code or in the azcode teams notebook, thx.

Added the test cases to the azcode notebook, under "P3 .NET Core User Secrets"

StephenWeatherford:
> Please provide testcases, either in code or in the azcode teams notebook, thx.

Just went through the testcase steps, and was able to run it to completion without any errors (Mac), even though I don't have your code.  What exactly do your changes do again?
 
BTW, my ConfigureServices looks completely different:


 public void ConfigureServices(IServiceCollection services)
 {
 services.Configure<CookiePolicyOptions>(options =>
 {
 // This lambda determines whether user consent for non-essential cookies is needed for a given request.
 options.CheckConsentNeeded = context => true;
 options.MinimumSameSitePolicy = SameSiteMode.None;
 });
 
 services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_2);
 
 Console.WriteLine("The secret is: " + Configuration["MySecretName"]);
 }

bwateratmsft:
> Just a suggestion... Have you guys considered writing out the debug configuration to the console for debugging purposes and for users to see what exactly you're doing?

@StephenWeatherford We hadn't but that's a good idea. We're considering a pretty significant refactor at the moment since the TaskProvider.resolveTask actually does something now. I'll keep that in mind as well.

PR #1213, opened 2019-08-15
Command to pull image from Images tree

bwateratmsft:
Commonly, I need to pull a bunch of base images. Doing so from the command line is a chore, and the registries view is unwieldy for e.g. the MCR repository, so it'd be nice to be able to pull specific images that are already on your machine, to update them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
         "onCommand:vscode-docker.images.inspect",
         "onCommand:vscode-docker.images.prune",
         "onCommand:vscode-docker.images.push",
+        "onCommand:vscode-docker.images.pull",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -319,10 +320,15 @@
                     "group": "images_2_general@2"
                 },
                 {
-                    "command": "vscode-docker.images.tag",
+                    "command": "vscode-docker.images.pull",
                     "when": "view == dockerImages && viewItem == image",
                     "group": "images_2_general@3"
```
bwateratmsft: I couldn't decide whether this should go next to Push, or down in the "destructive" group by Remove.

Issue #954, opened 2019-05-23
Improve default behavior in VS Code Remote scenarios


EricJizbaMSFT:
Currently, the Docker extension is defaulting to a UI extension in all scenarios (SSH, Dev Containers, and WSL). That means it will show local containers/images, but anything accessing the file system or a terminal will fail. If users want to access remote containers/images, they must run as a workspace extension by adding this to their settings:
```
"remote.extensionKind": {
    "ms-azuretools.vscode-docker": "workspace"
}
```

Here is my opinion of the ideal behavior for all three VS Code Remote scenarios:

# SSH

This should default to a workspace extension. This would display containers/images from the _remote_ machine and most things should "just work". It sounds like users want this as well (see https://github.com/microsoft/vscode-docker/issues/924).

# Dev Containers

Running as a workspace extension (aka "Docker in Docker") is generally confusing and so we want to avoid it. However, this scenario inherently deals with a "Dockerfile" so we can't completely ignore it. Ideally this defaults to a UI extension. The explorer will show containers running on the local machine and we should show an understandable error if the users tries to run a command that requires the terminal or needs access to the file system.

This scenario could be improved if VS Code gives us the following:
1. An API to create a terminal on the local machine (instead of in the container) to run commands like "Build Image"
1. An API to access the file system in the container so that we can run commands like "Add Docker Files"

If the user is running as a workspace extension, we should show a custom message instead of the default "Unable to connect to Docker.". It should direct users to either switch to a UI extension or set up docker in docker (if they're feeling adventurous).

### WSL

The ability to run docker inside WSL is pretty recent and it's unclear how well it's supported. If we were in an ideal world and docker worked great in WSL, I think this behavior should be similar to SSH. Otherwise, I think it makes sense to just pick a default and document how to change it if desired.

### Questions for VS Code Remote team:

1. Can we have a different default for "Dev Containers" (UI) vs "SSH" (Workspace)?
1. How do we detect if our extension is running in remote - and what kind of remote? (So that we can customize the error message)
1. Is there an official API to create a terminal on the local machine as a UI extension?
1. Is there an official API to access the file system inside a container when running as a UI extension?

egamma:
@EricJizbaMSFT 

Distinguishing between SSH and the container case makes sense.

>Can we have a different default for "Dev Containers" (UI) vs "SSH" (Workspace)?

There will be API that allows you distinguish between these two contexts. The context will be identifiable by the authority name that created the remote context. 

[...]

PR #1233, opened 2019-08-28
Change extension to run in remote workspace by default

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-docker/issues/954. See that issue for more info on _why_ we want to default to workspace extension instead of ui extension.

In addition to changing the default (which is the most important change), I added some helpful error messages to common scenarios.

## Scenario 1

User has Docker installed as "ui" extension. Currently they will see a bunch of random errors for unsupported actions like this:
Screen Shot 2019-08-27 at 11 36 12 AM
Screen Shot 2019-08-27 at 11 36 39 AM

I added this error message instead:
Screen Shot 2019-08-28 at 2 55 03 PM

## Scenario 2
User has Docker installed as "workspace" extension in dev container. "Add Docker Files to Workspace" should work, but any commands using the terminal won't work and the explorer won't show anything.

All I did was add a link in the explorer. The docs still need to be written, but we will likely direct them to one of these options:
1. Forward some ports so that the explorer works when running as a workspace extension
1. Just open a local VS Code instance if you want to see local containers in the explorer
1. Last resort docker in docker
Screen Shot 2019-08-27 at 2 31 48 PM

In terms of commands that require a terminal - I still think the long term fix is to move those over to Dockerode anyways.

### DIFF
--- a/src/commands/registerWorkspaceCommand.ts
+++ b/src/commands/registerWorkspaceCommand.ts
@@ -1,0 +1,104 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { commands, ConfigurationTarget, MessageItem, workspace, WorkspaceConfiguration } from 'vscode';
+import { callWithTelemetryAndErrorHandling, IActionContext, registerCommand, UserCancelledError } from 'vscode-azureextensionui';
+import { extensionId } from '../constants';
+import { ext } from '../extensionVariables';
+import { DockerExtensionKind, getVSCodeRemoteInfo, IVSCodeRemoteInfo, RemoteKind } from '../utils/getVSCodeRemoteInfo';
+
+/**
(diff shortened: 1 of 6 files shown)


EricJizbaMSFT:
ping @StephenWeatherford @sgreenmsft @bwateratmsft @nturinski @philliphoff

src/commands/registerWorkspaceCommand.ts
```diff
@@ -0,0 +1,104 @@
...
+    let value: {};
+
+    // If the setting is already defined as a workspace setting - overwrite that
+    if (typeof values.workspaceValue === 'object' && values.workspaceValue !== null && values.workspaceValue[extensionId]) {
+        target = ConfigurationTarget.Workspace;
+        value = values.workspaceValue;
+    } else { // otherwise update the global setting
+        target = ConfigurationTarget.Global;
+        if (typeof values.globalValue === 'object' && values.globalValue !== null) {
+            value = values.globalValue || {};
```
StephenWeatherford: nit: The ||{} is just defensive, right?  If it's an object and not null, then can it be falsy?

EricJizbaMSFT: You're right - shouldn't be possible to be falsy

Issue #1235, opened 2019-08-29
Add pickDockerContainer command


scabana:
Would be equivalent to command:pickDockerContainer from omnisharp. This would enable to have a single launch task that looks like this: 
```json
{
  "name": ".NET Core Docker Attach",
  "type": "coreclr",
  "request": "attach",
  "processId": "${command:pickRemoteProcess}",
  "pipeTransport": {
    "pipeProgram": "docker",
    "pipeArgs": [ "exec", "-i", "${command:pickDockerContainer}" ],
    "debuggerPath": "/vsdbg/vsdbg",
    "pipeCwd": "${workspaceRoot}",
    "quoteArgs": false
  }
}
```
instead of having 1 launch task per container.

scabana:
Would you take a contribution on this?

EricJizbaMSFT:
@scabana can you describe the scenario when you would use this launch.json? I'd like to hear more about the problem you're trying to solve before we go straight to the fix. Specifically @bwateratmsft is working on some .NET Core debugging features and I wonder if his work could simplify your debugging experience so that you don't have to pick the process/container each time

scabana:
@EricJizbaMSFT For a microservice architecture, where a lot of microservices are generated from a template. Most of the time, the remote attach is to debug docker-compose ran automated tests. In that case, we can have from 2-5 containers up for a single run into which we could want attach. If a command would exist, this would mean we could have a config without specific container names hard coded into launch.json. It also helps out when running a single image in the same microservice. The goal here is to run the same docker-compose as the pipeline would run reducing the "works on my machine, but not on the pipeline" scenario.

bwateratmsft:
At the moment I don't have any plans for .NET Core _attach_ (just launch) so I think this work is different from what I'm working on.

bwateratmsft:
Yes, I agree.

PR #1240, opened 2019-09-02
Adding vscode-docker.containers.select

scabana:
Resolves https://github.com/microsoft/vscode-docker/issues/1235.

Enables scenario like this: 
```json
{
    "name": ".NET Core Docker Attach",
    "type": "coreclr",
    "request": "attach",
    "processId": "${command:pickRemoteProcess}",
    "pipeTransport": {
      "pipeProgram": "docker",
      "pipeArgs": ["exec", "-i", "${command:vscode-docker.containers.select}"],
      "debuggerPath": "/vsdbg/vsdbg",
      "pipeCwd": "${workspaceRoot}",
      "quoteArgs": false
    }
}
```

Blocked by https://github.com/microsoft/vscode/issues/80219.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
         "onCommand:vscode-docker.containers.refresh",
         "onCommand:vscode-docker.containers.remove",
         "onCommand:vscode-docker.containers.restart",
+        "onCommand:vscode-docker.containers.select",
         "onCommand:vscode-docker.containers.start",
         "onCommand:vscode-docker.containers.stop",
         "onCommand:vscode-docker.containers.viewLogs",
@@ -1192,6 +1193,11 @@
             {
                 "command": "vscode-docker.containers.restart",
                 "title": "Restart",
+                "category": "Docker Containers"
+            },
+            {
+                "command": "vscode-docker.containers.select",
+                "title": "Select container",
                 "category": "Docker Containers"
             },
             {
--- a/package.json
+++ b/package.json
@@ -108,6 +108,10 @@
                     "when": "never"
                 },
                 {
+                    "command": "vscode-docker.containers.select",
+                    "when": "never"
+                },
+                {
                     "command": "vscode-docker.registries.azure.selectSubscriptions",
                     "when": "never"
                 }
--- a/src/commands/containers/selectContainer.ts
+++ b/src/commands/containers/selectContainer.ts
@@ -1,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from 'vscode-azureextensionui';
+import { ext } from '../../extensionVariables';
+import { ContainerTreeItem } from '../../tree/containers/ContainerTreeItem';
+
+export async function selectContainer(context: IActionContext): Promise<string> {
+
+    let node: ContainerTreeItem;
+
+    //Expecting running containers to change often as this is a debugging scenario.
+    await ext.containersTree.refresh();
+
+    node = await ext.containersTree.showTreeItemPicker(/^(running)Container$/i, { ...context });
+
+    return node.containerId;
+}
--- a/src/commands/containers/selectContainer.ts
+++ b/src/commands/containers/selectContainer.ts
@@ -14,7 +14,7 @@ export async function selectContainer(context: IActionContext): Promise<string> 
     //Expecting running containers to change often as this is a debugging scenario.
     await ext.containersTree.refresh();
 
-    node = await ext.containersTree.showTreeItemPicker(/^(running)Container$/i, { ...context });
+    node = await ext.containersTree.showTreeItemPicker(/^runningContainer$/i, { ...context });
 
     return node.containerId;
 }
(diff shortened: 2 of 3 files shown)


scabana:
Didn't know that was possible, added and tested :)

src/commands/containers/selectContainer.ts
```diff
@@ -0,0 +1,20 @@
...
+import { ContainerTreeItem } from '../../tree/containers/ContainerTreeItem';
+
+export async function selectContainer(context: IActionContext): Promise<string> {
+
+    let node: ContainerTreeItem;
+
+    //Expecting running containers to change often as this is a debugging scenario.
+    await ext.containersTree.refresh();
+
+    node = await ext.containersTree.showTreeItemPicker(/^(running)Container$/i, { ...context });
```
StephenWeatherford: Why the parens in the regex?  It's same as matching runningContainer.

StephenWeatherford: Verify what happens on a cancel.  Pretty sure Eric would have set it up to throw usercancelled, but haven't inspected the code.

scabana: Copied and altered from other command, I'll fix tonight.

package.json
```diff
@@ -1194,6 +1199,11 @@
                 "title": "Restart",
                 "category": "Docker Containers"
             },
+            {
+                "command": "vscode-docker.containers.select",
```
EricJizbaMSFT: So this command will be hard-coded in people's launch.json file, which means we should be very deliberate with the name because changing it in the future would break people.

I vote `docker.pickContainer` for two reasons:
1. I never understood why we had `vscode-docker` instead of just `docker` as the prefix. `vscode-` should be implied
1. `pickContainer` feels the most consistent with `pickRemoteProcess` and `pickProcess` from the C# extension

src/commands/containers/selectContainer.ts
EricJizbaMSFT: Yeah parens are only needed if multiple different states are allowed - like `/^(running|paused)Container$/i`

EricJizbaMSFT:
Also pending discussion here: https://github.com/microsoft/vscode-docker/issues/1235#issuecomment-528027428

package.json
scabana: 1. I decided to take this extension's current prefix so keep consistency. Do you still want the name changed?
2. pickProcess and pickRemoteProcess doesn't actually exist. They use the variables to redirect to these commands: https://github.com/OmniSharp/omnisharp-vscode/blob/cfec123909a05d88ac625ec72aa0351fe74435ef/package.json#L891-L894. I do like pickContainer though.

EricJizbaMSFT: 1. I'm not worried about consistency with the current command ids because those command ids aren't used directly by users (with a few minor exceptions).
1. Can we use a variable ourselves? Or would that have to be defined directly in the Omnisharp extension?

StephenWeatherford: > I never understood why we had vscode-docker instead of just docker as the prefix. vscode- should be implied

There can be other docker extensions, we should be careful about being too generic, we've conflicted with other extensions before.  vscode-docker ain't perfect, but it does somewhat imply an "official" docker extension.

scabana: @EricJizbaMSFT as I understand it, omnisharp would have to declare the variable for it to be a default for the omnisharp "clr" or "coreclr" debugger types. I would think this couples this command with omnisharp. Don't think that would be good.

EricJizbaMSFT: > as I understand it, omnisharp would have to declare the variable for it to be a default for the omnisharp "clr" or "coreclr" debugger types. I would think this couples this command with omnisharp. Don't think that would be good.

Agreed we shouldn't couple this with Omnisharp 

> There can be other docker extensions, we should be careful about being too generic, we've conflicted with other extensions before. vscode-docker ain't perfect, but it does somewhat imply an "official" docker extension.

Honestly I would expect the generic one to be the official one and other arbitrary prefixes (aka `vscode-`) to be the non-official ones

src/commands/containers/selectContainer.ts
scabana: Resolved!

scabana: Tried it, nothing exploded. Checked which code path was invoked, here's where the debugger lead me: https://github.com/microsoft/vscode/blob/dedf0de944fae56f0bf748913cbc1b2cc24987d2/src/vs/workbench/contrib/debug/browser/debugService.ts#L371-L374 
Returns false, all seem good!

package.json
scabana: Consensus doesn't seem to be here, I'll wait for a decision. As it stands, I would be inclined towards: vscode-docker.containers.pick. I used .select because it was the same term as the Omnisharp one (consistency).

scabana: For some reason, I though I had picked "select" because of Omnisharp... I was wrong, here are the command ids: 
- csharp.listProcess
- csharp.listRemoteProcess

I don't full agree with their ids since they return the selected process' id... which has nothing to do with listing in my mind.

bwateratmsft: @EricJizbaMSFT I'd be fine with using `vscode-docker` here, consistent with the old way. It's essentially harmless to have the longer name including `vscode-`, and I'm a big fan of consistency when there isn't a very compelling reason to go against it.

scabana:
hey @bwateratmsft, thanks for merging this in! First open source contribution to a ms repo!

PR #1245, opened 2019-09-04
Task-based debug build/run

### LINKED ISSUE #1242: Change to a tasks-based model for Docker Build / Docker Run
Historically the process of building and starting a debug container was done entirely during resolveDebugConfiguration, which at the time was the only way possible, but is a bit of an abuse of the API.

Now that resolveTask is supported in VSCode, it is possible for us to resolve `docker build` / `docker run` tasks just-in-time. This means we can use preLaunchTask + tasks to start the container for debugging.

In addition this gives us an opportunity to add debuggable languages/platforms more easily. Our top three appear to be Node.js, .NET Core, and Python.

This PR also enhances the previously-existing build command to use an available `docker-build` task from tasks.json, otherwise it falls back to the prior behavior.

This PR also adds a `pull` option to `docker build` options, equivalent to the `--pull` flag, to help mitigate issues caused by stale images.

Related items:
#764
#879
#38
#1094 
#1211 
#1202

bwateratmsft:
Resolves #1242


package.json
```diff
@@ -467,38 +470,99 @@
         },
         "debuggers": [
             {
-                "type": "docker",
-                "label": "Docker: Node.js",
+                "type": "docker-launch",
+                "label": "Docker: Launch",
                 "configurationSnippets": [
                     {
-                        "label": "Docker: Attach to Node",
```
sgreenmsft: >Attach to Node [](start = 42, length = 14)

We need to keep the Attach to Node functionality, though?

src/commands/debugging/initializeForDebugging.ts
```diff
@@ -0,0 +1,51 @@
...
+    const context: DockerDebugScaffoldContext = {
+        folder: folder,
+        platform: debugPlatform,
+        actionContext: actionContext,
+    }
+
+    switch (context.platform) {
+        case 'netCore':
+            const options: NetCoreScaffoldingOptions = {
+                appProject: (await quickPickProjectFileItem(undefined, context.folder, 'You must choose a .NET Core project file to set up for Docker debugging.')).absoluteFilePath,
```
sgreenmsft: >You [](start = 88, length = 3)

How about just "Choose a .NET Core project for Docker debugging"?

src/commands/debugging/initializeForDebugging.ts
```diff
@@ -0,0 +1,51 @@
...
+                platformOS: await quickPickOS(),
```
sgreenmsft: >quickPickOS [](start = 34, length = 11)

What percentage of people running vscode on macOS or Linux choose Windows as the platformOS for their container?

src/tasks/TaskHelper.ts
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+            arg = resolveWorkspaceFolderPath(folder, arg);
+            return arg === dockerfile;
+        }));
+
+    if (buildTasks.length === 1) {
+        return buildTasks[0]; // If there's only one build task, take it
+    } else if (buildTasks.length > 1) {
+        return buildTasks.find(t => t.name === 'docker-build: release') || buildTasks[0]; // If there's multiple try finding one with the name 'docker-build: release', else take first
```
sgreenmsft: >docker-build: release [](start = 48, length = 21)

We shouldn't take a dependency on the name of the task.

src/tasks/TaskHelper.ts
```diff
@@ -0,0 +1,166 @@
...
+    return folderPath.replace(/\$\{workspaceFolder\}/gi, folder.uri.fsPath);
+}
+
+export function unresolveWorkspaceFolderPath(folder: WorkspaceFolder, folderPath: string): string {
+    // tslint:disable-next-line: no-invalid-template-strings
+    return folderPath.replace(folder.uri.fsPath, '${workspaceFolder}').replace(/\\/g, '/');
+}
+
+// tslint:disable-next-line: no-any
+async function recursiveFindTaskByType(allTasks: TaskDefinition[], type: string, node: any): Promise<TaskDefinition | undefined> {
```
sgreenmsft: >recursiveFindTaskByType [](start = 15, length = 23)

This logic is very difficult to understand.  Why structure it in this way?

src/tasks/TaskHelper.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    return undefined;
+}
+
+async function findTaskByLabel(allTasks: TaskDefinition[], label: string): Promise<TaskDefinition | undefined> {
+    return allTasks.find(t => t.label === label);
+}
+
+async function findTaskByType(allTasks: TaskDefinition[], type: string): Promise<TaskDefinition | undefined> {
+    return allTasks.find(t => t.type === type);
```
sgreenmsft: >find [](start = 20, length = 4)

Intentionally case-sensitive?

src/tasks/TaskHelper.ts
```diff
@@ -0,0 +1,166 @@
...
+            return await recursiveFindTaskByType(allTasks, type, next);
+        }
+    }
+    // tslint:enable: no-unsafe-any
+
+    return undefined;
+}
+
+async function findTaskByLabel(allTasks: TaskDefinition[], label: string): Promise<TaskDefinition | undefined> {
+    return allTasks.find(t => t.label === label);
```
sgreenmsft: >find [](start = 20, length = 4)

Intentionally case-sensitive?

src/tasks/TaskHelper.ts
```diff
@@ -0,0 +1,166 @@
...
+        buildTasks.filter(t => t.execution.args.some(a => { // Find all build tasks where an argument to 'docker build' is this Dockerfile
+            let arg: string;
+            if (typeof a === 'string') {
+                arg = a;
+            } else {
+                arg = a.value;
+            }
+
+            arg = resolveWorkspaceFolderPath(folder, arg);
+            return arg === dockerfile;
```
sgreenmsft: >arg [](start = 19, length = 3)

Intentionally case-sensitive file path comparison?

src/tasks/TaskHelper.ts
```diff
@@ -0,0 +1,166 @@
...
+
+export async function getOfficialBuildTaskForDockerfile(dockerfile: string, folder: WorkspaceFolder): Promise<Task | undefined> {
+    let buildTasks = await tasks.fetchTasks({ type: 'docker-build' });
+    buildTasks =
+        buildTasks.filter(t => t.execution.args.some(a => { // Find all build tasks where an argument to 'docker build' is this Dockerfile
+            let arg: string;
+            if (typeof a === 'string') {
+                arg = a;
+            } else {
+                arg = a.value;
```
sgreenmsft: >value [](start = 24, length = 5)

Guaranteed to be defined?

src/debugging/DockerPlatformHelper.ts
```diff
@@ -0,0 +1,22 @@
...
+export type DockerPlatform = 'netCore' | 'node';
+
+interface DockerPlatformConfiguration {
+    platform?: DockerPlatform;
+    netCore?: {};
+    node?: {};
+}
+
+export function getPlatform<T extends DockerPlatformConfiguration>(configuration: T): DockerPlatform | undefined {
+    if (configuration.platform === 'netCore' || configuration.netCore !== undefined) {
```
sgreenmsft: >platform [](start = 22, length = 8)

Intentionally case-sensitive?

src/debugging/DockerPlatformHelper.ts
```diff
@@ -0,0 +1,22 @@
...
+        return 'netCore'
+    } else if (configuration.platform === 'node' || configuration.node !== undefined) {
```
sgreenmsft: >platform [](start = 29, length = 8)

Intentionally case-sensitive?

sgreenmsft:
>     let rootFolder: vscode.WorkspaceFolder = await quickPickWorkspaceFolder('To build Docker files you must first open a folder or workspace in VS Code.');

If the dockerFileUri !== undefined, then isn't the workspace folder implied by that uri?

---
Refers to: src/commands/images/buildImage.ts:20 in 3ca43d8. [](commit_id = 3ca43d8e4d0cabfbff4708f0f05c737b74be9b9b, deletion_comment = False)

src/commands/images/buildImage.ts
```diff
@@ -20,37 +21,42 @@ export async function buildImage(context: IActionContext, dockerFileUri: vscode.
...
+            contextPath = defaultContextPath;
+        }
+        let absFilePath: string = path.join(rootFolder.uri.fsPath, dockerFileItem.relativeFilePath);
+        let dockerFileKey = `buildTag_${absFilePath}`;
+        let prevImageName: string | undefined = ext.context.globalState.get(dockerFileKey);
+        let suggestedImageName: string;
+
+        if (!prevImageName) {
+            // Get imageName based on name of subfolder containing the Dockerfile, or else workspacefolder
+            suggestedImageName = path.basename(dockerFileItem.relativeFolderPath).toLowerCase();
```
sgreenmsft: >basename [](start = 38, length = 8)

DirectoryName.toLower isn't guaranteed to be a valid image name.

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -260,3 +266,18 @@ function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, o
 
     return contents;
 }
+
+async function initializeForDebugging(context: IActionContext, folder: WorkspaceFolder, platformOS: PlatformOS, { artifactName }: Partial<PackageInfo>): Promise<void> {
+    const scaffoldContext: DockerDebugScaffoldContext = {
+        folder: folder,
+        platform: 'netCore',
```
sgreenmsft: >platform [](start = 8, length = 8)

Using "platform" to refer to the programming language/framework (e.g. netCore, node) and "platformOS" to refer to the operating system (e.g. Windows and Linux) seems sub-ideal.

src/debugging/DebugHelper.ts
```diff
@@ -0,0 +1,77 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { CancellationToken, debug, DebugConfiguration, ExtensionContext, workspace, WorkspaceFolder } from 'vscode';
```
sgreenmsft: >import [](start = 0, length = 6)

Is there any standard practice around organizing imports?

src/debugging/DebugHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+    registerCommand('vscode-docker.debugging.initializeForDebugging', initializeForDebugging);
+}
+
+// TODO: This is stripping out a level of indentation, but the tasks one isn't
+export async function addDebugConfiguration(debugConfiguration: DockerDebugConfiguration): Promise<boolean> {
+    // Using config API instead of tasks API means no wasted perf on re-resolving the tasks, and avoids confusion on resolved type !== true type
+    const workspaceLaunch = workspace.getConfiguration('launch');
+    const allConfigs = workspaceLaunch && workspaceLaunch.configurations as DebugConfiguration[] || [];
+
+    if (allConfigs.some(c => c.name === debugConfiguration.name)) {
```
sgreenmsft: >name [](start = 31, length = 4)

Intentionally case-sensitive comparison?

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,91 @@
...
+        return undefined;
+    }
+
+    public resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration | undefined> {
+        const debugPlatform = getPlatform(debugConfiguration);
+        return callWithTelemetryAndErrorHandling(
+            `docker-launch/${debugPlatform || 'unknown'}`,
+            async (actionContext: IActionContext) => await this.resolveDebugConfigurationInternal(
+                {
+                    folder: folder || await quickPickWorkspaceFolder('To debug with Docker you must first open a folder or workspace in VS Code.'),
```
sgreenmsft: >quickPickWorkspaceFolder [](start = 44, length = 24)

When does this come up?

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,91 @@
...
+                    platform: debugPlatform,
+                    actionContext: actionContext
+                },
+                debugConfiguration));
+    }
+
+    private async resolveDebugConfigurationInternal(context: DockerDebugContext, originalConfiguration: DockerDebugConfiguration): Promise<DockerDebugConfiguration | undefined> {
+        context.actionContext.telemetry.properties.platform = context.platform;
+
+        context.folder = context.folder || await quickPickWorkspaceFolder('To debug with Docker you must first open a folder or workspace in VS Code.');
```
sgreenmsft: >quickPickWorkspaceFolder [](start = 49, length = 24)

When does this come up?

src/debugging/coreclr/CliDockerClient.ts
```diff
@@ -265,6 +276,38 @@ export class CliDockerClient implements DockerClient {
 
         return result.stdout;
     }
+
+    public async getContainerWebEndpoint(containerNameOrId: string): Promise<{ browserUrl: string | undefined, httpsPort: string | undefined }> {
+        let portMappingsString = await this.inspectObject(containerNameOrId, { format: '{{json .NetworkSettings.Ports}}' });
+        let portMappings = <IPortMappings>JSON.parse(portMappingsString);
```
sgreenmsft: >portMappingsString [](start = 53, length = 18)

What if portMappingsString is undefined?

src/debugging/coreclr/CliDockerClient.ts
```diff
@@ -265,6 +276,38 @@ export class CliDockerClient implements DockerClient {
...
+        }
+
+        return {
+            browserUrl: undefined,
+            httpsPort: undefined
+        };
+    }
+
+    public async getHostPort(containerNameOrId: string, containerPort: number): Promise<string> {
+        return (await this.inspectObject(containerNameOrId, { format: `{{(index (index .NetworkSettings.Ports \\"${containerPort}/tcp\\") 0).HostPort}}` })).trim();
```
sgreenmsft: >inspectObject [](start = 27, length = 13)

What if it's undefined?

src/utils/commandLineBuilder.ts
```diff
@@ -0,0 +1,156 @@
+import { ShellQuotedString, ShellQuoting } from "vscode";
+
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
```
sgreenmsft: >Copyright [](start = 3, length = 9)

Move copyright statement to top of file.  Also, where's the MIT license notice as in the other files?

src/utils/cloneObject.ts
```diff
@@ -0,0 +1,8 @@
+// tslint:disable-next-line: no-any
+export function cloneObject<T = any>(obj: T): T {
```
sgreenmsft: >export [](start = 0, length = 6)

Missing license comment

src/debugging/DockerDebugScaffoldingProvider.ts
```diff
@@ -0,0 +1,62 @@
+import { DockerBuildTaskDefinition } from '../tasks/DockerBuildTaskProvider';
```
sgreenmsft: >import [](start = 0, length = 6)

Missing license comment

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -0,0 +1,93 @@
...
+            task.source,
+            new ShellExecution(commandLine[0], commandLine.slice(1)),
+            task.problemMatchers);
+    }
+
+    private async resolveCommandLine(options: DockerBuildOptions): Promise<ShellQuotedString[]> {
+        return CommandLineBuilder
+            .create('docker', 'build', '--rm')
+            .withFlagArg('--pull', options.pull)
+            .withNamedArg('-f', options.dockerfile)
```
sgreenmsft: >options.dockerfile [](start = 32, length = 18)

Does validation of existence of this option happen elsewhere?

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -0,0 +1,93 @@
...
+            .withKeyValueArgs('--build-arg', options.args)
+            .withKeyValueArgs('--label', options.labels)
+            .withNamedArg('-t', options.tag)
```
sgreenmsft: >tag [](start = 40, length = 3)

Does validation of existence of this option happen elsewhere?

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -0,0 +1,259 @@
...
+            await this.vsDbgClientFactory().getVsDbgVersion('latest', 'win7-x64');
+        } else {
+            await this.vsDbgClientFactory().getVsDbgVersion('latest', 'linux-x64');
+            await this.vsDbgClientFactory().getVsDbgVersion('latest', 'linux-musl-x64');
+        }
+
+        const debuggerScriptPath = path.join(ext.context.asAbsolutePath('src/debugging/netcore'), 'vsdbg');
+        const destPath = path.join(NetCoreDebugHelper.getHostDebuggerPathBase(), 'vsdbg');
+        await fse.copyFile(debuggerScriptPath, destPath);
+        await fse.chmod(destPath, 0o755); // Give all read and execute permissions
```
sgreenmsft: >chmod [](start = 18, length = 5)

What does this do when running on Windows?

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -0,0 +1,259 @@
...
+    private async acquireDebuggers(platformOS: PlatformOS): Promise<void> {
+        if (platformOS === 'Windows') {
+            await this.vsDbgClientFactory().getVsDbgVersion('latest', 'win7-x64');
+        } else {
+            await this.vsDbgClientFactory().getVsDbgVersion('latest', 'linux-x64');
+            await this.vsDbgClientFactory().getVsDbgVersion('latest', 'linux-musl-x64');
+        }
+
+        const debuggerScriptPath = path.join(ext.context.asAbsolutePath('src/debugging/netcore'), 'vsdbg');
+        const destPath = path.join(NetCoreDebugHelper.getHostDebuggerPathBase(), 'vsdbg');
```
sgreenmsft: >destPath [](start = 14, length = 8)

Does this path conflict with any shipped version of the VS tools?

src/debugging/DebugHelper.ts
philliphoff: The linter enforces alphabetical order of import statements by imported filename, then alphabetical order of imported types.

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -0,0 +1,259 @@
...
+
+        const dockerServerReadyAction = await this.inferServerReadyAction(debugConfiguration, containerName, configureSsl);
+
+        return {
+            name: debugConfiguration.name,
+            type: 'coreclr',
+            request: 'launch',
+            program: debugConfiguration.program || 'dotnet',
+            args: debugConfiguration.args || [additionalProbingPathsArgs, containerAppOutput].join(' '),
+            cwd: debugConfiguration.cwd || platformOS === 'Windows' ? 'C:\\app' : '/app',
```
sgreenmsft: >platformOS [](start = 43, length = 10)

intentionally case-sensitive?

package.json
philliphoff: I could go either way; the snippet is so basic and makes so many assumptions about the environment in which its used that I'm not sure how much benefit it has.  That said, there's no real reason we *have* to remove it right now.

package.json
```diff
@@ -467,38 +470,99 @@
...
-                        "description": "Docker: Attach to Node",
+                        "label": "Docker: Launch",
+                        "description": "Docker: Launch",
```
philliphoff: We'll need to think more about how these snippets are expected to work in this new model; the one here won't work at all (in particular, because there's no platform specified), and it assumes a set of generated tasks (which we can't generate at this same time).  At best, we could have a bare-minimum snippet for each supported platform (i.e. .NET Core and Node.js) and leave off the `preLaunchTask`, or set it to something obviously requiring a change by the user (e.g. `set-this-your-docker-build-task-name`).

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -0,0 +1,308 @@
...
+                runOptions.env.ASPNETCORE_URLS = runOptions.env.ASPNETCORE_URLS || 'https://+:443;http://+:80';
+            }
+        }
+
+        runOptions.volumes = await this.inferVolumes(context.folder, runOptions, helperOptions, ssl, userSecrets); // Volumes specifically are unioned with the user input (their input does not override except where the container path is the same)
+
+        return runOptions;
+    }
+
+    public static async inferAppName(folder: WorkspaceFolder, helperOptions: NetCoreTaskOptions | NetCoreDebugOptions): Promise<string> {
```
sgreenmsft: >inferAppName [](start = 24, length = 12)

Contra this function's name, its purpose seems to be to generate a valid image name from a given project file name.  But it's also used to generate a container name from the project file name.

Why not separate these concerns into three functions createValidImageName, createValidContainerName, getProjectName?

(I also object to "AppName" being used instead of "ProjectName" as AppName doesn't correspond to an established concept in C# and will probably be confused for the OutputAssemblyName).

src/commands/debugging/initializeForDebugging.ts
bwateratmsft: It's not a prompt, rather it's the error message displayed if nothing is found. In fact I should probably change it to "No C# or F# project files were found."

src/commands/debugging/initializeForDebugging.ts
bwateratmsft: Does that even work?

src/tasks/TaskHelper.ts
bwateratmsft: Do you have an alternative suggestion? Adding another field just for this is overly verbose.

src/tasks/TaskHelper.ts
bwateratmsft: You need to recursively follow down your dependency chain in order to find the task you want. The dependency chain can have multiple dependsOn tasks.

src/tasks/TaskHelper.ts
bwateratmsft: Yes, VSCode is case-sensitive about task type.

src/tasks/TaskHelper.ts
bwateratmsft: Yes, VSCode is case-sensitive about task label.

src/tasks/TaskHelper.ts
bwateratmsft: Yes. Most of our users are on Unix. If they have incorrect casing in what they specify, it won't work for them anyway.

src/debugging/DockerPlatformHelper.ts
bwateratmsft: Yes. VSCode is case-sensitive in task and launch configs.

src/debugging/DockerPlatformHelper.ts
bwateratmsft: Yes. VSCode is case-sensitive in task and launch configs.

src/commands/images/buildImage.ts
bwateratmsft: No, but this isn't code I wrote and nobody's complained in the past.

src/configureWorkspace/configureDotNetCore.ts
bwateratmsft: How so?

src/debugging/DebugHelper.ts
bwateratmsft: Yes. VSCode is case-sensitive in task and launch configs.

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: It's technically possible to have debug configurations that apply even when a workspace folder is not open. Our extension is not one that supports this. So in that case, we prompt.

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: Never, since we prompt above. I'll remove this.

src/tasks/TaskHelper.ts
bwateratmsft: Yes, in ShellQuotedString can't have an undefined value.

src/debugging/netcore/NetCoreDebugHelper.ts
bwateratmsft: A whole lot of harmless nothing. :)

src/debugging/netcore/NetCoreDebugHelper.ts
bwateratmsft: Yes, VSCode is case-sensitive in launch and task configs.

package.json
bwateratmsft: I think we probably should not provide snippets at all...it's highly unlikely that the user is able to hand-craft an actual, working launch config without a lot of trial and error.

src/debugging/netcore/NetCoreDebugHelper.ts
bwateratmsft: It shouldn't, VSCode does this currently with the structure of vsdbg directory:

- User Folder
  - .vsdbg
    - linux-x64/linux-musl-x64/win7-x64
      - vs2017u5 (now, 'latest')
    - GetVsDbg.ps1/.sh
    - vsdbg (this script)

VS uses 'vsdbg', not '.vsdbg' as the folder name. The fact they aren't sharing bits is a problem of its own, but maybe not a huge one if we expect users don't switch much between VS and VSCode.

package.json
bwateratmsft: This snippet is a little odd. It's provided by us, but is strictly a Node Attach thing, i.e. we don't actually do the work of supporting it. I'll leave it in for now since it's probably harmless, but we should consider removing it in the future.

src/tasks/netcore/NetCoreTaskHelper.ts
bwateratmsft: I created a getImageName and getContainerName, and consequently there's no need at all of an appName.

src/tasks/TaskHelper.ts
bwateratmsft: Here's my suggestion: If there's more than one, and one is called "docker-build: release" then use it (the most common case); otherwise quickpick.

src/commands/images/buildImage.ts
```diff
@@ -20,37 +21,42 @@ export async function buildImage(context: IActionContext, dockerFileUri: vscode.
```
bwateratmsft: Copying in Steven's comment since it didn't work right:

If the dockerFileUri !== undefined, then isn't the workspace folder implied by that uri?

bwateratmsft: I'm not sure exactly. It might be possible (though horrible) to have two workspace folders open, one which is inside the other--in which case, both are equally valid as this Dockerfile's workspace folder.

I'll test it out.

src/debugging/netcore/NetCoreDebugHelper.ts
sgreenmsft: Note that older versions of VS use ".vsdbg"

bwateratmsft: Good point. I checked out the old source, and it does this:

- User folder
  - .vsdbg
    - vs2017u5
      -linux-x64

So this shouldn't conflict, since we're doing .vsdbg\*\latest, instead of .vsdbg\vs2017u5\*.

src/commands/images/buildImage.ts
bwateratmsft: Turns out, yes you can. So in this case I'd prefer to just leave the old code untouched.

src/configureWorkspace/configureDotNetCore.ts
sgreenmsft: Leads to confusion around whether a variable named "platform" means "Linux" or it means "netCore"

src/commands/images/buildImage.ts
sgreenmsft: Suggest to consolidate with the code you've written to generate a valid image name

src/configureWorkspace/configureDotNetCore.ts
bwateratmsft: Well, I don't think there's anywhere in the code where we call a PlatformOS 'platform'; it's always 'os' or 'platformOS'. I'd prefer not to refactor nearly every file in the source just for this.

src/tasks/TaskHelper.ts
sgreenmsft: macOS and Windows use (by default) case-insensitive file path comparison.  Those platforms are 80% of our usage.

src/commands/images/buildImage.ts
sgreenmsft: Perhaps we should notify the user that they've made bad life choices if we encounter this nested workspace scenario.

src/tasks/TaskHelper.ts
bwateratmsft: Very well. Made it case-insensitive.

src/commands/images/buildImage.ts
bwateratmsft: Probably, but this is a corner scenario already...quickPickWorkspaceFolder doesn't show a quick pick if there's only one choice, which is the overwhelming majority of cases.

src/tasks/TaskHelper.ts
philliphoff: Early on we had discussed adding a "config is for debugging" flag in the definition to allow us to more easily locate the likeliest configuration (as there's likely only two configurations per Dockerfile).  The Node.js `TaskHelper` already has such a flag (as it's needed for other reasons).  We could either make that a common flag or defer selection to the individual task helpers (which could use their own, potentially different logic).

I think I prefer the first option (the common debugging flag) myself.

package.json
philliphoff: If we're keeping this snippet, probably should keep the equivalent one provided by the dummy configuration provider.  That said, there's no real reason for us to keep/register that configuration provider; maybe we should fold that into our own.  We could even rename it `'docker'` as I'm thinking now that `'docker-launch'` somewhat conflates the idea that a debugger may operate in both `'launch'` and `'attach'` modes.  In the case of .NET Core, the configuration provider uses `'launch'` under the covers, but, in the case of Node.js, the configuration provider uses `'attach'` under the covers.  Technically, the configuration provider's not really "launching" Docker; the container is now being build/spun-up by the pre-launch tasks.

package.json
philliphoff: I think it's a matter of helping the user discover functionality; it the user expects to find possible debugging configurations via the "add configuration" entrypoint, then we've made it more difficult to use our features by only exposing them via the command palette.  Sure, we've perhaps mitigated that somewhat by adding the tasks/debug configuration as part of Dockerfile scaffolding, but it might still be useful to do both (even if I agree that we can't ensure an end-to-end working flow via the snippet mechanism--we should push VS Code to try to resolve that API inconsistency).

src/tasks/TaskHelper.ts
bwateratmsft: I don't agree with that approach. Allow me to explain.

If we add such a flag, we cannot use it to control logic within the task helpers. To do so would be to do "magic", something the users cannot look into (unless they go digging through our source), and they cannot control, etc. Magic === evil.

So that would leave the flag only useful for this one specific scenario of deciding which build config is the release one vs. debug one. If we think that users might change the name and take "release" out, they can just as easily change this hypothetical flag. If we fall back to using a quick-pick if there's no matching config of this name, what's the down side? We avoid an unnecessary flag and have equally functional behavior.

package.json
bwateratmsft: Consider also that it's not just the debug configuration that they must trial-and-error to figure out, they have to get both the docker-build and docker-run tasks as well. I can imagine an average user failing tens of times before things work. A snippet here is just going to lead them down a path to the Dark Side (TM).

package.json
bwateratmsft: Good point; I'd forgotten about the dummy debug config provider.

I'd rather not try to fold together that and this; I'm concerned there's too much risk of regressions to existing configs and breaking future work.

src/tasks/TaskHelper.ts
sgreenmsft: How would you feel about an 'isDefaultBuildTask' flag to indicate the task to use by default, with a fallback to quick pick when no default (or multiple defaults) are found?

bwateratmsft: VSCode has a notion for that already; you can see it with Ctrl+Shift+P > Configure default build task. It adds `"group": { "kind": "build", "isDefault" : true }`. I wouldn't really want to add something that has similar appearance but totally different effect.

sgreenmsft: What if we called it "isDefaultDockerTask"?

bwateratmsft: Yeah, that would be better. I'm going to think about it some and see if there's a clear, existing indicator. I still think a dedicated property may be a bit over-engineered, especially considering as far as we know this debug/release build scenario only really applies to .NET Core.

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    private async registerRemoveContainerAfterDebugging(resolvedConfiguration: ResolvedDebugConfiguration): Promise<void> {
+        if (resolvedConfiguration.dockerOptions !== undefined
+            && (resolvedConfiguration.dockerOptions.removeContainerAfterDebug === undefined || resolvedConfiguration.dockerOptions.removeContainerAfterDebug)
+            && resolvedConfiguration.dockerOptions.containerNameToKill !== undefined) {
+            const runningContainerName =
+                (await this.dockerClient.listContainers({ format: '{{.Names}}' }))
+                    .split('\n')
```
sgreenmsft: >split [](start = 21, length = 5)

Does this work on windows or do the strings end up with a trailing \r?

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+                    .find(name => name.toLowerCase() === resolvedConfiguration.dockerOptions.containerNameToKill.toLowerCase());
+
+            if (runningContainerName) {
+                await this.dockerClient.removeContainer(runningContainerName, { force: true });
```
sgreenmsft: >removeContainer [](start = 40, length = 15)

It's non-obvious to me why we do the removeContainer call and then register a callback to do the removeContainer call.  Perhaps a comment?

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -0,0 +1,298 @@
...
+    public async resolveDockerRunOptions(context: DockerRunTaskContext, runDefinition: NetCoreRunTaskDefinition): Promise<DockerRunOptions> {
+        const runOptions = runDefinition.dockerRun;
+        const helperOptions = runDefinition.netCore || {};
+
+        helperOptions.appProject = await NetCoreTaskHelper.inferAppProject(context.folder, helperOptions); // This method internally checks the user-defined input first
+
+        runOptions.containerName = runOptions.containerName || await NetCoreTaskHelper.getContainerName(helperOptions.appProject);
+        runOptions.labels = runOptions.labels || NetCoreTaskHelper.defaultLabels;
+        runOptions.os = runOptions.os || 'Linux';
+        runOptions.image = runOptions.image || await NetCoreTaskHelper.getImageName(helperOptions.appProject);
```
sgreenmsft: >getImageName [](start = 71, length = 12)

Just want to verify: if in my build task I specify a full image name (e.g. myapp:v1), and in my run task I do not specify any image name, then the run will still figure out that the image name is myapp:v1, right?

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: Agreed a comment would be good here. The reason is that, in the normal flow, we remove the container after debugging. But, if for some reason that doesn't happen, we can try removing in advance. One such example of the latter that we ran into is if you kill the experimental instance while it is debugging an app, it won't shut down the container, so it is left there.

src/tasks/netcore/NetCoreTaskHelper.ts
bwateratmsft: I'd intended to implement as such but I either forgot or it got removed at some point. Putting that in now. @philliphoff I'm adding it for Node as well.

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: I added some `.trim()`'s to the comparison to make sure any trailing/leading whitespace gets removed.

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: We could save a little bit of time during debug by simply doing this `this.dockerClient.removeContainer` without first listing the containers, and wrapping it in a try/catch {}, as a sort of best-effort kill. It would save us from one CLI execution. @philliphoff @sgreenmsft what do you think?

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+            .withNamedArg('--target', options.target)
+            .withQuotedArg(options.context)
+            .buildShellQuotedStrings();
+    }
+
+    private getHelper(platform: DockerPlatform): TaskHelper {
+        const helper = this.helpers[platform];
+
+        if (!helper) {
+            throw new Error(`The platform '${platform}' is not currently supported for Docker build tasks.`);
```
bwateratmsft: @philliphoff It occurs to me that if we allow for no helper, they can use our tasks to define any imaginable docker-build / docker-run. As long as they provide sufficient information to build/start the container, we don't technically need to infer anything extra.

So, I think we should _not_ fail if no helper is available. (For tasks that is, but not for debug)

bwateratmsft: The benefit isn't huge; they can already do this via any arbitrary commandline task, but it lets them take advantage of our intellisense on task definition to more easily create a functional command line.

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: I changed it to that as it was having issues if the container was created but not running.

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -0,0 +1,116 @@
...
+    constructor(private readonly helpers: { [key in DockerPlatform]: TaskHelper }) { }
+
+    public provideTasks(token?: CancellationToken): ProviderResult<Task[]> {
+        return []; // Intentionally empty, so that resolveTask gets used
+    }
+
+    public resolveTask(task: DockerBuildTask, token?: CancellationToken): ProviderResult<Task> {
+        const taskPlatform = getPlatform(task.definition);
+        return callWithTelemetryAndErrorHandling(
+            `docker-build-resolve/${taskPlatform || 'unknown'}`,
```
EricJizbaMSFT: We don't usually have dynamic event names like this. It's simpler to have one event name that you only have to register once with gdpr and then stick `taskPlatform` as a property on `context.telemetry.properties`

src/tasks/TaskHelper.ts
EricJizbaMSFT: Per Steven's comment, I'd prefer to see unit tests for this kind of difficult-to-understand logic

src/tasks/node/NodeTaskHelper.ts
```diff
@@ -0,0 +1,212 @@
...
+    public static inferPackagePath(packageFile: string | undefined, folder: WorkspaceFolder): string {
+        if (packageFile !== undefined) {
+            return resolveFilePath(packageFile, folder);
+        } else {
+            return path.join(folder.uri.fsPath, 'package.json');
+        }
+    }
+
+    public static async inferPackageName(packagePath: string): Promise<string> {
+        const packageJson = await fse.readFile(packagePath, 'utf8');
```
EricJizbaMSFT: Use `fse.readJSON` instead of `fse.readFile` and `JSON.parse`. In general it's more robust and handles some weird file encodings and whatnot

src/commands/images/buildImage.ts
```diff
@@ -20,37 +22,31 @@ export async function buildImage(context: IActionContext, dockerFileUri: vscode.
...
-        // Get imageName based on name of subfolder containing the Dockerfile, or else workspacefolder
-        suggestedImageName = path.basename(dockerFileItem.relativeFolderPath).toLowerCase();
-        if (suggestedImageName === '.') {
-            suggestedImageName = path.basename(rootFolder.uri.fsPath).toLowerCase();
-        }
-
-        suggestedImageName += ":latest"
+    const task = await getOfficialBuildTaskForDockerfile(dockerFileItem.absoluteFilePath, rootFolder);
+    if (task) {
+        await vscode.tasks.executeTask(task);
```
EricJizbaMSFT: Just a side note, but the user's default terminal can definitely cause issues. Make sure to test with cmd, PowerShell, and git bash as your default terminal on Windows

src/tasks/node/NodeTaskHelper.ts
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        return runOptions;
+    }
+
+    public static inferPackagePath(packageFile: string | undefined, folder: WorkspaceFolder): string {
+        if (packageFile !== undefined) {
+            return resolveFilePath(packageFile, folder);
+        } else {
+            return path.join(folder.uri.fsPath, 'package.json');
```
EricJizbaMSFT: One common place for problems is a lot of users will put their project in a subfolder of their workspace. It's common for people that had a C# project in VS and then open it up in VS Code for the first time. It affects other languages, too - some people just like having the folder structure like that. 

Honestly you may or may not support those scenarios already - didn't fully investigate in this PR, but just wanted to make sure you considered that scenario

src/utils/resolveFilePath.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Given a file path, tries to determine an absolute path. If relative, path is resolved relative to workspace folder.
+ * @param filePath The file path to resolve
+ * @param folder The workspace folder
+ */
+export function resolveFilePath(filePath: string, folder: WorkspaceFolder): string {
+    let replacedPath = filePath.replace(/\$\{workspaceFolder\}/gi, folder.uri.fsPath);
+
+    if (os.platform() === 'win32') {
```
EricJizbaMSFT: Is this stuff necessary? I thought `path.resolve` did some normalization of the separators

package.json
```diff
@@ -89,8 +90,10 @@
         "onCommand:vscode-docker.volumes.prune",
         "onCommand:vscode-docker.volumes.refresh",
         "onCommand:vscode-docker.volumes.remove",
+        "onCommand:workbench.action.tasks.runTask",
```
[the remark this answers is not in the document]
bwateratmsft: [Nope](https://code.visualstudio.com/api/references/activation-events) :(

package.json
```diff
@@ -676,6 +797,260 @@
                 }
             }
         ],
+        "taskDefinitions": [
+            {
+                "type": "docker-build",
+                "properties": {
+                    "dockerBuild": {
```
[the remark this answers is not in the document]
bwateratmsft: Here, `dockerBuild` is for all the options you'd pass to it, like --pull, args, etc.

package.json
```diff
@@ -676,6 +797,260 @@
...
+                        "description": "Options for building the Docker image used for debugging.",
+                        "properties": {
+                            "args": {
+                                "type": "object",
+                                "description": "Build arguments applied to the Docker image used for debugging.",
+                                "additionalProperties": {
+                                    "type": "string"
+                                }
+                            },
+                            "context": {
```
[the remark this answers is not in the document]
bwateratmsft: Context would be the Dockerfile build context, which is technically different I guess?

src/tasks/TaskHelper.ts
```diff
@@ -0,0 +1,204 @@
...
+            }
+
+            arg = resolveFilePath(arg, folder);
+            return arg.toLowerCase() === dockerfile.toLowerCase();
+        }));
+
+    if (buildTasks.length === 1) {
+        return buildTasks[0]; // If there's only one build task, take it
+    } else if (buildTasks.length > 1) {
+        const releaseTask = buildTasks.find(t => t.name === 'docker-build: release');
```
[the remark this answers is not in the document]
bwateratmsft: For debugging cases yes, but this is for the right-click Dockerfile > Build Image case, which in VS (and historically in VSCode) was for building a "release"y image.

src/configureWorkspace/configure.ts
```diff
@@ -344,7 +357,9 @@ export async function configure(context: IActionContext, rootFolderPath: string
         {
             rootPath: rootFolderPath,
             outputFolder: rootFolderPath,
-            openDockerFile: true
+            openDockerFile: true,
+            configureForDebugging: true,
```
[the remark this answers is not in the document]
bwateratmsft: For now no, but for any languages that don't have a debugging initializer, nothing will happen.

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+import { DebugHelper, DockerDebugContext, ResolvedDebugConfiguration } from './DebugHelper';
+import { DockerPlatform, getPlatform } from './DockerPlatformHelper';
+import { NetCoreDockerDebugConfiguration } from './netcore/NetCoreDebugHelper';
+import { NodeDockerDebugConfiguration } from './node/NodeDebugHelper';
+
+export interface DockerDebugConfiguration extends NetCoreDockerDebugConfiguration, NodeDockerDebugConfiguration {
+    platform?: DockerPlatform;
+}
+
+export class DockerDebugConfigurationProvider implements DebugConfigurationProvider {
```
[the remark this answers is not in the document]
bwateratmsft: I might rename the other to Legacy...Provider, since it's basically a do-nothing provider as it is...

src/debugging/node/NodeDebugHelper.ts
```diff
@@ -0,0 +1,101 @@
...
+            removeContainerAfterDebug: debugConfiguration.removeContainerAfterDebug
+        };
+
+        const resolvedConfiguration: NodeDebugConfiguration = {
+            ...options,
+            name: debugConfiguration.name,
+            dockerOptions,
+            preLaunchTask: debugConfiguration.preLaunchTask,
+            request: 'attach',
+            type: 'node2'
```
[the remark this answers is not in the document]
bwateratmsft: @philliphoff ?

src/configureWorkspace/configUtils.ts
```diff
@@ -95,3 +95,20 @@ export async function quickPickOS(): Promise<PlatformOS> {
...
 }
+
+export async function quickPickGenerateComposeFiles(): Promise<boolean> {
+    let opt: vscode.QuickPickOptions = {
+        matchOnDescription: true,
+        matchOnDetail: true,
+        placeHolder: 'Add Docker Compose files'
+    }
+
+    let response = await ext.ui.showQuickPick(
```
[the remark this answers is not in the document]
bwateratmsft: @philliphoff ?

src/configureWorkspace/configUtils.ts
```diff
@@ -95,3 +95,20 @@ export async function quickPickOS(): Promise<PlatformOS> {
     let response = await ext.ui.showQuickPick(items, opt);
     return response.data;
 }
+
+export async function quickPickGenerateComposeFiles(): Promise<boolean> {
+    let opt: vscode.QuickPickOptions = {
+        matchOnDescription: true,
+        matchOnDetail: true,
```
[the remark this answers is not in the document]
bwateratmsft: @philliphoff ?

src/tasks/DockerBuildTaskProvider.ts
bwateratmsft: Yeah, this was before I'd figured out IActionContext and attached the platform as a property of it

src/commands/images/buildImage.ts
bwateratmsft: Will do.

src/tasks/node/NodeTaskHelper.ts
bwateratmsft: Yes, I tried it out. By default VS makes such projects like this:

 - MySolution.sln
 - MyWebApp
   - MyWebApp.csproj
 - MyLibrary
   - MyLibrary.csproj

I tested with this config and it does work.

src/utils/resolveFilePath.ts
bwateratmsft: I thought so too! But it doesn't :(

src/debugging/DebugHelper.ts
```diff
@@ -0,0 +1,111 @@
...
```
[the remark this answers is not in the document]
bwateratmsft: Good idea.

src/commands/debugging/initializeForDebugging.ts
```diff
@@ -0,0 +1,53 @@
...
+    let debugPlatform: DockerPlatform;
+    switch (platform) {
+        case '.NET Core Console':
+        case 'ASP.NET Core':
+            debugPlatform = 'netCore';
+            break;
+        case 'Node.js':
+            debugPlatform = 'node';
+            break;
+        default:
```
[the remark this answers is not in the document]
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/tasks/TaskHelper.ts
```diff
@@ -0,0 +1,204 @@
...
+
+        if (releaseTask) {
+            return releaseTask;
+        }
+
+        const items: QuickPickItem[] = buildTasks.map(t => {
+            return { label: t.name }
+        });
+
+        const item = await ext.ui.showQuickPick(items, { placeHolder: 'Choose the official Docker Build definition.' });
```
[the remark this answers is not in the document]
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -260,3 +266,19 @@ function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, o
...
+async function initializeForDebugging(context: IActionContext, folder: WorkspaceFolder, platformOS: PlatformOS, dockerfile: string, { artifactName }: Partial<PackageInfo>): Promise<void> {
+    const scaffoldContext: DockerDebugScaffoldContext = {
+        folder: folder,
+        platform: 'netCore',
+        actionContext: context,
+        dockerfile: dockerfile,
+    }
+
+    const options: NetCoreScaffoldingOptions = {
+        appProject: `\${workspaceFolder}/${artifactName}`,
```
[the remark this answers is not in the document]
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+        return undefined;
+    }
+
+    public resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration | undefined> {
+        const debugPlatform = getPlatform(debugConfiguration);
+        return callWithTelemetryAndErrorHandling(
+            `docker-launch/${debugPlatform || 'unknown'}`,
+            async (actionContext: IActionContext) => await this.resolveDebugConfigurationInternal(
+                {
+                    folder: folder || await quickPickWorkspaceFolder('To debug with Docker you must first open a folder or workspace in VS Code.'),
```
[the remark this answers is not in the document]
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/debugging/DockerDebugConfigurationProvider.ts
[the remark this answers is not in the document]
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+                    if (sessionConfiguration
+                        && sessionConfiguration.dockerOptions
+                        && sessionConfiguration.dockerOptions.containerNameToKill === resolvedConfiguration.dockerOptions.containerNameToKill) {
+                        await this.dockerClient.removeContainer(resolvedConfiguration.dockerOptions.containerNameToKill, { force: true });
+                        disposable.dispose();
+                    } else {
+                        return; // Return without disposing--this isn't our debug session
+                    }
+                } catch {
+                    disposable.dispose();
```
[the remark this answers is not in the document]
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/debugging/DockerDebugScaffoldingProvider.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        for (const debugConfiguration of debugConfigurations) {
+            await addDebugConfiguration(debugConfiguration);
+        }
+    }
+}
+
+const dockerDebugScaffoldingProvider: IDockerDebugScaffoldingProvider = new DockerDebugScaffoldingProvider();
+
+export default dockerDebugScaffoldingProvider;
```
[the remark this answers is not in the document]
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/tasks/DockerBuildTaskProvider.ts
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/tasks/node/NodeTaskHelper.ts
bwateratmsft: Resolved in: https://github.com/microsoft/vscode-docker/pull/1278

src/debugging/DockerDebugScaffoldingProvider.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface IDockerDebugScaffoldingProvider {
+    initializeNetCoreForDebugging(context: DockerDebugScaffoldContext, options?: NetCoreScaffoldingOptions): Promise<void>;
+    initializeNodeForDebugging(context: DockerDebugScaffoldContext): Promise<void>;
+}
+
+export class DockerDebugScaffoldingProvider implements IDockerDebugScaffoldingProvider {
+    public async initializeNetCoreForDebugging(context: DockerDebugScaffoldContext, options?: NetCoreScaffoldingOptions): Promise<void> {
+        await this.initializeForDebugging(
+            context,
+            (_context: DockerDebugScaffoldContext) => netCoreDebugHelper.provideDebugConfigurations(_context, options),
```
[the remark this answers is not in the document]
bwateratmsft: Resolved in #1278

package.json
```diff
@@ -676,6 +797,260 @@
...
             }
         ],
+        "taskDefinitions": [
+            {
+                "type": "docker-build",
+                "properties": {
+                    "dockerBuild": {
+                        "description": "Options for building the Docker image used for debugging.",
+                        "properties": {
+                            "args": {
```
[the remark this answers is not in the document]
bwateratmsft: So Docker's --build-arg is a weird ugly thing that essentially is a key-value pairing of arguments. I'm actually going to change it from 'args' to 'buildArgs', but leave it as an object...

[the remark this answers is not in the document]
bwateratmsft: Resolved in #1278

src/configureWorkspace/configure.ts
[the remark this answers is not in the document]
EricJizbaMSFT: I like to keep APIs as simple as possible and generally only add things if they're requested. Makes it easier to maintain backwards compatibility and whatnot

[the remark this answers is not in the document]
bwateratmsft: That's fair.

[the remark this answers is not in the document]
bwateratmsft: Resolved in #1278

package.json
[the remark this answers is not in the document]
EricJizbaMSFT: So this is what a few example random tasks look like:
```
{
	"type": "shell",
	"label": "npm lint",
	"command": "npm run lint"
},
{
	"type": "npm",
	"script": "lint",
	"problemMatcher": "$tslint5"
},
```

I feel like this is what yours is doing:
```
{
	"type": "shell",
	"shell": {
		"command": "npm run lint"
	},
	"label": "npm lint",
},
{
	"type": "npm",
	"npm": {
		"script": "lint",
	},
	"problemMatcher": "$tslint5"
},
```

The task inherently _is_ the object where you put any options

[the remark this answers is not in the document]
bwateratmsft: I get that, but there's also things that aren't things that get fed to the `docker build` call--like our platform-specific options (`node`, `netCore`). On the debug launch config there's also meta-options like `dockerServerReadyAction`; we could end up with such meta-options for these tasks too.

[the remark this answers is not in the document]
EricJizbaMSFT: But the platform-specific options _do_ describe things that get fed to the `docker build` call, including the container name and command.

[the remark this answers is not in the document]
bwateratmsft: ...but also things like the .NET Core project, which is not directly part of `docker build`, and in the case of `docker-run` whether to configure SSL or not, which does a host things--volumes, commands run on the host, changes to the project file itself...

src/debugging/DebugHelper.ts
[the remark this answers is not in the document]
bwateratmsft: Resolved in #1278

src/debugging/DebugHelper.ts
```diff
@@ -0,0 +1,111 @@
...
+        )
+    );
+
+    registerServerReadyAction(ctx);
+
+    registerCommand('vscode-docker.debugging.initializeForDebugging', initializeForDebugging);
+}
+
+// TODO: This is stripping out a level of indentation, but the tasks one isn't
+export async function addDebugConfiguration(debugConfiguration: DockerDebugConfiguration): Promise<boolean> {
```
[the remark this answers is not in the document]
bwateratmsft: (Now is being used in #1278)

src/commands/images/buildImage.ts
bwateratmsft: CWD and PowerShell work fine but Git Bash is replacing `/dev/null` in our run command with `nul`. Stupid shell. Need to figure out how to override it.

bwateratmsft: Well, couldn't figure out how to override it, but I found that the entrypoint wasn't even needed. Tested all three across both Windows and Linux containers. Resolved in #1278.

Issue #1270, opened 2019-09-13
Invalid webhookName when deploying to app service


EricJizbaMSFT:
Saw this error popping up in telemetry and tried to repro myself. This is what I came up with

Repro steps:

1. Right click on a registry image -> Deploy image to azure app service
2. Enter a web app name with a hyphen and/or greater than 50 characters
3. Fill out the rest of the steps and wait for deployment to finish

Action: vscode-docker.registries.deployImageToAzure
Error type: Error
Error Message: "webhookName" should satisfy the constraint - "Pattern": /^[a-zA-Z0-9]*$/

Version: 0.8.0
OS: darwin
Product: Visual Studio Code - Insiders
Product Version: 1.39.0-insider
Language: en

<details>
<summary>Call Stack</summary>

```
Webhooks.f [as _beginCreate] extension.bundle.js:332:206660
Webhooks.beginCreate extension.bundle.js:332:224292
Webhooks.a [as _create] extension.bundle.js:332:189966
module.exports.e.exports.create.Promise extension.bundle.js:332:221705
Webhooks.create extension.bundle.js:332:221683
DockerWebhookCreateStep.createWebhookForApp extension.bundle.js:332:296664
process._tickCallback next_tick.js:68:7
```

</details>

PR #1271, opened 2019-09-13
Ensure webhookName meets requirements

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-docker/issues/1270

Per the portal, a webhook name has these requirements
Screen Shot 2019-09-13 at 10 22 14 AM

But a site name has looser requirements (allows hyphens and 3-64 characters I think). I scrapped the existing logic to generate a name and based it more closely off what the portal does:
Screen Shot 2019-09-13 at 10 15 57 AM

### DIFF
--- a/src/commands/registries/azure/DockerWebhookCreateStep.ts
+++ b/src/commands/registries/azure/DockerWebhookCreateStep.ts
@@ -17,6 +17,7 @@ import { DockerHubRepositoryTreeItem } from '../../../tree/registries/dockerHub/
 import { RemoteTagTreeItem } from '../../../tree/registries/RemoteTagTreeItem';
 import { nonNullProp } from "../../../utils/nonNull";
 import { openExternal } from '../../../utils/openExternal';
+import { randomUtils } from '../../../utils/randomUtils';
 
 export class DockerWebhookCreateStep extends AzureWizardExecuteStep<IAppServiceWizardContext> {
     public priority: number = 141; // execute after DockerSiteCreate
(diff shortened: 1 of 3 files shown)


philliphoff:
As the site name (minus invalid characters) passes 44 chars in length, there's an increasing chance for collision.  (An unlikely edge case, I know.)  What's the error message in the case of a colission?

PrashanthCorp:
I tried testing a collision (by hot-loading the same value for the `webhookName` variable). There's a plain overwrite. The `webhooks.create` method doesn't complain when the name is the same as before.

src/commands/registries/azure/DockerWebhookCreateStep.ts
```diff
@@ -67,19 +68,17 @@ export class DockerWebhookCreateStep extends AzureWizardExecuteStep<IAppServiceW
...
-        let baseName = `webapp${siteName}`;
-        let webhookName: string = baseName;
+        let webhookName: string = site.name;
+        // remove disallowed characters
+        webhookName = webhookName.replace(/[^a-zA-Z0-9]/g, '');
+        // add random chars for uniqueness and to ensure min length is met
+        webhookName += randomUtils.getRandomHexString(6);
+        // trim to max length
+        webhookName = webhookName.substr(0, 50);
+
```
PrashanthCorp: Is there a reason we removed the check on whether the name already exists?  I realize it may not be fully actionable by the code, but it feels like we can have info that can either be surfaced (possible collision, etc.). 

The current behavior in case of a namespace clash is too silent, IMO.

philliphoff: I think we could add random-ness to the generated webhook name while still comparing it to the known webhook names, until we get a non-match (similar to what was done before).  The difference would be to overlap the randomness with the base name when the base name is too long.  (I.e. don't just add the randomness to the end, only to then have it clipped due to length.)

philliphoff:
> The webhooks.create method doesn't complain when the name is the same as before.

@PrashanthCorp That's a little worrying, as there's always going to be the chance for a race.  I would have expected the "loser" to generate an error.  Is that worth filing a bug on the SDK?

src/commands/registries/azure/DockerWebhookCreateStep.ts
EricJizbaMSFT: > Is there a reason we removed the check on whether the name already exists? I realize it may not be fully actionable by the code, but it feels like we can have info that can either be surfaced (possible collision, etc.).

The call to list webhooks is not guaranteed to return all webhooks in the first batch. So the logic might not even prevent collisions. More importantly, the existing code added an underscore to the name, which is not allowed for webhook names. Which means the code was not even tested and thus didn't seem worth keeping.

EricJizbaMSFT: > The difference would be to overlap the randomness with the base name when the base name is too long.

Done

> I think we could add random-ness to the generated webhook name while still comparing it to the known webhook names, until we get a non-match (similar to what was done before).

The global uniqueness of the site name and the randomly generated chars should already be more than enough to prevent collisions IMO. With that it mind, it just doesn't seem worth it to make the code more complex because I feel like that also adds risk. For example, any while loop could cause the product to hang if not done properly. Also as mentioned above, the existing loop _did_ cause a bug which was never found.

philliphoff: This seems like it should work well enough.

philliphoff:
:shipit:

Issue #1217, opened 2019-08-20
Run/Run Interactive: do not specify -rm


bpasero:
I am using the Docker extension to create a container from an image and I picked the "Run Interactive" command. To my surprise all my data was lost when the container stopped. This seems to be due to the -rm option that is passed.

I suggest to not add this.

@chrmarti cc

xgdgsc:
I lost my 2 hours of config because of this!

bwateratmsft:
This will be fixed by #1596 so I will dupe this one to that.

chrmarti:
@bwateratmsft The change to fix #1596 seems to keep `--rm` in (https://github.com/microsoft/vscode-docker/pull/1649/files).

bwateratmsft:
@chrmarti Yes; by design. But it's trivial to customize now.

Issue #1001, opened 2019-06-18
Add port and volume options to 'Run interactive'


1138-4EB:
I find myself frequently writing commands as:

``` bash
$(command -v winpty) docker run --rm -itv /$(pwd)://src -p 8000:8000 image_name bash
```

or

``` bash
$(command -v winpty) docker run --rm -dv /$(pwd)://src -p 8000:8000 image_name bash -c "tail -f /dev/null"
```

in order to start a container to test/try some random project from GitHub. I then use Visual Studio Code's `Remote - Containers` extension along with this extension to attach to it, to run additional interactive sessions, and to remove the container. It would be great if this extension provided the feature to create these custom 'run' commands.

Ideally, a `Run Interactive Opts` option would be added to the context-menu of the images; which, instead of starting a session right away, would add an item to a (new) `Configurations` list below. Then, the context-menu of each item in the list would allow to add/remove volumes/ports, to start the session, and to rename/remove the item.

This would greatly improve UX thanks to the ability to select the paths graphically (taking the workspace as a reference) and to show a list of recently used paths/ports.

I know that multiple alternatives exist which use configuration files of different kinds. However, I want to precisely avoid having to type all the paths. These are just some temporal/stash configurations that the user might later want to save/export, but that will be frequently discarded.

1138-4EB:
As a side note, when a multi-root workspace is open and a new terminal is created, a list is shown for the user to select which of the folders is to be used as the root/workdir of the terminal. The same approach can be used here.

bwateratmsft:
Since it covers the same scenario, I'm going to resolve this one as a dupe of #1321

1138-4EB:
@bwateratmsft, on the one hand, I tried https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=10975 and https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=10908 in order to test #1245 nd #1327, which are referenced in #1321, but I couldn't find how to use them. I saw that command `Docker: initialize for Docker debugging` was added. The first step requires to select a directory, and then three options are shown: Node.js, ASP.NET Core, .NET Core Console. Then, I had a look at 'Tasks' and tried to create a `tasks.json` file. However, I don't understand why would anyone write such a file instead of a docker CLI command. Therefore, I must be missing something, because I don't see how #1321 and this issue are solved.

[...]

Issue #331, opened 2018-07-25
Support for multiple selection of items and actions on these items (batch)


wlsc:
It would be very helpful, if one would have an opportunity to select multiple, for instance, containers and then remove or stop containers with just one click. Current behavior: one is forced to select and remove/stop containers one by one.

StephenWeatherford:
Not sure this is possible in the views, need to find out.

StephenWeatherford:
@wlsc FYI, if you select "Stop Container" from the Containers node directly, there is an "All Containers" option.  Does that suffice for you?

If not and view  can't do it in the view, we could consider using a checklist when people do this from Containers node.

wlsc:
@StephenWeatherford thanks for your reply. I have a use case where I do need to stop/remove 3-4 containers from 9 in total. Unfortunately, I do not want to stop or remove them all.

EricJizbaMSFT:
@wlsc it's been a while - but I believe this was fixed for you in #966. If you run "Stop Container" from the command palette, you will see something like this:
![Screen Shot 2019-06-10 at 9 08 15 AM](https://user-images.githubusercontent.com/11282622/59209063-52ded500-8b5f-11e9-907e-854afe697970.png)

You can try out the latest from master with these instructions: https://aka.ms/azcodeInsiders

The only reason I didn't close this yet is we still don't allow users to multi-select directly in the tree. Not sure how necessary that part of the feature is.

1138-4EB:
A possible intermediate solution is to add a 'Action'/'Do' button in the CONTAINERS title, on the left of 'Prune'. It would be a dropdown with 'stop', 'remove', etc. Each of the items would point to the corresponding 'Docker Containers: ...' command.

EricJizbaMSFT:
Multi-select support for custom trees in VS Code is being tracked here: https://github.com/microsoft/vscode/issues/76941

@1138-4EB It's in the "August 2019" milestone so I guess I'd prefer to just wait for that

1138-4EB:
@bwateratmsft, is multi-select in the trees released already? I don't find the feature in `September 2019 (version 1.39)` and `0.8.1`.

bwateratmsft:
I was referring to this comment: https://github.com/microsoft/vscode-docker/issues/331#issuecomment-500477876

Is that insufficient? If so I will reopen this.

1138-4EB:
Precisely, the last part of that comment said/says:

[...]

Issue #38, opened 2016-11-28
Run command should allow specifying ports that need to be published


lostintangent:
The `Docker: Run` command that this extension adds to VS Code doesn't provide a way to specify which (if any) ports within the container should be exposed to the host. This makes it hard to use this command for many containers, and therefore, requires you to drop down to using the terminal directly. This isn't a big deal, but since the `Docker: Add docker files to workspace` command asks which port to expose, it would seemingly make sense for the `Docker: Run` command to provide a similar experience, for publishing ports at runtime.

kr428:
+1. This is definitely missing so far.

chrisdias:
I ran into this too, but there are potentially many questions to ask.  You need to know which port to expose, which port you want it exposed on, and if there are multiple ports exposed in the image, then you have to do it multiple times.  You could argue the "add files to workspace" command does not support exposing multiple ports.

As  result, I didn't do this and figured the easiest thing to do as a user was use a `docker-compose.yml` file instead, which can do all of the port mappings for you.

That said, if there are good ideas on how to ask the questions, I'm happy to do the work.

MicahZoltu:
As a short-term iterative step, perhaps provide a mechanism for having a VSCode setting that allows you to specify the specifics of `Docker Run`?  This way each project can setup `Docker Run` to be appropriate for that project by using a `.vscode/settings.json` file, and the default can remain `docker run <image-name>`.

lostintangent:
@MicahZoltu That's a great idea. As another potential solution, we also just merged a PR that enhances this command, so that it detects the ports that the selected image exposes, and automatically publishes them on the host machine (by adding the appropriate "-p" args to "docker run").

This behavior solves the primary scenario I had in mind when I initially opened this issue, but I'd love to get your thoughts on this and discuss any further customizations we might need to add in to enable additional scenarios/requirements.

MicahZoltu:
Does it publish the port to a dynamic port (like `-p 80`) or does it publish to a matching port on the host (`-p 80:80`)?  Both have their caveats, but both likely address the issue for the most common case as well.

One advantage of allowing for a setting that defines how `docker run` behaves on a per-project basis is that it allows for more than just port binding.  For example, if your project needs to map environment variables (e.g., user secrets) over this could be done via project settings as well.  For example, `docker run -e MY_SECRET` would map `MY_SECRET` environment variable on the host to `MY_SECRET` environment variable in the container and this may be required for the container to function properly.

[...]

PR #1274, opened 2019-09-16
:bug: Don't remove container created by "Run Image" automatically

xgdgsc:
solves #1217

### DIFF
--- a/src/commands/images/runImage.ts
+++ b/src/commands/images/runImage.ts
@@ -23,7 +23,7 @@ async function runImageCore(context: IActionContext, node: ImageTreeItem | undef
     const inspectInfo = await node.getImage().inspect();
     const ports: string[] = inspectInfo.Config.ExposedPorts ? Object.keys(inspectInfo.Config.ExposedPorts) : [];
 
-    let options = `--rm ${interactive ? '-it' : '-d'}`;
+    let options = `${interactive ? '-it' : '-d'}`;
     if (ports.length) {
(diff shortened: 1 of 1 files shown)


bwateratmsft:
Seems like we should have a setting controlling this? @philliphoff what do you think?

philliphoff:
@bwateratmsft While I empathize with losing work within an unintentionally-exited container--that sucks!--given the intentionally ephemeral nature of containers, I'm not sure simply removing `--rm` from the command is the best option (even aside from the change to today's behavior).  That said, I *would* support an extension-level setting that allows the user to disable use of `--rm` switch for those commands.

bwateratmsft:
> I've always wanted to have settings to control args for all these commands. It would need to be done in a well-planned way.

@StephenWeatherford That's fair. I think then it would be related to #38. If we make the container run a _task_, the user can specify anything as options to the task (including --rm and anything else). However, tasks only make sense in the context of a workspace so it's not a perfect solution...

eine:
Related to #1001 and #1494 (#331), I think that 'Run' should open a parater selection menu  with multiple select and text boxes boxes:

- Select boxes:
  - **daemon**, **interactive**: mutually exclusive.
  - **rm**
- Text boxes:
  - **port**
  - **volume**
  - any

Defaults might be set by the caller, so that different preset options are shown depending on 'Run', 'Run Interactive' or if the source is a Task. If the user does not want to change anything, just hitting enter will work. Those who want to do one-time modifications can do so without modifying files or manually writting the command in the terminal.

bwateratmsft:
This will be resolved by #1596 / #1649

PR #1278, opened 2019-09-16
PR feedback from Eric

bwateratmsft:
Addressing Eric's feedback on #1245. We had decided to go ahead and submit #1245 and address the feedback separately, because the PR was getting too large and unwieldy to effectively do feedback with.


bwateratmsft:
From Eric in #1245: package.json: I feel like [dockerBuild.args] should be consistent with "shell" and "process" tasks, which use an array instead of an object for "args"

bwateratmsft:
From Eric in #1245: package.json: I feel like [dockerBuild.args] should be consistent with "shell" and "process" tasks, which use an array instead of an object for "args"

bwateratmsft:
From Eric in #1245: initializeForDebugging.ts: I feel like this should show an error if they select a platform that isn't supported yet. Or even better quickPickPlatform should be modified to only show supported platforms

bwateratmsft:
From Eric in #1245: configUtils.ts: Huh I've never seen a quick pick for a "Yes" or "No" question. I feel like it's usually an info dialog or something. Also what's the most common case here and is it worth a prompt? Could it just be a setting instead?

bwateratmsft:
From Eric in #1245: matchOnDescription and matchOnDetail seems excessive for options on a simple quick pick

bwateratmsft:
From Eric in #1245: DebugHelper.ts: This means initializeForDebugging will silently do nothing if the config already exists. It's actually pretty useful to overwrite debug config with the fresh/clean config if you've been messing with it and broke things.

Might be best to prompt them saying "Do you want to overwrite?" or at the very least just show a warning saying "Config already exists" so that they know they need to delete it in order to get new config

bwateratmsft:
From Eric in #1245: NodeDebugHelper.ts: I thought node2 was an old type that was going away? Does node work? If so, that seems safer

bwateratmsft:
From Eric, Steven in #1245: TaskHelper.ts: Per Steven's comment, I'd prefer to see unit tests for this kind of difficult-to-understand logic

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -96,7 +100,7 @@ export class DockerBuildTaskProvider implements TaskProvider {
             .create('docker', 'build', '--rm')
             .withFlagArg('--pull', options.pull)
             .withNamedArg('-f', options.dockerfile)
-            .withKeyValueArgs('--build-arg', options.args)
+            .withKeyValueArgs('--build-arg', options.buildArgs)
```
philliphoff: This change means users of today's debugging configuration cannot just copy/paste their existing configuration into a task; they'll need to modify the property name?  (Not a bad thing, necessarily, but could complicate the migration story.)

src/tasks/node/NodeTaskHelper.ts
```diff
@@ -164,14 +164,15 @@ export class NodeTaskHelper implements TaskHelper {
     }
 
     public static async inferPackageName(packagePath: string): Promise<string> {
-        const packageJson = await fse.readFile(packagePath, 'utf8');
-        const packageContent = <NodePackage>JSON.parse(packageJson);
+        const packageContent = await fse.readJSON(packagePath);
```
philliphoff: Use cast here to avoid unsafe-any lint warning below?

src/tasks/DockerBuildTaskProvider.ts
bwateratmsft: Yeah, that is one effect. I was confused by `args` myself, I thought it was a way to pass arbitrary extra arguments to `docker build`, not a way of controlling the `--build-args`.

Now is the only opportunity to make a change like this so I'm inclined to do it even though it will impact the subset of users who want to copypasta the config from launch to tasks.

src/tasks/node/NodeTaskHelper.ts
bwateratmsft: D'oh!

bwateratmsft:
> From Eric in #1245: DebugHelper.ts: This means initializeForDebugging will silently do nothing if the config already exists. It's actually pretty useful to overwrite debug config with the fresh/clean config if you've been messing with it and broke things.
> 
> Might be best to prompt them saying "Do you want to overwrite?" or at the very least just show a warning saying "Config already exists" so that they know they need to delete it in order to get new config

@EricJizbaMSFT I added an overwrite prompt. If we encounter any duplicates it will prompt once, overwriting if instructed to. Note, even if you select Overwrite = No, it will still fill in any missing objects.

bwateratmsft:
> From Eric, Steven in #1245: TaskHelper.ts: Per Steven's comment, I'd prefer to see unit tests for this kind of difficult-to-understand logic

@EricJizbaMSFT Added some UTs and did some type refactoring.

philliphoff:
> NodeDebugHelper.ts: I thought node2 was an old type that was going away? Does node work? If so, that seems safer

Yes, from what I understand, the `'node2'` type is an implementation detail of VS Code's Node.js debugging and used when debugging applications targeting v6.3+.  It's possible the Node.js debugging infrastructure is refactored in a way that removes this specific provider/adaptor.  Unfortunately, there appears to be an [issue](https://github.com/microsoft/vscode/issues/81118) in VS Code that prevents us from using the more generic `'node'` type.

philliphoff:
>Huh I've never seen a quick pick for a "Yes" or "No" question. I feel like it's usually an info dialog or something. Also what's the most common case here and is it worth a prompt? Could it just be a setting instead?

The intent is to keep the user in the same, wizard-y prompt flow rather than have a random modal dialog be displayed between palette prompts. I don't believe this should be a setting as such scaffolding is not done frequently enough to warrant it--invocations are likely performed once per each workspace/application/service and may each elicit a different response.

PR #1279, opened 2019-09-16
Update dependencies.

philliphoff:
Update to the latest version of extension dependencies (in accordance with any semver restrictions), now that we're post-0.8.0 release. (The test changes were due to the latest TypeScript compiler complaining about some generalized test code; using `<any>` there was the only way I could think of reasonable accommodating it.)


bwateratmsft:
> (The test changes were due to the latest TypeScript compiler complaining about some generalized test code; using `<any>` there was the only way I could think of reasonable accommodating it.)

I'd noticed that earlier when I was trying to build on my Linux machine without luck...ended up just locally deleting that test file because I didn't need it there...

PR #1282, opened 2019-09-17
[Engineering] Fix incredibly obnoxious Problems window bug

bwateratmsft:


nturinski:
I've tried changing this setting and it seems to have resolved the issue.  At least it can't hurt to change it to "never".

StephenWeatherford:
Yep, had to do this in our repo, too.

PR #1288, opened 2019-09-19
Ensure vsdbg script is bundled with extension

### LINKED ISSUE #1286: Pop up an error when debugging a ASP.NET MVC web app
**OS:** Win10
**Build Version:** [20190918.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=10136)
**.NET Core Version**: 2.2.401
**Regression:** Not a Regression

**Repro Steps:**
1. Create a new ASP.NET MVC web app by "dotnet new mvc --no-https --name test1".
2. Open this project in VS Code.
3. Select "Yes" when asked whether add the required assets to build and debug or not.
image
4. F1 --> Type and select "Docker: Add Docker Files to Workspace" --> ASP.NET Core -->Linux -->Port : 80,443.
5. Navigate to "File Explorer" -->"Test1" project --> "Controllers" folder -->Open "HomeController.cs" -->Set a breakpoint in the Index() method.
6. Go to the Debug tab --> Select the Docker launch profile --> Debug.
image

**Expect:**
Succeed to debug and the breakpoint can be hit.

**Actual:**
Pop up below error.
image
image

bwateratmsft:
Resolves #1286


bwateratmsft:
@EricJizbaMSFT The other way to do this is to make sure this `vsdbg`  script file is copied in `webpack.config.js`. The (minor) disadvantage of that approach is that there will need to be a little extra logic in the NetCoreDebugHelper to look in `src/debugging/netcore` when doing extension development (i.e. I F5'd the extension), and otherwise look in `dist`.

The (minor) disadvantage of this approach is that in the extension as installed, in addition to the prior `dist` and `resources` folders we now have `src/debugging/netcore`.

Any feelings one way or the other?

PR #1290, opened 2019-09-19
Honor default Docker context for WSL 2

### LINKED ISSUE #1199: Honor the Docker context when running in WSL 2
I believe this is related to how dockerode works; however, I'm pretty sure the host/socket can be overridden when instantiating a new Docker instance so it may not actually be a bug with dockerode.

On Windows 10 (18950) running WSL 2 with the Docker Desktop WSL 2 Technical Preview. The TP separates the Windows and WSL contexts to prevent destructive operations and Dockerode only accesses the Windows context. For example, on Windows, with WSL 2 enabled I see this (notice that wsl is active)

image

Running `$ docker ps` and `$ docker images` shows the containers and images running in WSL no matter when shell I access Docker from. With the extension in UI mode, I get the containers running within the default Windows context.

This may appear to be working as expected since UI shows Windows and Workspace shows WSL, but it's a **damn** confusing experience for existing extension users who enable WSL 2 and install the technical preview expecting to see containers running in WSL (the same as they see from PowerShell). If the context were honored the same way, the workspace/ui separation wouldn't be an issue for Windows WSL 2 users.

bwateratmsft:
Resolves #1199


src/utils/getDefaultDockerContext.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import * as cp from 'child_process';
+import { DockerOptions } from 'dockerode';
+import * as util from 'util';
+
+const exec = util.promisify(cp.exec);
+
+export async function getDefaultDockerContext(): Promise<DockerOptions | undefined> {
+    try {
+        const { stdout, stderr } = await exec('docker context inspect');
```
philliphoff: Doesn't this assume the Docker CLI is installed (which seems a little weird to do in order to configure Dockerode, whose purpose is not need the Docker CLI)?  At the very least, can we catch the case where it isn't installed.  (I know it's an unlikely scenario, but I'd rather not unnecessarily add to our dependencies on the Docker CLI.)

Also, should we be checking the exit code as well as stderr?

src/utils/getDefaultDockerContext.ts
```diff
@@ -0,0 +1,47 @@
...
+            dockerContext[0].Endpoints &&
+            dockerContext[0].Endpoints.docker &&
+            dockerContext[0].Endpoints.docker.Host;
+        // tslint:enable: no-unsafe-any
+
+        if (host.indexOf('unix://') === 0) {
+            return {
+                socketPath: host.substring(7), // Everything after the unix:// (expecting unix:///var/run/docker.sock)
+            }
+        } else if (host.indexOf('npipe://') === 0) {
```
philliphoff: Make these constants and use their length in the the `substring()` call?

src/utils/getDefaultDockerContext.ts
```diff
@@ -0,0 +1,47 @@
...
+        if (stderr) {
+            throw new Error(stderr);
+        }
+
+        if (!stdout) {
+            return undefined;
+        }
+
+        const dockerContext = JSON.parse(stdout);
+        // tslint:disable: no-unsafe-any
```
philliphoff: Define an interface that matches the shape of the returned command, to avoid the linter error as well as help document our expectations?

src/utils/getDefaultDockerContext.ts
bwateratmsft: I thought about it, sounds good.

src/utils/getDefaultDockerContext.ts
bwateratmsft: If it's not installed we'll fall through the catch and down to the old behavior.

bwateratmsft: I'm going to rename it though to `tryGetDefaultDockerContext` since it's strictly a best-effort thing.

bwateratmsft:
What is going on with the build? A few succeed and a whole lot more time out, always on Windows.

src/extension.ts
```diff
@@ -308,7 +309,7 @@ function refreshDockerode(): void {
         process.env = { ...process.env }; // make a clone before we change anything
         addDockerSettingsToEnv(process.env, oldEnv);
         ext.dockerodeInitError = undefined;
-        ext.dockerode = new Dockerode();
+        ext.dockerode = new Dockerode(process.env.DOCKER_HOST ? undefined : tryGetDefaultDockerContext());
```
bwateratmsft: `tryGetDefaultDockerContext` cannot be async. VSCode does not appear to await activation of the extension; as a result if `refreshDockerode` is async then it completes at a somewhat unpredictable time later. The consequence of this is that you get a flash of "cannot connect" since the dockerode object isn't initialized yet, before it starts working. Also, tests fail miserably because each of the tree tests results in ~4 setting changes, thus four calls to `refreshDockerode`; eventually `ext.dockerode` would get overwritten by one of them _during_ a test, causing it to fail.

src/utils/tryGetDefaultDockerContext.ts
```diff
@@ -0,0 +1,60 @@
...
+interface IDockerContext {
+    Endpoints: { [key: string]: IDockerEndpoint }
+}
+
+let cachedDockerContext: DockerOptions;
+let cachedDockerContextTimestamp: number;
+
+export function tryGetDefaultDockerContext(): DockerOptions {
+    // First check the cached value--we will keep result from last 10 seconds
+    if (cachedDockerContextTimestamp && cachedDockerContextTimestamp + 10000 > Date.now()) {
```
philliphoff: Any particular reason for holding onto the cached context for only 10s?

bwateratmsft: We can make it longer (or shorter), but @Chuxel used 10s in his corresponding code for the Remote extension.

Making it longer would primarily benefit tests. The tests do `refreshDockerode` a lot, and they do take longer than 10s total, so they usually end up calling this 2-3 times.

Making it shorter would be bad for tests but better for users, if they are quickly changing between different settings.

src/extension.ts
EricJizbaMSFT: FWIW the tests don't change settings that affect dockerode. Perhaps we should be more selective in terms of which setting changes cause a refresh

bwateratmsft: I wanted to but I'm not sure we can, without redefining the settings, which I'm not sure we can do. If the settings for the `DOCKER_*` environment variables were in their own subgroup, like `docker.environment.*`, we could be more specific; but since they're just `docker.*` we're kinda stuck.

bwateratmsft: As discussed offline, we can filter on specific settings with `e.affectsConfiguration`, so I did that. Still has to be sync though to avoid the problem with the UI quickly flashing "cannot connect" on first activation.

src/utils/tryGetDefaultDockerContext.ts
bwateratmsft: I changed the settings watcher to only call `refreshDockerode` on relevant setting changes (which the tests do not do), so caching no longer matters for tests. I removed caching entirely since it could have adverse effects when users are changing things in the settings page.

Issue #1283, opened 2019-09-18
Move Node.js Docker debugging to task-based model


philliphoff:
Update the Node.js Docker scaffolding/debugging to be based on the new task-based model (by default).

philliphoff:
Basic support was added with #1245

PR #1291, opened 2019-09-19
Some tweaks for Node.js debugging within Docker

philliphoff:
- Tweak wording when prompting to include Docker compose files
- Scaffold `NODE_ENV` environment variable for Docker run tasks (to smooth debugging with Sails apps)
- Better support scaffolding debug command lines when app uses `nodejs` links
- Better support scaffolding debug command lines when preceded by environment variables
- Some refactoring to make testing easier


philliphoff:
Expands on #1283.

src/utils/nodeUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export type InspectMode = 'default' | 'break';
+
+export async function readPackage(packagePath: string): Promise<NodePackage> {
+    return <NodePackage> await fse.readJson(packagePath);
+}
+
+export async function inferPackageName(nodePackage: NodePackage | undefined, packagePath: string): Promise<string> {
+    if (nodePackage && nodePackage.name !== undefined) {
```
bwateratmsft: nit: If there's an empty but not-undefined name this will return it. Is that expected? As an alternative,

return (nodePackage && nodePackage.name) || path.basename(path.dirname(packagePath));

src/utils/nodeUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+const StartScriptName: string = 'start';
+
+export async function inferCommand(nodePackage: NodePackage | undefined, inspectMode: InspectMode, inspectPort: number): Promise<string> {
+    const inspectArg = inspectMode === 'break' ? '--inspect-brk' : '--inspect';
+    const inspectArgWithPort = `${inspectArg}=0.0.0.0:${inspectPort}`;
+
+    if (nodePackage) {
```
bwateratmsft: if (nodePackage && nodePackage.scripts)

test/utils/nodeUtils.test.ts
```diff
@@ -0,0 +1,112 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { inferPackageName, inferCommand, InspectMode, NodePackage } from "../../extension.bundle";
+
+ suite('utils/nodeUtils', () => {
+    suite('inferPackageName', () => {
```
bwateratmsft: Maybe add a UT for the above case of a defined, empty package name?

src/tasks/node/NodeTaskHelper.ts
```diff
@@ -109,7 +101,19 @@ export class NodeTaskHelper implements TaskHelper {
         const runOptions = runDefinition.dockerRun;
 
         const packagePath = NodeTaskHelper.inferPackagePath(helperOptions && helperOptions.package, context.folder);
-        const packageName = await NodeTaskHelper.inferPackageName(packagePath);
+
+        let nodePackage: NodePackage;
+
+        const getNodePackage: () => Promise<NodePackage> =
```
bwateratmsft: Maybe we should move Lazy into utils, and create an additional AsyncLazy.

src/utils/nodeUtils.ts
philliphoff: Yeah, not sure why I insisted on `undefined` there.  Updated to account for an empty string.

test/utils/nodeUtils.test.ts
philliphoff: Added.

src/tasks/node/NodeTaskHelper.ts
philliphoff: Moved `Lazy<T>` to reuse here.  Didn't go so far as to create `AsyncLazy<T>`--a PR for another day.

PR #1314, opened 2019-09-30
Better showing of ENOENT error for commands

### LINKED ISSUE #868: Improve error message when docker has not been found- ENOENT //./pipe/docker_engine
Repro steps:

==> Steps to reproduce the issue

1.CTRL+Shift+P in VScode 
2.typing "azur"
3.choose : "Doker: Azure CLI"

Action: vscode-docker.container.start.azurecli
Error type: ENOENT
Error Message: connect ENOENT //./pipe/docker_engine

Version: 0.6.1
OS: win32
VS Code Version: 1.32.3

<details>
<summary>Call Stack</summary>

```
PipeConnectWrap.afterConnect [as oncomplete] (net.js:1161:14)
```

</details>

bwateratmsft:
Resolves #868

### DIFF
--- a/src/commands/registerCommands.ts
+++ b/src/commands/registerCommands.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { commands } from "vscode";
-import { IActionContext, parseError, registerCommand } from "vscode-azureextensionui";
+import { registerCommand } from "vscode-azureextensionui";
 import { configure, configureApi } from "../configureWorkspace/configure";
 import { composeDown, composeRestart, composeUp } from "./compose";
 import { attachShellContainer } from "./containers/attachShellContainer";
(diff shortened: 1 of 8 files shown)


src/commands/registerCommands.ts
```diff
@@ -57,66 +57,90 @@ import { pruneVolumes } from "./volumes/pruneVolumes";
...
+    registerInternal(false, 'vscode-docker.volumes.inspect', inspectVolume);
+    registerInternal(false, 'vscode-docker.volumes.prune', pruneVolumes);
+    registerInternal(false, 'vscode-docker.volumes.remove', removeVolume);
+}
+
+// tslint:disable-next-line: no-any
+function registerInternal<T>(workspace: boolean, commandId: string, callback: (context: IActionContext, ...args: any[]) => T | PromiseLike<T>, debounce?: number): void {
+    let registerFunction = workspace ? registerWorkspaceCommand : registerCommand;
+
+    registerFunction(
```
bwateratmsft: @EricJizbaMSFT do you think this is an OK approach? Or might this mess up the registerCommand API?

EricJizbaMSFT: Can you refactor so that your new error is only shown when `ext.dockerode` is used? ENOENT is a pretty generic error type that can happen for a lot of reasons - also this new message would never make sense for registry commands or things like "Add Docker Files to Workspace"

bwateratmsft: Is there an easier way to do that than just wrapping all the touches to ext.dockerode?

EricJizbaMSFT: Well I don't think you want to wrap _all_ touches because the tree will already display the better message:
Screen Shot 2019-09-30 at 9 07 42 AM

Looks like prune and anything using `osUtils.getDockerOSType` might be the only cases to handle

bwateratmsft: Sounds good. I'll make it so.

PR #1326, opened 2019-10-07
Use ctx.subscriptions to manage disposables

bwateratmsft:
The whole of `deactivateInternal` can be refactored to be disposables pushed onto `ctx.subscriptions`. This is assuming that calling `dispose()` on the Disposable from `start()` is equivalent to, or at least also calls, `stop()`.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -185,11 +185,6 @@ async function setRequestDefaults(): Promise<void> {
     ext.request = requestWithDefaults;
 }
 
-export async function deactivateInternal(): Promise<void> {
-    // No-op
-    return;
-}
-
 namespace Configuration {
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -186,17 +186,11 @@ async function setRequestDefaults(): Promise<void> {
...
 
 export async function deactivateInternal(): Promise<void> {
-    if (!client) {
-        return undefined;
-    }
-    // perform cleanup
-    Configuration.dispose();
-    return await client.stop();
+    // No-op
+    return;
```
philliphoff: Can we just get rid of `deactivateInternal()`

bwateratmsft: Yeah, I don't have a strong opinion. We do have to keep `deactivate` inside main.js though.

PR #1327, opened 2019-10-07
Allow non-platform-specific tasks

### LINKED ISSUE #1321: Allow non-platform-specific build/run tasks
Allow the creation of Docker build/run tasks not tied to a given platform (e.g. Node or .NET Core) in order to allow tasks to be used for arbitrary configurations of images/containers.

For example, I may want to (repeatedly) create a MongoDB container with a specific set of environment variables, attached to a specific network, etc..  While this can be done via a shell script or Docker Compose YAML, it may not be as convenient as being able to execute a Docker run task in VS Code, or integrate as well into a sequence of tasks used for various purposes.

bwateratmsft:
Resolves #1321

### DIFF
--- a/src/tasks/DockerRunTaskProvider.ts
+++ b/src/tasks/DockerRunTaskProvider.ts
@@ -64,7 +64,10 @@ export class DockerRunTaskProvider implements TaskProvider {
 
         const helper = this.getHelper(context.platform);
 
-        definition.dockerRun = await helper.resolveDockerRunOptions(context, definition);
+        if (helper) {
+            definition.dockerRun = await helper.resolveDockerRunOptions(context, definition);
+        }
+
         await this.validateResolvedDefinition(context, definition.dockerRun);
 
         const commandLine = await this.resolveCommandLine(definition.dockerRun);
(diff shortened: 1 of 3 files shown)


src/tasks/DockerRunTaskProvider.ts
```diff
@@ -82,10 +85,6 @@ export class DockerRunTaskProvider implements TaskProvider {
         if (!dockerRun.image) {
             throw new Error('No Docker image name was provided or resolved.');
         }
-
-        if (!dockerRun.containerName) {
```
bwateratmsft: I could go either way on this. Upside, without this check users don't _have_ to specify container name. Downside, none of our debugging works without container name, so this serves as extra validation for that.

bwateratmsft: @philliphoff In your mongo example, do you have to provide container name anyway?

philliphoff: No, in the MongoDB example it was only necessary to set the network alias, not the name of the container.  This is potentially an advantage of having a "debugging" flag, where we can apply different validations in that case.

PR #1340, opened 2019-10-11
Don't fail on warning for insecure creds

### LINKED ISSUE #1296: Unable to pull image from registry
Repro steps:
Try to pull an image from docker registry from the command pallete or docker extension tab

```
Action: vscode-docker.registries.pullImage
Error type: string
Error Message: WARNING! Your password will be stored unencrypted in /home/anton/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store
```

The full output
```
docker login  --username nakamin --password-stdin
Login Succeeded
WARNING! Your password will be stored unencrypted in /home/anton/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Error: WARNING! Your password will be stored unencrypted in /home/anton/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store
```

Version: 0.8.1
OS: linux
Product: Visual Studio Code
Product Version: 1.38.1
Language: en

bwateratmsft:
Resolves #1296

### DIFF
--- a/src/commands/registries/logInToDockerCli.ts
+++ b/src/commands/registries/logInToDockerCli.ts
@@ -42,8 +42,6 @@ export async function logInToDockerCli(context: IActionContext, node?: RegistryT
                     reject(new Error(`In order to log in to the Docker CLI using tokens, you currently need to go to \nOpen your Docker config file and remove "credsStore": "wincred" from the config.json file, then try again. \nDoing this will disable wincred and cause Docker to store credentials directly in the .docker/config.json file. All registries that are currently logged in will be effectly logged out.`));
                 } else if (err) {
                     reject(err);
-                } else if (stderr) {
-                    reject(stderr);
                 } else {
                     resolve();
                 }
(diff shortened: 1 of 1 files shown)


src/commands/registries/logInToDockerCli.ts
```diff
@@ -42,8 +42,6 @@ export async function logInToDockerCli(context: IActionContext, node?: RegistryT
                     reject(new Error(`In order to log in to the Docker CLI using tokens, you currently need to go to \nOpen your Docker config file and remove "credsStore": "wincred" from the config.json file, then try again. \nDoing this will disable wincred and cause Docker to store credentials directly in the .docker/config.json file. All registries that are currently logged in will be effectly logged out.`));
                 } else if (err) {
                     reject(err);
-                } else if (stderr) {
```
philliphoff: Maybe include a comment about Docker writing ignorable warnings to stderr in this case?  Is there any other reason why Docker might write to stderr (that we know about) that we would *not* want to ignore?

bwateratmsft: None that I know of. For `docker build`, BuildKit pukes everything (not just errors) out to stderr, when I looked at telemetry of `docker build` / `docker run` there were no examples of a real failure but exit code == 0. If `docker login` is anything like `build` or `run`, we can expect real errors would always result in nonzero exit code.

PR #1346, opened 2019-10-15
Improved Dockerfile pattern matches, save event

bwateratmsft:


src/registerListeners.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+function onDidSaveTextDocument(doc: TextDocument): void {
+    // If it's not a Dockerfile, or last upload time is within an hour, skip
+    if (doc.languageId !== 'dockerfile' || lastUploadTime + hourInMilliseconds > Date.now()) {
+        return;
+    }
+
+    lastUploadTime = Date.now();
+    ext.reporter.sendTelemetryEvent('dockerfilesave', { "lineCount": doc.lineCount.toString() }, {});
```
rcjsuen: I guess the Docker extension team is curious as to how big people's Dockerfiles are? 🤔

bwateratmsft: Mostly, we want to see how much people are using the extension just to get the Dockerfile intellisense; because it seems to be the majority.

bwateratmsft: For example, in a month we had ~300K machines activate the extension, but ~60K that did anything else. There's more ways to trigger activation than just opening a Dockerfile but that is probably the most common one...

rcjsuen: Interesting. I'd be curious as to how useful people actually find the editor features and what they use the most of. I imagine most people don't actually **need** code completion given that Dockerfiles aren't complicated beasts.

src/registerListeners.ts
```diff
@@ -0,0 +1,24 @@
...
+    ctx.subscriptions.push(workspace.onDidSaveTextDocument(onDidSaveTextDocument));
+}
+
+function onDidSaveTextDocument(doc: TextDocument): void {
+    // If it's not a Dockerfile, or last upload time is within an hour, skip
+    if (doc.languageId !== 'dockerfile' || lastUploadTime + hourInMilliseconds > Date.now()) {
+        return;
+    }
+
+    lastUploadTime = Date.now();
```
philliphoff: Nit: capture `Date.now()` and use the same value in both places.

bwateratmsft: I wonder if an allocation wouldn't be more expensive?

PR #1347, opened 2019-10-16
Remove some unneeded direct dependencies

bwateratmsft:
None of these are direct dependencies and so can be removed. All three are still dependencies of dependencies, though...


bwateratmsft:
@EricJizbaMSFT I used [depcheck](https://www.npmjs.com/package/depcheck) to find these; it also had some other suggestions I knew were mistaken (e.g. `dockerfile-language-server-nodejs`). Nevertheless it does build with this set removed. The only one that I was less sure about was `azure-arm-resource`, but it builds fine without it.

bwateratmsft:
> @EricJizbaMSFT I used [depcheck](https://www.npmjs.com/package/depcheck) to find these; it also had some other suggestions I knew were mistaken (e.g. `dockerfile-language-server-nodejs`). Nevertheless it does build with this set removed. The only one that I was less sure about was `azure-arm-resource`, but it builds fine without it.

Incidentally, `azure-arm-resource` comes anyway since it's a dependency of `vscode-azureappservice`.

EricJizbaMSFT:
Looks like the docker repo already has the no-implicit-dependencies tslint rule turned on and I would trust that

https://palantir.github.io/tslint/rules/no-implicit-dependencies/

https://github.com/microsoft/vscode-docker/blob/master/tslint.json#L115

Issue #1322, opened 2019-10-04
Allow simple network creation in Docker explorer


philliphoff:
The Docker explorer allows the user to prune/remove networks that are no longer needed. It would be great if you could create a simple one through the explorer as well.

PR #1350, opened 2019-10-17
Add a command to create the simplest network

stuartthomson:
This was requested in https://github.com/microsoft/vscode-docker/issues/1322.

I could not find an appropriate icon for this action in the resources directory so I'm not sure that it can be merged until one is added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -239,7 +239,7 @@
                 {
                     "command": "vscode-docker.images.prune",
                     "when": "view == dockerImages",
-                    "group": "navigation@3"
+                    "group": "navigation@2"
                 },
                 {
                     "command": "vscode-docker.images.configureExplorer",
@@ -1708,8 +1708,8 @@
                 "title": "Create...",
                 "category": "Docker Networks",
                 "icon": {
-                    "light": "resources/light/connect.svg",
-                    "dark": "resources/dark/connect.svg"
+                    "light": "resources/light/add.svg",
+                    "dark": "resources/dark/add.svg"
                 }
             },
             {
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
         "onCommand:vscode-docker.images.runInteractive",
         "onCommand:vscode-docker.images.tag",
         "onCommand:vscode-docker.networks.configureExplorer",
+        "onCommand:vscode-docker.networks.create",
         "onCommand:vscode-docker.networks.inspect",
         "onCommand:vscode-docker.networks.prune",
         "onCommand:vscode-docker.networks.refresh",
@@ -221,9 +222,14 @@
                     "group": "navigation@9"
                 },
                 {
-                    "command": "vscode-docker.networks.prune",
+                    "command": "vscode-docker.networks.create",
                     "when": "view == dockerNetworks",
                     "group": "navigation@1"
+                },
+                {
+                    "command": "vscode-docker.networks.prune",
+                    "when": "view == dockerNetworks",
+                    "group": "navigation@2"
                 },
                 {
                     "command": "vscode-docker.networks.refresh",
@@ -233,7 +239,7 @@
                 {
                     "command": "vscode-docker.images.prune",
                     "when": "view == dockerImages",
-                    "group": "navigation@2"
+                    "group": "navigation@3"
                 },
                 {
                     "command": "vscode-docker.images.configureExplorer",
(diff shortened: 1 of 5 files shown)


src/commands/networks/createNetwork.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ext } from '../../extensionVariables';
+import { wrapDockerodeENOENT } from '../../utils/wrapError';
+
+export async function createNetwork(_context: IActionContext): Promise<void> {
+
+    const opts: InputBoxOptions = {
+        value: '',
+        prompt: 'Name of the network'
+    };
+    const name = await ext.ui.showInputBox(opts)
```
bwateratmsft: I think we should also prompt for [type](https://docs.docker.com/network/#network-drivers), i.e. host, bridge, etc. I'm not totally familiar with their different use cases, so we may not need to display all possible options.

package.json
```diff
@@ -233,7 +239,7 @@
                 {
                     "command": "vscode-docker.images.prune",
                     "when": "view == dockerImages",
-                    "group": "navigation@2"
+                    "group": "navigation@3"
```
bwateratmsft: This doesn't need to change does it? It's part of the Images view.

package.json
```diff
@@ -1697,6 +1703,15 @@
                     "dark": "resources/dark/settings.svg"
                 }
             },
+            {
+                "command": "vscode-docker.networks.create",
+                "title": "Create...",
+                "category": "Docker Networks",
+                "icon": {
+                    "light": "resources/light/connect.svg",
```
bwateratmsft: For icons I think we should use the network icon with a + superimposed, sort of like what is used for files:
image
I'm not entirely sure how we go about getting such a thing created though... @EricJizbaMSFT do you know?

src/commands/networks/createNetwork.ts
bwateratmsft: @philliphoff What do you think? Should we prompt at all?

stuartthomson: I think a prompt for the driver is probably a good idea, I'd be find to add it if people like the idea.

package.json
stuartthomson: That is true, not sure how that happened.

package.json
stuartthomson: I'm happy to make one like that if there is no more official process.

bwateratmsft: @misolori I saw you created most of the current icons, are you able to make one like this? The + sign superimposed over the network icon?

EricJizbaMSFT: Check here for icons, but not sure the one you're describing exists: https://github.com/microsoft/vscode-icons

I feel like you could just use the "plus" icon (which definitely exists in that repo) since "network" is implied and unlike the file view in the above screenshot, there's only one thing you can create

bwateratmsft: Yup, that's reasonable as well.

misolori: +1 on using the plus instead, otherwise it's a bit of a complicated icon:

image

src/commands/networks/createNetwork.ts
philliphoff: In all of my cases, the default settings were sufficient (i.e. `docker network create <name>`) so I'm fine starting with that unless/until we have feedback where more control is desired.  Adding an option now to specify the driver would be totally fine, however.

philliphoff: Should we do any validation of the entered name or just rely on the presumed error returned by Docker to inform the user?

stuartthomson: I'm not sure what limitations there are on the name, so wouldn't know what to check for. For example, multiple networks can have the same name. Incidentally, this is why I chose to display network id when creation succeeds.

stuartthomson: I've added a prompt for driver type in another commit.

src/commands/networks/createNetwork.ts
```diff
@@ -0,0 +1,34 @@
...
+        ],
+        {
+            canPickMany: false,
+            placeHolder: 'Select the network driver to use (default is "bridge").'
+        }
+    );
+
+    const result = <{ id: string }>await wrapDockerodeENOENT(() => ext.dockerode.createNetwork({ Name: name, Driver: driverSelection.label }));
+
+    window.showInformationMessage(`Network Created with ID ${result.id}`);
```
bwateratmsft: Is ID really long and should potentially be truncated?

stuartthomson: Yeah the ID is quite long. Docker truncates the Network ID to 12 characters when you run `docker network ls`, so I'll truncate it to 12 characters too.

src/commands/networks/createNetwork.ts
philliphoff: Yeah, I couldn't find any naming rules for networks either.  Should be fine for now.

bwateratmsft:
Resolves #1322 .

Issue #1300, opened 2019-09-24
Better prerequisite checks/errors


bwateratmsft:
Add prerequisite checks during tasks / debug.

Some possible ideas (not an exhaustive list):

- [x] .NET Core: build tasks are present
- [ ] All: container OS matches what is set in daemon (Windows only)
- [ ] All: Minimum Docker version installed
- [ ] All: Docker running
- [ ] Linux: user added to Docker users group
- [x] Python: extension installed
- [ ] .NET Core: project is structured in an understandable way (e.g. csproj exists, etc.)
- [ ] Node: project is structured in an understandable way (e.g. package.json exists, etc.)

~~.NET Core: drive sharing enabled~~ WSL2 makes this mostly unnecessary

bwateratmsft:
Blocked by https://github.com/microsoft/vscode/issues/80375

bwateratmsft:
Delaying this one to 0.10.0

philliphoff:
Let's use 0.10.0 to better understand what failures are occurring, in order to better understand what, specific, checks are needed.

ravipal:
The current behavior is 1.0, is F5 pop-ups this error
![image](https://user-images.githubusercontent.com/1775825/77207578-fd086780-6ab6-11ea-8389-55687fdaf06f.png)

bwateratmsft:
@ravipal could you add more context to that screenshot?

ravipal:
When user runs 'Add Docker Files to Workspace' without adding the Net Core build and launch tasks. Then F5, user will run into an error (the above screen shot)

ravipal:
Now the command 'Docker: Add Docker Files to Workspace' will check if the build task is present and fail if it is not present.

bwateratmsft:
I don't think this is resolved yet as we want to cover a lot more than just .NET Core build tasks.

ravipal:
Please add more information on what other issues that needs to be addressed.

bwateratmsft:
I've added some ideas to the original opening comment.

bwateratmsft:
I'm going to close this, it doesn't seem to be necessary--not really any volume of user reports that indicate confusion.

Issue #1275, opened 2019-09-16
Blazor apps using components cannot be debugged


bwateratmsft:
Because Blazor writes a manifest file based on the _host_ filesystem for where to find components (i.e. from NuGet packages and other projects in the solution), Blazor apps cannot be debugged, as our current model does build on the host and volume maps that into the container.

bwateratmsft:
Blocked by https://github.com/microsoft/vscode/issues/81007 (which is duped to https://github.com/microsoft/vscode/issues/2809)

PR #1360, opened 2019-10-22
Use CustomExecution instead of ShellExecution

### LINKED ISSUE #1388: Use child_process.spawn instead of child_process.exec
Docker Build operations can have very long output, especially in the case of Buildkit (which emits nearly everything to stderr).

According to [this](https://www.hacksparrow.com/nodejs/difference-between-spawn-and-exec-of-node-js-child-rocess.html), `child_process.exec()` uses 200KB buffers, whereas `child_process.spawn()` uses streams. I've seen some errors in telemetry of these buffers getting filled up and consequently, the exec fails. IMO this is highly likely to be Buildkit.

We should change our usage to `child_process.spawn`, especially for `docker build` commands which are likely to have very long output.

### LINKED ISSUE #1392: Add default VS labels to all task-built Docker images
Currently, the "old" .NET Core debugging logic as well as the new NET Core Docker build task will add the `com.microsoft.created-by` label, if no other labels are specified.  We should also add this label (by default) to all images created via tasks (.NET Core, Node.js, and otherwise).

bwateratmsft:
This PR is to use CustomExecution instead of ShellExecution. It will enable us to do a lot, like #1275, #1300, and more. Those other changes will be in later PRs.

Resolves #1388, resolves #1392


package.json
```diff
@@ -1952,8 +1952,9 @@
         }
     },
     "engines": {
-        "vscode": "^1.37.0"
+        "vscode": "^1.40.0"
     },
+    "enableProposedApi": true,
```
bwateratmsft: Remove when we can.

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -25,59 +25,37 @@ export interface DockerBuildTask extends Task {
...
-        // TODO : addDockerSettingsToEnv?
-        return new Task(
-            task.definition,
-            task.scope,
-            task.name,
-            task.source,
-            new ShellExecution(commandLine[0], commandLine.slice(1)),
-            task.problemMatchers);
+
+        // TODO: process errors from docker build so that warnings aren't fatal
```
bwateratmsft: Do?

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -25,59 +25,37 @@ export interface DockerBuildTask extends Task {
...
+        await context.shell.executeCommandInTerminal(commandLine, context.folder, /* rejectOnStdError: */ false, context.cancellationToken);
+        throwIfCancellationRequested(context);
+
+        if (helper && helper.postBuild) {
+            // TODO: attach results to context
```
bwateratmsft: Do.

src/tasks/DockerPseudoShell.ts
```diff
@@ -0,0 +1,106 @@
...
+                    cancelHandler.dispose();
+                });
+            }
+
+            process.stderr.on('data', (chunk: Buffer) => this.writeError(chunk.toString()));
+            process.stdout.on('data', (chunk: Buffer) => this.writeOutput(chunk.toString()));
+        });
+    }
+
+    public writeOutput(message: string): void {
```
bwateratmsft: Is there a more syntactically-sugary way to do all these methods?

src/tasks/DockerRunTaskProvider.ts
```diff
@@ -23,62 +23,42 @@ export interface DockerRunTask extends Task {
...
-        return new Task(
-            task.definition,
-            task.scope,
-            task.name,
-            task.source,
-            new ShellExecution(commandLine[0], commandLine.slice(1)),
-            task.problemMatchers);
+
+        // TODO: get container ID from output?
+        // TODO: process errors from docker run so that warnings aren't fatal
```
bwateratmsft: Do?

src/tasks/DockerRunTaskProvider.ts
```diff
@@ -23,62 +23,42 @@ export interface DockerRunTask extends Task {
...
+        await context.shell.executeCommandInTerminal(commandLine, context.folder, /* rejectOnStdError: */ false, context.cancellationToken);
+        throwIfCancellationRequested(context);
+
+        if (helper && helper.preRun) {
+            // TODO: attach results to context
```
bwateratmsft: Do.

src/tasks/DockerTaskProviderBase.ts
```diff
@@ -0,0 +1,63 @@
...
+import { DockerPseudoShell } from './DockerPseudoShell';
+import { DockerRunTask } from './DockerRunTaskProvider';
+import { DockerTaskExecutionContext, DockerTaskProviderName, TaskHelper } from './TaskHelper';
+
+export abstract class DockerTaskProviderBase implements TaskProvider {
+
+    protected constructor(private readonly telemetryName: DockerTaskProviderName, protected readonly helpers: { [key in DockerPlatform]: TaskHelper }) { }
+
+    public provideTasks(token?: CancellationToken): ProviderResult<Task[]> {
+        return []; // Intentionally empty, so that resolveTask gets used
```
bwateratmsft: Can we actually use this now?

src/tasks/TaskHelper.ts
```diff
@@ -107,6 +129,9 @@ export async function getAssociatedDockerBuildTask(runTask: DockerRunTask): Prom
 }
 
 export async function getOfficialBuildTaskForDockerfile(dockerfile: string, folder: WorkspaceFolder): Promise<Task | undefined> {
+    // TODO
```
bwateratmsft: Need to fix this method. We no longer resolve the Dockerfile as part of `fetchTasks` so a different approach is needed to find the "official" `docker-build` task.

src/utils/resolveVariables.ts
```diff
@@ -0,0 +1,68 @@
...
+            result[key] = resolveVariables(target[key], folder);
+        }
+
+        return result;
+    }
+}
+
+function resolveSingleVariable(variable: string, folder: WorkspaceFolder): string {
+    // tslint:disable: no-invalid-template-strings
+    switch (variable) {
```
bwateratmsft: More variables we want?

src/vscode.proposed.d.ts
```diff
@@ -0,0 +1,1035 @@
+/*---------------------------------------------------------------------------------------------
```
bwateratmsft: Remove this file when we can.

.azure-pipelines/common/test.yml
```diff
@@ -18,6 +18,7 @@ steps:
     SERVICE_PRINCIPAL_DOMAIN: $(SERVICE_PRINCIPAL_DOMAIN)
     DISPLAY: :10 # Only necessary for linux tests
     DOCKER_UNAVAILABLE: ${{ parameters.dockerUnavailable }}
+    CODE_VERSION: insiders # TODO : remove this line
```
bwateratmsft: Do.

rcjsuen: > Or do not. There is no try.

bwateratmsft: image

src/configureWorkspace/configure.ts
```diff
@@ -385,6 +386,7 @@ async function configureCore(context: IActionContext, options: ConfigureApiOptio
         os = await quickPickOS();
     }
     properties.configureOs = os;
+    properties.orchestration = 'single' as DockerOrchestration;
```
philliphoff: Is `as DockerOrchestration` needed here and below?

bwateratmsft: Strictly speaking no, but I can't do `properties.orchestration: DockerOrchestration = 'single'` and I didn't want it as typeless as a simple string...

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -25,59 +25,38 @@ export interface DockerBuildTask extends Task {
...
-            definition.dockerBuild = await helper.resolveDockerBuildOptions(context, definition);
+        if (helper && helper.preBuild) {
+            await helper.preBuild(context, definition);
+            throwIfCancellationRequested(context);
         }
 
-        await this.validateResolvedDefinition(context, definition.dockerBuild);
+        if (helper) {
+            definition.dockerBuild = await helper.getDockerBuildOptions(context, definition);
+            await this.validateResolvedDefinition(context, definition.dockerBuild);
```
philliphoff: Is there benefit in validating the definition even when there's no helper?  I.e. would we catch any mis-configuration by not specific to a platform?

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -25,59 +25,38 @@ export interface DockerBuildTask extends Task {
...
-            task.name,
-            task.source,
-            new ShellExecution(commandLine[0], commandLine.slice(1)),
-            task.problemMatchers);
+
+        // Because BuildKit outputs everything to stderr, we will not treat output there as a failure
+        await context.shell.executeCommandInTerminal(commandLine, context.folder, /* rejectOnStdError: */ false, context.cancellationToken);
+        throwIfCancellationRequested(context);
+
+        context.imageName = definition.dockerBuild.tag;
```
philliphoff: This is intended to set the default `imageName`?  Might `imageName` already be set at this point (or maybe could be in the future)?  Maybe only set if not already set?

src/tasks/DockerPseudoShell.ts
```diff
@@ -0,0 +1,109 @@
...
+import { DockerRunTask } from './DockerRunTaskProvider';
+import { DockerTaskProviderBase } from './DockerTaskProviderBase';
+import { DockerTaskExecutionContext } from './TaskHelper';
+
+const DEFAULT = '0m';
+const DEFAULTBOLD = '0;1m';
+const RED = '31m';
+const YELLOW = '33m';
+
+export class DockerPseudoShell implements Pseudoterminal {
```
philliphoff: Nit: maybe make naming consistent with `DockerPseudoterminal`?

src/tasks/DockerPseudoShell.ts
```diff
@@ -0,0 +1,109 @@
...
+    private readonly closeEmitter: EventEmitter<number> = new EventEmitter<number>();
+    private readonly writeEmitter: EventEmitter<string> = new EventEmitter<string>();
+    private readonly cts: CancellationTokenSource = new CancellationTokenSource();
+
+    public onDidWrite: Event<string> = this.writeEmitter.event;
+    public onDidClose: Event<number> = this.closeEmitter.event;
+
+    constructor(private readonly taskProvider: DockerTaskProviderBase, private readonly task: DockerBuildTask | DockerRunTask) { }
```
philliphoff: Nit: if other types refer to the base class `DockerTaskProviderBase`, it shouldn't end in the `Base` suffix.  Maybe just use `DockerTaskProvider`?

src/tasks/DockerPseudoShell.ts
```diff
@@ -0,0 +1,109 @@
...
+const DEFAULTBOLD = '0;1m';
+const RED = '31m';
+const YELLOW = '33m';
+
+export class DockerPseudoShell implements Pseudoterminal {
+    private readonly closeEmitter: EventEmitter<number> = new EventEmitter<number>();
+    private readonly writeEmitter: EventEmitter<string> = new EventEmitter<string>();
+    private readonly cts: CancellationTokenSource = new CancellationTokenSource();
+
+    public onDidWrite: Event<string> = this.writeEmitter.event;
```
philliphoff: `readonly`?  Or can expose as a property?

src/tasks/DockerPseudoShell.ts
```diff
@@ -0,0 +1,109 @@
...
+    public open(initialDimensions: TerminalDimensions | undefined): void {
+        const executeContext: DockerTaskExecutionContext = {
+            folder: this.task.scope as WorkspaceFolder,
+            cancellationToken: this.cts.token,
+            shell: this,
+        }
+
+        // Can't wait here
+        // tslint:disable-next-line: no-floating-promises
+        this.taskProvider.executeTask(executeContext, this.task).then(result => this.close(result));
```
philliphoff: Do we need to call `close()` in the error case (as the lambda will only be called on success)?

src/tasks/DockerPseudoShell.ts
```diff
@@ -0,0 +1,109 @@
...
+        const commandLine = resolveVariables(command.build(), folder);
+
+        // Output what we're doing, same style as VSCode does for ShellExecution/ProcessExecution
+        this.write(`> Executing task: ${commandLine} <\r\n\r\n`, DEFAULTBOLD);
+
+        // TODO: Maybe support remote Docker hosts and do addDockerSettingsToEnvironment?
+        return new Promise<{ stdout: string, stderr: string }>((resolve, reject) => {
+            const process = cp.exec(commandLine, (error, stdout, stderr) => {
+                if (error) {
+                    reject(error);
```
philliphoff: `return reject(error);`

src/tasks/DockerPseudoShell.ts
```diff
@@ -0,0 +1,109 @@
...
+                }
+
+                if (stderr && rejectOnStdError) {
+                    reject(stderr);
```
philliphoff: `return reject(stderr);`

src/tasks/DockerRunTaskProvider.ts
```diff
@@ -23,62 +23,41 @@ export interface DockerRunTask extends Task {
...
-            definition.dockerRun = await helper.resolveDockerRunOptions(context, definition);
+        if (helper && helper.preRun) {
+            await helper.preRun(context, definition);
+            throwIfCancellationRequested(context);
         }
 
-        await this.validateResolvedDefinition(context, definition.dockerRun);
+        if (helper) {
+            definition.dockerRun = await helper.getDockerRunOptions(context, definition);
+            await this.validateResolvedDefinition(context, definition.dockerRun);
```
philliphoff: As above, is there use if validating the definition regardless of whether there's a helper?

src/tasks/DockerTaskProviderBase.ts
philliphoff: I say keep it empty for now.

src/utils/resolveVariables.ts
philliphoff: Fine for now, I think.

src/tasks/DockerBuildTaskProvider.ts
bwateratmsft: I  meant to keep validation either way so this was an accident. I'll make this change.

src/tasks/DockerBuildTaskProvider.ts
bwateratmsft: The goal was to attach the image name to `context` before it gets passed to the post-build tasks. I suppose it's redundant with the config; we could just pass that.

The analogous is to attach the container ID to the context before it gets passed to the post-run tasks. Container ID is not redundant with config though since it isn't known until the container is started.

src/tasks/DockerPseudoShell.ts
bwateratmsft: Yeah, can do that. The reason I named it `DockerPseudoShell` is that we're using it more like a simple shell than a full-blown `Pseudoterminal`.

src/tasks/DockerRunTaskProvider.ts
bwateratmsft: (Same as above, made it this way by mistake--need to change to what you're describing)

src/tasks/DockerPseudoShell.ts
bwateratmsft: Yes and no. `DockerTaskProvider.executeTask()` cannot throw as all exceptions are caught and parsed, and the method returns some nonzero integer (error code when applicable, otherwise 1). It wouldn't be harmful to have additional catching here, but it's probably redundant. I'll leave a comment here to note that it assumes that `executeTask()` won't throw.

src/tasks/DockerTaskProvider.ts
```diff
@@ -0,0 +1,64 @@
...
+            task.name,
+            task.source,
+            new CustomExecution(() => Promise.resolve(new DockerPseudoterminal(this, task))),
+            task.problemMatchers
+        );
+    }
+
+    public async executeTask(context: DockerTaskExecutionContext, task: DockerBuildTask | DockerRunTask): Promise<number> {
+        try {
+            await callWithTelemetryAndErrorHandling(`${this.telemetryName}-execute`, async (actionContext: IActionContext) => {
```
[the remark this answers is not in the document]
bwateratmsft: I fixed it by adding `actionContext.errorHandling.suppressDisplay = true;`. @philliphoff Do you think there's any problem with losing the Report Issue button on the all-too-common `docker build` and `docker run` failures?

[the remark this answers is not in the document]
philliphoff: I could go either way.  As long as we have telemetry logging the frequency (and some sort of classification?) of the errors, I'm ok with losing the Report Issue button, as users tend not to give us any actionable context surrounding the issue.

src/tasks/DockerPseudoterminal.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    public async executeCommandInTerminal(command: CommandLineBuilder, folder: WorkspaceFolder, rejectOnStdError?: boolean, token?: CancellationToken): Promise<{ stdout: string, stderr: string }> {
+        const commandLine = resolveVariables(command.build(), folder);
+
+        // Output what we're doing, same style as VSCode does for ShellExecution/ProcessExecution
+        this.write(`> ${commandLine} <\r\n\r\n`, DEFAULTBOLD);
+
+        // TODO: Maybe support remote Docker hosts and do addDockerSettingsToEnvironment?
+        return new Promise<{ stdout: string, stderr: string }>((resolve, reject) => {
+            const process = cp.exec(commandLine, (error, stdout, stderr) => {
```
bwateratmsft: Per #1388 we should use `spawn` instead of `exec`.

src/utils/spawnAsync.ts
```diff
@@ -0,0 +1,105 @@
...
+    onStderr?: Progress,
+    stderrBuffer?: Buffer,
+    token?: CancellationToken): Promise<void> {
+
+    return await new Promise((resolve, reject) => {
+        let cancellationListener: Disposable;
+        let stdoutBytesWritten: number = 0;
+        let stderrBytesWritten: number = 0;
+
+        // Without the shell option, it pukes on arguments
```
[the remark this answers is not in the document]
bwateratmsft: Upon further research this is essentially what `child_process.exec` does--
> Spawns a shell then executes the `command` within that shell, buffering any generated output.

https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

Given that this has worked all along it is probably OK.

PR #1364, opened 2019-10-23
Ask some active users to take NPS survey

bwateratmsft:
After ten sessions on different days (i.e. multiple sessions in one day count as just one), choose 15% of users to prompt to take the NPS survey for the Docker extension. Prompt has Take Survey, Remind Me Later (will remind after 3 more sessions), and Don't Show Again.

image

### DIFF
--- a/src/utils/nps.ts
+++ b/src/utils/nps.ts
@@ -3,6 +3,8 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+// Loosely adapted from https://github.com/microsoft/vscode-azure-account/blob/2f497562cab5f3db09f983ab5101040f27dceb70/src/nps.ts
+
 import { env, Memento, MessageItem, Uri, window } from "vscode";
 import { ext } from "vscode-azureappservice/out/src/extensionVariables";
 
--- a/src/utils/nps.ts
+++ b/src/utils/nps.ts
@@ -1,0 +1,81 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { env, Memento, MessageItem, Uri, window } from "vscode";
+import { ext } from "vscode-azureappservice/out/src/extensionVariables";
+
+const PROBABILITY = 0.15;
+const MIN_SESSION_COUNT = 10;
+
+const SURVEY_NAME = 'nps1';
+const SURVEY_URL = 'https://www.surveymonkey.com/r/vscodedockernpsinproduct';
+
+const SESSION_COUNT_KEY = `${SURVEY_NAME}/sessioncount`;
+const LAST_SESSION_DATE_KEY = `${SURVEY_NAME}/lastsessiondate`;
+const IS_CANDIDATE_KEY = `${SURVEY_NAME}/iscandidate`;
+
+export async function nps(globalState: Memento): Promise<void> {
+    try {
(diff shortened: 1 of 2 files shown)


src/utils/nps.ts
```diff
@@ -0,0 +1,85 @@
...
+// Loosely adapted from https://github.com/microsoft/vscode-azure-account/blob/2f497562cab5f3db09f983ab5101040f27dceb70/src/nps.ts
+
+import { env, Memento, MessageItem, Uri, window } from "vscode";
+import { ext } from "vscode-azureappservice/out/src/extensionVariables";
+
+const PROBABILITY = 0.15;
+const MIN_SESSION_COUNT = 10;
+
+const SURVEY_NAME = 'nps1';
+const SURVEY_URL = 'https://www.surveymonkey.com/r/vscodedockernpsinproduct';
```
philliphoff: Should be `aka.ms` link?

bwateratmsft: I'm not sure. The Azure Account NPS survey used the direct link, and attached platform, extension version, and machine ID to the query string; I'm not sure if the aka.ms one will keep that query string for the redirected URL?

philliphoff: (Assuming such a link would work with the query parameters added later.)

src/utils/nps.ts
```diff
@@ -0,0 +1,85 @@
...
+        const result = (await window.showInformationMessage('Do you mind taking a quick feedback survey about the Docker Extension for VS Code?', take, remind, never)) || remind;
+
+        ext.reporter.sendTelemetryEvent('nps', { survey: SURVEY_NAME, response: result.title });
+
+        if (result === take) {
+            // If they hit Take, don't ask again (for this survey name), and open the survey
+            await globalState.update(IS_CANDIDATE_KEY, false);
+            await env.openExternal(Uri.parse(`${SURVEY_URL}?o=${encodeURIComponent(process.platform)}&m=${encodeURIComponent(env.machineId)}`));
+        } else if (result === remind) {
+            // If they hit the X or Remind Me Later, ask again in 3 sessions
```
philliphoff: The session count monotonically increases, so once above 13 sessions, the user still has a chance of being prompted on each subsequent session.  Perhaps set the session count to `MIN_SESSION_COUNT - 3`?

bwateratmsft: At session 10, we will decide a value for isCandidate, and save it (lines 56, 57). If it decides they are not a candidate they will never be prompted (line 31), otherwise, it goes through the take/remind/never flow. Take and never both result in isCandidate becoming false, so they will never be prompted. Remind me brings the session count down to 7 but leaves isCandidate true, so when it gets back up to session 10, we will prompt again.

bwateratmsft: X is treated as remind later, so how could the session count ever get above 10?

src/utils/nps.ts
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        const take: MessageItem = { title: 'Take Survey' };
+        const remind: MessageItem = { title: 'Remind Me Later' };
+        const never: MessageItem = { title: 'Don\'t Show Again' };
+
+        // Prompt, treating hitting X as Remind Me Later
+        const result = (await window.showInformationMessage('Do you mind taking a quick feedback survey about the Docker Extension for VS Code?', take, remind, never)) || remind;
+
+        ext.reporter.sendTelemetryEvent('nps', { survey: SURVEY_NAME, response: result.title });
```
philliphoff: Should we record the actual display string of the chosen selection or a non-display-string value (so we are free to tweak the display text without complicating telemetry queries)?

bwateratmsft: Yeah, can do that. I had it that way originally but wanted to simplify, down to one line for telemetry.

PR #1377, opened 2019-10-28
Remove docker.importCertificates

### LINKED ISSUE #854: [Discussion] Deprecate `docker.importCertificates` setting
VS Code is adding support for system certs, which will probably make the docker-specific setting obsolete. See here: https://github.com/Microsoft/vscode/issues/52880

Specifically, I was able to repro cert issues using Fiddler to decrpyt https traffic and those errors went away on the latest insiders.

https://github.com/Microsoft/vscode-docs/blob/vnext/release-notes/v1_33.md#network-proxy-support-for-extensions

bwateratmsft:
Resolves #854. Because VSCode added this (on by default) with https://github.com/microsoft/vscode/issues/52880 it is no longer needed in this product.


package.json
```diff
@@ -1366,38 +1366,6 @@
                     "default": "",
                     "description": "Equivalent to setting the DOCKER_MACHINE_NAME environment variable."
                 },
-                "docker.importCertificates": {
```
bwateratmsft: Is this the "proper" way to remove / deprecate a setting @EricJizbaMSFT ?

src/utils/registryRequestUtils.ts
```diff
@@ -4,10 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Response } from "request";
+import * as request from 'request-promise-native';
```
philliphoff: Nit: these two imports can be combined.

package.json
EricJizbaMSFT: Fine with me as long as it's mentioned in the changelog. Also as long as we verify there aren't any major differences between this support and VS Code support (I feel like minor differences could just be filed as issues on VS Code).

StephenWeatherford: That's not deprecating, it's removing.  vscode supports a "deprecationMessage" for a setting, you might consider using that for a release before removing.  Also I'm pretty sure I put the usage of this setting into telemetry, so you could see how many people are using it.

bwateratmsft: @StephenWeatherford I think deprecating for a release is reasonable. Part of the motivation for removing is to get rid of a component governance alert. Theoretically it's safe to remove as VSCode's setting is enabled-by-default, but it's also safe to merely deprecate for a release and sit on this PR until then. I'll discuss with PMs too.

bwateratmsft:
I validated that the VSCode setting works as expected, therefore it should be safe to remove this:

With the setting **on** (the default), and this `docker.importCertificates` removed:
image

With the setting **off** (_not_ default), and this `docker.importCertificates` removed:
image

bwateratmsft:
@grhm @MiguelTVMS About a year back you helped with getting this `docker.importCertificates` tested and added: https://github.com/microsoft/vscode-docker/issues/483

Since then, VSCode added https://github.com/microsoft/vscode/issues/52880 which theoretically makes the `docker.importCertificates` setting obsolete. Can you try turning off `docker.importCertificates` and turning on the VSCode setting (it's on by default):
image

I tested successfully with a private registry, with steps based on @grhm's https://github.com/grhm/testRegistry, but I would love more coverage if possible.

grhm:
I am running VSCode 1.39.2, with http.systemCertificates enabled.
@bwateratmsft, I can confirm that if I set docker.importCertificates false and restart VSCode (not sure if thats necessary), I can still content to my private registry with the certificate in the Windows Cert Store.

bwateratmsft:
Thanks @grhm, that's good news!

PR #1386, opened 2019-11-01
Add support for SSH remote Docker daemons

### LINKED ISSUE #646: Add support for ssh
As of Docker 18.09, you can now connect to remote daemons via SSH.   For example:

`docker -H ssh://server.example.com ps`

Also setting `DOCKER_HOST` works as well:
```
$ export DOCKER_HOST=ssh://server.example.com
$ docker ps
```

More info [here](https://docs.docker.com/engine/reference/commandline/dockerd/) at the end of the  Examples->Daemon socket option section.

At present, you cannot enter the protocol portion of the DOCKER_HOST setting in vscode-docker.  Also, setting the environment variable DOCKER_HOST outside of vscode does not work either.

This is a request to add support for ssh in DOCKER_HOST.

bwateratmsft:
Resolves #646


package.json
```diff
@@ -1979,6 +1979,7 @@
         "azure-storage": "^2.10.3",
         "deep-equal": "^1.1.0",
         "dockerfile-language-server-nodejs": "^0.0.21",
+        "docker-modem": "^2.0.3",
```
bwateratmsft: Need `docker-modem` 2.0.4+ when it is available.

src/utils/refreshDockerode.ts
```diff
@@ -0,0 +1,99 @@
...
+    return undefined;
+}
+
+async function getSshAuthSock(): Promise<string | undefined> {
+    if (isWindows()) {
+        return '\\\\.\\pipe\\openssh-ssh-agent';
+    } else {
+        // On Mac and Linux, if it isn't set there's nothing we can do
+        // Running ssh-agent will yield a new agent that doesn't have the needed keys
+        throw new Error('In order to use an SSH DOCKER_HOST on OS X and Linux, you must configure an ssh-agent.');
```
bwateratmsft: This error will be thrown and immediately eaten, we should instead show a warning/error message.

src/utils/exec.ts
```diff
@@ -0,0 +1,23 @@
...
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as cp from 'child_process';
+
+export async function exec(command: string, options?: cp.ExecOptions, progress?: (content: string, process: cp.ChildProcess) => void): Promise<{ stdout: string, stderr: string }> {
+    return await new Promise((resolve, reject) => {
+        const p = cp.exec(command, options, (error, stdout, stderr) => {
+            if (error) {
+                reject(error);
```
philliphoff: `return reject(error)`

src/utils/exec.ts
```diff
@@ -0,0 +1,23 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as cp from 'child_process';
+
+export async function exec(command: string, options?: cp.ExecOptions, progress?: (content: string, process: cp.ChildProcess) => void): Promise<{ stdout: string, stderr: string }> {
```
philliphoff: May consider renaming (e.g. `execAsync`) to avoid conflicts in the event that `child_process` gets imported for other reasons in the caller's code.

bwateratmsft: Good idea.

src/utils/refreshDockerode.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Dockerode parses and handles the well-known `DOCKER_*` environment variables, but it doesn't let us pass those values as-is to the constructor
+ * Thus we will temporarily update `process.env` and pass nothing to the constructor
+ */
+export async function refreshDockerode(): Promise<void> {
+    const oldEnv = process.env;
+    try {
+        process.env = { ...process.env }; // make a clone before we change anything
+        addDockerSettingsToEnv(process.env, oldEnv);
+        ext.dockerodeInitError = undefined;
+        ext.dockerode = new Dockerode(await getDockerodeOptions());
```
philliphoff: Since this method is now async, is there a chance that the now modified `process.env` is exposed more broadly than desired?

bwateratmsft: Good point. I'll look for a way to rearrange things so there's no async code inbetween the environment changes.

bwateratmsft: I think it is fixed now, double check my work.

src/utils/refreshDockerode.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    // Use default options
+    return undefined;
+}
+
+async function getSshAuthSock(): Promise<string | undefined> {
+    if (isWindows()) {
+        return '\\\\.\\pipe\\openssh-ssh-agent';
+    } else {
+        // On Mac and Linux, if it isn't set there's nothing we can do
```
karolz-ms: 'it' meaning DOCKER_HOST?

src/utils/refreshDockerode.ts
```diff
@@ -0,0 +1,104 @@
...
+// Also not exhaustive--only the properties we're interested in
+interface IDockerContext {
+    Endpoints: { [key: string]: IDockerEndpoint }
+}
+
+/**
+ * Dockerode parses and handles the well-known `DOCKER_*` environment variables, but it doesn't let us pass those values as-is to the constructor
+ * Thus we will temporarily update `process.env` and pass nothing to the constructor
+ */
+export async function refreshDockerode(): Promise<void> {
```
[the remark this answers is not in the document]
bwateratmsft: Generally speaking it seems the convention with TS is to not do `methodAsync`, but given that `exec` conflicts with `child_process.exec` there's a good reason to make it this way. That being said we could also name it something else but that seemed best to me.

src/utils/refreshDockerode.ts
bwateratmsft: SSH_AUTH_SOCK, but I can add that note.

src/utils/execAsync.ts
```diff
@@ -0,0 +1,23 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as cp from 'child_process';
+
+export async function execAsync(command: string, options?: cp.ExecOptions, progress?: (content: string, process: cp.ChildProcess) => void): Promise<{ stdout: string, stderr: string }> {
```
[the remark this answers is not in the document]
bwateratmsft: Yes. There is a TSLint rule that functions returning `Promise<>` must be marked async, and conversely, functions marked async must return `Promise<>`.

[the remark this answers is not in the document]
bwateratmsft: (Basically, `Promise<>` and `async` are 1:1)

src/utils/refreshDockerode.ts
```diff
@@ -0,0 +1,104 @@
...
+    const oldEnv = process.env;
+    try {
+        const newEnv: NodeJS.ProcessEnv = cloneObject(process.env); // make a clone before we change anything
+        addDockerSettingsToEnv(newEnv, oldEnv);
+
+        const dockerodeOptions = await getDockerodeOptions(newEnv);
+
+        ext.dockerodeInitError = undefined;
+        process.env = newEnv;
+        ext.dockerode = new Dockerode(dockerodeOptions);
```
philliphoff: Nit: the `finally` need only follow this statement (i.e. only reset `process.env` if it was actually changed) and is independent of the `catch`.

bwateratmsft: @philliphoff At a minimum, `getDockerodeOptions` must be in the `try` block, and `addDockerSettingsToEnv` should probably be, so we can't shrink the try block more than a line. How would you structure the `finally` to not change `process.env` unless needed, without making e.g. a boolean to keep track of the fact that we changed it?

philliphoff: It'd require two `try` blocks, one for the `catch` and the other for the `finally`.  Something like:

```typescript
const oldEnv = process.env;
try {
  ...
  process.env = newEnv;
  try {
    ext.dockerode = new Dockerode(dockerodeOptions);
  } finally {
    process.env = oldEnv;
  }
}
catch (error) {
  ...
}
```

src/utils/execAsync.ts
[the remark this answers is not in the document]
karolz-ms: :-/  What I was getting at is the async state machinery is not free and in this case it is unnecessary. 

But this is more general problem and orthogonal to what you are trying to do with this change, so never mind for now.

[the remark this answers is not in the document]
karolz-ms: We have a good reason--the rule is overly restrictive ☺️

PR #1396, opened 2019-11-05
Aggregate Docker build/run task labels with VS Code defaults

### LINKED ISSUE #1392: Add default VS labels to all task-built Docker images

philliphoff:
Moves the addition of the default VS Code labels to Docker images/containers to the base task providers in order to apply to all platforms (instead of just .NET Core).  Also updates the logic so that default labels are included (by default) even if the user explicitly adds their own labels, with an option (by setting `docker[Build|Run].labels.includeDefaults = false`) to *not* include the default labels.

Resolves #1392 (except for discussion related to ports, which should be a separate PR).

### DIFF
--- a/src/tasks/TaskDefinitionBase.ts
+++ b/src/tasks/TaskDefinitionBase.ts
@@ -13,3 +13,5 @@ export interface TaskDefinitionBase extends TaskDefinition {
     label?: string;
     dependsOn?: string[] | DependsOn;
 }
+
+export type DockerLabels = { includeDefaults?: boolean; } & { [key: string]: string; };
--- a/src/tasks/TaskDefinitionBase.ts
+++ b/src/tasks/TaskDefinitionBase.ts
@@ -15,3 +15,15 @@ export interface TaskDefinitionBase extends TaskDefinition {
 }
 
 export type DockerLabels = { includeDefaults?: boolean; } & { [key: string]: string; };
+
+export const defaultVsCodeLabels: { [key: string]: string } = {
+    'com.microsoft.created-by': 'visual-studio-code'
+};
+
+export function getAggregateLabels(labels: DockerLabels | undefined, defaultLabels: { [key: string]: string }): { [key: string]: string } {
+    const { includeDefaults, ...explicitLabels } = labels || {};
+
+    return (includeDefaults !== false)
+        ? { ...defaultLabels, ...explicitLabels }
+        : explicitLabels;
+}
--- a/package.json
+++ b/package.json
@@ -833,6 +833,13 @@
                             "labels": {
                                 "type": "object",
                                 "description": "Labels applied to the Docker image used for debugging.",
+                                "properties": {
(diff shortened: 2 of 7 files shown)


src/tasks/TaskDefinitionBase.ts
```diff
@@ -13,3 +13,17 @@ export interface TaskDefinitionBase extends TaskDefinition {
...
+export type DockerLabels = { includeDefaults?: boolean; } & { [key: string]: string; };
+
+export const defaultVsCodeLabels: { [key: string]: string } = {
+    'com.microsoft.created-by': 'visual-studio-code'
+};
+
+export function getAggregateLabels(labels: DockerLabels | undefined, defaultLabels: { [key: string]: string }): { [key: string]: string } {
+    const { includeDefaults, ...explicitLabels } = labels || {};
+
+    return (includeDefaults !== false)
```
karolz-ms: Just for my education: why `includeDefaults !== false` instead of just `includeDefaults`?

bwateratmsft: `includeDefaults` can be either `undefined`, which is the _real_ default, or `true` or `false`. `undefined` is considered "falsy", so using `if (includeDefaults !== false)` treats `undefined` as `true`, whereas `if (includeDefaults)` treats `undefined` as `false`.

bwateratmsft: [Here](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)'s a really interesting article on falsy-ness, which is definitely important for understanding the behavior of things like the above, along with how `||` and `&&` really behave in the context of assignments.

The wonderful `?.` operator doesn't exist until TypeScript 3.7 (due this month!), so instead of pretty things like `foo?.bar?.baz` we have to do `foo && foo.bar && foo.bar.baz` (which is also slightly different from the former).

philliphoff: The intent is `includeDefaults` be treated as `true` if omitted (i.e. if `undefined`), so we test whether it hasn't been set (explicitly) to `false`.

karolz-ms: In this case I suggest that we change the setting to `omitDefaults` so that we can treat it in code according to established JS conventions around truthiness

philliphoff: Personally, I don't like having to set things to `true` in order to turn behavior *off*, which is why I used `includeDefaults` (which defaults to true from the user's perspective).  I'm not sure whether the mechanics of JSON to JS to truthiness conversion matter to the user (or should matter).

karolz-ms: That's fair

PR #1400, opened 2019-11-06
Containerize Blazor static web assets manifest

### LINKED ISSUE #1275: Blazor apps using components cannot be debugged

bwateratmsft:
Resolves #1275

### DIFF
--- a/.azure-pipelines/common/test.yml
+++ b/.azure-pipelines/common/test.yml
@@ -18,7 +18,6 @@ steps:
     SERVICE_PRINCIPAL_DOMAIN: $(SERVICE_PRINCIPAL_DOMAIN)
     DISPLAY: :10 # Only necessary for linux tests
     DOCKER_UNAVAILABLE: ${{ parameters.dockerUnavailable }}
-    CODE_VERSION: insiders # TODO : remove this line
 
 - task: PublishTestResults@2
   displayName: 'Publish Test Results'
--- a/src/tasks/netcore/NetCoreTaskHelper.ts
+++ b/src/tasks/netcore/NetCoreTaskHelper.ts
@@ -144,7 +144,9 @@ export class NetCoreTaskHelper implements TaskHelper {
 
     public async postRun(context: DockerRunTaskContext, runDefinition: DockerRunTaskDefinition): Promise<void> {
         try {
-            await updateBlazorManifest(context, runDefinition);
+            if (await NetCoreTaskHelper.isWebApp(runDefinition.netCore.appProject)) {
+                await updateBlazorManifest(context, runDefinition);
+            }
         } catch (err) {
             context.terminal.writeWarningLine('Failed to update Blazor static web assets manifest. Static web assets may not work.');
             context.terminal.writeWarningLine(`The error was: ${err}`);
@@ -186,6 +188,12 @@ export class NetCoreTaskHelper implements TaskHelper {
         } catch { }
 
         return false;
+    }
+
+    public static async isWebApp(appProject: string): Promise<boolean> {
+        const projectContents = await fse.readFile(appProject);
+
+        return /Microsoft\.NET\.Sdk\.Web/i.test(projectContents.toString());
     }
 
     private async inferUserSecrets(helperOptions: NetCoreTaskOptions): Promise<boolean> {
--- a/src/tasks/netcore/NetCoreTaskHelper.ts
+++ b/src/tasks/netcore/NetCoreTaskHelper.ts
@@ -18,6 +18,7 @@ import { DockerBuildTaskDefinition } from '../DockerBuildTaskProvider';
 import { DockerContainerVolume, DockerRunOptions, DockerRunTaskDefinitionBase } from '../DockerRunTaskDefinitionBase';
 import { DockerRunTaskDefinition } from '../DockerRunTaskProvider';
 import { DockerBuildTaskContext, DockerRunTaskContext, DockerTaskScaffoldContext, getDefaultContainerName, getDefaultImageName, inferImageName, TaskHelper } from '../TaskHelper';
+import { updateBlazorManifest } from './updateBlazorManifest';
 
 export interface NetCoreTaskOptions {
     appProject?: string;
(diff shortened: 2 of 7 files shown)


.azure-pipelines/common/test.yml
```diff
@@ -18,7 +18,6 @@ steps:
     SERVICE_PRINCIPAL_DOMAIN: $(SERVICE_PRINCIPAL_DOMAIN)
     DISPLAY: :10 # Only necessary for linux tests
     DOCKER_UNAVAILABLE: ${{ parameters.dockerUnavailable }}
-    CODE_VERSION: insiders # TODO : remove this line
```
bwateratmsft: Will need them to release 1.40 before we can submit.

src/tasks/netcore/updateBlazorManifest.ts
```diff
@@ -0,0 +1,105 @@
...
+            throw new Error('Unable to determine Blazor manifest locations from output file.')
+        }
+    } finally {
+        if (await fse.pathExists(locationsFile)) {
+            await fse.unlink(locationsFile);
+        }
+    }
+}
+
+async function updateBlazorManifestInternal(context: DockerRunTaskContext, inputManifest: string, outputManifest: string, volumes: DockerContainerVolume[], os: PlatformOS): Promise<void> {
```
karolz-ms: transformBlazorManifest?

src/tasks/netcore/updateBlazorManifest.ts
```diff
@@ -0,0 +1,105 @@
...
+    const tempFileProvider = new OSTempFileProvider(new LocalOSProvider(), new ChildProcessProvider());
+
+    const locationsFile = tempFileProvider.getTempFilename();
+
+    const targetsFile = path.join(ext.context.asAbsolutePath('resources'), 'GetBlazorManifestLocations.targets');
+
+    const command = `dotnet build /r:false /t:GetBlazorManifestLocations /p:CustomAfterMicrosoftCommonTargets=${targetsFile} /p:BlazorManifestLocationsOutput=${locationsFile} ${runDefinition.netCore.appProject}`;
+
+    try {
+        await execAsync(command, { timeout: 5000 });
```
karolz-ms: Are you sure 5 seconds is enough?

src/tasks/netcore/updateBlazorManifest.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export async function updateBlazorManifest(context: DockerRunTaskContext, runDefinition: DockerRunTaskDefinition): Promise<void> {
+    const tempFileProvider = new OSTempFileProvider(new LocalOSProvider(), new ChildProcessProvider());
+
+    const locationsFile = tempFileProvider.getTempFilename();
+
+    const targetsFile = path.join(ext.context.asAbsolutePath('resources'), 'GetBlazorManifestLocations.targets');
+
+    const command = `dotnet build /r:false /t:GetBlazorManifestLocations /p:CustomAfterMicrosoftCommonTargets=${targetsFile} /p:BlazorManifestLocationsOutput=${locationsFile} ${runDefinition.netCore.appProject}`;
```
karolz-ms: Is this going to be executed for every .NET Core project?

src/tasks/netcore/updateBlazorManifest.ts
```diff
@@ -0,0 +1,105 @@
...
+
+interface StaticWebAssets {
+    ContentRoot: ContentRoot[];
+}
+
+interface Manifest {
+    StaticWebAssets: StaticWebAssets;
+}
+
+export async function updateBlazorManifest(context: DockerRunTaskContext, runDefinition: DockerRunTaskDefinition): Promise<void> {
```
bwateratmsft: Do we need to avoid running this for release-mode builds?

src/tasks/netcore/updateBlazorManifest.ts
bwateratmsft: Yes. The most we can do is restrict it to projects using the `Microsoft.NET.Sdk.Web` SDK, but I'm not entirely sure if that's an appropriate filter. We don't have any way of knowing more information about .NET Core projects.

src/tasks/netcore/updateBlazorManifest.ts
bwateratmsft: The Blazor thing is best-effort. Since we have no choice but to run it on nearly everything (see above comment), it would be better to timeout and kill the process than to take longer, since the projects actually needing it are probably a pretty small minority right now.

Additionally, the only way this takes longer than 5s would be if it's busted (in which case, we *should* terminate), or there's an absolutely massive amount of props/targets to evaluate.

In practice this only took ~300ms to run on a normal project, I think 5s is plenty.

src/tasks/netcore/updateBlazorManifest.ts
bwateratmsft: Then again, we do it only for AspNetCore in VS, so I'll filter on that SDK.

src/tasks/netcore/updateBlazorManifest.ts
bwateratmsft: We could do it only if the manifest path contains "debug" as that would be an accurate test the vast majority of the time.

bwateratmsft: Edit: We can't actually do release builds the way VS does. The method we use for obtaining the output path (via `netCoreProjectProvider`) doesn't take inputs related to configuration, so it would always return the debug path.

So, this is moot.

src/tasks/netcore/updateBlazorManifest.ts
karolz-ms: I think what you have is probably fine as a starting point; but I have seen cases when things completed within hundreds of milliseconds on my dev machine (not much else going on, content already cached in filesystem cache) vs took many seconds on a slow machine that just happened to do an anti-virus scan at the same time, completely saturating the storage subsystem.

We might consider augmenting our `execAsync` and `spawnAsync` methods to explicitly check for timeout condition and emit telemetry when that happens (that would be a separate PR of course), so that we know if this is a problem or not.

bwateratmsft: That makes sense. We can also monitor debugging success rates, errors, and other things to try to get hints to see if timeouts are a common occurrence.

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -178,6 +190,12 @@ export class NetCoreTaskHelper implements TaskHelper {
         return false;
     }
 
+    public static async isWebApp(appProject: string): Promise<boolean> {
+        const projectContents = await fse.readFile(appProject);
+
+        return /Microsoft\.NET\.Sdk\.Web/i.test(projectContents.toString());
```
karolz-ms: `/<\s*Project\s+Sdk=\"Microsoft.NET.Sdk.Web\"/` ?

(a bit more specific)

bwateratmsft: I thought about it, but we should maybe allow for some other things, like:
`<Project DefaultTargets="build" ToolsVersion="1234" Sdk="...">`, since there can be more things than just `Sdk` as attributes.

Maybe `/<\s*Project\s+.*Sdk\s*=\s*\"Microsoft.NET.Sdk.Web\"/`? Although maybe it's too complicated...

karolz-ms: Good point. Also the `<Project>` element can span multiple lines. Maybe checking just for `Sdk\s*=\s*\"Microsoft.NET.Sdk.Web\"` is a good compromise...

PR #1410, opened 2019-11-08
Update to 0.9.1-alpha

bwateratmsft:


.vscode/settings.json
```diff
@@ -13,5 +13,6 @@
 	"editor.formatOnSave": true,
 	"editor.codeActionsOnSave": {
 		"source.fixAll": true
-	}
+	},
+	"typescript.tsdk": "node_modules/typescript/lib"
```
bwateratmsft: Is this really needed?

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-docker",
-    "version": "0.9.0",
+    "version": "0.9.1-alpha",
```
bwateratmsft: Hold until 0.9.0 is released publicly

.vscode/settings.json
philliphoff: It sounds like it's only needed if VS Code cannot otherwise detect what version of TypeScript to use; did you find it necessary to get 3.7 features to work?

bwateratmsft: So this showed up when I selected on bottom right, where it has TS version, and changed from 3.6.3 ("What VSCode has") to 3.7.2 ("What the workspace has"). It may be until 1.41 that VSCode switches to 3.7?

bwateratmsft: I tested it out, and we do in fact need this line or the Intellisense starts disagreeing with the actual build. Presumably, when VSCode moves to 3.7+, this would no longer be needed.

PR #1422, opened 2019-11-13
Miscellaneous fixes

### LINKED ISSUE #1419: F5 without any .vscode folder has bad behavior
1. Create a .NET Core project with `dotnet new`
1. Open the folder in VSCode
1. When the prompt about required assets comes up, click "not now"
1. Hit F5
1. VSCode prompts on environment, showing .NET Core and the new and old Docker .NET Core debugging. Pick a .NET Core choice.

Both Docker options result in bad behavior. Instead we should throw an error saying to add it with "Add Docker Files..."

I think it calls provideInitialDebugConfigurations, which comes back with nothing in our extension.

### LINKED ISSUE #1421: Validate that spaces don't mess up .NET Core targets magic
For .NET Core projects we use a special project file to determine the output binary path, and a special targets file to determine the Blazor manifest input/output location.

Need to verify that spaces somewhere in those paths do not break the logic, since user directories may contain spaces.

bwateratmsft:
Resolves #1419
Resolves #1421

Also resolves the path issue @haniamr ran into.

### DIFF
--- a/src/debugging/DockerDebugConfigurationProvider.ts
+++ b/src/debugging/DockerDebugConfigurationProvider.ts
@@ -28,7 +28,7 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
 
         // Prompt them to add Docker files since they probably haven't
         window.showErrorMessage(
-            'To debug in a Docker container on supported platforms, use the command \"Docker: Add Docker Files to Workspace\", or click the button below.',
+            'To debug in a Docker container on supported platforms, use the command \"Docker: Add Docker Files to Workspace\", or click \"Add Docker Files\".',
             ...[add]).then((result) => {
                 if (result === add) {
                     commands.executeCommand('vscode-docker.configure');
--- a/src/debugging/DockerDebugConfigurationProvider.ts
+++ b/src/debugging/DockerDebugConfigurationProvider.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { CancellationToken, debug, DebugConfiguration, DebugConfigurationProvider, ProviderResult, window, WorkspaceFolder } from 'vscode';
+import { CancellationToken, commands, debug, DebugConfiguration, DebugConfigurationProvider, MessageItem, ProviderResult, window, WorkspaceFolder } from 'vscode';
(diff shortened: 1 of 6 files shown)


src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -24,7 +24,8 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
     ) { }
 
     public provideDebugConfigurations(folder: WorkspaceFolder | undefined, token?: CancellationToken): ProviderResult<DebugConfiguration[]> {
-        return undefined;
+        window.showErrorMessage("To debug in a Docker container on supported platforms, use the command \"Docker: Add Docker Files to Workspace\".");
```
philliphoff: Nit: use single quotes for strings.

philliphoff: Couldn't they also use the "initialize for debugging" command, too?

philliphoff: Should we show an error message with button(s) to immediately run either of those commands?

bwateratmsft: They could do both, but I'm guessing that most people running into this situation have not added Docker support in the first place. For future projects, if they had, it would already have the debug configuration and they wouldn't hit this.

The Add workflow does ask before overwriting, so IMO it's safe.

We can add a button, shouldn't be difficult.

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -24,7 +24,18 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
     ) { }
 
     public provideDebugConfigurations(folder: WorkspaceFolder | undefined, token?: CancellationToken): ProviderResult<DebugConfiguration[]> {
-        return undefined;
+        const add: MessageItem = { title: 'Add Docker Files' };
+
+        // Prompt them to add Docker files since they probably haven't
+        window.showErrorMessage(
+            'To debug in a Docker container on supported platforms, use the command \"Docker: Add Docker Files to Workspace\", or click the button below.',
```
philliphoff: Nit: if I recall, it's preferable not to refer to the relative positions of things in the UX, as it's not always obvious what that means when using assistive technologies (e.g. screen readers).  It may be better to refer to the button by its name.

PR #1429, opened 2019-11-14
Add `Open in Browser` command to running containers

philliphoff:
Adds an `Open in Browser` command to the context menu of running containers in the containers tree view.

Screen Shot 2019-11-14 at 11 51 43 AM

The command opens the browser to the first exposed port mapped to the host that likely represents a "web" endpoint.  The heuristic works as follows:

 - Choose the first port from a prioritized list of ports common to various web stacks with SSL ports preferred over standard HTTP ports (e.g. `443`, `80`, `3000`, `3001`, `5001`, `5000`, `8080`, `8081`)
 - If none found, and if only a single port was mapped, use that port
 - Otherwise, if more than one port was mapped, display a quick-pick where the user can select the port.

Screen Shot 2019-11-14 at 11 52 00 AM

The "possible" ports as well as the selected port is captured by telemetry, in order to understand how that heuristic could be improved.


package.json
```diff
@@ -293,20 +294,25 @@
                     "when": "view == dockerContainers && viewItem =~ /container$/i",
                     "group": "containers_1_general@3"
                 },
+                {
+                    "command": "vscode-docker.containers.browse",
```
bwateratmsft: Should we show this only in context menu or also in palette? I'm OK with showing in palette, just something to think about.

src/commands/containers/browseContainer.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    return undefined;
+}
+
+export async function browseContainer(context: IActionContext, node?: ContainerTreeItem): Promise<void> {
+    const telemetryProperties = <BrowseTelemetryProperties>context.telemetry.properties;
+
+    if (!node) {
+        node = await captureBrowseCancelStep('node', telemetryProperties, () => ext.containersTree.showTreeItemPicker<ContainerTreeItem>(ContainerTreeItem.allContextRegExp, context));
```
bwateratmsft: This will show all containers, not just running ones, right?

src/commands/containers/browseContainer.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    const inspectInfo = await node.getContainer().inspect();
+
+    const ports = inspectInfo.NetworkSettings && inspectInfo.NetworkSettings.Ports || {};
```
bwateratmsft: Though I'm pretty sure `inspectInfo` and `NetworkSettings` will always be present, should we do `inspectInfo && inspectInfo.NetworkSettings && ...` just to be safe?

src/commands/containers/browseContainer.ts
```diff
@@ -0,0 +1,114 @@
...
+              // Exclude ports that are non-TCP-based (which should be the default)...
+              .filter(port => port.containerPort.protocol === undefined || port.containerPort.protocol === 'tcp')
+              // Exclude ports not mapped to the host...
+              .filter(port => port.mappings && port.mappings.length > 0);
+
+    telemetryProperties.possiblePorts = possiblePorts.map(port => port.containerPort.port);
+
+    if (possiblePorts.length === 0) {
+        // tslint:disable-next-line: no-floating-promises
+        ext.ui.showWarningMessage('No valid ports were mapped from the container to the host.');
```
bwateratmsft: Do we want to leave a cancel step in telemetry here, and/or throw a `UserCancelledError`?

src/commands/containers/browseContainer.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        return;
+    }
+
+    // Prefer common port (in order of preference)...
+    let selectedPort = possiblePorts.find(port => commonWebPorts.find(commonPort => commonPort === port.containerPort.port) !== undefined);
```
bwateratmsft: Sets may not respect ordering so they may not work, but if they do this might be an easier way to do the intersection: https://2ality.com/2015/01/es6-set-operations.html#intersection

src/commands/containers/browseContainer.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    // Otherwise, if there's just a single port, assume that one...
+    if (selectedPort === undefined && possiblePorts.length === 1) {
```
bwateratmsft: Can't this check come before the above scan?

src/commands/containers/browseContainer.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    // Otherwise, ask the user which port to use...
+    if (selectedPort === undefined) {
+        const items = possiblePorts.map(port => ({ label: port.containerPort.port.toString(), port }));
+
+        // Sort ports in ascending order...
+        items.sort((a, b) => a.port.containerPort.port - b.port.containerPort.port);
+
+        const item  = await captureBrowseCancelStep('port', telemetryProperties, () => ext.ui.showQuickPick(items, { placeHolder: 'Select the container port to browse to.' }));
```
bwateratmsft: Nit: one extra space in `item  =`

src/commands/containers/browseContainer.ts
philliphoff: Couldn't hurt.

src/commands/containers/browseContainer.ts
philliphoff: I tried several variants and this seemed to work best (or was the least worst).  `UserCancelledError` doesn't seem technically correct, as the user didn't make a choice--the container just didn't have any ports to browse to.  Throwing any other kind of error results in the error dialog with the "report issue" button, which we clearly don't want.  You might argue that the command result should be `Failed` in this case, but I'm not sure it matters from the telemetry analysis point of view.  We can pick out this scenario by looking for a completed command, where the possible ports is the empty set and no port was selected.

src/commands/containers/browseContainer.ts
philliphoff: Logically yes, but the typing makes is slightly more awkward in my opinion.

src/commands/containers/browseContainer.ts
philliphoff: Surprised the linter didn't catch that, actually.

src/commands/containers/browseContainer.ts
bwateratmsft: For the somewhat similar case where the Azure registry already exists (the one Paul hit), I had it throw a `UserCancelledError` as a way of getting out of the execution; so it wouldn't be unprecedented.

package.json
philliphoff: Its peer commands seem to be designed to support the palette, so I'm ok with that, too.

bwateratmsft: Makes sense. It might also be the case that they're in the debugging window and don't want to switch.

src/commands/containers/browseContainer.ts
philliphoff: Good point, though its peer commands seem to not make a distinction either.  Still, it's simple enough to filter to running containers (just as the context menu command does).  I've also created #1430 to track looking at other commands to see if they should add filtering as well.

src/commands/containers/browseContainer.ts
philliphoff: We should try to make better use of the `Set` and `Map` types, but I'm not sure I want to do that here (quite yet).  (The spec indicates that sets keep values in insertion order, but I've not yet played with them.) . In any case, I discovered that my logic was inverted here anyway, so at least fixed that.

src/commands/containers/browseContainer.ts
```diff
@@ -0,0 +1,115 @@
...
+        node = await captureBrowseCancelStep('node', telemetryProperties, () => ext.containersTree.showTreeItemPicker<ContainerTreeItem>(ContainerTreeItem.runningContainerRegExp, context));
+    }
+
+    const inspectInfo = await node.getContainer().inspect();
+
+    const ports = inspectInfo && inspectInfo.NetworkSettings && inspectInfo.NetworkSettings.Ports || {};
+    const possiblePorts =
+        Object.keys(ports)
+              .map(portAndProtocol => ({ mappings: ports[portAndProtocol], containerPort: parsePortAndProtocol(portAndProtocol) }))
+              // Exclude port designations we cannot recognized...
```
haniamr: recognize*

src/commands/containers/browseContainer.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    let selectedPort =
+        possiblePorts.length === 1
+            // If there's just a single port, assume that one...
+            ? possiblePorts[0]
+            // Otherwise, prefer a common port (in order of preference)...
+            : commonWebPorts
+                .map(port => possiblePorts.find(possiblePort => port === possiblePort.containerPort.port))
+                .filter(possiblePort => possiblePort !== undefined)[0];
```
karolz-ms: Are sure the result of `filter()` will always have at least one element?

philliphoff: `filter()` will always return a valid array, but it may be empty.  If that's the case, `[0]` will return undefined, which is ok in this scenario, as it will get checked in the next `if` statement.

src/commands/containers/browseContainer.ts
philliphoff: Updated.

PR #1439, opened 2019-11-18
improve dockerfile pattern even more :sweat:

casz:
I have seen both `Dockerfile.*` and `Dockerfile-*` pattern in the wild.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1104,10 +1104,8 @@
                 "filenamePatterns": [
                     "*.dockerfile",
                     "Dockerfile",
-                    "Dockerfile.debug",
-                    "Dockerfile.dev",
-                    "Dockerfile.develop",
-                    "Dockerfile.prod"
+                    "Dockerfile.*",
+                    "Dockerfile-*"
                 ]
             },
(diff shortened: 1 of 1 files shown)


rcjsuen:
This has been raised numerous times and has gone back and forth I believe.

See #531, #248, #134, #192, and presumably many others.

casz:
Hmm. Dockerfile.windows and Dockerfile.linux than at least should be added as well 😓

philliphoff:
Wildcards for extensions tend to get us in trouble with other source files in the repo (e.g. `Dockerfile.cs`).  That said, I *think* we might be able to add a few more specific filenames to match more common conventions while avoiding some of the pitfalls of wildcards:

 - `Dockerfile-linux`
 - `Dockerfile-linux-amd64`
 - `Dockerfile-linux-arm64`
 - `Dockerfile-windows`
 - `Dockerfile.linux`
 - `Dockerfile.linux.amd64`
 - `Dockerfile.linux.arm64`
 - `Dockerfile.windows`

PR #1465, opened 2019-12-05
Allow fixed URI format string in dockerServerReadyAction.

### LINKED ISSUE #1448: dockerServerReadyAction does not work if verbatim URL is used (.NET Core debugging)
**Repro steps**

1. Create .NET Web API app (`dotnet new webapi --no-https`). When asked about port to listen on, specify 5000.
2. Add Docker files to project
3. In the Dockerfile, ensure that ASPNETCORE_URLS is set in the `base` stage:
```
FROM mcr.microsoft.com/dotnet/core/aspnet:3.0 AS base
WORKDIR /app
EXPOSE 5000
ENV ASPNETCORE_URLS=http://*:5000
...
```
4. In `tasks.json` make sure that port 5000 is used to expose the app on the host network:
```json
        {
            "type": "docker-run",
            "label": "docker-run: debug",
            "dependsOn": [
                "docker-build: debug"
            ],
            "dockerRun": {  
                "ports": [
                    {
                        "containerPort": 5000,
                        "hostPort": 5000
                    }
                ]              
            },
            "netCore": {
                "appProject": "${workspaceFolder}/netcorerest.csproj",
                "enableDebugging": true
            }
        },
```
5. In `launch.json` add the following (to use correct URL for the Web API)
```json
            "dockerServerReadyAction": {
                "uriFormat": "http://localhost:5000/WeatherForecast"
            }
```
6. F5

**Expected**

Browser opens using the URL specified in `launch.json`

**Actual**

Browser tries to open `http://localhost:5000/WeatherForecast%20http%205000` and fails

**Notes**

Seems like the `dockerServerReadyAction expects to always substitute something into the given value. The following works:
```json
            "dockerServerReadyAction": {
                "uriFormat": "%s://localhost:%s/WeatherForecast"
            }
```

Setting explicit startup URL for container debugging is a common need.

philliphoff:
Allows users to specify a fixed (i.e. with no substitutions) URI format string in the `dockerServerReadyAction` configuration object.  Also now displays an error if substitutions *are* made in the URI format string, but do not match the number of substitutions the extension expects (i.e. two, for protocol and port).

Resolves #1448

### DIFF
--- a/src/debugging/DockerServerReadyAction.ts
+++ b/src/debugging/DockerServerReadyAction.ts
@@ -14,8 +14,9 @@ import CliDockerClient from './coreclr/CliDockerClient';
 import { ResolvedDebugConfiguration } from './DebugHelper';
 
 // tslint:disable-next-line: no-any
-const localize = (message: string, ...param: any[]): string => {
-    return util.format(message, param);
+const localize = (message: string, formatString: string, ...param: any[]): string => {
+    // NOTE: This is a mock of the VS Code localize() method so its implementation may be subtly different.
+    return util.format(formatString.replace(/\{[0-9]+\}/g, '%s'), ...param);
 }
 
 const PATTERN = 'listening on.* (https?://\\S+|[0-9]+)'; // matches "listening on port 3000" or "Now listening on: https://localhost:5001"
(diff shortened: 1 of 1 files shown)


karolz-ms:
Thanks for taking time to fix this! ☺️

Issue #1444, opened 2019-11-22
error while creating a new network from vs code (HTTP code 404) driver not found


madkaikar-ashish:
Repro steps:

1. Select Docker from the left most vertical tabs 
2. In Networks section click on Create 
3. Or in command palette Docker Networks:Configure Explorer

Action: vscode-docker.networks.create
Error type: 404
Error Message: (HTTP code 404) driver not found - could not find plugin bridge in v1 plugin registry: plugin not found 

Version: 0.8.2
OS: win32
Product: Visual Studio Code
Product Version: 1.40.1
Language: en

<details>
<summary>Call Stack</summary>

```
extension.bundle.js:157:703098extension.bundle.js:157:703098
extension.bundle.js:157:703061extension.bundle.js:157:703061
m.buildPayload extension.bundle.js:157:703071
IncomingMessage.<anonymous> extension.bundle.js:157:702571
IncomingMessage.emit events.js:205:15
endReadableNT _stream_readable.js:1154:12
processTicksAndRejections task_queues.js:84:9
```

</details>

philliphoff:
@madkaikar-ashish Are you using Windows containers?  Also, did you try to create a `bridge` network?

I can reproduce the issue on Windows, using Windows containers, when creating a `bridge` network.  Apparently there are differing sets of Docker network types depending on whether using Linux containers vs. Windows containers.  We'll likely need to dynamically populate the network type quick-pick prompt based on target platform.

PR #1488, opened 2019-12-13
Use appropriate drives based on docker engine os

ravipal:
Earlier static set of drivers were used during create network. Some of these drivers are not applicable for windows os. Now based on engine os version, right set of drivers are provided to user.
This fixex #1444

### DIFF
--- a/src/commands/networks/createNetwork.ts
+++ b/src/commands/networks/createNetwork.ts
@@ -16,7 +16,7 @@ export async function createNetwork(_context: IActionContext): Promise<void> {
     });
 
     const engineVersion = await ext.dockerode.version();
-    const drivers = engineVersion.Os === "windows"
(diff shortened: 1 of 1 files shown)


src/commands/networks/createNetwork.ts
```diff
@@ -15,13 +15,22 @@ export async function createNetwork(_context: IActionContext): Promise<void> {
         prompt: 'Name of the network'
     });
 
-    const driverSelection = await ext.ui.showQuickPick(
-        [
+    const engineVersion = await ext.dockerode.version();
+    const drivers = engineVersion.Os === "windows"
```
philliphoff: Nit: use single quotes for strings.

ravipal: Fixed.

PR #1489, opened 2019-12-13
Migrate to ESLint

### LINKED ISSUE #1487: Migrate to ESLint
`tslint` [is being deprecated in 2019](https://github.com/palantir/tslint/issues/4534) in favor of `eslint` for both JavaScript and TypeScript. We should use the migration tools to migrate our use of `tslint`.

philliphoff:
[TSLint](https://palantir.github.io/tslint/) [is being deprecated by the end of 2019](https://github.com/palantir/tslint/issues/4534) in favor of [ESLint](), with TypeScript-specific rules being ported to an ESLint plugin ([`@typescript-eslint`](https://github.com/typescript-eslint/typescript-eslint)).

This PR migrates the repo's use of TSLint to ESLint.  A [tool](https://github.com/typescript-eslint/tslint-to-eslint-config) was used to port the `tslint.json` configuration file to ESLint `.eslintrc.js` configuration file.  As we derive our ruleset from the [`tslint-microsoft-contrib`](https://github.com/Microsoft/tslint-microsoft-contrib) ruleset which has not yet been (fully) ported, we make use of the `@typescript-eslint/tslint` plugin in the meantime to run those TSLint rules within the context of ESLint.

A number of minor code changes were made to better fit the standard set of ESLint rules.  A bunch of TSLint exclusion comments were also swapped with their ESLint equivalents.  Some additional exclusions were added for code that somehow didn't trigger warnings under TSLint.

Resolves #1487.


src/dockerHubSearch.ts
```diff
@@ -91,43 +91,50 @@ function invokeHubSearch(imageName: string, count: number, cache: boolean): Prom
         cache);
 }
 export interface IHubSearchResponse {
+    /* eslint-disable-next-line camelcase */
     num_pages: number;
```
ravipal: Looks like fixing this warning by rename seems simple.

src/tasks/DockerPseudoterminal.ts
```diff
@@ -22,10 +22,12 @@ export class DockerPseudoterminal implements Pseudoterminal {
     private readonly writeEmitter: EventEmitter<string> = new EventEmitter<string>();
     private readonly cts: CancellationTokenSource = new CancellationTokenSource();
 
+    /* eslint-disable-next-line no-invalid-this */
     public readonly onDidWrite: Event<string> = this.writeEmitter.event;
```
ravipal: Just for my understanding , why does eslint comlains about this in the member variable?

philliphoff: It's the use of `this`; the rule doesn't comprehend well setters for properties/fields of classes.

philliphoff: Luckily this construct isn't very common.

src/dockerHubSearch.ts
philliphoff: It likely matches the shape of the data coming back from an HTTP request, so I'm not inclined to rename it; unfortunately TS/JS doesn't have attributes that allow us to de-couple the JSON representation from its object representation.

src/debugging/coreclr/LocalOSProvider.ts
```diff
@@ -2,6 +2,8 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
+/* eslint-disable unicorn/filename-case */
```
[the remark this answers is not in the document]
philliphoff: The rule does allow regex's in an ignore list, though it would probably take longer to invent a regex that properly excludes such uses without *over* excluding files than just to add the few inline exclusions where needed.

src/commands/containers/browseContainer.ts
```diff
@@ -55,6 +55,7 @@ export async function browseContainer(context: IActionContext, node?: ContainerT
     const telemetryProperties = <BrowseTelemetryProperties>context.telemetry.properties;
 
     if (!node) {
+        /* eslint-disable-next-line @typescript-eslint/promise-function-async */
         node = await captureBrowseCancelStep('node', telemetryProperties, () => ext.containersTree.showTreeItemPicker<ContainerTreeItem>(ContainerTreeItem.runningContainerRegExp, context));
```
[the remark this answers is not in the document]
philliphoff: Perhaps, with a corresponding `await` in the body, but I don't like adding it for the purposes of satisfying the linter when it obviously isn't necessary.  Neither me nor @karolz-ms are fans of this rule in the first place; we might consider just turning it off.

src/commands/compose.ts
```diff
@@ -55,6 +55,7 @@ async function compose(commands: ('up' | 'down')[], message: string, dockerCompo
         // prompt for compose file
         const uris: vscode.Uri[] = await getDockerComposeFileUris(folder);
         if (!uris || uris.length === 0) {
+            /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
```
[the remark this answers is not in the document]
philliphoff: It's a common pattern for the `showInformationMessage()` type functions, as their returned promise completes only when the user actually responds to the message, which likely never happens (with the message eventually disappearing).

PR #1494, opened 2019-12-15
add remaining commands to view/title

eine:
Currently it is not possible to select multiple containers/images graphically and execute the same action with all of them (#331). However, it is possible to work around it by running some commands and selecting the items later. See https://github.com/microsoft/vscode-docker/issues/331#issuecomment-500477876.

As suggested in https://github.com/microsoft/vscode-docker/issues/331#issuecomment-518300409, this PR adds the commands which are not already being shown to the `view/title` of sections 'Containers' and 'Images'.

Some features work as expected, but others might need to be improved. Therefore, I'm keeping this PR as a draft for discussion.

---

- Containers:
  - **Attach Shell**: allows to select a single container. The 'select box' does NOT filter by container name; instead, it does by image name. This is unfortunate. Is there any other straighforward mechanism to search a container and attach to it?
  - **Inspect**, **View Logs**: allow to select a single container. Same issue as 'Attach Shell'. I think they should support selecting multiple containers.
  - **Remove**, **Restart**, **Stop**: allow to select multiple containers. OK.
  - **Browse**: is it sensible to add this too?
  - **Start**: produces a "No matching resources found" error.
  - **Select**: I am not sure about what this is expected to be used for.
- Images:
  - **Build**, **Run**, **Tag**: work as expected.
  - **Inspect**: same issue as with containers.
  - **Push**, **Remove**: allow to select a single image. Same issue as inspect.
  - **copyFullTag**: I am not sure about what this is expected to be used for.

### DIFF
--- a/package.json
+++ b/package.json
@@ -217,14 +217,38 @@
                     "group": "navigation@8"
                 },
                 {
-                    "command": "vscode-docker.networks.configureExplorer",
-                    "when": "view == dockerNetworks",
-                    "group": "navigation@8"
-                },
-                {
                     "command": "vscode-docker.containers.refresh",
                     "when": "view == dockerContainers",
                     "group": "navigation@9"
+                },
+                {
+                    "command": "vscode-docker.containers.attachShell",
+                    "when": "view == dockerContainers"
+                },
+                {
+                    "command": "vscode-docker.containers.inspect",
+                    "when": "view == dockerContainers"
+                },
+                {
+                    "command": "vscode-docker.containers.remove",
+                    "when": "view == dockerContainers"
+                },
+                {
+                    "command": "vscode-docker.containers.restart",
+                    "when": "view == dockerContainers"
+                },
+                {
+                    "command": "vscode-docker.containers.stop",
+                    "when": "view == dockerContainers"
+                },
+                {
+                    "command": "vscode-docker.containers.viewLogs",
+                    "when": "view == dockerContainers"
+                },
+                {
+                    "command": "vscode-docker.networks.configureExplorer",
+                    "when": "view == dockerNetworks",
+                    "group": "navigation@8"
                 },
                 {
                     "command": "vscode-docker.networks.create",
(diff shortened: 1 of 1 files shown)


bwateratmsft:
@eine, do you mind if I close this since we have added multi-select with #1623?

eine:
@bwateratmsft, I don't expect o make this "ready for review". However, I'd like to keep it as is, until the new release is deployed and I can test it and rethink the content of this placeholder. I will likely close it then, but there might be some subtle details that might be worth keeping visible. For example, some of the open questions/comments above are unanswered yet:

- Attach Shell, Inspect, View Logs, etc.: the 'select box' does NOT filter by container name; instead, it does by image name. This is unfortunate. Is there any other straighforward mechanism to search a container and do an action to/with it?
- Select: I am not sure about what this is expected to be used for.
- copyFullTag: I am not sure about what this is expected to be used for.

bwateratmsft:
@eine Sounds good to me.

- For Attach Shell / Inspect / etc. showing image name, not container name, it's sort of by design. The Explorer view for containers shows `FullTag (ContainerName - Status)` by default, but this is very configurable. If you set the `Label` property to be `ContainerName` (instead of the default `FullTag`), and set `Description` to be `FullTag` and `Status` (instead of the default `ContainerName` and `Status`), then the quick pick menu shows container names instead of image names: 
image
The Explorer view with those settings:
image
This is the settings JSON for the above (though you can change with the UI with the gear icon on each Explorer view):
```
{
    "docker.containers.label": "ContainerName",
    "docker.containers.description": [
        "FullTag",
        "Status"
    ]
}
```
To make a product change for this, it would be as simple as changing the default setting values. If you want, you can create an issue for that.
- Select is used as a substitution in debugging / task templates, where you can put e.g. `${command:vscode-docker.containers.select}` into the template. See [here](https://code.visualstudio.com/docs/editor/variables-reference#_command-variables) for more info.
- CopyFullTag puts the full image tag into the clipboard, and therefore can't be a multi-select command (since you'd just overwrite the clipboard until the last item).

bwateratmsft:
Closing this for now since multi-select seems to cover the scenario. Please let us know if you think multi-select should be enabled for more commands! Thank you!

Issue #592, opened 2018-10-24
Dockerfile should go in same folder as project


StephenWeatherford:

PrashanthCorp:
I might not be understanding this correctly. 
Do you have more info of cases when it does not place the dockerfile in the workspace folder?

StephenWeatherford:
If you have multiple projects in the workspace, Dockerfiles should go in same folder as the project file.

bwateratmsft:
I'm marking this one for milestone 0.10.0 since it may be essentially necessary for compose to work.

dbreshears:
.Net Core is fixed in 0.10.0.  Separate item will be created for Node.js in 0.11.0

PR #1502, opened 2019-12-19
Scaffold Dockerfile next to project file for .NET

philliphoff:
Updated Docker scaffolding to generate the `Dockerfile` next to the project file in the case of .NET applications (rather than the current workspace root folder).  The `.dockerignore` file continues to be generated at the root.  This brings scaffolding more in line with Visual Studio, as well as helps when scaffolding multiple services within a mono-repo.

The existing scaffolding (i.e. "configuring") logic, while pluggable to accommodate multiple stacks, was not very flexible on generating differing sets of Docker-related files (essentially limited to just `.dockerignore`, `Dockerfile`, and the Docker Compose files) nor was it flexible at all in where those files should be generated.  Therefore, a new `Scaffolder` type was added, where scaffolders (e.g. for each stack) are registered with `registerScaffolder()`, and then invoked on command via `scaffold()`.  The .NET "configure-er" was refactored to use this new mechanism directly, while an adaptor was created for the other stack "configure-ers".

`Scaffolder` implementors are invoked by `scaffold()` and passed context (e.g. the root folder), and expected to simply return a collection of `ScaffoldFile`s, each having a filename (relative to the root), its contents, and whether the file should be opened after scaffolding.  This enables a scaffolder to both generate arbitrary sets of files, as well as have them generated at arbitrary locations.  `scaffold()` then manages actually writing the file contents (and prompting for overrides).

> Note: once this PR is committed, a follow-on PR will refactor the Node.js scaffolder to ensure it generates its files at the same level as the `package.json`.

Related to #592.


src/configureWorkspace/scaffolding.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export interface ScaffolderContext extends ScaffoldContext {
+    captureStep<TReturn, TPrompt extends (...args: []) => Promise<TReturn>>(step: ConfigureTelemetryCancelStep, prompt: TPrompt): TPrompt;
+    folder: vscode.WorkspaceFolder;
+    initializeForDebugging: boolean;
+    platform: Platform;
+    promptForOS(): Promise<PlatformOS>;
+    promptForPorts(defaultPorts?: number[]): Promise<number[]>;
+    rootFolder: string;
+}
```
karolz-ms: @philliphoff for my education, can you explain what is the intention of defining `ScaffolderContext` interface that extends `ScaffoldContext`, but most of the members of the new interface seem to re-use the names of the parent interface members, just making them non-optional? Should I think about this as "the new interface members override the old interface members"?

Except down below (starting at line 118) you are calling the scaffolder by constructing an object that seems to de-structure the context object (of type `ScaffoldContext`), then adds a bunch of properties, some named, and some just by position?? How can you be sure that the data you are passing is going to end up populating the right properties of the resulting `ScaffolderContext` object?

It has been a while since I was looking at TS code that uses advanced typing mechanisms and it is quite confusing to me

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -267,20 +257,101 @@ function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, o
...
+    telemetryProperties.packageFileSubfolderDepth = getSubfolderDepth(context.rootFolder, rootRelativeProjectFileName);
+
+    const projectFilePath = path.posix.join(context.rootFolder, rootRelativeProjectFileName);
+    const workspaceRelativeProjectFileName = path.posix.relative(context.folder.uri.fsPath, projectFilePath);
+
+    let serviceNameAndPathRelative = rootRelativeProjectFileName.slice(0, -(path.extname(rootRelativeProjectFileName).length));
+    const projFileContents = (await fse.readFile(path.join(context.rootFolder, rootRelativeProjectFileName))).toString();
+
+    // Extract TargetFramework for version
+    const [version] = extractRegExGroups(projFileContents, /<TargetFramework>(.+)<\/TargetFramework/, ['']);
```
karolz-ms: What if the project is using `TargetFrameworks` property (multiple frameworks)?

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -267,20 +257,101 @@ function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, o
...
+
+    if (context.outputFolder) {
+        // We need paths in the Dockerfile to be relative to the output folder, not the root
+        serviceNameAndPathRelative = path.relative(context.outputFolder, path.join(context.rootFolder, serviceNameAndPathRelative));
+    }
+
+    serviceNameAndPathRelative = serviceNameAndPathRelative.replace(/\\/g, '/');
```
karolz-ms: Why is this necessary? Can you add a comment?

karolz-ms: Is there a standard library API that could be used for this?

src/configureWorkspace/scaffolding.ts
philliphoff: `scaffold()` can be invoked either via a user-executed command or programmatically (e.g. by the IoT extension).  In the latter case, in particular, we want to be lenient in what's specified, filling in defaults where it makes sense.  Hence, the `ScaffoldContext` input to `scaffold()` makes most properties optional.  However, we want the individual scaffolders (e.g. for .NET, Node.js, etc.) to be able to depend on *some* but not *every* one of those values being specified (or, rather, we don't want each scaffolder to have its own custom logic for determining the defaults).  So `ScaffolderContext` inherits from `ScaffoldContext`, to pass along the optional properties, but then "overrides" those that will have been provided common defaults.  (In C#, you do something equivalent by using the `new` keyword in the interface/class.)

In line 118, we construct a new `ScaffolderContext` by copying all properties of the existing `ScaffoldContext` (`...context,`), setting some of the properties that only exist on `ScaffolderContext` (e.g. `promptForOS`), and then overrides the properties optional in `ScaffoldContext` but no longer optional in `ScaffolderContext`.  Note that the copy/override is done by property name and never by position, apart from copying/overriding occurring in left-to-right order.  Also note that we create a new context rather than update the original `ScaffoldContext` context in the interests of not modifying an object we don't own, and we copy values from the original `ScaffoldContext` context to the new object in the interests of passing along properties to the individual scaffolder that we may not explicitly know about (e.g. an additional property to configure the Node.js scaffolder), but that a caller may know about.

src/configureWorkspace/configureDotNetCore.ts
philliphoff: That's a good point; we simply don't support that yet.  It should be something we can do, but it's worthy of its own issue and PR.  I've added #1532 to track that.

src/configureWorkspace/configureDotNetCore.ts
philliphoff: It's to ensure that the path (which may come from a Windows filesystem) uses POSIX separators in the resulting Dockerfile (which most likely targets Linux but, even if Windows, will still be valid).  You'd think that `path.posix.normalize()` would do that but, sadly, it doesn't convert Windows separators to POSIX separators.  I've added a comment.

src/configureWorkspace/scaffolding.ts
karolz-ms: Thanks Phil, this makes much more sense now.

Perhaps it is worth considering adding a comment that describes the intention between `ScaffoldContext` and `ScaffolderContex` (what you said above).

The last missing piece for me is the fact that, in Typescript object literals, `prop` is a shorthand for `prop: prop`, see https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#4.5

philliphoff: Added some more comments.

Issue #1510, opened 2019-12-25
There is no inputting validation for the "port input box"


v-ruizh:
**OS:** All
**Build Version:** [20191221.2](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=13288&view=results)
**Regression:** Not a Regression

**Steps to reproduce:**

1. Open a folder/workspace in VS Code.
2. F1->Docker: Add Docker Files to Workspace...->Node.js->Yes/No.
3. Input the port with "3;;'['']?\><" in the "port input box".

**Expect:**
There is an error like below.
![image](https://user-images.githubusercontent.com/45614674/71440494-3a6b8c00-2738-11ea-9af0-38c80f89127d.png)

**Actual:**
There is no inputting validation for the "port input box".
![image](https://user-images.githubusercontent.com/45614674/71440524-566f2d80-2738-11ea-9e6f-e9ced0ecea5c.png)

allanwright:
I've created #1518 to fix this issue.

PR #1518, opened 2019-12-30
There is no inputting validation for the "port input box" #1510

allanwright:
Input validation exists but its behaviour does not match the validation message:

_Ports must be a comma-separated list of positive integers (1 to 65535), or empty for no exposed port._

The current validation uses a global regex pattern to match on sequences of numbers, but does not care how they are separated. This has the advantage of being able to split the ports using any separator, but it does mean that the validation message isn't always displayed when it should be, which causes confusion when trying to determine what is or isn't valid input. I also found that entering a negative port number successfully validates and returns the port number as a positive value.

I've refactored the splitPorts function and added unit tests to cover different input variations. The behaviour of this function is now much more explicit in what is considers valid input and closely matches the validation message that is displayed.

I also had to export the splitPorts function in order to test it. I'm not sure if that is going to be an issue or if there is another way to achieve the same result without exporting it.


ravipal:
Thank you @allanwright for fixing this issue.

Issue #1461, opened 2019-12-04
[Suggestion] Auto refresh after installing Azure Account Extension


v-dantch:
**OS**: Mac
**Build Version:** [20191203.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=12805)
**Regression:** Not a Regression

**Repro Steps:**
1. Uninstall all dependencies of Azure Account Extension --> Uninstall Azure Account Extension.
2. Use "Connect Registry" button to connect Azure --> Prompt to install Azure Account Extension.
![image](https://user-images.githubusercontent.com/34729092/70120053-98abce80-16a6-11ea-9296-3f1661a1d8a8.png)
3, Install Azure Account Extension.

**Actual**:
Need to refresh manually after installing Azure Account Extension.

**Suggestion:**
Auto refresh after installing Azure Account Extension.

philliphoff:
That seems like a reasonable thing to do.

PR #1520, opened 2019-12-31
Auto refresh registries after installing the Azure ext.

ravipal:
When Azure extension is not installed, the Azure registry shows a node "Install Azure Account Extension...". Even after installing the extension, it still shows "Install Azure Account Extension..." until user manually refreshes the node. Now it is being auto refreshed.

Related to #1461.


src/tree/registries/azure/AzureAccountTreeItem.ts
```diff
@@ -28,4 +30,25 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase implements IR
     public createSubscriptionTreeItem(subContext: ISubscriptionContext): SubscriptionTreeItem {
         return new SubscriptionTreeItem(this, subContext);
     }
+
+    public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
+        const treeItems: AzExtTreeItem[] = await super.loadMoreChildrenImpl(_clearCache, context);
+
+        if (treeItems.length === 1 && treeItems[0].commandId === 'extension.open') {
```
karolz-ms: So this command causes the extension to be open in the gallery, but not necessarily to be installed. Should we have a time-out for the polling loop below, in case the user decides they don't want to install the extension after all?

ravipal: Added 10 min timeout.

src/tree/registries/azure/AzureAccountTreeItem.ts
```diff
@@ -28,4 +30,33 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase implements IR
...
+            // tslint:disable-next-line: no-floating-promises
+            this.refreshNodeWhenAzureExtensionIsInstalled();
```
philliphoff: This could get called multiple times, if the user manually refreshes this node, which means multiple waiters for the installation.  Moreover, if the user manually refreshes the parent node (e.g. the root tree), new instances of this class will be created and this function could then be called multiple times across several instances.  That means it's not enough to track waiting at the instance level; it needs to be statically tracked so there is at most a single waiter.

src/tree/registries/azure/AzureAccountTreeItem.ts
```diff
@@ -28,4 +30,33 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase implements IR
...
+        }
+
+        return treeItems;
+    }
+
+    private async refreshNodeWhenAzureExtensionIsInstalled(): Promise<void> {
+        let isExtensionInstalled: boolean = false;
+        let timedout: boolean = false;
+        const maxTimeout: number = 600000; //10 minutes
+        setTimeout(() => { timedout = true; }, maxTimeout);
```
philliphoff: Ideally this timeout timer will get canceled when an installation happens.

karolz-ms: Yes, if we can actually have a handler for the installation of this particular extension, then we do not need to poll at all. Can we?

philliphoff: Hmm, the [`vscode.extensions.onDidChange`](https://code.visualstudio.com/api/references/vscode-api#extensions) event looks promising.  We'd still need to ensure we're only listening at most once (see above comment), but would be better than polling.

ravipal: Now using event onDidChange. But still each instance need to subscribe to the event.

src/tree/registries/azure/AzureAccountTreeItem.ts
ravipal: Now using event

src/tree/registries/azure/AzureAccountTreeItem.ts
```diff
@@ -28,4 +29,18 @@ export class AzureAccountTreeItem extends AzureAccountTreeItemBase implements IR
     public createSubscriptionTreeItem(subContext: ISubscriptionContext): SubscriptionTreeItem {
         return new SubscriptionTreeItem(this, subContext);
     }
+
+    public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
+        const treeItems: AzExtTreeItem[] = await super.loadMoreChildrenImpl(_clearCache, context);
+
+        if (treeItems.length === 1 && treeItems[0].commandId === 'extension.open') {
+            const extensionUtils: ExtensionUtils = new ExtensionUtils('ms-vscode.azure-account');
```
philliphoff: This will still create `n` numbers of listeners of extension events, for `n` refreshes of this node or parent node, when there should just be, at most, one listener.  One of the tricks will be understanding when a node no longer needs to listen; the easy case is when the same instance is refreshed and the waited-for extension is now installed, but the hard case is when the node itself is thrown away in favor of a brand new node, as I don't think there's a disposal pattern in place for tree view nodes.

That said, I think there should only ever be just *one* of this type of node in place at any given time, so we might use construction of a new node to indicate that the existing node should stop listening (as, presumably, it's being thrown away).

src/utils/extensionUtils.ts
```diff
@@ -0,0 +1,34 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Event, EventEmitter, extensions } from "vscode";
+
+export class ExtensionUtils {
```
philliphoff: `ExtensionListener`?

src/utils/extensionUtils.ts
```diff
@@ -0,0 +1,34 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Event, EventEmitter, extensions } from "vscode";
+
+export class ExtensionUtils {
+    private extensionName: string;
+    private readonly extensionInstalledEmitter: EventEmitter<boolean> = new EventEmitter<boolean>();
```
philliphoff: I believe an `EventEmitter<T>` is disposable, which means this type must also be disposable (and call the emitter's `dispose()` function).

src/utils/extensionUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+        this.extensionName = extensionName;
+        this.subscribeToExtensionChange();
+    }
+
+    private subscribeToExtensionChange(): void {
+        if (this.isExtensionInstalled(this.extensionName)) {
+            this.extensionInstalledEmitter.fire(true);
+        }
+
+        extensions.onDidChange(() => {
```
philliphoff: This returns a disposable object (with which to unsubscribe from the event), which should be disposed of when we no longer care about listening.

src/tree/registries/azure/AzureAccountTreeItem.ts
ravipal: When the node gets refreshes, the same instance is not used instead the old instance is thrown away. AzureAccountTreeItemBase which is the base class for this type implements disposable. But unfortunately when the node/parent gets refreshed, the dispose is not getting called, so can't  use the dispose() method to dispose the listener.

ravipal: Now listening only once.

src/extension.ts
```diff
@@ -131,6 +132,13 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
     });
 }
 
+export async function deactivateInternal(ctx: vscode.ExtensionContext): Promise<void> {
+    await callWithTelemetryAndErrorHandling('docker.deactivate', async (activateContext: IActionContext) => {
+        activateContext.telemetry.properties.isActivationEvent = 'true';
```
philliphoff: Should we consider this an activation event? (I'm not sure what we use this property for in our telemetry analysis.)

src/utils/AzureAccountExtensionListener.ts
```diff
@@ -0,0 +1,52 @@
...
+            AzureAccountExtensionListener.extensionInstalledEmitter.dispose();
+        }
+        AzureAccountExtensionListener.extensionInstalledEmitter = new EventEmitter<boolean>();
+        return this.extensionInstalledEmitter.event;
+    }
+
+    private static subscribeToExtensionsChange(): Disposable {
+        const listener: Disposable = extensions.onDidChange(() => {
+            if (this.isExtensionInstalled(AzureAccountExtensionListener.extensionName)) {
+                // Once the extension is installed, no need to continue listening for the event.
```
philliphoff: We may need to restart listening in the event that the user uninstalls the Azure Account extension.  Perhaps reset `extensionsChangeEventListener` after disposing of `listener`?.

src/extension.ts
ravipal: This property is to indicate whether this telemetry is as part of extension activation and deactivation related Vs user operation.

src/utils/AzureAccountExtensionListener.ts
```diff
@@ -0,0 +1,52 @@
...
+        return extensions.getExtension(AzureAccountExtensionListener.extensionName) !== undefined;
+    }
+
+    public static dispose(): void {
+        if (AzureAccountExtensionListener.extensionInstalledEmitter) {
+            AzureAccountExtensionListener.extensionInstalledEmitter.dispose();
+        }
+        if (AzureAccountExtensionListener.extensionsChangeEventListener) {
+            AzureAccountExtensionListener.extensionsChangeEventListener.dispose();
+        }
```
philliphoff: Reset to `undefined`, so that it's setup to go again in the event that the extension is reactivated?

Issue #1430, opened 2019-11-14
Review execution of container commands via palette


philliphoff:
Many (all?) of the container-related contextual commands are also available via the command palette.  In that scenario, the command will typically display a quick-pick prompt so the user can select a specific container.  However, several of these commands are valid only when a container is in a given state, which means they may fail if the wrong container is selected.

We should review each command to (1) determine whether it makes sense to be available via the command palette and (2) whether the quick-pick prompt should be filtered to show only containers relevant to the command.

v-dantch:
#1576 This issue also reproduces.

PR #1527, opened 2020-01-07
Filter and show only the container groups valid for that command

ravipal:
Container related commands like start and stop is valid for certain containers based on state. If the containers are grouped (default), then the first level tree item picker shows all container groups. Now this is filtered to show the group that has a valid container.

Related to #1430

### DIFF
--- a/src/tree/LocalGroupTreeItemBase.ts
+++ b/src/tree/LocalGroupTreeItemBase.ts
@@ -11,7 +11,7 @@ export abstract class LocalGroupTreeItemBase<TItem extends ILocalItem, TProperty
     public parent: LocalRootTreeItemBase<TItem, TProperty>;
     public group: string;
     private _items: TItem[];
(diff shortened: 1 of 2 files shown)


src/tree/LocalGroupTreeItemBase.ts
```diff
@@ -41,4 +44,15 @@ export abstract class LocalGroupTreeItemBase<TItem extends ILocalItem, TProperty
     public compareChildrenImpl(ti1: AzExtTreeItem, ti2: AzExtTreeItem): number {
         return this.parent.compareChildrenImpl(ti1, ti2);
     }
+
+    public get ChildrenTreeItems(): AzExtTreeItem[] {
```
philliphoff: Nit: `childTreeItems` or `children`?

ravipal: renamed to childTreeItems

Issue #1383, opened 2019-10-30
Error rates: Disallow removing built-in networks


bwateratmsft:
A relatively common issue is users trying to remove the built-in networks (bridge, host, none), which is not allowed by Docker. As a result the attempt fails.

#1175, #1248, and it's a relatively common issue from telemetry (over 36% of all Remove Network attempts fail for this).

PR #1563, opened 2020-01-17
Remove ability to remove built-in networks

stuartthomson:
This relates to issue: https://github.com/microsoft/vscode-docker/issues/1383

There is now a warning message shown if you try to remove a built-in network, and the extension will not actually try to remove them. Additionally, I've added a setting to allow these networks to be hidden from the explorer.


package.json
```diff
@@ -1307,6 +1307,11 @@
                         ]
                     }
                 },
+                "docker.networks.hideBuiltInNetworks": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Hide the built in networks from the explorer"
```
philliphoff: Nit: `built-in`

src/commands/networks/removeNetwork.ts
```diff
@@ -16,6 +16,11 @@ export async function removeNetwork(context: IActionContext, node: NetworkTreeIt
         nodes = await ext.networksTree.showTreeItemPicker(NetworkTreeItem.contextValue, { ...context, canPickMany: true, suppressCreatePick: true });
     }
 
+    if (nodes.some(node => ['bridge', 'host', 'none'].includes(node.networkName))) {
+        ext.ui.showWarningMessage("It's not possible to remove the built-in networks 'bridge', 'host', or 'none'");
```
philliphoff: Nit: avoid contractions in messages shown to users (`'It is not possible...'`).  Alternatively, `"The built-in networks 'bridge', 'host', and 'none' cannot be removed."`

src/commands/networks/removeNetwork.ts
```diff
@@ -16,6 +16,11 @@ export async function removeNetwork(context: IActionContext, node: NetworkTreeIt
         nodes = await ext.networksTree.showTreeItemPicker(NetworkTreeItem.contextValue, { ...context, canPickMany: true, suppressCreatePick: true });
     }
 
+    if (nodes.some(node => ['bridge', 'host', 'none'].includes(node.networkName))) {
```
philliphoff: I wonder if clever use of tree item context values would allow us to filter out the built-in networks from being chosen.  Ideally, they'd be listed but disabled, but I'm not sure the VS Code picker API allows that so just filtering them out completely may be better than warning them after the fact.

We *should* akso be able to remove/disable (preferably the latter) the remove command from the context menus of the built-in nodes.

src/commands/networks/removeNetwork.ts
```diff
@@ -16,6 +16,11 @@ export async function removeNetwork(context: IActionContext, node: NetworkTreeIt
...
+        ext.ui.showWarningMessage("It's not possible to remove the built-in networks 'bridge', 'host', or 'none'");
+        nodes = nodes.filter((node) => !['bridge', 'host', 'none'].includes(node.networkName));
```
philliphoff: As the list of built-in networks is used in at least 3 places, perhaps refactor it as a single constant?

src/commands/networks/removeNetwork.ts
stuartthomson: I have added a new commit which filters them out. I could not figure out how to disable the command but still have it visible (in either case). I agree it would be preferable and perhaps somebody more knowledgeable can figure out if that's even possible.

bwateratmsft: I don't think it is in the current framework. In any case, I'm happy with hiding `Remove` entirely from built-in networks.

package.json
```diff
@@ -1307,6 +1307,11 @@
                         ]
                     }
                 },
+                "docker.networks.hideBuiltInNetworks": {
```
bwateratmsft: Nit: Generally we try to use the positive instead of the negative, i.e. instead of `hideBuiltInNetworks`, default `false`, it would be `showBuiltInNetworks`, default `true`. Negatives can get confusing when they are false (i.e. it's a double-negative).

src/tree/networks/NetworkTreeItem.ts
```diff
@@ -5,20 +5,27 @@
...
+
     private readonly _item: LocalNetworkInfo;
 
     public constructor(parent: AzExtParentTreeItem, itemInfo: LocalNetworkInfo) {
         super(parent);
         this._item = itemInfo;
     }
 
+    public get contextValue(): string {
+        const prefix = builtInNetworks.includes(this._item.networkName) ? 'unremoveable' : 'removeable';
```
bwateratmsft: Should we consider `defaultNetwork` and `customNetwork`, instead of `unremoveableNetwork` and `removeableNetwork`? I'm ok with either, just a thought. @philliphoff as well.

package.json
```diff
@@ -1420,6 +1420,11 @@
                         ]
                     }
                 },
+                "docker.networks.showBuiltInNetworks": {
+                    "type": "boolean",
+                    "default": false,
```
bwateratmsft: IMO the default should be `true` for this.

src/commands/networks/removeNetwork.ts
```diff
@@ -13,11 +14,14 @@ export async function removeNetwork(context: IActionContext, node?: NetworkTreeI
...
         { ...context, suppressCreatePick: true, noItemFoundErrorMessage: 'No networks are available to remove' },
         ext.networksTree,
-        new RegExp(NetworkTreeItem.contextValue, 'i'),
+        NetworkTreeItem.customNetworkRegExp,
         node,
         nodes
     );
-
+    if (nodes.some(n => builtInNetworks.includes(n.networkName))) {
+        await ext.ui.showWarningMessage("The built-in networks 'bridge', 'host', and 'none' cannot be removed.");
```
bwateratmsft: Probably shouldn't await this one as it will hold up until the user clicks X. You can remove the `await` and add `/* eslint-disable-next-line @typescript-eslint/no-floating-promises */` above that to prevent the linter error.

src/tree/networks/NetworkTreeItem.ts
```diff
@@ -4,22 +4,30 @@
...
+
     private readonly _item: LocalNetworkInfo;
 
     public constructor(parent: AzExtParentTreeItem, itemInfo: LocalNetworkInfo) {
         super(parent);
         this._item = itemInfo;
     }
 
+    public get contextValue(): string {
+        const prefix = builtInNetworks.includes(this._item.networkName) ? 'default' : 'custom';
```
bwateratmsft: I think it would be faster to do this:

`return builtInNetworks.includes(this._item.networkName) ? 'defaultNetwork' : 'customNetwork';`

This avoids a string concat (and the associated memory alloc, concat, return, etc.). There are a *ton* of things that reference `contextValue` in both the extension code and VS Code itself, so it would be getting called a lot.

src/commands/networks/removeNetwork.ts
bwateratmsft: Nitpick: by convention at least in this repo, we try to use single-quotes on strings. `'The built-in networks \'bridge\', \'host\', and \'none\' cannot be removed.'`

src/commands/networks/removeNetwork.ts
bwateratmsft: I'll resolve this comment, the current implementation will not show the `Remove` button when right-clicking a built-in network. If you multi-select and do `Remove` on a non-built-in network, it will filter the built-in ones out of the list to prevent errors.

src/commands/networks/removeNetwork.ts
```diff
@@ -13,11 +14,15 @@ export async function removeNetwork(context: IActionContext, node?: NetworkTreeI
     nodes = await multiSelectNodes(
         { ...context, suppressCreatePick: true, noItemFoundErrorMessage: 'No networks are available to remove' },
         ext.networksTree,
-        NetworkTreeItem.contextValue,
+        NetworkTreeItem.customNetworkRegExp,
         node,
         nodes
     );
-
+    if (nodes.some(n => builtInNetworks.includes(n.networkName))) {
```
bwateratmsft: The call to `multiSelectNodes` with `NetworkTreeItem.customNetworkRegExp` might automatically filter off all the built-in networks, making this redundant. I'll investigate.

bwateratmsft:
Thanks for your contribution @stuartthomson!

PR #1564, opened 2020-01-17
Improve extension configuration property descriptions

karolz-ms:
As part of documentation revamp I did a quick pass over our extension property descriptions

- Tried to make the descriptions somewhat more direct and understandable.
- Removed references to default values (these are displayed in a separate column in the extension "Contributions" tab, so putting them in the description is redundant.
- Added list of possible values for enum-type configuration properties. Without them one needs to look at the extension source/manifest to learn what is possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1137,12 +1137,12 @@
                 "docker.explorerRefreshInterval": {
                     "type": "number",
                     "default": 2000,
-                    "description": "Explorer refresh interval, default is 2000ms"
+                    "description": "Docker view refresh interval (milliseconds)"
                 },
                 "docker.containers.groupBy": {
                     "type": "string",
                     "default": "None",
-                    "description": "The property to use when grouping containers.",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1137,12 +1137,12 @@
...
                     "type": "number",
                     "default": 2000,
-                    "description": "Explorer refresh interval, default is 2000ms"
+                    "description": "Docker view refresh interval (milliseconds)"
                 },
                 "docker.containers.groupBy": {
                     "type": "string",
                     "default": "None",
-                    "description": "The property to use when grouping containers.",
+                    "description": "The property to use to group containers in Docker view: ContainerId, ContainerName, CreatedTime, FullTag, ImageId, Networks, Ports, Registry, Repository, RepositoryName, RepositoryNameAndTag, State, Status, Tag, or None",
```
ravipal: None is not specified in the enum value below. Is this an implicit value? Line 1290 defines a enum with None value.

ravipal: Never mind. It is defined, but I was looking at the bottom of the list.

Issue #1528, opened 2020-01-07
Differentiate unexpected error/bug Vs expected error/user error


ravipal:
In general, any error thrown during command execution is reported as error and 'Report an issue' button is offered to user. But some of the errors are expected like user tries to delete an image that is in use or try to start/stop container but no container matches the required state exists and so on.
We need to a better way to categorize the error and offer 'Report an issue' button only on unknown/unexpected error.

PR #1565, opened 2020-01-17
Provide a useful error message and suppress report issue button

ravipal:
If no matching item found while selecting an item in tree item picker, then a generic error message 'No resource found' with 'report issue' button is offered to user.
Now a customized message is given (like no running container are available to stop) and the 'report issue' button is not shown to user. The azuretools extension behavior is changed to not show 'report issue' when a custom message is provided (refer https://github.com/microsoft/vscode-azuretools/pull/644).

related to #1528

### DIFF
--- a/src/commands/containers/browseContainer.ts
+++ b/src/commands/containers/browseContainer.ts
@@ -59,7 +59,7 @@ export async function browseContainer(context: IActionContext, node?: ContainerT
         node = await captureBrowseCancelStep('node', telemetryProperties, () =>
             ext.containersTree.showTreeItemPicker<ContainerTreeItem>(ContainerTreeItem.runningContainerRegExp, {
                 ...context,
-                noItemFoundErrorMessage: 'No running containers are available to open in browser.'
+                noItemFoundErrorMessage: 'No running containers are available to open in a browser'
             }));
     }
 
--- a/src/commands/containers/browseContainer.ts
+++ b/src/commands/containers/browseContainer.ts
@@ -56,7 +56,11 @@ export async function browseContainer(context: IActionContext, node?: ContainerT
 
     if (!node) {
         /* eslint-disable-next-line @typescript-eslint/promise-function-async */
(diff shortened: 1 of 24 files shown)


src/commands/containers/browseContainer.ts
```diff
@@ -56,7 +56,11 @@ export async function browseContainer(context: IActionContext, node?: ContainerT
 
     if (!node) {
         /* eslint-disable-next-line @typescript-eslint/promise-function-async */
-        node = await captureBrowseCancelStep('node', telemetryProperties, () => ext.containersTree.showTreeItemPicker<ContainerTreeItem>(ContainerTreeItem.runningContainerRegExp, context));
+        node = await captureBrowseCancelStep('node', telemetryProperties, () =>
+            ext.containersTree.showTreeItemPicker<ContainerTreeItem>(ContainerTreeItem.runningContainerRegExp, {
+                ...context,
+                noItemFoundErrorMessage: 'No running containers are available to open in browser.'
```
karolz-ms: ... in a browser?

src/commands/containers/inspectContainer.ts
```diff
@@ -9,7 +9,10 @@ import { ContainerTreeItem } from "../../tree/containers/ContainerTreeItem";
 
 export async function inspectContainer(context: IActionContext, node?: ContainerTreeItem): Promise<void> {
     if (!node) {
-        node = await ext.containersTree.showTreeItemPicker<ContainerTreeItem>(ContainerTreeItem.allContextRegExp, context);
+        node = await ext.containersTree.showTreeItemPicker<ContainerTreeItem>(ContainerTreeItem.allContextRegExp, {
+            ...context,
+            noItemFoundErrorMessage: 'No containers are available to inspect'
```
karolz-ms: Nit: some error messages end with a period, some don't. Let's be consistent. Suggest no period

ravipal: fixed

Issue #1585, opened 2020-01-24
Update CI VM images


philliphoff:
On March 23, 2020, Azure DevOps is retiring several pipeline VM images. In particular, the MacOS 10.13 image used by our Mac pipeline.  We need to update our pipelines to use the latest VM images before that date.

PR #1586, opened 2020-01-24
Update vm images

EricJizbaMSFT:
Got an email saying the images we're using will be removed in March. Other option is to hardcode to a later image, but using latest seemed easier and _should_ work fine.


philliphoff:
Resolves #1585.

philliphoff:
Agree, `latest` seems best.  Will merge once 0.10.0 goes out the door.

PR #1603, opened 2020-02-04
use default files in docker-compose

### LINKED ISSUE #370: Compose - Don't prompt if only one yml file in CWD
Possibly the same feature request as #223, but relating to workspaces which contain a single compose file.

If your project contains only a single `docker-compose.yml` file, could the extension please default to just running compose against that file, instead of prompting for which file you want to run. 

This is by far my favourite VSCode extension, but it's very annoying that every time I have to run a compose command via the command palette, that I have to endure 2 levels of prompting before the command runs. This would be a nice quality of life improvement for many small projects.

ravipal:
If the workspace container only the default docker-compose files 
1. Just one file and it is named 'docker-compose.y(a)ml'
2. Two files docker-compose.y(a)ml and docker-compose.override.y(a)ml

then -f is not used and let the docker-compose pick these files.

Fixes #370

### DIFF
--- a/src/commands/compose.ts
+++ b/src/commands/compose.ts
@@ -78,11 +78,15 @@ async function compose(commands: ('up' | 'down')[], message: string, dockerCompo
 
     terminal.sendText(`cd "${folder.uri.fsPath}"`);
     for (let command of commands) {
-        const fileParameter: string = selectedItems.map(i => `-f "${i.file}"`).join(' ');
-        terminal.sendText(command.toLowerCase() === 'up' ? `docker-compose ${fileParameter} ${command} ${detached} ${build}` : `docker-compose ${fileParameter} ${command}`);
+        if (selectedItems.length === 0) {
+            terminal.sendText(command.toLowerCase() === 'up' ? `docker-compose ${command} ${detached} ${build}` : `docker-compose ${command}`);
+        } else {
+            selectedItems.forEach((item: Item) => {
+                terminal.sendText(command.toLowerCase() === 'up' ? `docker-compose -f "${item.file}" ${command} ${detached} ${build}` : `docker-compose -f "${item.file}" ${command}`);
+            });
+        }
         terminal.show();
     }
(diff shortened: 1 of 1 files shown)


src/commands/compose.ts
```diff
@@ -61,7 +61,14 @@ async function compose(commands: ('up' | 'down')[], message: string, dockerCompo
...
 
         const items: vscode.QuickPickItem[] = computeItems(folder, uris);
-        selectedItems = [<Item>await ext.ui.showQuickPick(items, { placeHolder: `Choose Docker Compose file ${message}` })];
+
+        if ((items.length === 1 && isDefaultDockerComposeFile(items[0].label))
+            || (items.length === 2 && items.some(i => isDefaultDockerComposeFile(i.label)) && items.some(i => isDefaultDockerComposeOverrideFile(i.label)))) {
+            // if the current set of docker files contain only docker-compose.yml or docker-compose.yml with override file,
+            // don't ask user for a docker file and let docker-compose automatically pick these files.
+        } else {
+            selectedItems = [<Item>await ext.ui.showQuickPick(items, { placeHolder: `Choose Docker Compose file ${message}` })];
```
karolz-ms: From this prompt it is not clear for the user that files will be used in order of selection, but it does matter for Docker Compose. How about

"Select Docker Compose file(s). Files will be used in the order in which they were selected."

It is a bit verbose, but I cannot think of anything better/clearer 🤔

ravipal: The compose command can be executed in 2 ways. Right click on the docker-compose*.yml files and choose the context menu or using command palette.
At this time multi selection is allowed only during context menu and not during command palette execution. The order of files selected are preserved while executing from context menu.

kzadora: Got it, thanks. Please disregard my suggestion then

ravipal: updated the description to make it clear.

PR #1611, opened 2020-02-10
Python debugging support

haniamr:
This PR adds debugging support to Python apps (django, flask and general). 
Included in this change:

- Scaffolding for Docker artifacts (Dockerfile, .dockerignore, docker-compose*.yml) for the different Python project types
- Scaffolding Python requirements file
- Changes for the naming and ordering of the platforms
- Scaffold a "Python Attach" configuration if compose is selected
- Ensure that only the "debugging" port is assigned to a host port when scaffolding the compose debug file (other ports will be randomly assigned by Docker)

There is known issue with the Python debugger that it doesn't retry if the connection fails, so we have to do a workaround to know if the debugger (running in the container) is ready to receive connections. The idea is to instruct the debugger to "log" to a certain directory, then look for a line that says "Starting server daemon on".

We're mounting a r/w volume with a temp directory for the logs, so we can read the file on the host machine without an expensive "docker exec/cp" operation. We check every half a second to a maximum of 10 seconds.

The workaround is not optimal, since it requires us to run our logic in "resolveLaunchConfiguration", which is something that we went away from with the Task-based model. We still execute the tasks but manually before resolving the launch configuration to be able to wait for the debugger (tried implicit tasks but it didn't work).

Some open issues for improving the Python debugger and possibly have an API for the Python extension:
https://github.com/microsoft/vscode-python/issues/7331
https://github.com/microsoft/vscode-python/issues/7282


package.json
```diff
@@ -626,6 +627,60 @@
                                         ]
                                     }
                                 }
+                            },
+                            "python": {
```
karolz-ms: We need to schedule work to update the documentation. We probably need a Python "tutorial" topic plus update the documentation for build & debug tasks.

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+        args = args || [];
+
+        return `/pydbg/ptvsd --host ${options.host} --port ${options.port} ${options.wait ? "--wait" : ""} ${fullTarget} ${args.join(" ")}`;
+    }
+
+    export function getLauncherFolderPath(): string {
+        const pyExt = vscode.extensions.getExtension("ms-python.python");
+
+        if (!pyExt) {
+            throw new Error("For debugging Python apps in a container to work, the Python extension must be installed.");
```
rcjsuen: I suggest either stating the name and/or the id (`ms-python.python` in this error message? When I search for "python" in the [VS Code Marketplace](https://marketplace.visualstudio.com/vscode), there are quite a few hits out there.

src/configureWorkspace/configurePython.ts
```diff
@@ -3,68 +3,224 @@
...
+        prompt: launchFilePrompt.prompt,
+        value: launchFilePrompt.defaultFile,
+        validateInput: (value: string): string | undefined => { return value && value.trim().length > 0 ? undefined : "Enter a valid Python file path/module." }
+    };
+
+    const file = await ext.ui.showInputBox(opt);
+
+    // If the input has the .py extension, then assume it is a file.
+    // TODO: is there a more robust way to check?
+    if (file.toLowerCase().endsWith(".py")) {
```
philliphoff: [path.extname(name)](https://nodejs.org/dist/latest-v12.x/docs/api/path.html#path_path_extname_path)?

src/configureWorkspace/configurePython.ts
```diff
@@ -3,68 +3,224 @@
...
+    files.forEach(file => {
+    // Remove multiple empty lines with single empty lines, as might be produced
+    // if $expose_statements$ or another template variable is an empty string.
+        file.contents = file.contents
+            .replace(/(\r\n){3,4}/g, "\r\n\r\n")
+            .replace(/(\n){3,4}/g, "\n\n");
+    });
+
+    if (context.initializeForDebugging) {
+        await initializeForDebugging(context, path.join(outputFolder, "Dockerfile"), ports, generateComposeFiles, launchFile, projectType);
```
philliphoff: Nit: (here and elsewhere) prefer single-quotes (`'`) to double-quotes (`"`) for strings.

package.json
haniamr: Thanks Karol. Absolutely, I'll work with Uche to get that work tracked.

src/tasks/python/PythonExtensionHelper.ts
haniamr: Thanks for the feedback. I agree the message should include more guidance.

package.json
```diff
@@ -1100,6 +1161,67 @@
                                 "description": "The path to the package.json for the application."
                             }
                         }
+                    },
+                    "python": {
+                        "oneOf": [
```
philliphoff: There may be a clever way of using `oneOf` and `required` to eliminate the repetition in these properties.

src/configureWorkspace/configurePython.ts
```diff
@@ -3,68 +3,224 @@
...
-};
+interface LaunchFilePrompt {
+    prompt: string,
+    defaultFile: string
+}
 
-function genDockerFile(serviceNameAndRelativePath: string, platform: string, os: string | undefined, ports: number[], { cmd, author, version, artifactName }: Partial<PackageInfo>): string {
-    let exposeStatements = getExposeStatements(ports);
+const defaultLaunchFile: Map<PythonProjectType, LaunchFilePrompt> = new Map<PythonProjectType, LaunchFilePrompt>([
+    ["django", { prompt: "Enter the relative path to the application (e.g. manage.py)", defaultFile: "manage.py" }],
```
philliphoff: Nit: make prompts more consistent (parens vs. no parens).  Also, we've been encouraged not to use "e.g." in user text, preferring "for example".

src/configureWorkspace/configurePython.ts
```diff
@@ -3,68 +3,224 @@
...
+    const scaffoldContext: DockerDebugScaffoldContext = {
+        folder: context.folder,
+        platform: "python",
+        actionContext: context,
+        dockerfile: dockerfile,
+        generateComposeTask: generateComposeFiles,
+        ports: ports
+    }
+
+    const pyOptions: PythonScaffoldingOptions = {
```
philliphoff: Nit:

```js
const pyOptions: PythonScaffoldingOptions = {
    projectType,
    target
};
```

src/debugging/DockerDebugScaffoldingProvider.ts
```diff
@@ -9,17 +9,25 @@ import { DockerBuildTaskDefinition } from '../tasks/DockerBuildTaskProvider';
...
 import { addDebugConfiguration, DockerDebugScaffoldContext } from './DebugHelper';
 import { DockerDebugConfiguration } from './DockerDebugConfigurationProvider';
 import { netCoreDebugHelper, NetCoreDebugScaffoldingOptions } from './netcore/NetCoreDebugHelper';
 import { nodeDebugHelper } from './node/NodeDebugHelper';
+import { pythonDebugHelper } from './python/PythonDebugHelper';
 
 export type NetCoreScaffoldingOptions = NetCoreDebugScaffoldingOptions | NetCoreTaskScaffoldingOptions;
+export interface PythonScaffoldingOptions {
+    projectType?: PythonProjectType;
+    target?: PythonFileTarget | PythonModuleTarget
```
philliphoff: To reduce repetition, consider: 

```js
type PythonTarget = PythonFileTarget | PythonModuleTarget;
```

src/configureWorkspace/configurePython.ts
```diff
@@ -3,68 +3,224 @@
...
+    const rootFolderPath: string = context.rootFolder;
+    const outputFolder = context.outputFolder ?? rootFolderPath;
+
+    const generateComposeFiles = await context.captureStep("compose", quickPickGenerateComposeFiles)();
+    const projectType = getPythonProjectType(context.platform);
+
+    const defaultPort = PythonDefaultPorts.get(projectType);
+    let ports = [];
+
+    if (defaultPort) {
```
philliphoff: We only prompt the user if there's a default port for a project type?  That is, we don't prompt for project types that don't have a well known port?  Isn't that exactly when we *should* prompt?

src/configureWorkspace/configurePython.ts
```diff
@@ -3,68 +3,224 @@
...
+        ports = await context.promptForPorts([ defaultPort ]);
+    }
+
+    const launchFile = await context.captureStep("pythonFile", promptForLaunchFile)(projectType);
+    const dockerFileContents = genDockerFile(serviceName, launchFile, projectType, ports);
+
+    const files: ScaffoldFile[] = [
+        { fileName: 'Dockerfile', contents: dockerFileContents, open: true },
+        { fileName: '.dockerignore', contents: genCommonDockerIgnoreFile(context.platform) },
+        { fileName: 'requirements.txt', contents: genRequirementsFile(projectType) }
```
philliphoff: Should we always generate this file?  If the user already has this file, isn't it likely it that it has the correct content for the project?  While we *do* prompt the user to overwrite any scaffolded file, that's under the assumption that we may be adding specific content for the purposes of the building/debugging flows and the user should be generally ok with overwriting it.  I'm not sure the same can be said of this file, though.

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+                ],
+                projectType: options.projectType
+            }
+        }];
+
+        // If we generated compose files, then we should generate a Python attach configuration.
+        if (context.generateComposeTask) {
+            configs.push(
+                {
+                    name: "Python: Remote Attach",
```
philliphoff: We use colon here in the name but not in the other configuration?

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+
+        return configs;
+    }
+
+    public async resolveDebugConfiguration(context: DockerDebugContext, debugConfiguration: PythonDockerDebugConfiguration): Promise<ResolvedDebugConfiguration | undefined> {
+        const containerName = inferContainerName(debugConfiguration, context, context.folder.name);
+
+        // Since Python is a special case, we need to ensure the container is removed before attempting to resolve
+        // the debug configuration.
+        try {
```
philliphoff: Do we need to test that `removeContainerAfterDebug` is set?

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+            await this.cliDockerClient.removeContainer(containerName, { force: true });
+        } catch {}
+
+        const debuggerLogFilePath = PythonExtensionHelper.getDebuggerLogFilePath(context.folder.name);
+        await fse.remove(debuggerLogFilePath);
+
+        let debuggerReadyPromise = new Promise((resolve) => resolve());
```
philliphoff: `let debuggerReadyPromise = Promise.resolve()`?

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+        if (!prelaunchTaskName) {
+            return undefined;
+        }
+
+        const tasks = await vscode.tasks.fetchTasks();
+
+        if (tasks) {
+            const results = tasks.filter(t => t.name.localeCompare(prelaunchTaskName) === 0);
+
+            if (results && results.length > 0) {
```
philliphoff: Nit: `filter()` always returns a valid (but just potentially empty) array.

src/configureWorkspace/configurePython.ts
haniamr: The thinking is that for projects under the "general" category, we shouldn't assume a port has to be exposed (i.e. it could be a console-like app). While for django and flask, there has to be a port since it's a web app essentially.

Initially, I had it prompting for a port all the time, but based on one of the feedback meetings there was a vote to not do that for "general" and the user can easily expose a port if needed.

src/configureWorkspace/configurePython.ts
haniamr: That was one of the requirements, for covering the "new project and F5" scenario there has to be a requirements file. If the user happens to have one, then they would probably say "no" to our prompt of replacing it.

So from our side, instead of trying to get the actual project dependencies (i.e. do a "pip freeze" command), we agreed it is safer just to have a minimal list of requirements for the known frameworks, given we explicitly mention that file in the Dockerfile.

src/debugging/python/PythonDebugHelper.ts
haniamr: The thinking is that for the Docker configuration, the vote was not to use the colon. However, for the Python attach one, that is how it gets generated by default from the extension so we thought we'd use their defaults.

src/debugging/python/PythonDebugHelper.ts
haniamr: At this point, we're removing the container _before_ debugging not after. Which is the same logic we have here: https://github.com/microsoft/vscode-docker/blob/22abd6cf00ad132c2f9af0af3768eb1b6ec9a681/src/debugging/DockerDebugConfigurationProvider.ts#L109

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+    export function getDebuggerLogFilePath(folderName: string) : string {
+    // The debugger generates the log file with the name in this format: ptvsd-{pid}.log,
+    // So given that we run the debugger as the entry point, then the PID is guaranteed to be 1.
+        return path.join(os.tmpdir(), folderName, "ptvsd-1.log");
+    }
+
+    export async function ensureDebuggerReady(prelaunchTask: vscode.Task, debuggerSemaphorePath: string, containerName: string, cliDockerClient: CliDockerClient): Promise<void> {
+        // tslint:disable-next-line:promise-must-complete
+        return new Promise((resolve, reject) => {
+            vscode.tasks.onDidEndTask(async e => {
```
philliphoff: Should unsubscribe from the event when the debugger's attached or on timeout.

src/configureWorkspace/configurePython.ts
haniamr: The idea is to be similar to the prompts in the Python extension, so as to make the experience more familiar to the user. I'll chat with Uche on that since he suggested the format above.

src/tasks/python/PythonExtensionHelper.ts
philliphoff: Should this handler have a `try`/`catch` block to ensure `reject()` is called if an inspect or file read operation fails?

haniamr: Which timeout? You mean the 10 seconds or a different one? I wanted to have a way to unsubscribe but the thing is, when should we call this a failure given the prelaunch task could be executing N number of other tasks in unknown time?

haniamr: Yes, makes perfect sense.

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+                    }
+
+                    const maxRetriesCount = 20;
+                    let retries = 0;
+                    let created = false;
+
+                    // Look for the magic string below in the log file with a retry every 0.5 second for a maximum of 10 seconds.
+                    // TODO: Should be gone as soon as the retry logic is part of the Python debugger/extension.
+                    while (++retries < maxRetriesCount && !created) {
+                        if (fse.existsSync(debuggerSemaphorePath)) {
```
philliphoff: Any reason for the synchronous functions over the asynchronous ones?

haniamr: The only reason is that the asynchronous version accepts a callback as a parameter and returns void. So I couldn't find a way to "wait" for it's completion in a non-messy way, thoughts?

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+                        reject("Failed to attach the debugger within the alotted timeout.");
+                    }
+                }});
+        })
+    }
+
+    export function getRemotePtvsdCommand(target: PythonFileTarget | PythonModuleTarget, args?: string[], options?: DebugLaunchOptions): string {
+        let fullTarget: string;
+
+        if ((target as PythonFileTarget).file) {
```
philliphoff: Are the casts needed?  Usually TS is smart enough to notice that if a property exclusive to a type exists, it then "becomes" that type.  Sometimes a ["type guard"](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-the-in-operator) is used to help TS identify a type.  You may be able to do something like:

```typescript
if ('file' in target) {
  // Target is now of type PythonFileTarget
} else if ('module' in target) {
  // Target is now of type PythonModuleTarget
} else {
  // Target is of unknown type
}
```

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+            fullTarget = (target as PythonFileTarget).file;
+        } else if ((target as PythonModuleTarget).module) {
+            fullTarget = `-m ${(target as PythonModuleTarget).module}`;
+        } else {
+            throw new Error("One of either module or file must be provided.");
+        }
+
+        options = options || {};
```
philliphoff: Note: we can now use `??` instead of `||` in many (most?) circumstances.  (Though note the implications around, for example, the empty string `''` being used as to indicate the default value.

src/tasks/python/PythonExtensionHelper.ts
philliphoff: Alternatively, ensure that we display an error notification with an action that takes the user to the extension in the marketplace (or the extension in the extensions tab, which I believe there's an API for).

src/tasks/python/PythonExtensionHelper.ts
bwateratmsft: Isn't it `?.` most of the time?

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+
+        const debuggerPath = path.join(pyExt.extensionPath, "pythonFiles", "lib", "python");
+        const oldDebugger = path.join(debuggerPath, "old_ptvsd");
+        const newDebugger = path.join(debuggerPath, "new_ptvsd");
+
+        // Always favor the old_ptvsd debugger since it will work in all cases.
+        // If it is not found, then look for the new instead.
+        // TODO: This should be revisited when the Python extension releases the new debugger since it might have a different name.
+
+        if (fse.existsSync(oldDebugger)) {
```
philliphoff: Can we use the async forms?

src/tasks/python/PythonExtensionHelper.ts
philliphoff: Well, yes, we can use `?.` now too, and they can often be used together (e.g. `obj?.prop ?? defaultValue`).

src/tasks/python/PythonTaskHelper.ts
```diff
@@ -0,0 +1,193 @@
...
+    wait?: boolean;
+    debugPort?: number
+}
+
+export interface PythonRunTaskDefinition extends DockerRunTaskDefinitionBase {
+    python?: PythonTaskRunOptions;
+}
+
+export class PythonTaskHelper implements TaskHelper {
+    private static readonly defaultLabels: { [key: string]: string } = {
```
philliphoff: We have some common handling for default labels in `TaskDefinitionBase`.

src/tasks/python/PythonExtensionHelper.ts
haniamr: Same comment about it accepting a callback and returns void instead of a Promise, plus for some reason "fse.exists()" is deprecated.

src/tasks/python/PythonTaskHelper.ts
```diff
@@ -0,0 +1,193 @@
...
+            context,
+            context.folder.name
+        );
+
+        runOptions.containerName = runOptions.containerName || getDefaultContainerName(context.folder.name);
+
+        const dbgLogsFolder = path.join(os.tmpdir(), context.folder.name);
+
+        // The debugger will complain if the logs directory does not exist.
+        if (!fse.existsSync(dbgLogsFolder)) {
```
philliphoff: Async versions?

src/tasks/python/PythonTaskHelper.ts
```diff
@@ -0,0 +1,193 @@
...
+        return [{
+            type: "docker-run",
+            label: "docker-run: debug",
+            dependsOn: ["docker-build"],
+            python: runOptions
+        }];
+    }
+
+    private inferVolumes(runOptions: DockerRunOptions, launcherFolder: string, dbgLogsFolder: string): DockerContainerVolume[] {
+        const volumes: DockerContainerVolume[] = [];
```
philliphoff:
```typescript
const volumes = runOptions?.volumes ? [...runOptions.volumes] : [];
```
or
```typescript
const volumes = [...(runOptions?.volumes ?? [])];
```

...I think.  :-)

philliphoff: Similarly for ports below.

src/tasks/python/PythonExtensionHelper.ts
haniamr: It makes sense to use `??` in the above case where we have an object, but wouldn't `||` still be better for strings? Example:
`buildOptions.context = buildOptions.context || '${workspaceFolder}';`

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+            if (!task) {
+                throw new Error(`Unable to find the prelaunch task with the name: ${debugConfiguration.preLaunchTask}`);
+            }
+
+            debugConfiguration.preLaunchTask = undefined;
+
+            /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
+            vscode.tasks.executeTask(task);
+
+            debuggerReadyPromise = PythonExtensionHelper.ensureDebuggerReady(task, debuggerLogFilePath, containerName, this.cliDockerClient);
```
philliphoff: Since this subscribes to the task end event, should it be created *before* the task starts executing?

src/configureWorkspace/configurePython.ts
philliphoff: Ok, I'll buy that.  We can see if we get any feedback from users on that.

src/configureWorkspace/configurePython.ts
philliphoff: Totally understand generating the file if it doesn't exist; I'm more concerned about the case where it's already there.  In that case, what we generate is probably not correct--that is, won't at least match what's already there--even if the user was ok with us overwriting it.

src/debugging/python/PythonDebugHelper.ts
philliphoff: Yeah, I get that and, if we were generating them one-by-one, that would be reasonable.  However, we generate these at the *same* time and, if they're the only ones the user ever adds to `launch.json`, it just looks odd to me.

src/tasks/python/PythonExtensionHelper.ts
philliphoff: I was referring to the 10s timeout--basically when `ensureDebuggerReady()` "completes", whether successfully or not.

src/configureWorkspace/configurePython.ts
haniamr: I see your point, do you suggest we look if it exists, find if the minimum requirements are there and add them if not? It would be a good point to bring up in our meeting with Dan today since I'm sure he'll have valuable feedback.

src/debugging/python/PythonDebugHelper.ts
haniamr: Yep, make sense.

src/tasks/python/PythonExtensionHelper.ts
philliphoff: `fs-extra` has overloads for most functions; if you pass a callback they return `void` and act as a traditional Node asynchronous functions.  But, if you *don't* pass a callback, they return a `Promise` which can then be awaited.

package.json
```diff
@@ -626,6 +627,60 @@
...
+                                                    "description": "The container path."
+                                                }
+                                            }
+                                        }
+                                    },
+                                    "justMyCode": {
+                                        "type": "boolean",
+                                        "description": "Debug only user-written code."
+                                    },
+                                    "django": {
```
bwateratmsft: Should `django` / `jinja` be an enum rather than two booleans? Or can both be true at the same time?

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+        await fse.remove(debuggerLogFilePath);
+
+        let debuggerReadyPromise = new Promise((resolve) => resolve());
+        if (debugConfiguration.preLaunchTask) {
+            // There is this limitation with the Python debugger where we need to ensure it's ready before allowing VSCode to attach,
+            // if attach happens too soon then it will fail silently. The workaround here is to set the preLaunchTask to undefined,
+            // then execute it ourselves with a listener to when it is finished, then wait for the debugger to be ready and return
+            // the resolved launch configuration.
+
+            const task = await this.tryGetPrelaunchTask(debugConfiguration.preLaunchTask);
```
bwateratmsft: Nit: tryGetPre**L**aunchTask?

src/tasks/python/PythonExtensionHelper.ts
philliphoff: Depends on whether you want `''` to indicate "use the default value" or "literally use an empty string".  If the former, then use `||`.  If the latter, use `??`.  The former has traditionally been used as it was just a consequence of using the `||` form, which was too convenient *not* to use and the empty string edge case was just glossed over given that in both JS and JSON, 'undefined' is far more likely to be used for unspecified strings rather than the empty string.

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+              pattern: this.getServerReadyPattern(projectType),
+              uriFormat: "%s://localhost:%s"
+          },
+          true);
+
+        // These properties are required by the old debugger, should be changed to normal properties in the configuration
+        // as soon as the new debugger is released to 100% of the users.
+        const debugOptions = ["FixFilePathCase", "RedirectOutput", "ShowReturnValue"];
+
+        if (this.localOsProvider.os === "Windows") {
```
bwateratmsft: Since `localOsProvider` and most of its siblings will eventually be removed or refactored, I used the `os` package, along with `os.platform() === 'win32'` etc. for most of these checks. Should we do the same here?

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    private getServerReadyPattern(projectType: PythonProjectType) : string | undefined {
+        switch (projectType) {
+            case "django":
+                return "Starting development server at (https?://\\S+|[0-9]+)";
+            case "flask":
+                return "Running on (https?://\\S+|[0-9]+)";
+            default:
+                return undefined
```
bwateratmsft: Nit: missing semicolon at the end

src/configureWorkspace/configurePython.ts
philliphoff: As a start, I'd just check whether it exists and, if so, *not* scaffold a new one, and not be concerned with its contents (for now).  That is, assume that, if it exists, it likely has all of the content needed.

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export namespace PythonExtensionHelper {
+    export interface DebugLaunchOptions {
+        host?: string;
+        port?: number;
+        wait?: boolean;
+    }
+
+    export function getDebuggerEnvironmentVars(): { [key: string]: string } {
+        return { ["PTVSD_LOG_DIR"]: `/dbglogs` };
```
bwateratmsft: Nit: I think this should read `return { 'PTVSD_LOG_DIR': '/dbglogs' };`. In any case, can use single quotes, but I don't think `PTVSD_LOG_DIR` needs to be boxed.

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    export function getDebuggerLogFilePath(folderName: string) : string {
+    // The debugger generates the log file with the name in this format: ptvsd-{pid}.log,
```
bwateratmsft: Nit: indentation

src/tasks/python/PythonExtensionHelper.ts
bwateratmsft: I think it's `fse.pathExists(...)` rather than `fse.exists(...)`

src/tasks/python/PythonExtensionHelper.ts
bwateratmsft: Same as above, I think it's `fse.pathExists()`

src/tasks/python/PythonTaskHelper.ts
```diff
@@ -0,0 +1,193 @@
...
+export class PythonTaskHelper implements TaskHelper {
+    private static readonly defaultLabels: { [key: string]: string } = {
+        "com.microsoft.created-by": "visual-studio-code"
+    };
+
+    public async getDockerBuildOptions(context: DockerBuildTaskContext, buildDefinition: DockerBuildTaskDefinition): Promise<DockerBuildOptions> {
+        const buildOptions = buildDefinition.dockerBuild;
+
+        /* eslint-disable no-template-curly-in-string */
+        buildOptions.context = buildOptions.context || "${workspaceFolder}";
```
bwateratmsft: Nit: single quotes

src/tasks/python/PythonTaskHelper.ts
```diff
@@ -0,0 +1,193 @@
...
+
+        const target: PythonFileTarget | PythonModuleTarget = helperOptions.file
+            ? { file: helperOptions.file, }
+            : { module: helperOptions.module };
+
+        const launcherCommand : string = PythonExtensionHelper.getRemotePtvsdCommand(
+            target,
+            helperOptions.args,
+            {
+                host: "0.0.0.0",
```
bwateratmsft: Nit: single quotes

src/tasks/python/PythonTaskHelper.ts
```diff
@@ -0,0 +1,193 @@
...
+        runOptions.entrypoint = runOptions.entrypoint || "python";
+        runOptions.command = runOptions.command || launcherCommand;
+
+        runOptions.env = this.addDebuggerEnvironmentVars(runOptions.env);
+        runOptions.portsPublishAll = runOptions.portsPublishAll || true;
+
+        return runOptions;
+    }
+
+    public async provideDockerBuildTasks(context: DockerTaskScaffoldContext): Promise<DockerBuildTaskDefinition[]> {
```
bwateratmsft: Nit: the other platforms have `provideDocker*Tasks` above the `getDocker*Options`. Conceptually, you scaffold before you run.

src/tasks/python/PythonTaskHelper.ts
```diff
@@ -0,0 +1,193 @@
...
+            },
+            {
+                localPath: dbgLogsFolder,
+                containerPath: "/dbglogs",
+                permissions: "rw"
+            }];
+
+        dbgVolumes.map(dbgVol => {
+            if (volumes.find(volume => volume.containerPath === dbgVol.containerPath) === undefined) {
+                volumes.push(dbgVol);
```
bwateratmsft: Over in `NetCoreTaskHelper` I have an `addVolumeWithoutConflicts` which will add volumes to a list, while never overriding something the user specified. Should we refactor that method out of `NetCoreTaskHelper` and use it here?

src/utils/pythonUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+            ];
+        default:
+            return undefined;
+    }
+}
+
+export function getPythonProjectType(platform: Platform): PythonProjectType | undefined {
+    switch (platform) {
+        case 'Python: Django':
+            return "django";
```
bwateratmsft: Nit: single quotes

package.json
haniamr: Technically they both can be true at the same time, no idea how will that behave though. I followed the same syntax in the Python extension just to ensure parity.

src/configureWorkspace/configure.ts
```diff
@@ -57,8 +57,14 @@ export function getExposeStatements(ports: number[]): string {
     return ports ? ports.map(port => `EXPOSE ${port}`).join('\n') : '';
 }
 
-export function getComposePorts(ports: number[]): string {
-    return ports && ports.length > 0 ? '    ports:\n' + ports.map(port => `      - ${port}:${port}`).join('\n') : '';
+export function getComposePorts(ports: number[], debugPort?: number): string {
+    let portMappings: string[] = ports.map(port => `      - ${port}`);
```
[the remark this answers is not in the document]
haniamr: How about `ports?.map(port => `      - ${port}`)`? Since merging both in the ports array won't really work given we want the debugPort to be mapped to the same host port.

src/debugging/python/PythonDebugHelper.ts
haniamr: Sure, if it will be removed then yeah it makes sense.

src/configureWorkspace/configurePython.ts
haniamr: I don't have much preference to break the tie, so which style is more consistent with the repo?

src/tasks/python/PythonExtensionHelper.ts
haniamr: Cool, for strings I left it as "||" and for objects I changed it to "??".

src/configureWorkspace/configurePython.ts
haniamr: I believe only attempt to create it if it doesn't exist makes sense to me. If the user has a requirements.txt file then it would definitely more complete than the one we generate. 

I'll go ahead with that approach and bring it to discussion later with Uche.

src/configureWorkspace/configurePython.ts
haniamr: Alright, I looked in a random set of files in the repo and it seems the longer form is more dominant. Sorry, Phil :).

src/configureWorkspace/configurePython.ts
haniamr: I added the parentheses for consistency, but will keep the wording till we get Uche's feedback.

src/tasks/python/PythonExtensionHelper.ts
haniamr: I changed that. to Phil's suggestion, by showing an error message with an action that takes the user to the extension page in VSCode.

src/configureWorkspace/configurePython.ts
bwateratmsft: `string.Method` or `String.Method`? 😄

src/configureWorkspace/configure.ts
```diff
@@ -57,8 +57,14 @@ export function getExposeStatements(ports: number[]): string {
     return ports ? ports.map(port => `EXPOSE ${port}`).join('\n') : '';
 }
 
-export function getComposePorts(ports: number[]): string {
-    return ports && ports.length > 0 ? '    ports:\n' + ports.map(port => `      - ${port}:${port}`).join('\n') : '';
+export function getComposePorts(ports: number[], debugPort?: number): string {
+    let portMappings: string[] = ports?.map(port => `      - ${port}`);
```
bwateratmsft: Since `?.` would return `undefined` if `ports` is `undefined`, we should end this line with `?? []`, so that we can be certain `portMappings` will be defined (even if empty).

src/configureWorkspace/configurePython.ts
```diff
@@ -3,68 +3,235 @@
...
+        files.push({ fileName: 'requirements.txt', contents: genRequirementsFile(projectType) });
+    }
+
+    if (generateComposeFiles) {
+        properties.orchestration = 'docker-compose';
+
+        const dockerComposeFile = genDockerCompose(serviceName, ports);
+        const dockerComposeDebugFile = await genDockerComposeDebug(serviceName, projectType, ports, launchFile);
+
+        files.push(
```
ravipal: Do you have special logic to create the file with *.(n) if the file exists?

haniamr: No, that wouldn't be part of the PR. I'm not sure 100% if we need it for Python but we can discuss it at some point.

ravipal: I believe we want the same compose behavior across all platform. it is possible, use might have had another service added to the existing compose and we don't want to overwrite it. In any case, I am adding this functionality to my PR and you could make use of it at later point.

haniamr: Cool, thanks for the heads up, @ravipal. I'd definitely re-use whatever you have at that point.

src/configureWorkspace/configure.ts
philliphoff: At some point we should start adding the strict-null-check TS compiler flag; this might help flag such situations.

src/configureWorkspace/configurePython.ts
philliphoff: I don't know if I'd look in this repo for any sort of consistent guidance on style.  It's gone through at least three significant changes in ownership, each with their own particular style. (I'm going to use the short form whenever/wherever I can.) :-)

PR #1613, opened 2020-02-10
Add help button to Docker view

karolz-ms:


src/commands/help.ts
```diff
@@ -0,0 +1,49 @@
...
+    let items: HelpMenuItem[] = [
+        { label: 'Get started with Docker...', handler: getStarted },
+        { label: 'Review Docker extension issues...', handler: reviewIssues },
+        { label: 'Report Docker extension issue...', handler: reportIssue },
+        { label: 'Edit settings...', handler: editSettings }
+    ];
+
+    const options: IAzureQuickPickOptions = { canPickMany: false, suppressPersistence: true }
+    let selectedItem: HelpMenuItem = await ext.ui.showQuickPick(items, options);
+    context.telemetry.properties.helpItem = selectedItem.handler.name;
```
philliphoff: As the name effectively becomes a soft contract in telemetry analysis, I'd suggest using a constant so that future telemetry is unaffected by any refactoring (i.e. renaming) that might occur by someone not realizing where that name gets used.

karolz-ms: Good suggestion, fixed

src/utils/extension.ts
```diff
@@ -0,0 +1,21 @@
...
+    publisher: string;
+}
+
+let packageJson: IDockerExtensionPackageJson;
+export function getDockerExtensionPackageJson(): IDockerExtensionPackageJson {
+    if (!packageJson) {
+        // Don't remove the following constant, or webpack will complain about nonexistent package.
+        const packageJsonPath = '../../../package.json';
+        packageJson = require(packageJsonPath) as IDockerExtensionPackageJson;
+    }
```
philliphoff: `vscode.extensions.getExtension('<id>')` should provide access to the same information (canonical name and package JSON); not sure if that's any simpler.  We already embed our extension ID as a constant for that purpose.

package.json
```diff
@@ -124,6 +124,10 @@
                 {
                     "command": "vscode-docker.registries.azure.selectSubscriptions",
                     "when": "never"
+                },
+                {
+                    "command": "vscode-docker.help",
+                    "when": "never"
```
philliphoff: Any reason *not* to allow the command from the palette?

karolz-ms: Because Morty said we have a lot of commands already so let's hold off on this for now ☺️

src/utils/extension.ts
karolz-ms: I missed that--will use the hard-coded extension ID.

karolz-ms:
image

karolz-ms:
image

Issue #1543, opened 2020-01-10
[.NET Core] Add Docker Compose support


dbreshears:
"Add Dockerfiles to Workspace...." command palette should include a prompt to optionally add "Compose" files.

dbreshears:
This should include integrated debugging

PR #1616, opened 2020-02-12
debug dotnet by attaching to container

ravipal:
The existing commands 'Add Docker Files to Workspace...' and 'Initialize for Docker Debugging' will add a new debug launch config 'Docker .NET Core Attach'.
This launch config does the following as part of resolve
1. Ask user to select a running container, if it is not provided in the launch config.
2. Install the remote debugger, if the remote debug path is not provided in the launch config.
3. Generate a coreclr attach launch config which will handle the actual attach.

related to #1543

### DIFF
--- a/package.json
+++ b/package.json
@@ -694,10 +694,10 @@
                         }
                     },
                     {
-                        "label": "Docker: .NET Core Attach",
+                        "label": "Docker: .NET Core Attach (Preview)",
                         "description": "Docker: Attach to a .NET Core process in a container",
                         "body": {
-                            "name": "Docker .NET Core Attach",
+                            "name": "Docker .NET Core Attach (Preview)",
                             "type": "docker",
                             "request": "attach",
                             "platform": "netCore",
--- a/package.json
+++ b/package.json
@@ -653,10 +653,6 @@
                                     }
                                 ]
                             },
-                            "debuggerPath": {
-                                "type": "string",
-                                "description": "enter the path for the debugger on the target machine, for example /remote_debugger/vsdbg."
-                            },
                             "platform": {
                                 "type": "string",
                                 "description": "The target plaform for the application.",
@@ -672,6 +668,15 @@
                                 },
                                 "default": {
                                     "<insert-source-path-here>": "<insert-target-path-here>"
+                                }
+                            },
+                            "netCore": {
+                                "description": "Options for debugging .NET Core projects in Docker.",
+                                "properties": {
+                                    "debuggerPath": {
+                                        "type": "string",
+                                        "description": "enter the path for the debugger on the target machine, for example /remote_debugger/vsdbg."
+                                    }
                                 }
                             }
                         }
--- a/package.json
+++ b/package.json
@@ -628,6 +628,53 @@
                                 }
                             }
                         }
+                    },
+                    "attach": {
+                        "properties": {
+                            "containerName": {
+                                "type": "string",
+                                "description": "The container name to attach to. If not specified, then user will be prompted to pick a container."
+                            },
+                            "processName": {
+                                "type": "string",
+                                "description": "The process name to attach to. If this is used, 'processId' should not be used."
+                            },
+                            "processId": {
+                                "anyOf": [
+                                    {
+                                        "type": "string",
+                                        "description": "The process id to attach to. Use \"${command:pickProcesss}\" to get a list of running processes to attach to. If 'processId' used, 'processName' should not be used.",
+                                        "default": "${command:pickProcess}"
+                                    },
+                                    {
+                                        "type": "integer",
+                                        "description": "The process id to attach to. Use \"${command:pickProcesss}\" to get a list of running processes to attach to. If 'processId' used, 'processName' should not be used.",
+                                        "default": 0
+                                    }
+                                ]
+                            },
+                            "debuggerPath": {
+                                "type": "string",
+                                "description": "enter the path for the debugger on the target machine, for example /remote_debugger/vsdbg."
+                            },
+                            "platform": {
+                                "type": "string",
+                                "description": "The target plaform for the application.",
+                                "enum": [
+                                    "netCore"
+                                ]
+                            },
+                            "sourceFileMap": {
+                                "type": "object",
+                                "description": "Optional source file mappings passed to the debug engine. Example: '{ \"/src\":\"${workspaceFolder}\" }'",
+                                "additionalProperties": {
+                                    "type": "string"
+                                },
+                                "default": {
+                                    "<insert-source-path-here>": "<insert-target-path-here>"
+                                }
+                            }
+                        }
                     }
                 },
                 "configurationSnippets": [
--- a/package.json
+++ b/package.json
@@ -692,6 +692,19 @@
                             "name": "Docker: Attach to Node",
                             "remoteRoot": "/usr/src/app"
                         }
+                    },
+                    {
+                        "label": "Docker: .NET Core Attach",
+                        "description": "Docker: Attach to a .NET Core process in a container",
+                        "body": {
+                            "name": "Docker .NET Core Attach",
+                            "type": "docker",
+                            "request": "attach",
+                            "platform": "netCore",
+                            "sourceFileMap": {
+                                "/src": "^\"\\${workspaceFolder}\""
+                            }
+                        }
                     }
                 ]
             },
--- a/src/debugging/DockerDebugConfigurationProvider.ts
+++ b/src/debugging/DockerDebugConfigurationProvider.ts
@@ -17,6 +17,12 @@ export interface DockerDebugConfiguration extends NetCoreDockerDebugConfiguratio
     platform?: DockerPlatform;
 }
 
+export interface DockerAttachConfiguration extends NetCoreDockerDebugConfiguration, NodeDockerDebugConfiguration {
+    debuggerFilePath?: string;
+    processName?: string;
+    processId?: string | number;
+}
+
 export class DockerDebugConfigurationProvider implements DebugConfigurationProvider {
     public constructor(
         private readonly dockerClient: DockerClient,
@@ -42,7 +48,7 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
 
     public resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DockerDebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration | undefined> {
         return callWithTelemetryAndErrorHandling(
-            'docker-launch',
+            debugConfiguration.request === 'attach' ? 'docker-attach' : 'docker-launch',
             async (actionContext: IActionContext) => {
                 if (!folder) {
                     folder = workspace.workspaceFolders[0];
@@ -56,6 +62,10 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
                     // If type is undefined, they may be doing F5 without creating any real launch.json, which won't work
                     // VSCode subsequently will call provideDebugConfigurations which will show an error message
                     return null;
+                }
+
+                if (!debugConfiguration.request) {
+                    throw new Error('The property "request" must be specified in the debug config.')
                 }
 
                 const debugPlatform = getPlatform(debugConfiguration);
--- a/src/debugging/DockerDebugConfigurationProvider.ts
+++ b/src/debugging/DockerDebugConfigurationProvider.ts
@@ -18,7 +18,7 @@ export interface DockerDebugConfiguration extends NetCoreDockerDebugConfiguratio
 }
 
 export interface DockerAttachConfiguration extends NetCoreDockerDebugConfiguration, NodeDockerDebugConfiguration {
-    debuggerFilePath?: string;
+    debuggerPath?: string;
     processName?: string;
     processId?: string | number;
 }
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -628,6 +628,53 @@
...
+                                        "default": "${command:pickProcess}"
+                                    },
+                                    {
+                                        "type": "integer",
+                                        "description": "The process id to attach to. Use \"${command:pickProcesss}\" to get a list of running processes to attach to. If 'processId' used, 'processName' should not be used.",
+                                        "default": 0
+                                    }
+                                ]
+                            },
+                            "debuggerPath": {
```
bwateratmsft: Is "debuggerPath" applicable to platforms other than .NET Core?

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -94,11 +97,34 @@ export class NetCoreDebugHelper implements DebugHelper {
                 netCore: {
                     appProject: unresolveWorkspaceFolder(options.appProject, context.folder)
                 }
+            },
+            {
+                name: 'Docker .NET Core Attach',
```
bwateratmsft: Are we sure we want to add this by default? It stuffs more into launch.json; I thought we wanted to have _less_.

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -228,6 +285,39 @@ export class NetCoreDebugHelper implements DebugHelper {
 
         return pathNormalize(result, platformOS);
     }
+
+    private async InstallDebuggerOnContainer(containerName: string): Promise<string> {
+        const debuggerPath: string = '/remote_debugger';
+        const installUnZip: string = '/bin/sh -c "apt-get update && apt-get install unzip"';
+        const installDebugger: string = `/bin/sh -c "curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${debuggerPath}"`;
```
bwateratmsft: The performance of this is going to be awful. Unlike in VS, we do not reuse containers in most cases, meaning we'll add several seconds (up to minutes) of launch time to every single container.

Even if we did reuse containers, it would still be seconds (up to minutes) per session.

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -228,6 +285,39 @@ export class NetCoreDebugHelper implements DebugHelper {
...
+        );
+
+        return `${debuggerPath}/vsdbg`;
+    }
+
+    private async getContainerNameToAttach(): Promise<string> {
+        const context: IActionContext = { telemetry: { properties: {}, measurements: {} }, errorHandling: { issueProperties: {} } };
+        const containerItem: ContainerTreeItem = await ext.containersTree.showTreeItemPicker(ContainerTreeItem.runningContainerRegExp, {
+            ...context,
+            noItemFoundErrorMessage: 'No running containers are availble to attach'
```
bwateratmsft: Nit: availble => available

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -17,6 +17,12 @@ export interface DockerDebugConfiguration extends NetCoreDockerDebugConfiguratio
     platform?: DockerPlatform;
 }
 
+export interface DockerAttachConfiguration extends NetCoreDockerDebugConfiguration, NodeDockerDebugConfiguration {
+    debuggerFilePath?: string;
```
bwateratmsft: This would have to be `debuggerPath` to match what is in package.json; else in there make it `debuggerFilePath`.

package.json
philliphoff: If these properties are specific only to .NET Core, then they should be housed in a `netCore` property.

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -167,6 +193,37 @@ export class NetCoreDebugHelper implements DebugHelper {
...
+        const containerName: string = debugConfiguration.containerName ?? await this.getContainerNameToAttach();
+
+        // If debugger path is not specified, install the debugger
+        const debuggerFilePath: string = debugConfiguration.debuggerFilePath ?? await this.InstallDebuggerOnContainer(containerName);
+
+        return {
+            ...debugConfiguration, // Gets things like name, preLaunchTask, serverReadyAction, etc.
+            type: 'coreclr',
+            request: 'attach',
+            "justMyCode": false,
```
philliphoff: Nit: prefer single-quotes to double-quotes (here and elsewhere).

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -228,6 +285,39 @@ export class NetCoreDebugHelper implements DebugHelper {
 
         return pathNormalize(result, platformOS);
     }
+
+    private async InstallDebuggerOnContainer(containerName: string): Promise<string> {
```
philliphoff: Nit: `installDebuggerInContainer`.

src/debugging/netcore/NetCoreDebugHelper.ts
philliphoff: This is basically what VS does--I don't see a way around it if the user *really* wants to debug an arbitrary container (i.e. one that we didn't specifically create for the purposes of debugging).  I don't think the use case is repeated build/attach of a container, though, so the performance aspect is likely tolerable.

bwateratmsft: If the intent is for debugging arbitrary containers, then I'm even more certain that we should not add this launch profile by default...

philliphoff: Does this cover all of the basic .NET Core image flavors (e.g. slim, alpine, bionic)?  For example, I recall having to install `curl` on some base images.

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -228,6 +285,39 @@ export class NetCoreDebugHelper implements DebugHelper {
...
+        const debuggerPath: string = '/remote_debugger';
+        const installUnZip: string = '/bin/sh -c "apt-get update && apt-get install unzip"';
+        const installDebugger: string = `/bin/sh -c "curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${debuggerPath}"`;
+
+        const outputManager = new DefaultOutputManager(ext.outputChannel);
+        const dockerClient = new CliDockerClient(new ChildProcessProvider());
+        const options = { interactive: true };
+
+        await outputManager.performOperation(
```
philliphoff: Should we warn the user that we're modifying the running container (by installing additional packages and the debugger)?  Since we're scaffolding the launch configuration for this attach process, making it immediately available, the user may not understand the implications of using it.

package.json
ravipal: Not sure until we know how the other platform debugger works. If other platform supports such debugger, then it can be at top level instead of specific to platform.

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: Yes that is the plan we agreed on when we discussed this feature. Otherwise user has know about this and do one extra step to generate this config.

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: The InstallDebugger command will skip if the latest version is already on the machine. But installing the unzip could be optimized in the future.

src/debugging/DockerDebugConfigurationProvider.ts
ravipal: Yes, Last minute I changed the property from debuggerFilePath to debuggerPath, but forgot to update in all places.

src/debugging/netcore/NetCoreDebugHelper.ts
bwateratmsft: I wasn't here when we discussed this feature 😄 
I offer a vote of dissent. We're placing front and center an experience we know is not great, meant only as an arbitrary solution for those who can't use the "proper" experience.

src/debugging/netcore/NetCoreDebugHelper.ts
bwateratmsft: Where is the logic to skip install if it's already there? Did I miss it?

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: Are you suggesting just a warning on the output window or some kind of popup that user has to confirm?
Just showing a warning doesn't offer an option to stop the installation, but the pop-up will be annoying.

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: I haven't tested all flavors. I added a TODO item to add validation on target OS and flavor and stop the debugging.

ravipal: The script internally does it.

bwateratmsft: Ah, yes. We should see how much time is also added by the `apt-get` call, including when `unzip` is already present.

src/debugging/netcore/NetCoreDebugHelper.ts
philliphoff: I'd suggest a pop-up on the first attempt to attach to a container (i.e. if we detect that the debugger is not yet installed), to minimize the pop-up, perhaps with a "don't prompt again" option for users that find even that annoying.  While containers are assumed to be ephemeral, they're also assumed to be (relatively) identical.  Since we're not creating the container (so there's no notion of ownership of the container), and we're actively changing that running container, and it'll still be different when we're done debugging, it seems like we should make users aware of that fact.

src/debugging/netcore/NetCoreDebugHelper.ts
bwateratmsft: Additionally, aside from automatically installing the debugger for .NET Core, I'm not really sure we're offering anything that they couldn't accomplish themselves with a simple `coreclr` `attach` configuration...

ravipal: This simplifies the user experience of preparing the container for debugging and constructing the coreclr attach with right settings.

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: The installUnZip command takes 5.5 seconds if the unzip is not present. Second time it takes 2 seconds. The perf can be improved later. Not part of this iteration.

src/debugging/netcore/NetCoreDebugHelper.ts
bwateratmsft: I'm not sure I understand what we're trying to solve here. IMO the value add is not simplicity--attach is already off the happy path and therefore is the advanced case, consequently doing a Docker attach is not really any simpler than doing an CoreCLR attach. There _is_ a clear value add in terms of setting up and installing the debugger, but unfortunately it takes a lot of time even in the best case--whereas, with proper setup, a CoreCLR attach config could include debugger setup out-of-band so as not to slow down F5.

Honestly, I think we would be better off providing a guide on how to use the CoreCLR attach, and guidance or even scripts that would assist in setting up the debugger. I'm concerned about the added complexity to our debugging experience, and especially concerned about scaffolding this by default.

src/debugging/netcore/NetCoreDebugHelper.ts
bwateratmsft: I had a thought a while back about possibly pre-installing the debugger on new debug versions of the .NET Core Docker images, I wonder if we need to revisit that to save perf here...

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: Now showing an information message to user asking to allow or not.

package.json
ravipal: Moved it inside the netCore

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: Now it is added only thru add new config.

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: The new script will handle ubuntu, debian and alpine flavors.
But no validation yet for windows or other flavors and will be handled later.

package.json
```diff
@@ -692,6 +692,19 @@
                             "name": "Docker: Attach to Node",
                             "remoteRoot": "/usr/src/app"
                         }
+                    },
+                    {
+                        "label": "Docker: .NET Core Attach",
```
bwateratmsft: "Preview", per the email thread

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -228,6 +276,42 @@ export class NetCoreDebugHelper implements DebugHelper {
 
         return pathNormalize(result, platformOS);
     }
+
+    private async installDebuggerInContainer(containerName: string): Promise<string> {
+        const yesItem: MessageItem = DialogResponses.yes;
+        const install = (yesItem === await window.showInformationMessage('Attaching to container requiers .NET Core debugger in the container. Do you want to install debugger in the container?', ...[DialogResponses.yes, DialogResponses.no]));
```
haniamr: typo: "requires".

package.json
ravipal: added Preview

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: fixed.

Issue #1492, opened 2019-12-14
WORKDIR paths should be absolute


doggy8088:
1. F1 > `Docker: Add Docker Files to Workspace`
2. `ASP․NET Core`

The generated `Dockerfile` has an error on `WORKDIR "/src/."`. See below:

![image](https://user-images.githubusercontent.com/88981/70853276-b4a14280-1ee6-11ea-8bd1-65b59eae0201.png)

I have to remove double quote manually to remove the warning.

![image](https://user-images.githubusercontent.com/88981/70853281-c71b7c00-1ee6-11ea-9562-42f610ad9083.png)

Is this kind of bug in vscode-docker?

rcjsuen:
@doggy8088 This is an issue with the linter where it isn't taking the quotation marks into account correctly. I have opened rcjsuen/dockerfile-utils#67 to track this.

Thank you for your bug report and have a great day!

doggy8088:
@rcjsuen Thanks!

bwateratmsft:
@rcjsuen Would we need a new version of `dockerfile-utils` published to consume the fix in https://github.com/rcjsuen/dockerfile-utils/commit/11ef5d225d05bc79cb6435a78c89123e8efc03e1?

rcjsuen:
@bwateratmsft You'll need a new version of the language server to get the fix. When are you planning to cut a new release?

bwateratmsft:
The powers that be tell me sometime in the next couple weeks. @BigMorty to keep me honest.

BigMorty:
We are actually planning a small release fairly soon, how soon could we integrate a new version of the language server with this fix? Otherwise the next release will be a number weeks out; we don't have a specific date at this time.

rcjsuen:
@BigMorty The size of the pull request is often rather miniscule (as the change is really a simple change to the `package.json` file). Let me tie up some loose ends and then I'll push out a new release for you.

rcjsuen:
@bwateratmsft @BigMorty Opened #1619 to upgrade the language server.

bwateratmsft:
Should be fixed now by https://github.com/microsoft/vscode-docker/pull/1619.

bwateratmsft:
Verified this is fixed.

BigMorty:
@bwateratmsft, I assume we will integrate the new language server in our next major release?

bwateratmsft:
@BigMorty Yep!

bwateratmsft:
[...]

PR #1619, opened 2020-02-13
Upgrade language server version to 0.0.22

### LINKED ISSUE #1337: FROM image suggestions from store
Very often use images like `store/intersystems/iris-community` in `FROM` and it does not have latest version, I should remember the version number. 
Would it be possible if this extension could suggest a version image in such cases?

rcjsuen:
This pull request updates the language server to the newly released version 0.0.22. This update helps fix #1337 and #1492.

Otherwise, there aren't a lot of big changes here outside of newly added support for the `--platform` flag in `FROM` instructions which was introduced in Docker CE 18.04. However, judging by the fact that no one ever reported this, this clearly doesn't bother anyone here.

Please use the Dockerfile below to compare the before/after behaviours.
```Dockerfile
# escape=\
# escape=\
# declaring the escape parser directive two times will now be flagged as an error

# invoke code completion between FROM and scratch,
# you'll get the new --platform flag suggested
FROM       scratch

# hover over the --platform flag to get documentation
FROM --platform=arm64 scratch

# scratch is incorrectly flagged as an error
FROM --platform=arm64 scratch

# this incorrect PLATFORM flag name will be flagged as an error
# as it must be lowercase
FROM --PLATFORM= alpine

# this WORKDIR argument used to be a warning because of the quotes, see #1492
# https://github.com/microsoft/vscode-docker/issues/1492
WORKDIR "/src/."

# invoke code completion after the : to get Docker tag suggestions, see #1337
# https://github.com/microsoft/vscode-docker/issues/1337
FROM store/intersystems/iris-community:
```
This update also fixes a very, very obscure bug that you can replicate with the Dockerfile below. If you have this file open and use the "Output" panel to look at the logs of the "Dockerfile Language Server", you should see some errors. Feel free to take a look at rcjsuen/dockerfile-ast#66 if you are interested in the details (though as aforementioned it is very much an edge case that I don't expect the average user to encounter).
```Dockerfile
RUN alpin\
e
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3156,39 +3156,40 @@
       }
     },
     "dockerfile-ast": {
-      "version": "0.0.16",
-      "resolved": "https://registry.npmjs.org/dockerfile-ast/-/dockerfile-ast-0.0.16.tgz",
-      "integrity": "sha512-+HZToHjjiLPl46TqBrok5dMrg5oCkZFPSROMQjRmvin0zG4FxK0DJXTpV/CUPYY2zpmEvVza55XLwSHFx/xZMw==",
+      "version": "0.0.20",
+      "resolved": "https://registry.npmjs.org/dockerfile-ast/-/dockerfile-ast-0.0.20.tgz",
+      "integrity": "sha512-VGFMBT0Av1Sk4SCjafsv2S/MCVy6+AuhLW+958wS2df86wb90JPl+WJXhTTGHk5DQVwQX0NoQQBzQQP1U7GaWg==",
       "requires": {
         "vscode-languageserver-types": "^3.5.0"
       }
     },
     "dockerfile-language-server-nodejs": {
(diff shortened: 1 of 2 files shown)


bwateratmsft:
I'll complete this now, since we have already forked for 0.10.1.

rcjsuen:
Thank you @philliphoff for reviewing and @bwateratmsft for merging.

Apologies for not getting this in sooner for your 0.10.1 release.

bwateratmsft:
No worries! We kinda preferred to save it for 0.11.0; we're reluctant to do dependency updates in minor releases.

rcjsuen:
> No worries! We kinda preferred to save it for 0.11.0; we're reluctant to do dependency updates in minor releases.

Makes sense. Better safe than sorry. Have a nice day!

PR #1622, opened 2020-02-13
Show a warning with Learn More link if SSH is in use without a working ssh-agent

### LINKED ISSUE #1458: Remote Containers using SSH doesn't work in Windows
Related to microsoft/vscode-remote-release#1935

- I confirmed that I can successfully connect to the remote machine using SSH via command line
- There is no error in the entire UI, just loading progress bars when opening the Docker viewlet
- When running `Attach to Running Container`, nothing happens
- When I open devtools, this is what I see:

image

bwateratmsft:
Resolves #1458 with @zifik's suggestion. Because an ssh-agent is required by the `ssh2` Node package (used by Dockerode) when using an SSH connection to a remote Docker daemon, we will show a warning with a Learn More link if SSH is in use without a working ssh-agent.

The check is done by looking at the value for `SSH_AUTH_SOCK` (which on Windows can be defaulted to `\\.\pipe\openssh-ssh-agent`), and then trying to connect to that pipe to ensure it is actually working.

The warning:
image

The link: https://aka.ms/AA7assy

### DIFF
--- a/src/utils/refreshDockerode.ts
+++ b/src/utils/refreshDockerode.ts
@@ -9,7 +9,7 @@ import { Socket } from 'net';
 import { ext } from '../extensionVariables';
 import { addDockerSettingsToEnv } from './addDockerSettingsToEnv';
 import { cloneObject } from './cloneObject';
-import { delayWithResult } from './delay';
+import { delay } from './delay';
 import { isWindows } from './osUtils';
 import { execAsync } from './spawnAsync';
 
@@ -103,7 +103,7 @@ async function validateSshAuthSock(newEnv: NodeJS.ProcessEnv): Promise<boolean> 
     });
 
     // Unfortunately Socket.setTimeout() does not actually work when attempting to establish a connection, so we need to race
-    const result = await Promise.race([connectPromise, delayWithResult(1000, false)]);
+    const result = await Promise.race([connectPromise, delay(1000).then(() => false)]);
     authSock.end();
 
     return result;
--- a/src/utils/refreshDockerode.ts
+++ b/src/utils/refreshDockerode.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { Socket } from 'net';
 import { DockerOptions } from 'dockerode';
 import Dockerode = require('dockerode');
 import { ext } from '../extensionVariables';
@@ -56,11 +57,13 @@ async function getDockerodeOptions(newEnv: NodeJS.ProcessEnv): Promise<DockerOpt
 
     try {
         if (newEnv.DOCKER_HOST &&
-            SSH_URL_REGEX.test(newEnv.DOCKER_HOST) &&
-            !newEnv.SSH_AUTH_SOCK) {
-            // If DOCKER_HOST is an SSH URL, we need to configure SSH_AUTH_SOCK for Dockerode
+            SSH_URL_REGEX.test(newEnv.DOCKER_HOST)) {
+            // If DOCKER_HOST is an SSH URL, we need to configure / validate SSH_AUTH_SOCK for Dockerode
             // Other than that, we use default settings, so return undefined
-            newEnv.SSH_AUTH_SOCK = await getSshAuthSock();
+            if (!await validateSshAuthSock(newEnv)) {
+                await ext.ui.showWarningMessage('In order to use an SSH DOCKER_HOST, you must configure an ssh-agent.', { learnMoreLink: 'https://aka.ms/AA7assy' });
+            }
+
             return undefined;
         } else if (!newEnv.DOCKER_HOST) {
             // If DOCKER_HOST is unset, try to get default Docker context--this helps support WSL
@@ -72,14 +75,31 @@ async function getDockerodeOptions(newEnv: NodeJS.ProcessEnv): Promise<DockerOpt
     return undefined;
 }
 
-async function getSshAuthSock(): Promise<string | undefined> {
-    if (isWindows()) {
-        return '\\\\.\\pipe\\openssh-ssh-agent';
-    } else {
+async function validateSshAuthSock(newEnv: NodeJS.ProcessEnv): Promise<boolean> {
+    if (!newEnv.SSH_AUTH_SOCK && isWindows()) {
+        // On Windows, we can use this one by default
+        newEnv.SSH_AUTH_SOCK = '\\\\.\\pipe\\openssh-ssh-agent';
+    } else if (!newEnv.SSH_AUTH_SOCK) {
         // On Mac and Linux, if SSH_AUTH_SOCK isn't set there's nothing we can do
         // Running ssh-agent would yield a new agent that doesn't have the needed keys
-        await ext.ui.showWarningMessage('In order to use an SSH DOCKER_HOST on OS X and Linux, you must configure an ssh-agent.');
+        return false;
     }
+
+    // Validate the pipe actually exists
+    return await new Promise<boolean>((resolve, reject) => {
+        const authPipe = new Socket();
+
+        authPipe.on('error', () => {
+            resolve(false);
+        });
+
+        authPipe.on('connect', () => {
+            authPipe.end();
+            resolve(true);
+        });
+
+        authPipe.connect(newEnv.SSH_AUTH_SOCK);
+    });
 }
 
 async function getDefaultDockerContext(): Promise<DockerOptions | undefined> {
--- a/src/utils/refreshDockerode.ts
+++ b/src/utils/refreshDockerode.ts
@@ -9,6 +9,7 @@ import { Socket } from 'net';
 import { ext } from '../extensionVariables';
 import { addDockerSettingsToEnv } from './addDockerSettingsToEnv';
 import { cloneObject } from './cloneObject';
+import { delayWithResult } from './delay';
 import { isWindows } from './osUtils';
 import { execAsync } from './spawnAsync';
 
@@ -61,7 +62,9 @@ async function getDockerodeOptions(newEnv: NodeJS.ProcessEnv): Promise<DockerOpt
             // If DOCKER_HOST is an SSH URL, we need to configure / validate SSH_AUTH_SOCK for Dockerode
             // Other than that, we use default settings, so return undefined
             if (!await validateSshAuthSock(newEnv)) {
-                await ext.ui.showWarningMessage('In order to use an SSH DOCKER_HOST, you must configure an ssh-agent.', { learnMoreLink: 'https://aka.ms/AA7assy' });
+                // Don't wait
+                /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
+                ext.ui.showWarningMessage('In order to use an SSH DOCKER_HOST, you must configure an ssh-agent.', { learnMoreLink: 'https://aka.ms/AA7assy' });
             }
 
             return undefined;
@@ -85,21 +88,25 @@ async function validateSshAuthSock(newEnv: NodeJS.ProcessEnv): Promise<boolean> 
         return false;
     }
 
-    // Validate the pipe actually exists
-    return await new Promise<boolean>((resolve, reject) => {
-        const authPipe = new Socket();
+    const authSock = new Socket();
 
-        authPipe.on('error', () => {
+    const connectPromise = new Promise<boolean>(resolve => {
+        authSock.on('error', (err) => {
             resolve(false);
         });
 
-        authPipe.on('connect', () => {
-            authPipe.end();
+        authSock.on('connect', () => {
             resolve(true);
         });
 
-        authPipe.connect(newEnv.SSH_AUTH_SOCK);
+        authSock.connect(newEnv.SSH_AUTH_SOCK);
     });
+
+    // Unfortunately Socket.setTimeout() does not actually work when attempting to establish a connection, so we need to race
+    const result = await Promise.race([connectPromise, delayWithResult(1000, false)]);
+    authSock.end();
+
+    return result;
 }
 
 async function getDefaultDockerContext(): Promise<DockerOptions | undefined> {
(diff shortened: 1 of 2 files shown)


src/utils/refreshDockerode.ts
```diff
@@ -56,11 +57,13 @@ async function getDockerodeOptions(newEnv: NodeJS.ProcessEnv): Promise<DockerOpt
...
         if (newEnv.DOCKER_HOST &&
-            SSH_URL_REGEX.test(newEnv.DOCKER_HOST) &&
-            !newEnv.SSH_AUTH_SOCK) {
-            // If DOCKER_HOST is an SSH URL, we need to configure SSH_AUTH_SOCK for Dockerode
+            SSH_URL_REGEX.test(newEnv.DOCKER_HOST)) {
+            // If DOCKER_HOST is an SSH URL, we need to configure / validate SSH_AUTH_SOCK for Dockerode
             // Other than that, we use default settings, so return undefined
-            newEnv.SSH_AUTH_SOCK = await getSshAuthSock();
+            if (!await validateSshAuthSock(newEnv)) {
+                await ext.ui.showWarningMessage('In order to use an SSH DOCKER_HOST, you must configure an ssh-agent.', { learnMoreLink: 'https://aka.ms/AA7assy' });
```
bwateratmsft: Add a screenshot.

src/utils/refreshDockerode.ts
```diff
@@ -72,14 +75,31 @@ async function getDockerodeOptions(newEnv: NodeJS.ProcessEnv): Promise<DockerOpt
...
+        authPipe.on('error', () => {
+            resolve(false);
+        });
+
+        authPipe.on('connect', () => {
+            authPipe.end();
+            resolve(true);
+        });
+
+        authPipe.connect(newEnv.SSH_AUTH_SOCK);
```
bwateratmsft: So this takes forever, and `setTimeout` doesn't work--need to work out some other way.

src/utils/refreshDockerode.ts
```diff
@@ -72,14 +79,35 @@ async function getDockerodeOptions(newEnv: NodeJS.ProcessEnv): Promise<DockerOpt
...
+        newEnv.SSH_AUTH_SOCK = '\\\\.\\pipe\\openssh-ssh-agent';
+    } else if (!newEnv.SSH_AUTH_SOCK) {
         // On Mac and Linux, if SSH_AUTH_SOCK isn't set there's nothing we can do
         // Running ssh-agent would yield a new agent that doesn't have the needed keys
-        await ext.ui.showWarningMessage('In order to use an SSH DOCKER_HOST on OS X and Linux, you must configure an ssh-agent.');
+        return false;
     }
+
+    const authSock = new Socket();
+    const cts = new CancellationTokenSource();
```
philliphoff: A `CancellationTokenSource` is disposable, so should be disposed before the function returns, either in the `finally()` handler or you could add `try`/`finally` blocks.

diablodale:
I recommend the documentation linked from that dialog box remove ambiguity for the use of WSL. As this warning dialog is written, it would be ambiguous to a setup that uses SSH and SSH-AGENT running within WSL. In addition, that linked documentation https://code.visualstudio.com/docs/containers/ssh#_directly-via-ssh step 2.3. could be read to further support WSL Ubuntu Linux (and WSL2 which is imminent).

bwateratmsft:
@karolz-ms, could you look at adjusting the docs per @diablodale's feedback?

We also currently have the ability to automagically work with WSL2 when using Docker Desktop's [WSL2 backend feature](https://docs.docker.com/docker-for-windows/wsl-tech-preview/). The extension uses the `docker context` command to detect whichever context is active when the extension activates.

karolz-ms:
@diablodale what would you like https://code.visualstudio.com/docs/containers/ssh documentation topic to say about WSL?

The whole topic is about "connecting to remote Docker daemon". With WSL there is no "remote Docker daemon" in the picture, so it should be pretty clear it does not apply to WSL. And with regards to WSL specifically we have this: https://code.visualstudio.com/docs/containers/choosing-dev-environment#_windows-subsystem-for-linux

diablodale:
Users that use WSL for managing their secure key store and ssh toolset...we use the local client tools within WSL. WSL runs `ssh`, `ssh-agent`, `gpg-agent`, `ssh-add`, etc. And it is usually WSL Ubuntu Linux.

I have no comments about docker daemon. The docker daemon is a server process running on a remote server. I'm not discussing the daemon. Instead, for clarity/simplicity my feedback is only about the local computer client, running client tools on that client, etc.

@bwateratmsft , yes I'm aware of some the integration with Docker Desktop's server. However, none of what I suggest involves Docker Desktop server. I have no comments about that 3rd party toolset. Instead, I'm focused on clarifying this 2nd of the two SSH connectivity methods listed in this documentation that use client tools (not server). [Caveat, perhaps there is better support in WSL2 to run an ELF binary directly from Win32, last I heard still required prepend with `wsl`]
 
@karolz-ms , my suggestion is to be more clear with the documentation (perhaps also adjusting the dialog if it can be compact enough) such that using WSL for client tools `ssh`, `ssh-agent`, `ssh-add`, `gpg-agent`, etc. or dependencies that use such tools (dockernode, ssh2, etc.) is not supported.

All the tools and commands in this section are from Linux and BSD. Around Aug 2016 WSL had `ssh`. A year later there was a disabled native Win32 `ssh`. Took yet another year for that to not be disabled. Therefore, the context (the thinking orientation) for a technical reader of this section might readily be *nix.

Here are some ideas:

* At https://code.visualstudio.com/docs/containers/ssh#_directly-via-ssh in the very first paragraph, write something like "WSL is not supported for this SSH method."
* Adjust 2.1. and 2.2. under there. The term "Windows" is ambiguous. "Win32" is great, but rather techy. But then the audience for this is techy...so perhaps that's the right choice.
* Adjust 2.3. to be "Linux (not WSL). Ubuntu was tested; you might have different results on other distributions:"

I acknowledge this is not absolute. I distinguish between what is supported and what is possible. My suggests are to clarify that WSL is not supported for this 2nd of two SSH connectivity methods.

karolz-ms:
@diablodale thanks. We can certainly clarify in the doc which parts do not apply to WSL (or WSL 2). 

Full disclosure: I haven't tried the steps listed under https://code.visualstudio.com/docs/containers/ssh#_directly-via-ssh  with WSL, but I would expect them to work, so I would like to understand why you are saying they are not supported. Are you aware of any known issues with this setup and WSL? Or maybe you tried and found the steps not working/sufficient? Thanks in advance for clarification!

bwateratmsft:
@karolz-ms You can use [Remote - WSL](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) to attach VSCode to WSL (also, install the Docker extension within WSL; by default it runs as a UI extension). No configuration of the "host" (i.e. the Windows machine) is needed in that case. `bash` is automatically used as the terminal, and `dockerode` and `docker *` commands happen in the same context, so the setting `docker.host` (or env var `DOCKER_HOST`) will work for `ssh://...` values (as long as `SSH_AUTH_SOCK` is also set in WSL).

When I tried with the Ubuntu distro, I had to call `eval $(ssh-agent -s)` and `ssh-add keyfile` from within `.bashrc` so that when the VSCode server-side process ran in WSL it would have the needed `SSH_AUTH_SOCK` environment variable, along with a running `ssh-agent` process, but I'm sure there are better ways to accomplish that. Once I did that everything worked as expected.

If you want to _not_ attach with Remote - WSL, then it is necessary to do the `ssh-agent` configuration in both Windows and WSL, because `dockerode` and `docker *` commands are not in the same context.

karolz-ms:
@bwateratmsft thank you, that is exactly what I would expect. Based on that I would consider WSL supported.

diablodale:
And that exposes the distinction between what's supported vs. what's possible.
My feedback is not about method 1 (Remote-WSL). I have no comments about that method within this github issue. It is only the last sentence from @bwateratmsft which relates to my feedback.

If this team wants to support duplicate setups of `ssh-agent` keystores, duplicate installs of `ssh` client tools, etc. -> then more documentation is needed for that duplicate WSL+Win32 setup. Personally, I wouldn't support it. It creates a computer scenario where the user has to constantly keep keys, passwords, and other details exactly the same using two different toolsets, two different drive locations, etc. Otherwise...save time/effort and state WSL is unsupported for method 2.

karolz-ms:
@diablodale thanks for clarification. I think I understand where you are coming from. The difference is in the mental model of what WSL really is. Is it *a shell*? This is how it was originally advertised ("bash on Windows"), but it evolved into something that nowadays most people here at Microsoft would describe it as *virtual environment for running Linux programs*. That has been made very prominent with WSL 2 and its native Linux kernel. And with this the expected, default level of integration between Win32 and WSL environment (as well as between different WSL instances) is really none. They are completely isolated.

Arguably complete isolation is not always the most practical choice, so WSL (and tools that leverage it) have multiple integration points with Win32 (file system mounts, socket mounts, the new Docker implementation for WSL2, VS Code Remote-WSL extension are good examples). The fundamental principle remains though that WSL environments and Win32 are isolated by default.

I can see your point that for SSH connectivity (and probably other things) there is not enough integration and I agree. It is a pain to manage all the SSH identities separately for Win32 vs WSL. But this is largely an issue outside of the VS Code Docker extension control. @bwateratmsft  confirmed that  the setup described in the doc you referenced IS working with WSL, with the design of WSL as it is today. If anybody reports that it is not working, we will investigate. Hence it is supported.

Thank you for taking time to share your thoughts!

diablodale:
You and I are definitely reading different posts. I see nothing written from @bwateratmsft that confirms method 2 works with WSL when someone follows the docs at https://code.visualstudio.com/docs/containers/ssh#_directly-via-ssh. Instead, BW at a high-level describes some duplicate work needed in two diffenent subsystems (Win32, WSL) + some eval() things in WSL + manual edits to .bashrc in WSL. None of that exists in the doc at https://code.visualstudio.com/docs/containers/ssh#_directly-via-ssh. Perhaps you want to update the doc and point it to this issue so that someone can try and understand what's needed.

Good to hear that WSL is supported. Does this mean that vscode-docker now supports WSL? Last I heard from BW in another issue, WSL was not confirmed to be supported by this team.

I continue to confirm that method 1 is not in discussion and I have no comments regarding that.

karolz-ms:
Understood that we are not talking about method 1

WSL is supported via Remote-WSL extension. When that extension is used, a headless instance of VS Code is installed into WSL environment. Then one can use that VS Code instance to install headless instance of Docker extension into WSL environment. That headless extension instance will see WSL environment configuration, including SSH/Docker configuration, and will drive the Docker extension UI running on Win32 side. All the data comes from the WSL environment and all the Docker commands are executed in that environment. The only caveat that Brandon describes is that ssh-agent configuration does not survive WSL environment restart, so it needs to be re-created via `.bashrc` file. This probably needs to be added to the doc. Everything else in https://code.visualstudio.com/docs/containers/ssh#_directly-via-ssh doc applies to WSL via Remote-WSL unchanged. 

What is not supported is running VS Code and Docker extension entirely on Win32 side, without Remote-WSL extension. At least not without extra manual work of keeping Win32 and WSL configuration in sync. This goes back to the point they are isolated environments.

Hope this clarifies things but please let me know if I missed something.

diablodale:
Ah, I think we are using different numbering/naming. Whiteboards are so good for this. ;-)
On the 1st paragraph of page https://code.visualstudio.com/docs/containers/ssh are two bullets. These two bullets describe the two methods discussed on that page's text
```
"method 1" -> bullet 1 -> Remote dev using SSH -> aka VSCode Remote-SSH extension
"method 2" -> bullet 2 -> Manually set up an SSH connection
```

The post by BW at https://github.com/microsoft/vscode-docker/pull/1622#issuecomment-587723325 writes about a 3rd method that is undocumented. This 3rd method uses the `Remote-WSL` extension.

Also in that same post by BW he writes "If you want to not attach with `Remote-WSL`, then it is necessary to do the ssh-agent configuration in both Windows and WSL...". Written another way, he writes if a user doesn't use `Remote-WSL`, then that user must use an undocumented variation of method 2 that needs additional/duplicate `ssh-agent` configuration in both Windows and WSL. This variation of method 2 is not documented except at a high-level in that post by BW.

Give the above, there are two gaps in the documentation:

* undocumented method 3
* undocumented variation of method 2

bwateratmsft:
I agree that those are gaps in the documentation, but I'm not convinced that we _should_ document them. There's probably an infinite number of ways and combinations of ways to make remote connectivity work, but I think that our docs should stick to the beaten path(s), so to speak:

A - Want to talk to remote Docker via HTTP? Use the appropriate `DOCKER_*` env vars or `docker.*` settings.

B - Want to talk to remote Docker via SSH? Use Remote - SSH with Docker extension installed inside (preferred), or use the SSH features the Docker extension has (`DOCKER_HOST` / `docker.host` and `SSH_AUTH_SOCK`).

C - Want to talk to remote Docker in WSL? Use Remote - WSL with Docker extension installed inside (preferred), or stand up SSH within WSL and use the SSH features the Docker extension has.

D - Want to talk to remote Docker in WSL2? Use Remote - WSL with Docker extension installed inside (preferred), or the Docker WSL2 backend feature (preferred, but somewhat less so) (no extension config needed), or stand up SSH within WSL2 and use the SSH features the Docker extension has.

Method 3 (Remote - WSL + SSH) is a combination of options C and B, IMO better solved with just option B alone, so I'd prefer to not document it. The variant of method 2 (WSL as shell + SSH) is essentially option B but since it uses WSL as a shell, requires additional config work due to the disparate environments.

diablodale:
💯 with you documenting all combinations isn't the right direction. I do like the organization that you are surfacing. That can help users choose among so many possibilities and steer them towards those preferred/supported. Perhaps the doc writes of only 3 methods -- for brevity and focus. One method for each "via/dest".

* A
* B
* Combine C&D and only describe "Use Remote - WSL with Docker extension installed inside (preferred)". There is a bit of ambiguity that might surface when the text is written due to WSL/WSL2 client tools work, WSL2 capable of Docker daemon, yet WSL is not capable of Docker daemon so that subset-configuration would always have the Daemon on a different computer -or- back in Win32 with Docker Desktop and opening the TCP port. Though hopefully the text can be written in a way that avoids ambiguity and is clear to read.

And not write anything about Method 3

diablodale:
Hi. Circling back around I didn't see any updates from our above conversation to https://code.visualstudio.com/docs/containers/ssh
Is that in the work queue?

PR #1620, opened 2020-02-13
Start, stop and restart multiple containers

allanwright:
Addresses #1453.

Enabled `CanSelectMany` option on `containersTreeView`. Modified `startContainers`, `stopContainers` and `restartContainers` commands to support multi selection.

PR #1623, opened 2020-02-13
Add multi-select to context menu commands that most need it

### LINKED ISSUE #331: Support for multiple selection of items and actions on these items (batch)

### LINKED ISSUE #1453: Add "start all containers" option
Add "start all containers" option

bwateratmsft:
Fixes #331, fixes #1453.

This PR is similar to #1620 (and also covers what #1494 does) but adds multi-select support to all of the explorer groups (except Registries), with multi-select support for the following commands:

Containers, Images, Networks, Volumes: Remove
Containers only: Start, Stop, Restart

It also will filter off nodes for which the action taken is not eligible (for example, parent nodes, or stopping an already-stopped container). Unfortunately it shows actions for whatever was the specific item in the selection set that was right-clicked (not actions that are valid for _all_ selected items) but we think that for the most part that will not be an issue.

Some of the remaining commands do not make sense as a multi-select (for example, Image > Copy full tag), some of the others _maybe_ make sense but we are leaving them as single-select at this time (for example, Inspect).

### DIFF
--- a/src/commands/containers/removeContainer.ts
+++ b/src/commands/containers/removeContainer.ts
@@ -4,28 +4,23 @@
  *--------------------------------------------------------------------------------------------*/
 
 import vscode = require('vscode');
-import { IActionContext, UserCancelledError } from 'vscode-azureextensionui';
+import { IActionContext } from 'vscode-azureextensionui';
 import { ext } from '../../extensionVariables';
 import { ContainerTreeItem } from '../../tree/containers/ContainerTreeItem';
+import { multiSelectNodes } from '../../utils/multiSelectNodes';
 
-export async function removeContainer(context: IActionContext, node: ContainerTreeItem | undefined): Promise<void> {
-    let nodes: ContainerTreeItem[] = [];
-    if (node) {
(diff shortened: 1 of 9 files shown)


eine:
> Some of the remaining commands do not make sense as a multi-select (for example, Image > Copy full tag), some of the others maybe make sense but we are leaving them as single-select at this time (for example, Inspect).

If this is applied, I'd request to please NOT close #331 as part of this PR.

bwateratmsft:
> > Some of the remaining commands do not make sense as a multi-select (for example, Image > Copy full tag), some of the others maybe make sense but we are leaving them as single-select at this time (for example, Inspect).
> 
> If this is applied, I'd request to please NOT close #331 as part of this PR.

I think it would be appropriate to close #331 but create a new one for Inspect/View Logs/etc.). This would help us gauge the importance of taking such a change.

PR #1625, opened 2020-02-14
.Net Core compose support

ravipal:
Now the dotnet core scaffolding will ask whether to generate compose files or not similar to node.js.
If an existing compose file is present, it will create docker-compose(n).yaml

### DIFF
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -166,9 +166,8 @@ services:
   $service_name$:
     image: $image_name$
     build:
-      context: $build_context$
-      dockerfile: $dockerfile$$ports$
-`;
+      context: .
+      dockerfile: $dockerfile$$ports$`;
 
 const dotNetComposeDebugTemplate = `${dotNetComposeTemplate}$environment$
     volumes:
@@ -276,11 +275,10 @@ function validateForUnresolvedToken(contents: string): void {
 }
 
 function generateComposeFiles(dockerfileName: string, platform: Platform, os: PlatformOS | undefined, ports: number[], artifactName: string): ScaffoldFile[] {
-    const serviceName = path.basename(artifactName, path.extname(artifactName));
+    const serviceName = path.basename(artifactName, path.extname(artifactName)).toLowerCase();
     const imageName = serviceName;
-    const buildContext = path.dirname(dockerfileName);
     let jsonPorts: string = '';
-    if (ports) {
+    if (ports && ports.length > 0) {
         const portsArray = ports.map(p => `      - ${p}`).join("\r\n");
         jsonPorts = `
     ports:
@@ -308,16 +306,17 @@ ${portsArray}`;
         volumesList += ':rw'
     }
 
+    // Ensure the path scaffolded in the Dockerfile uses POSIX separators (which work on both Linux and Windows).
+    dockerfileName = dockerfileName.replace(/\\/g, '/');
+
     let composeFileContent = dotNetComposeTemplate.replace('$service_name$', serviceName)
         .replace(/\$image_name\$/g, imageName)
-        .replace(/\$build_context\$/g, buildContext)
         .replace(/\$dockerfile\$/g, dockerfileName)
         .replace(/\$ports\$/g, jsonPorts);
     validateForUnresolvedToken(composeFileContent);
 
     let composeDebugFileContent = dotNetComposeDebugTemplate.replace('$service_name$', serviceName)
         .replace(/\$image_name\$/g, imageName)
-        .replace(/\$build_context\$/g, buildContext)
         .replace(/\$dockerfile\$/g, dockerfileName)
         .replace(/\$ports\$/g, jsonPorts)
         .replace(/\$environment\$/g, environmentVariables)
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -324,11 +324,23 @@ ${portsArray}`;
     validateForUnresolvedToken(composeDebugFileContent);
 
     return [
-        { fileName: 'docker-compose.yml', contents: composeFileContent, createNewNameIfExist: true },
-        { fileName: 'docker-compose.debug.yml', contents: composeDebugFileContent, createNewNameIfExist: true }
-    ]
+        { fileName: 'docker-compose.yml', contents: composeFileContent, onConflict: async (filePath) => { return await generateNonConflictFileName(filePath) } },
+        { fileName: 'docker-compose.debug.yml', contents: composeFileContent, onConflict: async (filePath) => { return await generateNonConflictFileName(filePath) } }
+    ];
 }
 
+async function generateNonConflictFileName(filePath: string): Promise<string> {
+    let newFilepath = filePath;
+    let i = 1;
+    const extName = path.extname(filePath);
+    const extNameRegEx = new RegExp(`${extName}$`);
+
+    while (await fse.pathExists(newFilepath)) {
+        newFilepath = filePath.replace(extNameRegEx, i + extName);
+        i++;
+    }
+    return newFilepath;
+}
 // Returns the relative path of the project file without the extension
 async function findCSProjOrFSProjFile(folderPath?: string): Promise<string> {
     const opt: vscode.QuickPickOptions = {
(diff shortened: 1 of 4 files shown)


src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -303,6 +375,7 @@ async function initializeForDebugging(context: IActionContext, folder: Workspace
 // tslint:disable-next-line: export-name
 export async function scaffoldNetCore(context: ScaffolderContext): Promise<ScaffoldFile[]> {
     const os = context.os ?? await context.promptForOS();
+    const IsCompose = await context.promptForCompose();
```
philliphoff: Nit: `isCompose`

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -344,11 +417,16 @@ export async function scaffoldNetCore(context: ScaffolderContext): Promise<Scaff
...
 
-    const files: ScaffoldFile[] = [
+    const composeFiles = IsCompose ? generateComposeFiles(dockerFileName, context.platform, os, ports, workspaceRelativeProjectFileName) : [];
+
+    let files: ScaffoldFile[] = [
         { fileName: dockerFileName, contents: dockerFileContents, open: true },
         { fileName: dockerIgnoreFileName, contents: genCommonDockerIgnoreFile(context.platform) }
     ];
 
+    // TODO test if adding an empty array is no-op
```
philliphoff: Empty arrays passed to `concat()` will no-op.  Undefined values (e.g. an `undefined` "array") passed to `concat()` will be result in the `undefined` value being added to the array.

src/configureWorkspace/scaffolding.ts
```diff
@@ -36,6 +36,7 @@ export interface ScaffolderContext extends ScaffoldContext {
     platform: Platform;
     promptForOS(): Promise<PlatformOS>;
     promptForPorts(defaultPorts?: number[]): Promise<number[]>;
+    promptForCompose(): Promise<boolean>;
```
philliphoff: Update generic-configure-scaffolding and Python scaffolder to use this new function?

src/configureWorkspace/scaffolding.ts
```diff
@@ -138,14 +141,32 @@ export async function scaffold(context: ScaffoldContext): Promise<ScaffoldedFile
...
     return writtenFiles;
+
+    async function generateNonConflictFileName(filePath: string): Promise<string> {
+        let newFilepath = filePath;
+        let i = 1;
+        const extName = path.extname(filePath);
+        const extNameRegEx = new RegExp(`${extName}$`);
+
+        while (await fse.pathExists(newFilepath)) {
+            newFilepath = filePath.replace(extNameRegEx, i + extName);
```
philliphoff: It might be better to split apart the base file name from the directory and extension, and only update the base file name.  This regex has a risk of replacing a pattern elsewhere in the path.  For example, imagine a filename like: `/repo/config.yamls/docker-compose.yaml`.

ravipal: The $ looks for ext name at the end. So it is safe.

src/configureWorkspace/configureDotNetCore.ts
ravipal: Thanks, I tested it, but forgot to remove the comment.

src/configureWorkspace/scaffolding.ts
```diff
@@ -48,6 +49,7 @@ export type ScaffoldFile = {
     contents: string;
     fileName: string;
     open?: boolean;
+    createNewNameIfExist?: boolean;
```
philliphoff: It seems conceivable that we may end up adding more configurable behavior on name conflicts (e.g. `promptForOverwrite`) which would end up complicating this type (e.g. which property would have precedence?) . It might be better to consolidate that behavior into a single "conflict resolution method" handler.

One option would be a simple enumeration:

```typescript
type ConflictResolutionMethod = 'prompt' | 'overwrite' | 'unique-name'
type ScaffoldFile = {
   ...
   onConflict?: ConflictResolutionMethod; // If omitted, defaults to 'prompt'
};
```

Another option would be to allow the individual scaffolders the flexibility to do whatever is needed (as it may be difficult to use a single unique naming scheme across all types, or the scaffolder type may want to use a more specific prompt to overwrite).  Something like:

```typescript
type ScaffoldFile = {
   ...
   onConflict?: (path: string, defaultHandler: (path: string): => Promise<string>) => Promise<string>;
};
```

In this case, the scaffolding type can attach a callback which the scaffolder will invoke, which will return the (possibly new) filename to use.  Also passed could be the "default" handler (that would normally prompt for overwrite) that could be called if the specific handler wanted a fallback in certain cases.

In either case, we might also consider how a scaffolder type could indicate that a file be "skipped" altogether when there's a conflict.

ravipal: I think the second suggestion is more open. But Why not use
`  onConflict?: (path: string) => Promise<string|undefined>;`

Since OnConflict is optional, then the default behavior of overwrite can be used if onConflict is undefined. No need to pass defaultHandler to callback.

src/configureWorkspace/scaffolding.ts
philliphoff: Yes, sorry, missed that trailing `$`.

src/configureWorkspace/scaffolding.ts
philliphoff: My original thinking was that returning `undefined` didn't make clear (or allow for) whether the default behavior should be used (e.g. prompt for overwrite) or whether the file should just be skipped (as you suggest).  In the latter case, the platform-specific scaffolder is then responsible for prompting for override, if that's ultimately the behavior they want for that file in that case.  (Passing the default allows the scaffolder the option to perform completely custom logic, or some combination of custom and default behavior.)

It only really becomes an issue if/when that default behavior gets more sophisticated.  Since, today, it only prompts for overwriting the file, it may be overkill to add that extra flexibility.

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -249,14 +262,85 @@ function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, o
...
     }
+}
 
-    return contents;
+function generateComposeFiles(dockerfileName: string, platform: Platform, os: PlatformOS | undefined, ports: number[], artifactName: string): ScaffoldFile[] {
+    const serviceName = path.basename(artifactName, path.extname(artifactName)).toLowerCase();
+    const imageName = serviceName;
+    let jsonPorts: string = '';
+    if (ports && ports.length > 0) {
+        const portsArray = ports.map(p => `      - ${p}`).join("\r\n");
```
bwateratmsft: Isn't there a common port mappings generator function?

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -249,14 +262,85 @@ function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, o
...
+        jsonPorts = `
+    ports:
+${portsArray}`;
+    }
+
+    let environmentVariables: string = '';
+    if (platform === 'ASP.NET Core') {
```
bwateratmsft: Hani's code might change this `ASP.NET Core` value, make sure these get merged right.

src/configureWorkspace/scaffolding.ts
ravipal: I think this should work for most cases and can be improved when needed.

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -249,14 +262,85 @@ function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, o
...
+            // eslint-disable-next-line @typescript-eslint/tslint/config
+            let aspNetCoreUrl: string = `      - ASPNETCORE_URLS=http://+:${ports[0]}`;
+            if (ports.length >= 2) {
+                aspNetCoreUrl += `;https://+:${ports[1]}`;
+            }
+            environmentVariables += `\r\n${aspNetCoreUrl}`
+        }
+    }
+
+    let volumesList = '      - ~\\.vsdbg:/remote_debugger';
```
bwateratmsft: Does this get added to both Compose files? We should make sure the debugger mapping only is added to the `docker-compose.debug.yml` file.

src/configureWorkspace/configureDotNetCore.ts
ravipal: yes.

src/configureWorkspace/configureDotNetCore.ts
ravipal: Not that I know of. During launch, it uses the hardcoded value of 80, 443.

src/configureWorkspace/configureDotNetCore.ts
ravipal: Yes it is added only to debug.yml

src/configureWorkspace/configureDotNetCore.ts
bwateratmsft: I mean `getComposePorts` from configure.ts.

PR #1626, opened 2020-02-14
Adds "active use" tracking for survey purposes

philliphoff:
Adds tracking of telemetry for gaps in "active use". (Active use is determined by the reporting of a non-activation-event telemetry.) If the gap is large enough (22 days), a selection of users (25%) will be chosen to receive a prompt asking them to take a Docker extension related survey.


src/telemetry/surveys/activeUseSurvey.ts
```diff
@@ -0,0 +1,193 @@
...
+                                }
+
+                                // Is the user still considered a candidate?
+                                if (isCandidate) {
+                                    sendSurveyTelemetryEvent('surveys.activeUseSurvey.eligibleCandidate');
+
+                                    const response = await promptForSurvey();
+
+                                    // Regardless of the response, user is no longer a candidate.
+                                    await state.update(isCandidateKey, false);
```
bwateratmsft: So no Ask Me Later like the NPS one?

philliphoff: Correct.  The mechanics of how to track when "later" was was more complicated than the need for that option, so we just went with a simple yes/no.

bwateratmsft:
@philliphoff will you merge this forward into `master`?

Issue #1629, opened 2020-02-16
TesteDocker


antoniopereirasilva:
Does this occur consistently? 
Repro steps:

1.
2.

Action: vscode-docker.configure
Error type: Error
Error Message: To generate Docker files you must first open a folder or workspace in VS Code.

Version: 0.10.0
OS: linux
OS Release: 5.3.0-28-generic
Product: Visual Studio Code - Insiders
Product Version: 1.43.0-insider
Language: en

<details>
<summary>Call Stack</summary>

```
Object.t.quickPickWorkspaceFolder extension.bundle.js:1:427813
c extension.bundle.js:317:80786
extension.bundle.js:31:190971
Object.t.scaffold extension.bundle.js:317:81418
t.configure extension.bundle.js:1:246670
extension.bundle.js:317:348968
```

</details>

bwateratmsft:
Same as #1559. @antoniopereirasilva , can you open a workspace folder in VSCode before running the `Configure Workspace...` command?

Issue #1628, opened 2020-02-14
teste docker


wesleyrossetti:
Does this occur consistently? 
Repro steps:

1.
2.

Action: vscode-docker.configure
Error type: Error
Error Message: To generate Docker files you must first open a folder or workspace in VS Code.

Version: 0.10.0
OS: win32
OS Release: 10.0.17763
Product: Visual Studio Code
Product Version: 1.41.1
Language: en

<details>
<summary>Call Stack</summary>

```
Object.t.quickPickWorkspaceFolder extension.bundle.js:1:427875
c extension.bundle.js:317:80786
extension.bundle.js:31:190971extension.bundle.js:31:190971
Object.t.scaffold extension.bundle.js:317:81418
t.configure extension.bundle.js:1:246732
extension.bundle.js:317:348968extension.bundle.js:317:348968
```

</details>

bwateratmsft:
Same as #1559. @wesleyrossetti , can you open a folder or workspace before trying to generate Docker files?

Issue #1559, opened 2020-01-15
bug


daklinux:
Repro steps:

1.
2.

Action: vscode-docker.configure
Error type: Error
Error Message: To generate Docker files you must first open a folder or workspace in VS Code.

Version: 0.9.0
OS: linux
Product: Visual Studio Code
Product Version: 1.41.1
Language: fr

<details>
<summary>Call Stack</summary>

```
Object.t.quickPickWorkspaceFolder extension.bundle.js:1:423182
extension.bundle.js:1:305105
Object.t.captureCancelStep extension.bundle.js:157:373232
S extension.bundle.js:1:299085
t.configure extension.bundle.js:1:305086
extension.bundle.js:324:343471
```

</details>

philliphoff:
@daklinux Did you have a folder open when invoking the command?  Or did you have a workspace open?  (There were a couple known issues with the latter case, that should be fixed in the next release.) . Can you provide repro steps?

PR #1632, opened 2020-02-18
Suppress report issue on Add Docker Files when no workspace is open

bwateratmsft:
We get a fairly large number of spurious issues (e.g. #1559, #1628, #1629, and more) that occur because the user has not opened a workspace folder. This is not a bug and therefore should not have the Report Issue button.


bwateratmsft:
The more I read the text the more I feel like it's already correct. I admit it's hard to convey what a workspace is to absolute beginners (and we only kinda maybe support workspaces), but opening a folder seems pretty straightforward to me, especially when the second item on the welcome page is to open a folder:
image

karolz-ms:
@bwateratmsft sounds good, thanks for sharing your thoughts. Approved.

PR #1645, opened 2020-02-21
Bump compose version to 3.4

haniamr:
Switch from scaffolding 2.1 to 3.4 to be consistent with the VS tooling.


haniamr:
@bwateratmsft I tested Node and Python was already using this version so I suppose it's a safe change, but please let me know if there are certain cases you'd want me to test.

Issue #1496, opened 2019-12-16
How to use vscode-docker as a dependency of other extensions


umarcor:
I'm writing a TS extension that provides some features on top of vscode-docker and I would like to use some of the types and functions defined here.

Precisely, a command is added to `"category": "Docker Images"` and to the context menu (`"when": "view == dockerImages && viewItem == image", "group": "images_1_run@2"`). This is currently handled as follows:

```ts
registerCommand('dockerTemplates.runImage',
	(item) => showWarningMessage(`'runImage' not implemented yet [${item}].`));
```

I would like to define the type of `item` as `ImageTreeItem` instead. The type is defined in subdir `src/tree/images/ImageTreeItem` of this repo. However, it seems not to be possible to add `vscode-docker` as a dependency through `npm` or `yarn`.

Apart from that, when the command is executed directly (not from the context menu), the image (`item`) is `undefined`. For other commands, vscode-docker shows a list with all the available images so that users can pick one. It is a multi-level menu that asks for the name first, and then shows the tags. I'd like to import and reuse this "menu to select an image" feature. Is it possible?

philliphoff:
@umarcor I'd be interested to know what kind of features you are looking to add.

You can define an extension as a required dependency of your extension via the `extensionDependencies` of the `package.json` file.  However, while extensions *can* provide APIs to be consumed or extended by other extensions, `vscode-docker` does not really do so (yet).  VS Code's declarative command definitions also allow arbitrary extensions to "extend" the context menus of other extensions provided you know where to look for views/groups, even if those extensions are not necessarily designed for such extension.  In that case, the extender needs to accept that updates to the underlying extension may change those views/groups in ways that break the dependent extension.

> I would like to define the type of item as ImageTreeItem instead. The type is defined in subdir `src/tree/images/ImageTreeItem` of this repo. However, it seems not to be possible to add `vscode-docker`  as a dependency through `npm` or `yarn`.

Given that `vscode-docker` was not specifically designed to be extended in that way, we don't ship the definitions for types used within the extension.  You could certainly create your own type definitions to match, however.

> Apart from that, when the command is executed directly (not from the context menu), the image (`item`) is `undefined`. For other commands, vscode-docker shows a list with all the available images so that users can pick one. It is a multi-level menu that asks for the name first, and then shows the tags. I'd like to import and reuse this "menu to select an image" feature. Is it possible?

[...]

PR #1649, opened 2020-02-24
Add command customization templates

### LINKED ISSUE #1596: Docker Command Customization
## Overview

The Docker extension executes a number of Docker CLI commands on behalf of its users, such as to build images, run containers, attach to containers, and view container logs. Some of these commands have a large number of optional arguments, often used in very specific scenarios, and our users have stated a need be able to augment or override arguments passed to commands executed by the extension.

To date, the ability to customize the commands executed by the extension has been added piecemeal, with the customization often just appended to the command line generated by the extension. While this is often sufficient, it doesn't allow the user to significantly alter the command, for example, to change arguments or flags set by the extension itself, which may be necessary in a number of user scenarios.

This proposal is intended to:

 - Support greater flexibility in customizion of Docker commands.
 - Ensure such customization is consistent across the breadth of the commands supported by the extension.

## Non-goals

Some Docker interaction is performed via an SDK rather than via the command lines. For example, container start/stop/restart/remove/inspect/prune. Customization of those commands may be considered in a future proposal as it would likely require a different customization scheme, if and when there is demand for it.

## Related issues (not an exhaustive list)
#794
#807
#1197
#1274
#1349
#1505
#1590
#1593

## General Proposal

For each Docker command (and variant), for example, build, run, attach, and view logs, the extension will:

 - Define a configuration setting that represents a "template" of the command line to be executed.
 - The default value of this setting will represent a complete command line for that Docker command (generally corresponding to that currently generated and executed by the extension).
 - The setting may support a set of "tokens" that represent values generated by the extension, which, if present, are substituted with their representative value at the location of the token within the template.

   > These tokens are conceptually similar to those used in `launch.json` and `tasks.json`, for example, `${workspaceFolder}`.

 - The set of tokens supported by a template may vary based on the Docker command.
 - More than one template may be defined for each command; in such cases, each `template` property may be associated with a `label` and/or a `match` property.
    - If a `match` property is specified, those commands supporting heuristic-based selection will use that value to match against values appropriate to the command (for example, the container or image name).

      > The `match` property is interpreted as a JavaScript regular expression.

    - If multiple templates specify a `label` property and no template was otherwise selected based on their `match` properties, those templates will be included in the list (built from the templates' `label` properties) from which the user will be prompted to select a template.
    - If no templates were selected based on their `match` properties and fewer than 2 remaining templates have an associated `label` property, the first such template will be selected.
    - If no templates were selected based on their `match` properties and no other templates have been specified, then the default template for that setting will be selected.

### Settings JSON Schema

VS Code will prompt the user to edit the `settings.json` file when adding or editing these settings. Customizing a command, at a minimum, is just assigning the corresponding setting name a string value that represents the Docker command line to execute.  For example, to customize the Docker build command to set the build argument `NODE_VERSION` to `12.14.1`:

```json
{
    "docker.commands.build": "docker build -f \"${dockerfile}\" -r ${tag} --build-arg NODE_VERSION=12.14.1 \"${context}\""
}
```

In cases where a single template is not sufficient, multiple templates can be specified by assigning the corresponding setting name an array of template objects. For example, to provide two customizations of the Docker build command, to set the `NODE_VERSION` build argument to differing values:

```json
{
    "docker.commands.build": [
        {
            "label": "Build with Node 12",
            "template": "docker build -f \"${dockerfile}\" -r ${tag} --build-arg NODE_VERSION=12.14.1 \"${context}\""
        },
        {
            "label": "Build with Node 13",
            "template": "docker build -f \"${dockerfile}\" -r ${tag} --build-arg NODE_VERSION=13.7.0 \"${context}\""
        }
    ]
}
```

In that scenario, upon invoking the build command, the extension will prompt the user to select a template to use from a list of the template labels.

Suppose the user wants to show all logs for Redis containers but otherwise limit logs for other containers to the last 10 lines.  The user could define two templates, one matching `redis` and another template that serves as the default for all others (by not specifying a `match` or `label` property).

```json
{
    "docker.commands.logs": [
        {
            "match": "redis",
            "template": "docker logs -f ${containerId}"
        },
        {
            "template": "docker logs -tail 10 ${containerId}"
        }
    ]
}
```

### Current Behavior

When building an image, the following command is used:

```bash
docker build --rm -f "<dockerfile>" -t <tag> "<context>"
```

The `<dockerfile>` value is the workspace-relative path of the selected `Dockerfile`.  The `<context>` value is derived from the `docker.imageBuildContextPath` setting or, if not set, defaults to the workspace-relative folder in which the `Dockerfile` resides.

The `<tag>` value is ultimately entered/confirmed by the user when invoking the build command.  However, if previously built, it defaults to the previously-entered value for that `Dockerfile`. Otherwise, it defaults
to `<folder>:latest` where `<folder>` is the parent folder of the `Dockerfile`.

### Proposed Behavior

The new behavior will be to offer the following setting:

| Setting | Default |
|--|--|
| `docker.commands.build` | `docker build --rm -f "${dockerfile}" -t ${tag} "${context}"`

This setting will support the following tokens:

| Token | Description |
| -- | -- |
| `${context}` | If set, the value of the `docker.imageBuildContextPath` setting. Otherwise, the workspace-relative folder in which the `Dockerfile` resides. |
| `${dockerfile}` | The workspace-relative path of the selected `Dockerfile`. |
| `${tag}` | The value entered/confirmed by the user upon invoking the build command. If previously built, defaults to the previously-entered value for that `Dockerfile`. Otherwise, defaults to `<folder>:latest` where `<folder>` is the parent folder of the `Dockerfile`. |

> NOTE: if the `docker.commands.build` setting does not contain the `${tag}` token, the user will *not* be prompted to enter/confirm a tag.

### Current Behavior

When running an image, the following commands are used:

```bash
# Run
docker run --rm -d <ports> <tag>

# Run Interactive
docker run --rm -it <ports> <tag>

# Run Azure CLI
docker run <network> <volumes> -it --rm azuresdk/azure-cli-python:latest
```

The `<tag>` value is the full tag of the selected image. The `<ports>` value is generated from the list of exposed ports in the image (i.e. ultimately from the `Dockerfile`), where each exposed port is mapped to the same port on the local machine. For example, `"EXPOSE 5000 5001"` would generate `"-p 5000:5000 -p 5001:5001"`.

For the Azure CLI, the `<network>` value is `--net=host` on Linux but not set on other platforms. The `<volumes>` value is a mapping of volumes in the container to the user's local `.azure`, `.ssh`, and `.kube` folders (so simplify remote connections to Azure resources). For example, `"-v /Users/<user>/.azure:/root/.azure -v /Users/<user>/.ssh:/root/.ssh -v /Users/<user>/.kube:/root/.kube"`.

### Proposed Behavior

The new behavior will be to offer the following settings:

| Setting | Default |
|--|--|
| `docker.commands.run` | `docker run --rm -d ${exposedPorts} ${tag}` |
| `docker.commands.runInteractive` | `docker run --rm -it ${exposedPorts} ${tag}` |
| `docker.commands.runAzureCli` | `docker run ${network} ${userVolumes} -it --rm azuresdk/azure-cli-python:latest` |

These settings will support the following tokens:

| Token | Description |
| -- | -- |
| `${exposedPorts}` | Generated from the list of exposed ports in the image (i.e. ultimately from the `Dockerfile`), where each exposed port is mapped to the same port on the local machine.  For example, `"EXPOSE 5000 5001"` would generate `"-p 5000:5000 -p 5001:5001"`. |
| `${tag}` | The full tag of the selected image. |
| `${network}` | (Only for Run Azure CLI) On Linux, set to `--net=host`. Not set for other platforms. |
| `${userVolumes}` | (Only for Run Azure CLI) A mapping of volumes in the container to the user's local `.azure`, `.ssh`, and `.kube` folders (so simplify remote connections to Azure resources). For example, `"-v /Users/<user>/.azure:/root/.azure -v /Users/<user>/.ssh:/root/.ssh -v /Users/<user>/.kube:/root/.kube"`. |

### Template Matching

For non-Azure CLI run commands, if there are multiple templates associated with the command, the extension will first look for a template with a `label` property matching the tag of the image being run. If found, that template will be used. If not found, then the user will be prompted to select from the set of templates.

### Current Behavior

When attaching to a running container, the following command is used:

```bash
docker exec -it <containerId> <shellCommand>
```

The `<shellCommand>` value is derived from the value of either the `docker.attachShellCommand.windowsContainer` or `docker.attachShellCommand.linuxContainer` settings.

The defaults for those settings are:

| Setting | Default |
|--|--|
| `docker.attachShellCommand.linuxContainer` | `/bin/sh -c \"[ -e /bin/bash ] && /bin/bash || /bin/sh\"` |
| `docker.attachShellCommand.windowsContainer` | `powershell` |

### Proposed Behavior

The new behavior will be to offer the following setting:

| Setting | Default |
|--|--|
| `docker.commands.attach` | `docker exec -it ${containerId} ${shellCommand}`

This setting will support the following tokens:

| Token | Description |
| -- | -- |
| `${containerId}` | The ID of the container to attach to. |
| `${shellCommand}` | The value of the `docker.attachShellCommand.linuxContainer` or `docker.attachShellCommand.windowsContainer` setting, as appropriate. |

### Template Matching

If there are multiple templates associated with the command, the extension will first look for a template with a `label` property matching the name of the container being attached to. If found, that template will be used. If not found, the extension will then look for a template with a `label` property matching the tag of the image of the container being attached to. If not found, then the user will be prompted to select from the set of templates.

### Current Behavior

When viewing logs of a container, the following command is used:

```bash
docker logs -f <containerId>
```

### Proposed Behavior

The new behavior will be to offer the following setting:

| Setting | Default |
|--|--|
| `docker.commands.logs` | `docker logs -f ${containerId}`

This setting will support the following tokens:

| Token | Description |
| -- | -- |
| `${containerId}` | The ID of the container to view the logs for. |

### Template Matching

If there are multiple templates associated with the command, the extension will first look for a template with a `label` property matching the name of the container being attached to. If found, that template will be used. If not found, the extension will then look for a template with a `label` property matching the tag of the image of the container being attached to. If not found, then the user will be prompted to select from the set of templates.

### Current Behavior

When bringing up a composition, the following command is used:

```bash
docker-compose -f "<file>" up <detached> <build>
```

The `<file>` value is the workspace-relative path to the selected Docker Compose YAML file. The `<detached>` value is set to `-d` if the setting `docker.dockerComposeDetached` is set to `true`. The `<build>` value is set to `--build` if the setting `docker.dockerComposeBuild` is set to `true`.

### Proposed Behavior

The new behavior will be to offer the following setting:

| Setting | Default |
|--|--|
| `docker.commands.composeUp` | `docker-compose ${configurationFile} up ${detached} ${build}` |

This setting will support the following tokens:

| Token | Description |
| -- | -- |
| `${configurationFile}` | Set to `-f ` plus the workspace-relative path to the selected Docker Compose YAML file. |
| `${detached}` | Set to `-d` if the setting `docker.dockerComposeDetached` is set to `true`. Otherwise, set to `""`. |
| `${build}` | Set to `--build` if the setting `docker.dockerComposeBuild` is set to `true`. Otherwise, set to `""`. |

### Current Behavior

When taking down a composition, the following command is used:

```bash
docker-compose -f "<file>" down
```

### Proposed Behavior

The new behavior will be to offer the following setting:

| Setting | Default |
|--|--|
| `docker.commands.composeDown` | `docker-compose ${configurationFile} down` |

This setting will support the following tokens:

| Token | Description |
| -- | -- |
| `${configurationFile}` | Set to `-f ` plus the workspace-relative path to the selected Docker Compose YAML file. |

bwateratmsft:
Resolves #1596, maybe also #1496. This adds fairly unlimited command customization capabilities for the commands that need it most:

- Docker build
- Docker run
- Docker run (interactive)
- Attach
- View logs
- Compose up
- Compose down


bwateratmsft:
The changes aren't fully tested, but since there's a fair bit I wanted to get eyes on them as early as possible.

package.json
```diff
@@ -2193,5 +2410,24 @@
         "vscode-azureextensionui": "^0.29.5",
         "vscode-languageclient": "^6.1.0",
         "xml2js": "^0.4.22"
+    },
+    "commandsItemTemplate": {
```
bwateratmsft: Need to remove this temp code

src/commands/selectTemplate.ts
```diff
@@ -0,0 +1,108 @@
...
+            detail: template.template,
+            data: template,
+        }
+    });
+
+    const selection = await ext.ui.showQuickPick(items, {
+        placeHolder: 'Choose a command template to execute'
+    });
+
+    if (!selection) {
```
bwateratmsft: `ext.ui.showQuickPick` automatically throws `UserCancelledError` so this isn't needed.

src/commands/selectTemplate.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+    });
+
+    let selectedTemplate: CommandTemplate;
+
+    if (templates.length === 0) {
+        // No templates matched, so fall back to extension default
+        selectedTemplate = defaults[command];
+    } else {
+        selectedTemplate = await quickPickTemplate(context, templates);
```
philliphoff: What if the user has templates that are both "matching" as well as "labeled"?  In this implementation, matching templates that don't match aren't considered part of the set of selectable templates.  I can see a scenario where one template has a match property as a kind of "hint" for a well known set of images/containers, but that same template might still be desired for non-matching images/containers.

I think we should attempt to match first and, if no match is found, prompt the user to select a template from the set of all explicitly labeled templates (perhaps also adding the extension/user default).  (I.e. if the user wants a template to *only* be used for a specific match, then she would omit the label property.)

src/commands/selectTemplate.ts
```diff
@@ -0,0 +1,107 @@
...
+    'build': { label: 'Docker Build', template: 'docker build --rm -f "${dockerfile}" -t ${tag} "${context}"' },
+    'run': { label: 'Docker Run', template: 'docker run --rm -d ${exposedPorts} ${tag}' },
+    'runInteractive': { label: 'Docker Run (Interactive)', template: 'docker run --rm -it ${exposedPorts} ${tag}' },
+    'attach': { label: 'Docker Attach', template: 'docker exec -it ${containerId} ${shellCommand}' },
+    'logs': { label: 'Docker Logs', template: 'docker logs -f ${containerId}' },
+    'composeUp': { label: 'Compose Up', template: 'docker-compose up ${configurationFile} ${detached} ${build}' },
+    'composeDown': { label: 'Compose Down', template: 'docker-compose down ${configurationFile}' },
+};
+
+export async function selectTemplate(context: IActionContext, command: TemplateCommand, matchContext?: string, folder?: vscode.WorkspaceFolder, additionalVariables?: { [key: string]: string }): Promise<string> {
```
philliphoff: Should `matchContext` be an array of strings?  While unlikely, there's a chance that a matching template could match content that overlaps individual context values.  It seems like it'd be better to test each piece of context individually.

src/utils/resolveVariables.ts
```diff
@@ -4,44 +4,78 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as path from 'path';
-import { window, WorkspaceFolder } from 'vscode';
+import { window, workspace, WorkspaceFolder } from 'vscode';
 import { cloneObject } from '../utils/cloneObject';
 
-export function resolveVariables<T>(target: T, folder: WorkspaceFolder): T {
+const variableMatcher: RegExp = /\$\{[a-z\.\-_:]+\}/ig;
+const configVariableMatcher: RegExp = /\$\{config:([a-z\.\-_]+)\}/i;
```
philliphoff: No `g` flag?

src/commands/containers/viewContainerLogs.ts
```diff
@@ -15,7 +16,15 @@ export async function viewContainerLogs(context: IActionContext, node?: Containe
         });
     }
 
+    const terminalCommand = await selectTemplate(
+        context,
+        'attach',
```
philliphoff: Should this be `'logs'`?

src/commands/selectTemplate.ts
philliphoff: `selectCommandTemplate()`?  (Even then, it's not really selecting a template--it's returning a command line derived from a selected command template.)

src/commands/containers/viewContainerLogs.ts
```diff
@@ -15,7 +16,15 @@ export async function viewContainerLogs(context: IActionContext, node?: Containe
...
+        `${node.containerName} ${node.fullTag}`,
+        undefined,
+        { 'containerId': node.containerId }
```
philliphoff: There's a strong coupling between the variables provided for replacement and those documented/expected for each command.  Having the clients of `selectTemplate()` pass in the variables by name allows for easy misspelling or accidental omission, as well as requires every use of the template to pass in the (hopefully) same dictionary.

Why not create functions for each command (e.g. `selectLogsTemplate(containerId: string): Promise<string>`) that wraps `selectTemplate()`, allowing easy reuse, keeps the translation to variables closer to their substitution (to minimize the chance of misspelling), as well as enforces passing the expected variable values.

src/commands/selectTemplate.ts
bwateratmsft: I don't think using the label for anything other than display makes sense (i.e. it should not be used for making decisions on whether or not to display a template, users will find that confusing). Also, I think the `match` property _should_ be used to exclude non-matching templates. If they want the same command used for non-matching images, they should define another template with different / no `match` property.

I can see an example where my implementation gives sub-ideal behavior. Suppose the user has two templates, one for MySQL (with a `match`), and one for everything else (without a `match`). If the user right-clicks and runs a MySQL image, they will get prompted to choose between the two templates, even though it's obvious that they would want the MySQL one to be run.

So, I think the right behavior is this:

1. Look for any user-defined matching templates with `match` specified. If multiple, prompt; if none, move to step 2.
2. Look for any user-defined templates with `match` unspecified. If multiple, prompt; if none, move to step 3.
3. Fall back to default (i.e. hard-coded template). (my implementation does this part now)

src/commands/containers/viewContainerLogs.ts
bwateratmsft: Sure; though I think that `resolveVariables` should keep the behavior it has of accepting a dictionary. Would these `selectLogsTemplate` etc. methods be in this selectTemplate.ts file?

src/utils/resolveVariables.ts
bwateratmsft: For this, yes. `variableMatcher` is used to find all instances of anything that looks like `${something}`, including `${config:something.something}`, within the entire input string. `configVariableMatcher` is used only to match against a specific single variable--so there's nothing before or after it--and also, to separate the `something.something` config parameter into a regex group. `g` isn't needed here since we aren't looking for multiple `${config:something.something}` entries within the string.

src/commands/selectTemplate.ts
philliphoff: I think that's ok behavior.  Should we just require label properties to help in the case where multiple, matching templates are found (rather than, say, defaulting to the template itself as the label, which won't necessarily make for the nicest looking prompt)?

src/commands/containers/viewContainerLogs.ts
philliphoff: Agreed on `resolveVariables()` itself accepting a dictionary.  I think it'd be fine for `selectLogsTemplate()` and company to be in the same file.  (You could just keep `selectTemplate()` as non-exported to prevent nefarious use outside of the file.)

src/commands/selectTemplate.ts
bwateratmsft: I'm fine with that, I'll make it so.

src/commands/selectCommandTemplate.ts
```diff
@@ -0,0 +1,157 @@
...
+    let selectedTemplate: CommandTemplate;
+    if (matchedTemplates.length > 0) {
+        selectedTemplate = await quickPickTemplate(context, matchedTemplates);
+    } else if (universalTemplates.length > 0) {
+        selectedTemplate = await quickPickTemplate(context, universalTemplates);
+    } else {
+        selectedTemplate = defaults[command];
+    }
+
+    context.telemetry.properties.isDefaultCommand = selectedTemplate.template === defaults[command].template ? 'true' : 'false';
```
philliphoff: Note this would be `true` both if the user did not customize the command (and is using the default command) and if the user *did* customize the command but happened to set it to the same value as the default command.  I'm not sure if we care about that distinction.

bwateratmsft: Yeah, that was intentional on my part, as it is not possible to distinguish whether the setting value came from extension defaults, or if it was a user-provided value that happened to be identical to extension defaults. I don't imagine many users will be providing values that are identical to defaults as that is redundant.

src/commands/selectCommandTemplate.ts
```diff
@@ -0,0 +1,157 @@
...
+    let portsString: string = '';
+    if (exposedPorts) {
+        portsString = Object.keys(exposedPorts).reduce((partialPortsString: string, portAndProtocol: string) => {
+            return `${partialPortsString} -p ${portAndProtocol.split('/')[0]}:${portAndProtocol}`
+        }, portsString);
+    }
+
+    return await selectCommandTemplate(
+        context,
+        interactive ? 'runInteractive' : 'run',
```
philliphoff: Do we care about the "run Azure CLI" flavor?

bwateratmsft: I didn't implement it in settings either; I'm not sure why users would need to customize it...

src/commands/images/runImage.ts
```diff
@@ -19,21 +20,21 @@ async function runImageCore(context: IActionContext, node: ImageTreeItem | undef
...
 
-    let options = `--rm ${interactive ? '-it' : '-d'}`;
-    if (ports.length) {
-        const portMappings = ports.map((port) => `-p ${port.split("/")[0]}:${port}`); // 'port' is of the form number/protocol, eg. 8080/udp.
-        // In the command, the host port has just the number (mentioned in the EXPOSE step), while the destination port can specify the protocol too
-        options += ` ${portMappings.join(' ')}`;
-    }
+    const terminalCommand = await selectRunCommand(
+        context,
+        [node.fullTag],
```
philliphoff: The contexts are part of the customization contract, so I think it makes sense to simply make them a required argument to the `selectXxxCommand()` function.  (This is especially easy if the context is already a required argument, such as here.)

src/commands/selectCommandTemplate.ts
philliphoff: Agree it's not a P1 scenario, but I can imagine that users may want to, for example, volume mount some directory full of ARM templates/scripts.

bwateratmsft: I guess a potential workaround would be to pull the Azure CLI image and define a custom Run Interactive command scoped to that image. You would have to go through the explorer / command palette + select, rather than just command palette, but it's something users will be able to do immediately upon release.

FelipeMartini:
how use this? where  i configure these commands options?

karolz-ms:
@FelipeMartini https://code.visualstudio.com/docs/containers/reference#_command-customization

FelipeMartini:
> @FelipeMartini https://code.visualstudio.com/docs/containers/reference#_command-customization

thanks brow, i got this working for build and for logs command, also is possible to add another command like start and stop? or only these are configurable, build, logs, attach... but stop and start i not see if its possible too, do you know that @karolz-ms ???

karolz-ms:
> thanks brow, i got this working for build and for logs command, also is possible to add another command like start and stop? or only these are configurable, build, logs, attach... but stop and start i not see if its possible too, do you know that @karolz-ms ???

Only those commands that are documented are customizable through templates, sorry. What I do for the others, if necessary, is to create a VS Code custom task https://code.visualstudio.com/docs/editor/tasks#_custom-tasks Together with a keyboard shortcut for `workbench.actions.tasks.runTask` (I use `alt+t`) that works quite well.  

Also note that Docker extensions adds special tasks for Docker build, Docker run, and Docker Compose, it is all described in https://code.visualstudio.com/docs/containers/reference

PR #1651, opened 2020-02-24
Fix GitLab repository missing name

### LINKED ISSUE #1643: Missing name of repository in GitLab Docker Registry
I'm using the GitLab registry and I noticed that the name of the current repository from a given registry is missing.

And this behavior can be observed only for GitLab, DockerHub works like a charm.

I'll provide some images so that it's more clear.

Observation:
Image of observed defect

Current state in GitLab:
Current state in GitLab

The same thing, but in DockerHub:
Current state in GitLab

danielgospodinow:
Closes #1643

### DIFF
--- a/src/tree/registries/gitLab/GitLabRepositoryTreeItem.ts
+++ b/src/tree/registries/gitLab/GitLabRepositoryTreeItem.ts
@@ -17,7 +17,9 @@ export class GitLabRepositoryTreeItem extends RemoteRepositoryTreeItemBase {
     private _nextLink?: string;
 
     public constructor(parent: GitLabProjectTreeItem, id: string, name: string) {
-        super(parent, name);
+        // GitLab returns an empty repository name,
+        // if the project's namespace is the same as the repository
+        super(parent, name || parent.label);
         this.repoId = id;
     }
 


src/tree/registries/gitLab/GitLabRepositoryTreeItem.ts
```diff
@@ -17,6 +17,10 @@ export class GitLabRepositoryTreeItem extends RemoteRepositoryTreeItemBase {
     private _nextLink?: string;
 
     public constructor(parent: GitLabProjectTreeItem, id: string, name: string) {
+        if (!name) {
+            name = parent.label;
+        }
+
         super(parent, name);
```
bwateratmsft: An alternative to the `if` block would be this:
`super(parent, name || parent.label)`

This is saying that if `name` is any falsy value (`undefined`, empty string), then use `parent.label`.

src/tree/registries/gitLab/GitLabRepositoryTreeItem.ts
```diff
@@ -17,6 +17,10 @@ export class GitLabRepositoryTreeItem extends RemoteRepositoryTreeItemBase {
     private _nextLink?: string;
 
     public constructor(parent: GitLabProjectTreeItem, id: string, name: string) {
+        if (!name) {
```
bwateratmsft: Can you add a comment as to why? e.g.
`// GitLab returns repositories with empty names, if the project's namespace is used as the repository`

src/tree/registries/gitLab/GitLabRepositoryTreeItem.ts
danielgospodinow: Oh, it really looks much better. :+1: I'll refactor this.

bwateratmsft:
I've merged it in, thanks @danielgospodinow!

danielgospodinow:
It was a pleasure, thanks you! :smile:

PR #1652, opened 2020-02-25
Create Attach config for dotnet core compose

### LINKED ISSUE #1543: [.NET Core] Add Docker Compose support

ravipal:
When compose is selected during scaffolding, the 'Docker .NET Core Attach (Preview)' debug config is added automatically.

fixes #1543

### DIFF
--- a/src/debugging/netcore/NetCoreDebugHelper.ts
+++ b/src/debugging/netcore/NetCoreDebugHelper.ts
@@ -88,18 +88,32 @@ export class NetCoreDebugHelper implements DebugHelper {
     public async provideDebugConfigurations(context: DockerDebugScaffoldContext, options?: NetCoreDebugScaffoldingOptions): Promise<DockerDebugConfiguration[]> {
         options = options || {};
         options.appProject = options.appProject || await NetCoreTaskHelper.inferAppProject(context.folder); // This method internally checks the user-defined input first
-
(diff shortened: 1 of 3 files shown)


src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -88,18 +88,32 @@ export class NetCoreDebugHelper implements DebugHelper {
...
-        return [
-            {
-                name: 'Docker .NET Core Launch',
-                type: 'docker',
-                request: 'launch',
-                preLaunchTask: 'docker-run: debug',
-                netCore: {
-                    appProject: unresolveWorkspaceFolder(options.appProject, context.folder)
-                }
+        let debugConfigurations : DockerDebugConfiguration[] = [{
```
bwateratmsft: I would `if`/`else` this. The .NET Core launch config probably does not work in the expected way if a compose file is present (i.e. it will not use compose at all, users might be confused by this). Also, if compose I would also not add the `docker-build` / `docker-run` tasks.

ravipal: I don't get it. Why would single not work if compose file is present?

PR #1658, opened 2020-02-25
Look in Docker logs to determine if the server is ready

haniamr:
- Adds a `DockerLogsTracker` class that looks for the pattern in "docker logs"
- Adds `MultiOutputDockerServerReadyManager` which is essentially a combination of looking in the debug console and docker logs
- Some simple refactoring

For the most (if not all) cases, the pattern will be found in "docker logs" before it ever reaches the debug console. However, it might be safer to keep looking in the debug console just to cover cases that we might not know about.

These are pretty much @philliphoff's changes from that branch: https://github.com/microsoft/vscode-docker/compare/philliphoff-docker-logs

### DIFF
--- a/src/debugging/DockerServerReadyAction.ts
+++ b/src/debugging/DockerServerReadyAction.ts
@@ -25,28 +25,17 @@ const URI_FORMAT = 'http://localhost:%s';
 /* eslint-disable-next-line no-template-curly-in-string */
 const WEB_ROOT = '${workspaceFolder}';
 
-export class ServerReadyDetector extends vscode.Disposable implements DockerServerReadyManager {
+export class ServerReadyDetector implements DockerServerReadyDetector {
     private hasFired: boolean = false;
     private regexp: RegExp;
-    private disposables: vscode.Disposable[] = [];
 
     public constructor(private session: vscode.DebugSession) {
-        super(() => this.internalDispose());
-
         const configuration = <ResolvedDebugConfiguration>session.configuration;
 
         this.regexp = new RegExp(
-            (configuration
-                && configuration.dockerOptions
-                && configuration.dockerOptions.dockerServerReadyAction
-                && configuration.dockerOptions.dockerServerReadyAction.pattern)
+            (configuration?.dockerOptions?.dockerServerReadyAction?.pattern)
             || PATTERN,
             'i');
-    }
-
-    private internalDispose(): void {
-        this.disposables.forEach(d => { d.dispose(); });
-        this.disposables = [];
     }
 
     public detectPattern(s: string): boolean {
@@ -56,7 +45,6 @@ export class ServerReadyDetector extends vscode.Disposable implements DockerServ
                 /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
                 this.openExternalWithString(this.session, matches.length > 1 ? matches[1] : '');
                 this.hasFired = true;
-                this.internalDispose();
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/debugging/DockerServerReadyAction.ts
```diff
@@ -213,54 +190,187 @@ export class ServerReadyDetector extends vscode.Disposable {
...
-                };
+class DockerServerReadyDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
+    private readonly trackers: Set<string> = new Set<string>();
+    private readonly trackerFactory: DockerDebugAdapterTrackerFactory = new DockerDebugAdapterTrackerFactory();
+
+    public constructor(private readonly context: vscode.ExtensionContext) {
+    }
+
+    public resolveDebugConfiguration?(folder: vscode.WorkspaceFolder | undefined, debugConfiguration: ResolvedDebugConfiguration, token?: vscode.CancellationToken): vscode.ProviderResult<vscode.DebugConfiguration> {
+        if (debugConfiguration && debugConfiguration.type && debugConfiguration.dockerOptions && debugConfiguration.dockerOptions.dockerServerReadyAction) {
```
philliphoff: We should be able to now consolidate these long expressions (here and elsewhere), e.g.:

`if (debugConfiguration?.type && debugConfiguration?.dockerOptions?.dockerServerReadyAction) {`

src/debugging/DockerServerReadyAction.ts
```diff
@@ -213,54 +190,187 @@ export class ServerReadyDetector extends vscode.Disposable {
...
+                    this.logStream = <LogStream>stream;
+                    this.logStream.on('data', (data) => {
+                        // tslint:disable-next-line:no-unsafe-any
+                        detector.detectPattern(data.toString());
+                    });
+                });
+        } catch { }
+    }
+
+    private disposeStream() : void {
```
philliphoff: Nit: Since this is only called from one place (in `super()`), the implementation could just be moved there.

src/debugging/DockerServerReadyAction.ts
```diff
@@ -24,37 +25,12 @@ const URI_FORMAT = 'http://localhost:%s';
 /* eslint-disable-next-line no-template-curly-in-string */
 const WEB_ROOT = '${workspaceFolder}';
 
-export class ServerReadyDetector extends vscode.Disposable {
-    private static detectors: Map<vscode.DebugSession, ServerReadyDetector> = new Map<vscode.DebugSession, ServerReadyDetector>();
-
+export class ServerReadyDetector extends vscode.Disposable implements DockerServerReadyManager {
```
philliphoff: I don't think this type needs to be disposable; it doesn't seem to store any private disposable fields.

src/debugging/DockerServerReadyAction.ts
```diff
@@ -213,54 +190,187 @@ export class ServerReadyDetector extends vscode.Disposable {
...
+type DebugAdapterMessage = {
+    body?: {
+        category?: string;
+        output?: string;
+    };
+    type?: string;
+    event?: string;
+};
+
+interface DockerServerReadyManager {
```
philliphoff: I kind of want to rename this something like `DockerServerReadyDetector`; "manager" doesn't really help me understand the difference between the two implementations.

src/debugging/DockerServerReadyAction.ts
haniamr: Yep, turns out the disposable[] is always empty so removed all of that.

Issue #1634, opened 2020-02-18
Unable to debug ASP.NET Core app on Ubuntu 18.04 LTS (client: Windows 10)


karolz-ms:
Repro steps:

One needs an Ubuntu 18.04 machine with Docker installed (used Docker engine version 19.03.6). Machine should be accessible via SSH. Also install .NET SDK version 3.1 or newer (used SDK 3.1.101).

1. Use Remote-SSH extension to connect to the Ubuntu machine via SSH
2. Follow ASP.NET quickstart https://code.visualstudio.com/docs/containers/quickstart-aspnet-core and try debugging

Expected: debugging should work
Actual: an error message that says "Unable to acquire the .NET Core debugger"

bwateratmsft:
I'm also able to reproduce; even with having installed the C# and Docker extensions within the SSH session.

bwateratmsft:
I found that the debug download code correctly identified the OS as Linux, and used the following command:
`/home/bwaterloo/.vsdbg/getvsdbg.sh -v latest -r linux-x64 -l "/home/bwaterloo/.vsdbg/linux-x64/latest"`
When I ran that manually it gave an error that `unzip` was not installed. Once I installed `unzip` (and separately, reinstalled the not-snap version of Docker due to https://github.com/moby/moby/issues/34427), everything worked.

IMO this issue probably belongs to the vsdbg team; they can decide whether to auto-install `unzip` or what the official guidance should be.

bwateratmsft:
I sent an email to the debugger folks to see what their guidance is.

karolz-ms:
The guidance is the user should ensure that unzip utility is installed. 

Per @bwateratmsft 
> The “unzip is not here” message was really clear from the output of GetVsDbg.sh when I ran it manually, but our extension does not emit much of anything—no command line, no install output, etc. I think there’s definitely room for us to improve with that. IMO we should move that command execution to terminal, if possible.

PR #1671, opened 2020-02-26
Show vsdbg install progress

ravipal:
Capture and output of docker exec and show it in the output during vsdbg installation.

### DIFF
--- a/src/debugging/coreclr/CliDockerClient.ts
+++ b/src/debugging/coreclr/CliDockerClient.ts
@@ -4,7 +4,7 @@
 
 import { CommandLineBuilder } from "../../utils/commandLineBuilder";
 import { spawnAsync } from "../../utils/spawnAsync";
-import { ProcessProvider } from "./ChildProcessProvider";
+import { ProcessProvider, ProcessProviderExecOptions } from "./ChildProcessProvider";
 import { LineSplitter } from "./lineSplitter";
 
 export type DockerBuildImageOptions = {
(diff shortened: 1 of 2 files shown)


bwateratmsft:
I think this also fixes #1634. @ravipal is the goal to get it in for 1.0 or OK to wait for 1.1?

ravipal:
> I think this also fixes #1634. @ravipal is the goal to get it in for 1.0 or OK to wait for 1.1?

I don't think fixes #1634, this will show the actual error to user.
Yes, this should go into 1.0

src/debugging/coreclr/CliDockerClient.ts
```diff
@@ -264,18 +264,18 @@ export class CliDockerClient implements DockerClient {
         return id.substring(0, 12);
     }
 
-    public async exec(containerNameOrId: string, args: string, options?: DockerExecOptions): Promise<string> {
-        options = options || {};
+    public async exec(containerNameOrId: string, args: string, dockerExecOptions?: DockerExecOptions, processExecOptions?: ProcessProviderExecOptions): Promise<string> {
```
philliphoff: I'm not a fan of pushing the `ProcessProviderExecOptions` through the Docker client (as that's a leaky abstraction--who says the Docker client implementation even uses child processes as opposed to, say, Dockerode?).

What's the underlying issue?  Is it that `processProvider.exec()` is throwing on failure rather than returning the `stdout`?  If we don't want to change the semantics of `dockerClient.exec()` (in that it throws on error rather than return the `stdout`) then I think we should add an optional `progress` property to the `DockerExecOptions` for callers which want the `stdout` up to the point of failure.

ravipal: Yes the issue is that processProvider.exec() throws on failure. I will make the progress as property of DockerExecOptions

bwateratmsft:
The issue in #1634 was really that no useful output whatsoever was provided by the extension if `GetVsDbg.sh` failed. This is related but since this is to fix the .NET Core attach scenario, (where `GetVsDbg.sh` is running in the container, as opposed to locally), you're correct that this PR won't fix #1634.

Issue #1665, opened 2020-02-26
Fail to run 'Docker .Net Core Attach (Preview)' configuration on Mac


v-ruizh:
**OS:** Mac
**Build Version:** [20200225.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=15184&view=artifacts&type=publishedArtifacts)
**.NET Core SDK:** 2.1.607
**Regression:** Not a Regression (0.10.0)

**Repro Steps:**
1. Create a .NET Core Console app by "dotnet new mvc --name mvctest".
2. Open the project in VS Code --> Prompt to add the required assets --> Yes
3. F1 --> Type and select "Docker: Add Docker Files to Workspace" --> ASP.NET Core --> Linux --> Include optional Docker Compose files.
4. Right click the docker-compose.yml --> Select "Compose Up".
5. Open launch.json and click 'Add Configration…', and choose 'Docker .NET Core Attach (Preview)'.
6. Run the above configuration, select the running container from step4 and click yes to allow installing vsdbg on the container.

**Expect:**
Run the configuration successfully without any error.

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/45614674/75346967-9fd11b80-5854-11ea-83b6-c5e8413da166.png)

**More Info:**
This issue does not reproduce on Windows with .NET Core 3.1 SDK installed.

philliphoff:
@v-ruizh I'm not able to reproduce this on Mac using .NET Core 2.1 (2.1.16 in the container) nor with the latest .NET Core 3.1.  Does this reproduce every time for you?

v-ruizh:
This issue does not reproduce on the new build 20200227.1 with Linux image Ubuntu/Debian. But it also reproduces with Linux image 2.1-alpine. The error is "Process exited with code 127".

ravipal:
I tried both 3.1-alpine and 2.1-alpine from windows host machine and it works for me.
@v-ruizh can you please share the logs from output window? The latest build should log the output of the debugger installation on container. Also did you create the mvc project targeting 2.1 framework using  `-f netcoreapp2.1` ?

philliphoff:
I'm seeing the "127" error on Mac.  I'm continuing to look into it.

philliphoff:
I think it's a Mac vs Windows shell quoting issue.  This doesn't work (what we try to run):

```bash
[...]

PR #1676, opened 2020-02-27
Use proper quote based on host os while installing vsdbg

ravipal:
Mac and Linux require single quote while executing the script on the target container. Handling it property now.

Fixes: #1665

### DIFF
--- a/src/debugging/netcore/NetCoreDebugHelper.ts
+++ b/src/debugging/netcore/NetCoreDebugHelper.ts
@@ -285,7 +285,11 @@ export class NetCoreDebugHelper implements DebugHelper {
         }
 
         const debuggerPath: string = '/remote_debugger';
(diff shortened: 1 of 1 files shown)


src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -285,7 +285,11 @@ export class NetCoreDebugHelper implements DebugHelper {
         }
 
         const debuggerPath: string = '/remote_debugger';
-        const installDebugger: string = `/bin/sh -c "ID=default; if [ -e /etc/os-release ]; then . /etc/os-release; fi; echo $ID; if [ $ID == alpine ]; then apk --no-cache add curl && curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${debuggerPath} ; else apt-get update && apt-get install unzip && curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${debuggerPath}; fi"`
+        // Windows require double quotes and Mac and Linux require single quote.
+        const osProvider = new LocalOSProvider();
+        const installDebugger: string = osProvider.os === 'Windows' ?
```
bwateratmsft: Would prefer `os.platform() === 'win32'` to be consistent with the other newer code on tasks/debugging.

ravipal: I was following this debugging code https://github.com/microsoft/vscode-docker/blob/aa268845b800e2d833dc03dbc8eb23b2c8cb97e7/src/debugging/coreclr/debuggerClient.ts#L40.

PR #1682, opened 2020-02-28
Replace hardcoded strings

### LINKED ISSUE #1612: Remove hardcoded strings to enable localization
Lots and lots of hardcoded strings; we should move these out to enable localization.

bwateratmsft:
Fixes #1612


bwateratmsft:
@ravipal @haniamr @philliphoff @karolz-ms I hope you guys enjoy code-reviewing these mindless, repetitive changes as much as I enjoyed making them. 🤣

package.nls.json
```diff
@@ -0,0 +1,246 @@
+{
+    "vscode-docker.description": "Makes it easy to create, manage, and debug containerized applications.",
```
bwateratmsft: Need to validate that there are no duplicate IDs, here and in the source files themselves.

philliphoff: Yeah, there is a definite need for a NLS extension to help with these sorts of editing and validation tasks; it's still very much a mysterious process to me.  Maybe if we can find a way to tie that to containers, management will let us work on one.  :-)

karolz-ms:
😲 ☺️

src/debugging/coreclr/debugSessionManager.ts
```diff
@@ -28,7 +29,7 @@ export class DockerDebugSessionManager implements DebugSessionManager, vscode.Di
                 () => {
                     this.dockerManager
                         .cleanupAfterLaunch()
-                        .catch(reason => console.log(`Unable to clean up Docker images after launch: ${reason}`));
+                        .catch(reason => console.log(localize('vscode-docker.debug.coreclr.cleanupFailed', 'Unable to clean up Docker images after launch: {0}', reason as string)));
```
philliphoff: Are we sure this is a string?

package.nls.json
bwateratmsft: Validated and found two duplicate IDs, now fixed.

PR #1690, opened 2020-03-02
Switch to using the Flask CLI as the entrypoint for flask apps

haniamr:
Switch to using the [Flask CLI](https://flask.palletsprojects.com/en/1.1.x/cli/) since it seems to be the modern way to run Flask apps and to ensure consistency with the Python extension.


haniamr:
> @haniamr does this need to go into 1.0.0? If so we'll need to make a rel/1.0.0 branch and do a dual-checkin.

Yep, I've created the release branch and will cherry-pick the changes there.

PR #1703, opened 2020-03-03
[Python] Fix for the compose image name + temp dir on Mac

haniamr:
- Fixes the generated image name in the main compose file
- Fixes the temp directory path issue on Mac

### DIFF
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -276,8 +276,7 @@ function validateForUnresolvedToken(contents: string): void {
 }
 
 function generateComposeFiles(dockerfileName: string, platform: Platform, os: PlatformOS | undefined, ports: number[], artifactName: string): ScaffoldFile[] {
-    const serviceName = path.basename(artifactName, path.extname(artifactName)).toLowerCase();
-    const imageName = serviceName;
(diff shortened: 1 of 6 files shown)


bwateratmsft:
Python changes all look good to me. Holding off until .NET Core thing added. :)

src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -306,14 +307,17 @@ function generateComposeFiles(dockerfileName: string, platform: Platform, os: Pl
     // Ensure the path scaffolded in the Dockerfile uses POSIX separators (which work on both Linux and Windows).
     dockerfileName = dockerfileName.replace(/\\/g, '/');
 
-    let composeFileContent = dotNetComposeTemplate.replace('$service_name$', serviceName)
-        .replace(/\$image_name\$/g, imageName)
+    const normalizedServiceName = getValidImageName(serviceName);
```
ravipal: Line 279 has toLower() which is not needed anymore.
Also you can simply reuse the normalizedServiceName for the normalizedImagename

haniamr: Oh I see, I was going to ask you if they're any different. Thanks, I'll do that change!

Issue #1696, opened 2020-03-03
The .NET Core docker-compose.debug.yml template maps the debugger but the attach config still installs it anyway


bwateratmsft:
When compose-up'ing the docker-compose.debug.yml, it will map vsdbg into the container. However, it still gets installed by the attach configuration which is not necessary. It also pollutes the directory on the host, because it is mapped with `rw` permissions.

bwateratmsft:
Fixed by #1704

PR #1704, opened 2020-03-04
Reuse an existing debugger if present

ravipal:
Check if the debugger is already installed on a container by checking the presence of the file vsdbg. If so, then skip the debugger installation.

Fixes: #1696

### DIFF
--- a/src/debugging/netcore/NetCoreDebugHelper.ts
+++ b/src/debugging/netcore/NetCoreDebugHelper.ts
@@ -189,8 +189,17 @@ export class NetCoreDebugHelper implements DebugHelper {
         // Get Container Name if missing
         const containerName: string = debugConfiguration.containerName ?? await this.getContainerNameToAttach();
 
+        let debuggerPath: string = debugConfiguration.netCore?.debuggerPath;
+
         // If debugger path is not specified, install the debugger
-        const debuggerPath: string = debugConfiguration.netCore?.debuggerPath ?? await this.installDebuggerInContainer(containerName);
+        if (!debuggerPath) {
+            const debuggerDirectory = '/remote_debugger';
+            debuggerPath = `${debuggerDirectory}/vsdbg`;
+            const isDebuggerInstalled: boolean = await this.isDebuggerInstalled(containerName, debuggerPath);
+            if (!isDebuggerInstalled) {
+                await this.installDebuggerInContainer(containerName, debuggerDirectory);
+            }
+        }
 
         return {
             ...debugConfiguration, // Gets things like name, preLaunchTask, serverReadyAction, etc.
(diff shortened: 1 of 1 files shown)


src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -277,19 +286,17 @@ export class NetCoreDebugHelper implements DebugHelper {
         return pathNormalize(result, platformOS);
     }
 
-    private async installDebuggerInContainer(containerName: string): Promise<string> {
+    private async installDebuggerInContainer(containerName: string, installPath: string): Promise<string> {
         const yesItem: MessageItem = DialogResponses.yes;
         const install = (yesItem === await window.showInformationMessage('Attaching to container requires .NET Core debugger in the container. Do you want to install debugger in the container?', ...[DialogResponses.yes, DialogResponses.no]));
         if (!install) {
             throw new UserCancelledError("User didn't grand permission to install .NET Core debugger.");
```
haniamr: Typo: grant

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -300,15 +307,25 @@ export class NetCoreDebugHelper implements DebugHelper {
...
         );
 
-        return `${debuggerPath}/vsdbg`;
+        return `${installPath}/vsdbg`;
+    }
+
+    private async isDebuggerInstalled(containerName: string, debuggerPath: string): Promise<boolean> {
+        const dockerClient = new CliDockerClient(new ChildProcessProvider());
+        const osProvider = new LocalOSProvider();
+        const command: string = osProvider.os === 'Windows' ?
```
haniamr: So the only difference on a Windows host is to use double quotes instead of single? There's no way to that one quoting technique works on both?

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -189,8 +189,17 @@ export class NetCoreDebugHelper implements DebugHelper {
         // Get Container Name if missing
         const containerName: string = debugConfiguration.containerName ?? await this.getContainerNameToAttach();
 
+        let debuggerPath: string = debugConfiguration.netCore?.debuggerPath;
+
         // If debugger path is not specified, install the debugger
```
haniamr: Nit: I'd modify this comment to say: ".... install the debugger if it doesn't exist in the container".

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: Yes that is right. Single quote isn't working on windows.

PR #1707, opened 2020-03-04
(CherryPick) Reuse an existing debugger if present (#1704)

ravipal:
* Reuse an existing debugger if present

* updated a comment

### DIFF
--- a/src/debugging/netcore/NetCoreDebugHelper.ts
+++ b/src/debugging/netcore/NetCoreDebugHelper.ts
@@ -190,8 +190,17 @@ export class NetCoreDebugHelper implements DebugHelper {
         // Get Container Name if missing
         const containerName: string = debugConfiguration.containerName ?? await this.getContainerNameToAttach();
 
-        // If debugger path is not specified, install the debugger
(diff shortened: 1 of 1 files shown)


src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -278,19 +287,17 @@ export class NetCoreDebugHelper implements DebugHelper {
         return pathNormalize(result, platformOS);
     }
 
-    private async installDebuggerInContainer(containerName: string): Promise<string> {
+    private async installDebuggerInContainer(containerName: string, installPath: string): Promise<string> {
         const yesItem: MessageItem = DialogResponses.yes;
         const install = (yesItem === await window.showInformationMessage(localize('vscode-docker.debug.netcore.attachingRequiresDebugger', 'Attaching to container requires .NET Core debugger in the container. Do you want to install debugger in the container?'), ...[DialogResponses.yes, DialogResponses.no]));
         if (!install) {
-            throw new UserCancelledError();
```
bwateratmsft: The message from UserCancelledError will never be shown; which is why I'd removed the string when doing the loc work. Is it needed?

bwateratmsft: Nit if we keep it: single quotes, wrap in localize call

Issue #1693, opened 2020-03-03
docker: Error response from daemon: mkdir C:\Program Files\dotnet\sdk\NuGetFallbackFolder: Access is denied.


dbreshears:
Extension assumes that "C:\Program Files\dotnet\sdk\NuGetFallbackFolder" exists and asks to volume mount.   I am uncertain when this folder is actually created but I don't have it.  I have dotnet sdk 3.1 + dotnet sdk 5.0 Preview 1 installed and ".Net Core Launch (web)" works

> docker run -dt -P --name "mymvc-dev" -e "DOTNET_USE_POLLING_FILE_WATCHER=1" -e "ASPNETCORE_ENVIRONMENT=Development" -e "ASPNETCORE_URLS=https://+:443;http://+:80" --label "com.microsoft.created-by=visual-studio-code" -v "c:\Users\devinb\mymvc:/app:rw" -v "c:\Users\devinb\mymvc:/src:rw" -v "C:\Users\devinb\.vsdbg:/remote_debugger:ro" -v "C:\Users\devinb\.nuget\packages:/root/.nuget/packages:ro" **-v "C:\Program Files\dotnet\sdk\NuGetFallbackFolder**:/root/.nuget/fallbackpackages:ro" -v "C:\Users\devinb\AppData\Roaming\Microsoft\UserSecrets:/root/.microsoft/usersecrets:ro" -v "C:\Users\devinb\AppData\Roaming\ASP.NET\Https:/root/.aspnet/https:ro" "mymvc:dev" <

**docker: Error response from daemon: mkdir C:\Program Files\dotnet\sdk\NuGetFallbackFolder: Access is denied.**

bwateratmsft:
Should be easy to validate that the folder exists before doing a volume mount. I think probably the right place for it is our `addVolumeWithoutConflicts` (spelling?) method.

PR #1719, opened 2020-03-04
validate nuget package back up volume mount

ravipal:
Add the optional volume mount only when the local path is present.
Fixes: #1693


src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -280,7 +280,9 @@ export class NetCoreTaskHelper implements TaskHelper {
             addVolumeWithoutConflicts(volumes, srcVolume);
             addVolumeWithoutConflicts(volumes, debuggerVolume);
             addVolumeWithoutConflicts(volumes, nugetVolume);
-            addVolumeWithoutConflicts(volumes, nugetFallbackVolume);
+            if (fse.pathExistsSync(nugetFallbackVolume.localPath)) {
```
haniamr: Any reason not to use the "async" version?

ravipal: No reason. Will use async version

PR #1726, opened 2020-03-05
don't install unzip

ravipal:
The latest version of GetVsDbg.sh doesn't require unzip. So not installing unzip in non-alpine flavor OS

Fixes: #1727

### DIFF
--- a/src/debugging/netcore/NetCoreDebugHelper.ts
+++ b/src/debugging/netcore/NetCoreDebugHelper.ts
@@ -295,9 +295,11 @@ export class NetCoreDebugHelper implements DebugHelper {
         }
         // Windows require double quotes and Mac and Linux require single quote.
         const osProvider = new LocalOSProvider();
+        const installDebuggerOnAlpine = `apk --no-cache add curl && curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${installPath}`;
+        const installDebuggerOnNonAlpine = `curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${installPath}`;
         const installDebugger: string = osProvider.os === 'Windows' ?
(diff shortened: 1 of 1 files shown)


src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -295,9 +295,11 @@ export class NetCoreDebugHelper implements DebugHelper {
...
+        const installDebuggerOnNonAlpine = `curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${installPath}`;
         const installDebugger: string = osProvider.os === 'Windows' ?
-            `/bin/sh -c "ID=default; if [ -e /etc/os-release ]; then . /etc/os-release; fi; echo $ID; if [ $ID == alpine ]; then apk --no-cache add curl && curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${installPath} ; else apt-get update && apt-get install unzip && curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${installPath}; fi"`
-            : `/bin/sh -c 'ID=default; if [ -e /etc/os-release ]; then . /etc/os-release; fi; echo $ID; if [ $ID == alpine ]; then apk --no-cache add curl && curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${installPath} ; else apt-get update && apt-get install unzip && curl -sSL https://aka.ms/getvsdbgsh | /bin/sh /dev/stdin -v latest -l ${installPath}; fi'`
+            `/bin/sh -c "ID=default; if [ -e /etc/os-release ]; then . /etc/os-release; fi; echo $ID; if [ $ID == alpine ]; then ${installDebuggerOnAlpine}; else ${installDebuggerOnNonAlpine}; fi"`
```
philliphoff: Since you're refactoring anyway, if the only difference between these strings are the double/single quotes, can the entire contents between the quotes be extracted as well?

PR #1733, opened 2020-03-06
Disable some faulty tests on Windows

bwateratmsft:
These tests have been failing an awful lot lately; I'm going to disable them on Windows.

The root cause is the same for all of them, permissions error trying to open `testOutput\Dockerfile`.

### DIFF
--- a/test/configure.test.ts
+++ b/test/configure.test.ts
@@ -7,6 +7,7 @@ import * as assert from 'assert';
 import * as assertEx from './assertEx';
 import * as vscode from 'vscode';
 import * as fse from 'fs-extra';
+import * as os from 'os';
 import * as path from 'path';
 import { Suite } from 'mocha';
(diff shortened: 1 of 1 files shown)


test/configure.test.ts
```diff
@@ -1041,6 +1042,12 @@ suite("Configure (Add Docker files to Workspace)", function (this: Suite): void
     // Java
 
     suite("Java", () => {
+        if (os.platform() === 'win32') {
+            // Skip tests that are faulty on Windows
+            // TODO: re-enable?
```
ravipal: Do we need an issue to track this? otherwise we will forget this.

bwateratmsft: The tests are sorely in need of some serious TLC. We should consider a much bigger test improvement task...

philliphoff: I think we should add a task to revisit the test infrastructure, including whether we bother with some/all of the automated tests).

bwateratmsft: #1737

PR #1734, opened 2020-03-06
Don't install unzip

ravipal:
Do not install Unzip on non-alpine flavor


src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -332,7 +336,7 @@ export class NetCoreDebugHelper implements DebugHelper {
         const context: IActionContext = { telemetry: { properties: {}, measurements: {} }, errorHandling: { issueProperties: {} } };
         const containerItem: ContainerTreeItem = await ext.containersTree.showTreeItemPicker(ContainerTreeItem.runningContainerRegExp, {
             ...context,
-            noItemFoundErrorMessage: 'No running containers are available to attach'
+            noItemFoundErrorMessage: localize('vscode-docker.debug.netcore.noContainers', 'No running containers are available to attach.')
```
ravipal: This needs to be reverted

PR #1736, opened 2020-03-06
Fix missing os: Windows in dockerRun config

### LINKED ISSUE #1735: .NET Core Debug fails on Windows containers
Debugging .NET Core on Windows containers is failing. The error is in docker-run-execute, `docker: Error response from daemon: invalid volume specification: 'd:\Sandbox\NetWindows:/src:rw'.`

image

The issue is that `"dockerRun": { "os": "Windows" }` is missing, probably since 0.10.0.

**Workaround**
In the `"dockerRun"` section of the `tasks.json` config, put `"os": "Windows"`. For example:
```json
        {
            "type": "docker-run",
            "label": "docker-run: debug",
            "dependsOn": [
                "docker-build: debug"
            ],
            "dockerRun": {
                "os": "Windows"
            },
            "netCore": {
                "appProject": "${workspaceFolder}/NetWindows.csproj",
                "enableDebugging": true
            }
        },
```

bwateratmsft:
Fixes #1735.

### DIFF
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -386,7 +386,7 @@ async function initializeForDebugging(context: IActionContext, folder: Workspace
 
 // tslint:disable-next-line: export-name
 export async function scaffoldNetCore(context: ScaffolderContext): Promise<ScaffoldFile[]> {
-    const os = context.os ?? await context.promptForOS();
+    const os = context.os ?? (context.os = await context.promptForOS());
     const isCompose = await context.promptForCompose();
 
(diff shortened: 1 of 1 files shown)


src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -386,7 +386,7 @@ async function initializeForDebugging(context: IActionContext, folder: Workspace
 
 // tslint:disable-next-line: export-name
 export async function scaffoldNetCore(context: ScaffolderContext): Promise<ScaffoldFile[]> {
-    const os = context.os ?? await context.promptForOS();
```
philliphoff: An alternative fix would be to not use `context.os` after this point (in favor of the local `os` value), which was likely my intent at the time.

Issue #1662, opened 2020-02-26
Fail to debug with "Docker: .NET Core Attach (Preview)" configuration for ASP .NET Core web apps in Windows Containers mode


v-dantch:
**OS:** Windows
**Build Version**: [20200225.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=15184&view=artifacts&type=publishedArtifacts)
**Regression:** Not a Regression

**Repro Steps:**
1. Switch to Windows Containers mode.
2. Create a ASP .NET Core web app by "dotnet new mvc".
3. Open this project in VS Code --> Prompt to add the required assets --> Yes.
4. F1 --> Type and select "Docker: Add Docker Files to Workspace" --> ASP .NET Core --> Windows--> Include optional Docker Compose files --> Enter ports: the default 80,443.
5. Right click the docker-compose.yml --> Select "Compose Up".
6. Open launch.json file --> Add "Docker: .NET Core Attach (Preview)" configuration.
7. Debug with "Docker: .NET Core Attach (Preview)" configuration.

**Expect:**
Succeed to debug.

**Actual:**
Fail to debug with error "Process exited with code 126".
![image](https://user-images.githubusercontent.com/34729092/75346055-b8d8cd00-5852-11ea-896e-22daf5df2f7f.png)

ravipal:
Attach to windows container is not supported. But we should improve the experience by validating the target container OS and provide appropriate error message to user.

bwateratmsft:
Is it not possible, period, or could users set up the debugger themselves and make it work?

ravipal:
We don't support windows out of the box. But user can still debug with the below steps.
1. Run 'Add Docker Files to Workspace' to create the docker-compose.debug.yml
2. Add the "Docker .NET Core Attach (Preview)" to launch.json
3. Download (https://aka.ms/getvsdbgps1) and install the vsdbg on the host machine at ~/.vsdbg
4 Update the launch config "Docker .NET Core Attach (Preview)" by adding debuggerPath as shown below.
        "netCore": {
            "debuggerPath": "C:\\remote_debugger\\win7-x64\\latest\\vsdbg"
        }
5. Now "Docker .NET Core Attach (Preview)" will attach the windows container.

dbreshears:
Let's emit a better error message and the above steps should be added to the troubleshooting page.  Confirm and/or add telemetry for Windows attach vs. Linux

ravipal:
[...]

PR #1738, opened 2020-03-07
use correct volume mount path on windows

ravipal:
Scaffold docker-compose with the right volume mount path for windows container.
Related: #1662

### DIFF
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -302,10 +302,9 @@ function generateComposeFiles(dockerfileName: string, platform: Platform, os: Pl
         }
     }
 
-    let volumesList = '      - ~/.vsdbg:/remote_debugger';
-    if (os === 'Linux') {
-        volumesList += ':rw'
-    }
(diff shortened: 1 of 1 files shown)


src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -302,10 +302,9 @@ function generateComposeFiles(dockerfileName: string, platform: Platform, os: Pl
         }
     }
 
-    let volumesList = '      - ~/.vsdbg:/remote_debugger';
-    if (os === 'Linux') {
-        volumesList += ':rw'
-    }
+    let volumesList = os === 'Windows' ?
+        '      - ~/.vsdbg:c:\\remote_debugger'
```
bwateratmsft:
```suggestion
        '      - ~/.vsdbg:c:\\remote_debugger:rw'
```

The `:rw` works on Windows without issue. We should keep it for consistency.

bwateratmsft: (By the way, I just learned about this suggest changes feature and it's really neat! Makes it easy to incorporate changes from feedback)

PR #1745, opened 2020-03-09
Better error message output

### LINKED ISSUE #1744: Cannot debug aspnet core projects
Repro steps:

1. Create a new web api project (cli): `dotnet new webapi`
2. Make sure OmniSharp server is running and run (command palette): `.NET: Generate Assets for Build and Debug`
3. Run (command palette): `Docker: Add Docker Files to Workspace`
4. Select `ASP.NET Core`
5. Select `Linux`
6. Keep default ports: `80, 443`
7. Start debugging with the newly created `Docker .NET Core Launch` launch task.

This is happening suddenly on all my docker asp.net core projects.

Action: docker-launch
Error type: 9
Error Message: Process exited with code 9

Version: 0.9.0
OS: darwin
Product: Visual Studio Code
Product Version: 1.42.1
Language: en

<details>
<summary>Call Stack</summary>

```
ChildProcess.<anonymous> extension.bundle.js:1:530610
ChildProcess.emit events.js:200:13
maybeClose child_process.js:1021:16
Process.ChildProcess._handle.onexit child_process.js:283:5
```

</details>

bwateratmsft:
Fixes #1744 

We will output the command line that was executed along with the stderr output.

### DIFF
--- a/src/utils/spawnAsync.ts
+++ b/src/utils/spawnAsync.ts
@@ -55,7 +55,14 @@ export async function spawnAsync(
                 return reject(new UserCancelledError());
             } else if (code) {
                 // Replicate the error object of child_process.exec().
-                const error = <ExecError>new Error(localize('vscode-docker.utils.spawn.exited', 'Process exited with code {0}', code));
+                let errorMessage = localize('vscode-docker.utils.spawn.exited', 'Process \'{0}\' exited with code {1}', command, code);
+
+                if (stderrBuffer) {
+                    errorMessage += localize('vscode-docker.utils.spawn.exitedError', '\nError: {0}', bufferToString(stderrBuffer));
+                }
+
+                let error = <ExecError>new Error(errorMessage);
+
                 error.code = code;
                 error.signal = signal;
                 return reject(error);
--- a/src/utils/spawnAsync.ts
+++ b/src/utils/spawnAsync.ts
@@ -54,14 +54,14 @@ export async function spawnAsync(
                 // If cancellation is requested we'll assume that's why it exited
                 return reject(new UserCancelledError());
             } else if (code) {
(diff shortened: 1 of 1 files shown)


src/utils/spawnAsync.ts
```diff
@@ -54,8 +54,15 @@ export async function spawnAsync(
                 // If cancellation is requested we'll assume that's why it exited
                 return reject(new UserCancelledError());
             } else if (code) {
-                // Replicate the error object of child_process.exec().
-                const error = <ExecError>new Error(localize('vscode-docker.utils.spawn.exited', 'Process exited with code {0}', code));
+                let errorMessage = localize('vscode-docker.utils.spawn.exited', 'Process \'{0}\' exited with code {1}', command.length > 50 ? `${command.substring(0, 50)}...` : command, code);
+
+                if (!onStderr && stderrBuffer) {
```
philliphoff: Is there a reason for not just *always* adding the entire error output to the error message, if we have it anyway?

bwateratmsft: The main reason was the tasks. For them, `onStderr` already results in the standard error output being emitted to the task pseudoterminal window in nice red font. If we put it in the error message as well in this scenario, the same stderr output gets re-emitted to the terminal by https://github.com/microsoft/vscode-docker/blob/master/src/tasks/DockerTaskProvider.ts#L56, which is redundant.

bwateratmsft: Generally speaking, the assumption is that if the caller specifies `onStderr`, they are handling error output themselves and putting it in the error message would be redundant; whereas if `onStderr` is `undefined`, then the output would be lost entirely if we _don't_ put it in the error message.

Issue #869, opened 2019-03-27
Support third-party registries


StephenWeatherford:
Specifically requested (but goal is generic support), in no particular order:

Open:
- [ ] goharbor.io (see #878, #3449, #3965)
- [ ] ECR (Amazon, see #3202, #3604)
- [ ] GCR (Google) (+3 see #394, #2890)
- [ ] quay.io (+4 see #390)
- [ ] DTR (Enterprise Docker trusted registry) (see #3325)
- [ ] Cloudsmith (see #1887)
- [ ] Aliyun (see #1964)
- [ ] Portus (see #2735)
- [ ] Digital Ocean (see #2879)
- [ ] Generic registries with self-signed certificate (#3639)
- [ ] Artifactory (#4011)

**Blocked** on no/partial Registry V2 support:
- [ ] V2 registries that don't use `/v2/` endpoint (see #3562)
- [ ] GitHub (see #1525, #1938, #3686)
- [ ] GitLab self-hosted (+6 see #1832, #2457, #2735, #3119, #3271, #3813)
- [ ] `harbor.korporate.org` (see #1927)
- [ ] GitLab Groups (see #1970)
- [ ] IBM Container Registry (see #3961)

Done:
- [x] DockerHub
- [x] Azure
- [x] GitLab (+25-ish see #162)
- [x] Generic V2 registry (+2)

gauravlakhani:
while connecting using private registry option, is private Docker DTR (Enterprise Docker trusted registry) supported at all?

StephenWeatherford:
@gauravlakhani Support for registries beyond Docker Hub is fairly new, we haven't tested on DTR. I've added it to the list.

shaneholloman:
@here champion!

imkhairulikhwan:
Hi, do we have an estimated date for this to complete?

Nhimself:
Following this...

bwateratmsft:
@BigMorty I think that we should take this one for 0.12.0...

BigMorty:
Agree we should try to get it into .12 or .14 (we are skipping .23 😉)

bwateratmsft:
This could potentially require and/or take advantage of https://github.com/microsoft/vscode/issues/88309

bwateratmsft:
I've spent a lot of time researching in the past few days. Unfortunately this is going to be extremely difficult, due to two main issues:

[...]

PR #1748, opened 2020-03-11
First stab at OAuth implementation

bwateratmsft:
Related to #869.

**Goals:**

This supports generic V2 registries where the registry requires OAuth, **and the token endpoint accepts basic (password) auth**. It also maintains the existing basic auth directly to the registry.

It also refactors out the auth from Azure and the above into auth helpers which can be implemented on a per-registry basis without needing to redefine the entire tree structure. Other registries should be able to use `GenericDockerV2RegistryTreeItem` as their root, the only difference will be icons and the auth helper. For example, GCR would be able to use `GenericDockerV2RegistryTreeItem` but would need a purpose-built GCR auth helper.

Trying also to reduce duplication and maximize code reuse.

**Non-goals:**

Actually supporting any additional registry types (ECR, GCR etc.) is out-of-scope for this PR.


src/tree/registries/dockerHub/DockerHubAccountTreeItem.ts
```diff
@@ -11,24 +11,24 @@ import { nonNullProp } from "../../../utils/nonNull";
...
     public label: string = 'Docker Hub';
     public childTypeLabel: string = 'namespace';
     public baseUrl: string = dockerHubUrl;
-    public cachedProvider: ICachedRegistryProvider;
 
     private _token?: string;
     private _nextLink?: string;
 
-    public constructor(parent: AzExtParentTreeItem, cachedProvider: ICachedRegistryProvider) {
+    public constructor(parent: AzExtParentTreeItem, protected readonly provider: IRegistryProvider, public readonly cachedProvider: ICachedRegistryProvider) {
```
bwateratmsft: This isn't strictly necessary to refactor all the way down, the child nodes don't actually use the `provider`. Could just accept it in the constructor (to fit `IRegistryProvider`) and ignore it.

src/tree/registries/gitLab/GitLabAccountTreeItem.ts
```diff
@@ -21,14 +22,12 @@ export class GitLabAccountTreeItem extends AzExtParentTreeItem implements IRegis
     public label: string = 'GitLab';
     public childTypeLabel: string = 'project';
     public baseUrl: string = 'https://gitlab.com/';
-    public cachedProvider: ICachedRegistryProvider;
 
     private _token?: string;
     private _nextLink?: string;
 
-    public constructor(parent: AzExtParentTreeItem, provider: ICachedRegistryProvider) {
+    public constructor(parent: AzExtParentTreeItem, protected readonly provider: IRegistryProvider, public readonly cachedProvider: ICachedRegistryProvider) {
```
bwateratmsft: This isn't strictly necessary to refactor all the way down, the child nodes don't actually use the `provider`. Could just accept it in the constructor (to fit `IRegistryProvider`) and ignore it.

karolz-ms:
Looking...

src/tree/registries/IRegistryProvider.ts
```diff
@@ -54,5 +55,10 @@ export interface IRegistryProvider {
     /**
      * The tree item class to instantiate after a provider is connected
      */
-    treeItemType: new (parent: AzExtParentTreeItem, cached: ICachedRegistryProvider) => AzExtParentTreeItem & IRegistryProviderTreeItem;
+    treeItemType: new (parent: AzExtParentTreeItem, provider: IRegistryProvider, cachedProvider: ICachedRegistryProvider) => AzExtParentTreeItem & IRegistryProviderTreeItem;
```
philliphoff: It seems like it would've been easier had `treeItemType` been a factory function rather than a type; that would have allowed each registry provider to instantiate whatever object it wanted, with whatever arguments were needed (e.g. an auth provider), without having to make those dependencies known to *all* registry tree item types (e.g. those that don't need any auth).

src/tree/registries/RegistriesTreeItem.ts
```diff
@@ -132,8 +132,10 @@ export class RegistriesTreeItem extends AzExtParentTreeItem {
             cachedProvider.url = wizardContext.url;
             cachedProvider.username = wizardContext.username;
 
-            if (wizardContext.password) {
-                await setRegistryPassword(cachedProvider, wizardContext.password);
+            if (wizardContext.secret && provider.authHelper) {
+                await provider.authHelper.persistAuth(cachedProvider, wizardContext.secret);
+            } else if (wizardContext.secret) {
+                await setRegistryPassword(cachedProvider, wizardContext.secret);
```
philliphoff: Would it make sense to push this into an `authHelper` on the other registry providers, so that the parent type doesn't have to make this distinction?

src/tree/registries/auth/IAuthHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as request from 'request-promise-native';
+import { URL } from 'url';
+import { ICachedRegistryProvider } from '../ICachedRegistryProvider';
+import { IDockerCliCredentials } from '../RegistryTreeItemBase';
+
+export interface IAuthHelper {
+    addAuth(cachedProvider: ICachedRegistryProvider, options: request.RequestPromiseOptions, authContext?: IOAuthContext): Promise<void>;
```
philliphoff: I'm not a fan of methods that modify incoming data.  Would it be better to return a new `request.RequestPromiseOptions` that's a (at least a shallow) copy the incoming `options` but with, for example, the auth properties set?  It's a pattern that request options are created once and then reused between requests; callers may not be expecting that object to be modified.

src/tree/registries/auth/oAuthUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+import { URL } from 'url';
+import { parseError } from 'vscode-azureextensionui';
+import { IOAuthContext } from './IAuthHelper';
+
+const realmRegExp = /realm=\"([^"]+)\"/i;
+const serviceRegExp = /service=\"([^"]+)\"/i;
+const scopeRegExp = /scope=\"([^"]+)\"/i;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/tslint/config
+export function getWwwAuthenticateHeader(error: any): IOAuthContext {
```
philliphoff: Is this specific to the Docker V2 API, or a more general OAuth mechanism?

philliphoff: Nit: should return `IOAuthContext | undefined`?

karolz-ms:
```suggestion
export function getWwwAuthenticateContext(error: any): IOAuthContext {

```

src/tree/registries/auth/IAuthHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as request from 'request-promise-native';
+import { URL } from 'url';
+import { ICachedRegistryProvider } from '../ICachedRegistryProvider';
+import { IDockerCliCredentials } from '../RegistryTreeItemBase';
+
+export interface IAuthHelper {
```
karolz-ms:
```suggestion
export interface IAuthProvider {

```
'Helper' to me conveys the notion of a utility class with a few stateless methods, e.g. https://docs.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.helpers.packageversionhelper?view=win-comm-toolkit-dotnet-stable This is not the case here, hence `Provider`

bwateratmsft:
> I like how this refactoring reduces code duplication, good job!

Thanks! That was definitely one of my goals.

src/tree/registries/auth/IAuthHelper.ts
bwateratmsft: For this we can return `AuthOptions` or whatever it's called. To change `addAuth` on tree nodes would require refactoring the other providers; we certainly can do that but we should consider that.

src/tree/registries/auth/oAuthUtils.ts
bwateratmsft: The `Www-Authenticate` header is supposed to be a generic OAuth thing, so I put it over here.

src/tree/registries/auth/AzureOAuthHelper.ts
```diff
@@ -0,0 +1,35 @@
...
+import { acquireAcrAccessToken, acquireAcrRefreshToken } from '../../../utils/azureUtils';
+import { ICachedRegistryProvider } from '../ICachedRegistryProvider';
+import { IDockerCliCredentials } from '../RegistryTreeItemBase';
+import { IAuthProvider, IOAuthContext } from './IAuthProvider';
+
+export interface IAzureOAuthContext extends IOAuthContext {
+    subscriptionContext: ISubscriptionContext
+}
+
+class AzureOAuthHelper implements IAuthProvider {
```
karolz-ms:
```suggestion
class AzureOAuthProvider implements IAuthProvider {

```
😬

src/tree/registries/auth/BasicOAuthHelper.ts
```diff
@@ -0,0 +1,64 @@
...
+import * as vscode from 'vscode';
+import { ICachedRegistryProvider } from '../ICachedRegistryProvider';
+import { getRegistryPassword } from '../registryPasswords';
+import { IDockerCliCredentials } from '../RegistryTreeItemBase';
+import { IAuthProvider, IOAuthContext } from './IAuthProvider';
+
+/**
+ * Performs basic auth and password-grant-type OAuth
+ */
+class BasicOAuthHelper implements IAuthProvider {
```
karolz-ms:
```suggestion
class BasicOAuthProvider implements IAuthProvider {

```

src/tree/registries/auth/AzureOAuthHelper.ts
bwateratmsft: Heheh, that should have been obvious given the change to IAuthProvider 😆

Issue #1600, opened 2020-02-04
Add "Docker: Add Docker Compose Files to Workspace..."  command


dbreshears:
Provide a command that will add docker-compose.yml and docker-compose.debug.yml files to the workspace.

ravipal:
The current implementation generates only docker-compose.yml.

PR #1752, opened 2020-03-13
Add Docker Compose files to Workspace

ravipal:
Adding a new command 'Add Docker Compose Files to Workspace...'

related: #1600

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "onCommand:vscode-docker.compose.restart",
         "onCommand:vscode-docker.compose.up",
         "onCommand:vscode-docker.configure",
-        "onCommand:vscode-docker.configureDocker",
+        "onCommand:vscode-docker.configureCompose",
         "onCommand:vscode-docker.containers.attachShell",
         "onCommand:vscode-docker.containers.browse",
         "onCommand:vscode-docker.containers.configureExplorer",
@@ -1991,8 +1991,8 @@
                 "category": "%vscode-docker.commands.category.docker%"
             },
             {
-                "command": "vscode-docker.configureDocker",
-                "title": "%vscode-docker.commands.configureDocker%",
+                "command": "vscode-docker.configureCompose",
+                "title": "%vscode-docker.commands.configureCompose%",
                 "category": "%vscode-docker.commands.category.docker%"
             },
             {
@@ -2444,9 +2444,9 @@
         "@types/string-replace-webpack-plugin": "^0.1.0",
         "@types/tar": "^4.0.3",
         "@types/xml2js": "^0.4.5",
-        "@typescript-eslint/eslint-plugin": "^2.22.0",
-        "@typescript-eslint/eslint-plugin-tslint": "^2.22.0",
-        "@typescript-eslint/parser": "^2.22.0",
+        "@typescript-eslint/eslint-plugin": "^2.23.0",
+        "@typescript-eslint/eslint-plugin-tslint": "^2.23.0",
+        "@typescript-eslint/parser": "^2.23.0",
         "adm-zip": "^0.4.14",
         "copy-webpack-plugin": "^5.1.1",
         "eslint": "^6.8.0",
--- a/package.json
+++ b/package.json
@@ -2475,7 +2475,6 @@
         "azure-arm-containerregistry": "^3.0.0",
         "azure-arm-website": "^5.3.0",
         "azure-storage": "^2.10.3",
-        "docker-modem": "^2.1.1",
         "dockerfile-ast": "^0.0.24",
         "dockerfile-language-server-nodejs": "^0.0.22",
         "dockerode": "^3.1.0",
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
         "onCommand:vscode-docker.compose.restart",
         "onCommand:vscode-docker.compose.up",
         "onCommand:vscode-docker.configure",
+        "onCommand:vscode-docker.configureDocker",
         "onCommand:vscode-docker.containers.attachShell",
         "onCommand:vscode-docker.containers.browse",
         "onCommand:vscode-docker.containers.configureExplorer",
@@ -1990,6 +1991,11 @@
                 "category": "%vscode-docker.commands.category.docker%"
             },
             {
+                "command": "vscode-docker.configureDocker",
+                "title": "%vscode-docker.commands.configureDocker%",
+                "category": "%vscode-docker.commands.category.docker%"
+            },
+            {
                 "command": "vscode-docker.containers.attachShell",
                 "title": "%vscode-docker.commands.containers.attachShell%",
                 "category": "%vscode-docker.commands.category.dockerContainers%"
@@ -2470,6 +2476,7 @@
         "azure-arm-website": "^5.3.0",
         "azure-storage": "^2.10.3",
         "docker-modem": "^2.1.1",
+        "dockerfile-ast": "^0.0.24",
         "dockerfile-language-server-nodejs": "^0.0.22",
         "dockerode": "^3.1.0",
         "fs-extra": "^8.1.0",
--- a/resources/template.docker-compose.yml
+++ b/resources/template.docker-compose.yml
@@ -1,0 +1,19 @@
+# Please refer https://docs.docker.com/compose to learn more about Docker Compose.
+
+# This is a sample docker-compose file with two services
+# 1. yourwebapp is an sample web service where the docker container image will be built from the Dockerfile as
+#    part of starting the compose.
+# 2. redis is an existing image hosted in docker hub.
+version: '3.4'
+
+services:
+  # yourwebapp:
+  #   image: yourwebapp
+  #   build:
+  #     context: .
+  #     dockerfile: Dockerfile
+  #   ports:
+  #       - 80
+
+  redis:
+    image: redis
--- a/src/configureWorkspace/configureCompose.ts
+++ b/src/configureWorkspace/configureCompose.ts
@@ -69,7 +69,7 @@ export async function configureCompose(context: IActionContext): Promise<void> {
         if (hasDuplicateService) {
             updateWithUniqueServiceName(composeScaffoldContexts, uniqueServiceNames);
         }
-        composeFile = await generateComposeFiles(composeScaffoldContexts);
+        composeFile = generateComposeFile(composeScaffoldContexts);
     } else {
         // No dockerfile is present in the workspace. Create a template docker-compose file.
         const templateComposeFile = path.join(ext.context.asAbsolutePath('resources'), 'template.docker-compose.yml');
@@ -101,7 +101,7 @@ async function getDockerFilesInWorkspace(context: IActionContext, rootFolder: vs
     return items;
 }
 
-async function generateComposeFiles(contexts: ComposeScaffoldContext[]): Promise<ScaffoldFile> {
+function generateComposeFile(contexts: ComposeScaffoldContext[]): ScaffoldFile {
     let services: string = '';
     contexts.forEach(context => {
         let ports: string = context.ports?.length > 0 ? `\n${getComposePorts(context.ports)}` : '';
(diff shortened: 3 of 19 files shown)


package.json
```diff
@@ -1989,6 +1990,11 @@
                 "title": "%vscode-docker.commands.configure%",
                 "category": "%vscode-docker.commands.category.docker%"
             },
+            {
+                "command": "vscode-docker.configureDocker",
```
bwateratmsft: I think we should name it `vscode-docker.configureCompose`. The existing `vscode-docker.configure` is already too vague of a name but not something we can change, let's make this one more specific.

src/configureWorkspace/configureCompose.ts
```diff
@@ -0,0 +1,109 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
+import * as path from "path";
```
bwateratmsft: Some of these imports use single and some double quotes, can we standardize them all to single quotes?

src/configureWorkspace/configureCompose.ts
```diff
@@ -0,0 +1,109 @@
...
+    let composeFile: ScaffoldFile;
+    // Get list of dockerfiles (services) to add to compose
+    const dockerFiles: Item[] = await getDockerFilesInWorkspace(context, rootFolder);
+
+    // Add the services to docker-compose.yaml
+    if (dockerFiles) {
+        const composeScaffoldContexts: ComposeScaffoldContext[] = [];
+        await Promise.all(
+            dockerFiles.map(async dockerfile => {
+                const serviceName = path.basename(path.dirname(dockerfile.absoluteFilePath));
```
bwateratmsft: Need to normalize `serviceName` with `getValidImageName`

src/configureWorkspace/configureCompose.ts
```diff
@@ -0,0 +1,109 @@
...
+                const dockerfileInfo: DockerfileInfo = await parseDockerfile(rootFolder.uri.fsPath, dockerfile.absoluteFilePath);
+                AddToServiceNames(serviceName);
+                composeScaffoldContexts.push({
+                    ...dockerfileInfo,
+                    serviceName: serviceName,
+                });
+            })
+        );
+        if (hasDuplicateService) {
```
bwateratmsft: If we already have a list of service names, why wait until now to update `composeScaffoldContexts` with a unique one? Why not make it unique before pushing into `composeScaffoldContexts`?

resources/template.docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+# Please refer https://docs.docker.com/compose to learn more about Docker Compose.
+
+# This is a sample docker-compose file with two services
+# 1. yourwebapp is an sample web service where the docker container image will be built from the Dockerfile as
+#    part of starting the compose.
+# 2. redis is an existing image hosted in docker hub.
+version: '3.4'
+
+services:
```
bwateratmsft: Note that the extension gets packaged pretty differently for prod (webpack) vs. when you F5. I think it's OK, but double check that this file ends up in the VSIX like we expect.

src/configureWorkspace/configureCompose.ts
```diff
@@ -0,0 +1,109 @@
...
+        if (processedServices.includes(context.serviceName)) {
+            context.serviceName = generateUniqueName(context.serviceName, existingNames, i => `.${i}`);
+            existingNames.push(context.serviceName);
+        }
+        processedServices.push(context.serviceName);
+    })
+}
+
+async function promptForFolder(): Promise<vscode.WorkspaceFolder> {
+    return await quickPickWorkspaceFolder(localize('vscode-docker.scaffolding.dockerCompose', 'To generate docker-compose files you must first open a folder or workspace in VS Code.'));
```
bwateratmsft: To make duplicate IDs less likely:
```suggestion
    return await quickPickWorkspaceFolder(localize('vscode-docker.scaffolding.dockerCompose.noWorkspaceFolder', 'To generate docker-compose files you must first open a folder or workspace in VS Code.'));
```

src/utils/generateUniqueName.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export function generateUniqueName(name: string, existingNames: string[], generateSuffix: (currentIndex: number) => string): string {
```
bwateratmsft: This is fairly similar to `generateNonConflictFileName`, is there a way we can refactor them to share code?

package.json
```diff
@@ -2469,6 +2475,7 @@
         "azure-arm-containerregistry": "^3.0.0",
         "azure-arm-website": "^5.3.0",
         "azure-storage": "^2.10.3",
+        "dockerfile-ast": "^0.0.24",
```
bwateratmsft: Need to run `npm update` and get the changes to `package-lock.json`. Also need to verify that the language service isn't affected since it uses 0.0.20.

src/utils/quickPickFile.ts
```diff
@@ -51,14 +52,27 @@ export async function resolveFileItem(rootFolder: vscode.WorkspaceFolder, fileUr
...
+            return items;
         } else {
-            const res: vscode.QuickPickItem = await ext.ui.showQuickPick(items, { placeHolder: message });
-            return <Item>res;
+            return await ext.ui.showQuickPick<Item>(items, { placeHolder: message, canPickMany: true });
         }
     }
 }
 
+export async function resolveFileItems(rootFolder: vscode.WorkspaceFolder, fileUri: vscode.Uri | undefined, globPatterns: string[], message: string): Promise<Item[] | undefined> {
```
bwateratmsft: I don't think this actually needs to be exported, since there's no external callers.

src/utils/quickPickFile.ts
```diff
@@ -51,14 +52,27 @@ export async function resolveFileItem(rootFolder: vscode.WorkspaceFolder, fileUr
...
+    return await resolveFileItemsInternal(rootFolder, fileUri, globPatterns, message, true);
+}
+
+export async function resolveFileItem(rootFolder: vscode.WorkspaceFolder, fileUri: vscode.Uri | undefined, globPatterns: string[], message: string): Promise<Item | undefined> {
```
bwateratmsft: This also I don't think needs to be exported.

src/utils/quickPickFile.ts
```diff
@@ -79,7 +93,7 @@ export async function quickPickDockerFileItem(context: IActionContext, dockerFil
     return dockerFileItem;
 }
 
-function getDockerFileGlobPatterns(): string[] {
+export function getDockerFileGlobPatterns(): string[] {
```
bwateratmsft: Same here, I don't think it's needed to export.

src/utils/dockerfileUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+    ports?: number[];
+}
+
+export async function parseDockerfile(rootFolderPath: string, dockerfilePath: string): Promise<DockerfileInfo> {
+    const content = await readFileContent(dockerfilePath);
+    const dockerFile = DockerfileParser.parse(content);
+    const ports: number[] = await getExposedPorts(dockerFile);
+    let dockerFilenameRelativeToRoot = path.relative(rootFolderPath, dockerfilePath);
+    // Ensure the path uses POSIX separators (which work on both Linux and Windows).
+    dockerFilenameRelativeToRoot = dockerFilenameRelativeToRoot.replace(/\\/g, '/');
```
bwateratmsft: I think `pathNormalize(dockerFilenameRelativeToRoot, 'Linux')` can be used instead of the regex replace.

src/utils/dockerfileUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+    const ports = exposes.map(e => {
+        const port = e.getArgumentsContent();
+        const index = port.indexOf('/');
+        return index > 0 ? +port.substr(0, index) : +port;
+    });
+    return ports;
+}
+
+async function readFileContent(dockerfilePath: string): Promise<string> {
+    if (dockerfilePath && fse.pathExists(dockerfilePath)) {
```
bwateratmsft: `pathExists` is async
```suggestion
    if (dockerfilePath && await fse.pathExists(dockerfilePath)) {
```

src/utils/dockerfileUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+        return (await fse.readFile(dockerfilePath)).toString();
+    }
+    throw new Error(localize('vscode-docker-utils.dockerfileUtils.invalidDockerfile', 'Either the dockerfile "{0}" is not provided or does not exist', dockerfilePath));
```
bwateratmsft:
```suggestion
    throw new Error(localize('vscode-docker-utils.dockerfileUtils.invalidDockerfile', 'The Dockerfile "{0}" was not provided or does not exist', dockerfilePath));
```

package.json
ravipal: Not sure how I came up with ConfigureDocker while the functions are named configureCompose

src/configureWorkspace/configureCompose.ts
ravipal: At this point we don't have all the existing names. For examples if the input has (web1, web1, web2, web1.1) then while processing the second web1, it will come up with a unique name of web1.1, which is an actual servicename that wasn't processed yet and then web1.1 will be replaced with web1.2.

resources/template.docker-compose.yml
ravipal: We already use GetTargetPath.proj from resources folder. This is no different than the other file right?

bwateratmsft: For the vsdbg script, at one point I had to put it in a manifest or move it or something like that; I can't recall.

package.json
ravipal: language service seems to be working with this change.

src/utils/dockerfileUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    return {
+        rootFolder: rootFolderPath,
+        dockerfileNameRelativeToRoot: dockerFilenameRelativeToRoot,
+        ports: ports
+    };
+}
+
+async function getExposedPorts(dockerFile: Dockerfile): Promise<number[]> {
+    const insts = dockerFile.getInstructions();
```
philliphoff: Do we care about multi-stage Dockerfiles?  (I.e. not all stages will contribute to an end set of exposed ports.)

src/configureWorkspace/configureCompose.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+async function getDockerFilesInWorkspace(context: IActionContext, rootFolder: vscode.WorkspaceFolder): Promise<Item[]> {
+    const message = localize('vscode-docker.scaffolding.dockerCompose.selectDockerFiles', 'Choose Dockerfiles to include in docker-compose.')
+    const items: Item[] = await quickPickDockerFileItems(context, undefined, rootFolder, message);
+    return items;
+}
+
+async function generateComposeFiles(contexts: ComposeScaffoldContext[]): Promise<ScaffoldFile> {
+    let services: string = '';
```
philliphoff: Nit: this method doesn't need async.

src/utils/dockerfileUtils.ts
ravipal: I can't think of a reason why the intermediate (not part of final stage) stages would expose ports. Exposing additional ports is not blocking compose up to bring up the container. My preference is go with this until we hear from customer.

PR #1753, opened 2020-03-13
OAuth implementation and refactoring

bwateratmsft:
Related to #869.

### DIFF
--- a/src/tree/registries/azure/azureRegistryProvider.ts
+++ b/src/tree/registries/azure/azureRegistryProvider.ts
@@ -3,7 +3,9 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { AzExtParentTreeItem } from "vscode-azureextensionui";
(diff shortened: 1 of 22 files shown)


src/tree/registries/azure/azureRegistryProvider.ts
```diff
@@ -15,5 +17,7 @@ export const azureRegistryProvider: IRegistryProvider = {
     api: RegistryApi.DockerV2,
     onlyOneAllowed: true,
     connectWizardOptions: undefined,
-    treeItemType: AzureAccountTreeItem
+    treeItemFactory: (parent: AzExtParentTreeItem, cachedProvider: ICachedRegistryProvider) => new AzureAccountTreeItem(parent, cachedProvider),
+    persistAuth: undefined,
```
philliphoff: Nit: the `undefined` properties can just be omitted?

bwateratmsft: I guess I favor being explicit here so it's clear that they aren't forgotten but intentionally left out...

PR #1762, opened 2020-03-17
Emit application host ports to output window

### LINKED ISSUE #1602: List Application Host ports that are in use to the output window
When  debugging , output the Application Host ports that are currently in use.

bwateratmsft:
Resolves #1602

The output looks like this:
image

### DIFF
--- a/src/debugging/DockerDebugConfigurationProvider.ts
+++ b/src/debugging/DockerDebugConfigurationProvider.ts
@@ -6,6 +6,7 @@
 import { CancellationToken, commands, debug, DebugConfiguration, DebugConfigurationProvider, MessageItem, ProviderResult, window, workspace, WorkspaceFolder } from 'vscode';
 import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
 import { DockerOrchestration } from '../constants';
+import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { getAssociatedDockerRunTask } from '../tasks/TaskHelper';
 import { DockerClient } from './coreclr/CliDockerClient';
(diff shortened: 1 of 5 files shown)


src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -111,27 +113,60 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
...
+                const sessionConfiguration = <ResolvedDebugConfiguration>session.configuration;
+
+                if (sessionConfiguration?.dockerOptions?.containerName === resolvedConfiguration.dockerOptions.containerName) {
+                    try {
+                        const inspectInfo = await ext.dockerode.getContainer(resolvedConfiguration.dockerOptions.containerName)?.inspect();
+                        const portMappings: string[] = [];
+
+                        if (inspectInfo?.NetworkSettings?.Ports) {
+                            for (const containerPort of Object.keys(inspectInfo.NetworkSettings.Ports)) {
+                                const mapping = inspectInfo.NetworkSettings.Ports[containerPort][0];
```
philliphoff: Do we care about more than just the first mapping (which would be unusual)?

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -111,27 +113,60 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
...
-                    && sessionConfiguration.dockerOptions
-                    && sessionConfiguration.dockerOptions.containerNameToKill === resolvedConfiguration.dockerOptions.containerNameToKill) {
+                if (sessionConfiguration?.dockerOptions?.containerName === resolvedConfiguration.dockerOptions.containerName) {
                     try {
-                        await this.dockerClient.removeContainer(resolvedConfiguration.dockerOptions.containerNameToKill, { force: true });
+                        await this.dockerClient.removeContainer(resolvedConfiguration.dockerOptions.containerName, { force: true });
+                    } finally {
+                        disposable.dispose();
+                    }
+                } else {
```
karolz-ms: Nit: I would just put a comment above the if statement and not use an `else` clause at all. E.g.

`// Don't do anything if this is not our debug session`

Makes things a bit shorter and easier to understand
Your call though

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: Good thought, I made that change.

PR #1764, opened 2020-03-17
Ensure stdErr is written to telemetry

haniamr:
- Pass the stdErr buffer for both docker-build and docker-run task providers to ensure the error message makes its way to telemetry
- Add a new flag to the error from spawnAsync the determines if the error is already written through the handler to avoid duplication

### DIFF
--- a/src/tasks/DockerRunTaskProvider.ts
+++ b/src/tasks/DockerRunTaskProvider.ts
@@ -54,14 +54,14 @@ export class DockerRunTaskProvider extends DockerTaskProvider {
         const commandLine = await this.resolveCommandLine(definition.dockerRun);
 
         const stdoutBuffer = Buffer.alloc(4 * 1024); // Any output beyond 4K is not a container ID and we won't deal with it
-        const stdrrrBuffer = Buffer.alloc(4 * 1024)
+        const stderrBuffer = Buffer.alloc(10 * 1024);
 
         await context.terminal.executeCommandInTerminal(
             commandLine,
             context.folder,
             true, // rejectOnStderr
             stdoutBuffer,
-            stdrrrBuffer,
+            stderrBuffer,
             context.cancellationToken
         );
         throwIfCancellationRequested(context);
--- a/src/tasks/DockerRunTaskProvider.ts
+++ b/src/tasks/DockerRunTaskProvider.ts
@@ -54,12 +54,14 @@ export class DockerRunTaskProvider extends DockerTaskProvider {
         const commandLine = await this.resolveCommandLine(definition.dockerRun);
 
         const stdoutBuffer = Buffer.alloc(4 * 1024); // Any output beyond 4K is not a container ID and we won't deal with it
+        const stdrrrBuffer = Buffer.alloc(4 * 1024)
+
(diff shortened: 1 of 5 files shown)


src/utils/spawnAsync.ts
```diff
@@ -13,7 +13,7 @@ const DEFAULT_BUFFER_SIZE = 10 * 1024; // The default Node.js `exec` buffer size
 export type Progress = (content: string, process: cp.ChildProcess) => void;
 
 /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-type ExecError = Error & { code: any, signal: any };
+export type ExecError = Error & { code: any, signal: any, stdErrHandled: boolean };
```
ravipal: The caller is passing the stdErr and we should leave the caller to make this call, than passing this information in the error.

src/utils/spawnAsync.ts
```diff
@@ -56,15 +56,16 @@ export async function spawnAsync(
...
-                    // If there was no progress handler for stderr, then the error output would be lost, so help mitigate that by putting the error output into the Error we throw
+                if (stderrBuffer) {
                     errorMessage += localize('vscode-docker.utils.spawn.exitedError', '\nError: {0}', bufferToString(stderrBuffer));
                 }
 
                 const error = <ExecError>new Error(errorMessage);
 
                 error.code = code;
                 error.signal = signal;
+                error.stdErrHandled = onStderr != null;
```
ravipal: If we are going with this approach, why consider only onStderr and ignoring stderrBuffer.

src/tasks/DockerRunTaskProvider.ts
```diff
@@ -54,12 +54,14 @@ export class DockerRunTaskProvider extends DockerTaskProvider {
         const commandLine = await this.resolveCommandLine(definition.dockerRun);
 
         const stdoutBuffer = Buffer.alloc(4 * 1024); // Any output beyond 4K is not a container ID and we won't deal with it
+        const stdrrrBuffer = Buffer.alloc(4 * 1024)
```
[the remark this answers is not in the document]
haniamr: I think it makes sense to make that 10K as well.

src/utils/spawnAsync.ts
haniamr: The caller here can choose to provide an strErrHandler or not (for example, DockerTaskProvider provides one while CliDockerClient doesn't). The thing is, the caller sometimes decides to write back the errorMessage which is thrown from here.

This flag simply lets the caller know that somewhere along the process lifetime, stdErrors were sent to the handler so no need to write them again.

src/utils/spawnAsync.ts
haniamr: I don't see that it adds any safety to include stdErrBuffer in this flag, since technically it is optional. The flag's purpose is only regarding the existence of the handler.

bwateratmsft: We have to ignore stderrBuffer, because if the process exits with a non-zero code we reject rather than return (i.e. we throw an exception). The caller will have never gotten the stderr result.

PR #1775, opened 2020-03-19
Convert the platform to a string in tasks telemetry

haniamr:
An attempt to fix the "platform" not showing up in the tasks telemetry.

### DIFF
--- a/src/tasks/DockerTaskProvider.ts
+++ b/src/tasks/DockerTaskProvider.ts
@@ -47,7 +47,7 @@ export abstract class DockerTaskProvider implements TaskProvider {
                 context.actionContext = actionContext;
                 context.platform = getPlatform(task.definition);
 
-                context.actionContext.telemetry.properties.platform = context.platform;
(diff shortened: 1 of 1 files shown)


src/tasks/DockerTaskProvider.ts
```diff
@@ -47,7 +47,7 @@ export abstract class DockerTaskProvider implements TaskProvider {
                 context.actionContext = actionContext;
                 context.platform = getPlatform(task.definition);
 
-                context.actionContext.telemetry.properties.platform = context.platform;
+                context.actionContext.telemetry.properties.platform = context.platform.toString();
```
bwateratmsft: We need to validate that this still works as expected when the extension is webpacked.

haniamr: I can see it with Fiddler. However, I see it even before this change so not sure what's happening...

bwateratmsft:
Resetting the reviews; it appears to be due to the name being `platform`, not the values being a typedef.

PR #1780, opened 2020-03-20
Validate dependency build task

### LINKED ISSUE #1423: Improve the flow to reach an F5'able project
Currently in order to F5 a project successfully, they need to add Docker files. For .NET Core, they need to add the .NET Core build/debug assets _before_ adding Docker debug assets (either via "Add Docker Files..." or "Initialize for debugging", due to https://github.com/OmniSharp/omnisharp-vscode/issues/3403.

Considering only native .NET Core debugging, if the user hits F5 it will add those assets if they are missing, and abort the launch. The next F5 will succeed. We should try to get close to that.

For Node, if they hit F5 and at the Select Environment pick they choose Docker, we should begin the Add Docker Files flow. For .NET Core, we should execute the C# extension's "Generate build/debug assets..." command, and then begin the Add Docker Files flow.

ravipal:
This change will prevent user from creating docker specific tasks and config without the net core tasks.
During Net Core scaffolding, validate if the dotnet core build task is present or not and block scaffolding the docker specific tasks.

Fixes #1423, #1300

### DIFF
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -20,6 +20,7 @@ import { DockerDebugScaffoldContext } from '../debugging/DebugHelper';
 import { dockerDebugScaffoldingProvider, NetCoreScaffoldingOptions } from '../debugging/DockerDebugScaffoldingProvider';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
+import { hasTask } from '../tasks/TaskHelper';
 import { extractRegExGroups } from '../utils/extractRegExGroups';
 import { getValidImageName } from '../utils/getValidImageName';
 import { globAsync } from '../utils/globAsync';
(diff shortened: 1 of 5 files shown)


src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -462,3 +464,11 @@ export async function scaffoldNetCore(context: ScaffolderContext): Promise<Scaff
 
     return files;
 }
+
+export function ensureDotNetCoreDependencies(workspaceFolder: WorkspaceFolder, context: IActionContext): void {
+    if (!hasTask('build', workspaceFolder)) {
+        context.errorHandling.suppressReportIssue = true;
+        const message = localize('vscode-docker.configureDotNetCore.missingDependencies', 'Dependency build task is missing. Please generate build task by running \'.NET: Generate Assets for Build an Debug\' before running this command');
```
karolz-ms: The "dependency" word here might be understood as "we are missing a build task for a dependency", which is not the case. Suggest saying simply "A build task is missing... " (rest of the error message unchanged).

bwateratmsft:
```suggestion
        const message = localize('vscode-docker.configureDotNetCore.missingDependencies', 'Dependency build task is missing. Please generate build task by running \'.NET: Generate Assets for Build and Debug\' before running this command');
```

PR #1791, opened 2020-03-26
Set scope to machine for imageBuildContextPath

haniamr:
This will essentially avoid reading this setting (if set in the User settings) in remote scenarios, which makes sense since the path might not make sense when running in a container/WSL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1798,7 +1798,7 @@
                     "type": "string",
                     "default": "",
                     "description": "%vscode-docker.config.docker.imageBuildContextPath%",
-                    "scope": "machine"
+                    "scope": "machine-overridable"
                 },
                 "docker.truncateLongRegistryPaths": {
                     "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -1797,7 +1797,8 @@
                 "docker.imageBuildContextPath": {
                     "type": "string",
                     "default": "",
-                    "description": "%vscode-docker.config.docker.imageBuildContextPath%"
+                    "description": "%vscode-docker.config.docker.imageBuildContextPath%",
+                    "scope": "machine"
                 },
                 "docker.truncateLongRegistryPaths": {
                     "type": "boolean",


karolz-ms:
According to [the docs](https://code.visualstudio.com/api/references/contribution-points#Configuration-schema) the "machine" scope can be set in "user or remote settings". So changing the scope does not really prevent setting this at user level and messing up the remote scenario? In general, anything seems to be configurable at user level; various scopes are there to define what is the *lowest* possible level of configuration where a setting might be overriden.

Am I reading the docs wrong?

haniamr:
> According to [the docs](https://code.visualstudio.com/api/references/contribution-points#Configuration-schema) the "machine" scope can be set in "user or remote settings". So changing the scope does not really prevent setting `imageBuildContextPath` at user level and messing up the remote scenario? In general, anything seems to be configurable at user level; various scopes are there to define what is the _lowest_ possible level of configuration where a setting might be overriden.
> 
> Am I reading the docs wrong?

Well, that's how it is written in the docs:

> Machine specific settings that can be set in user or remote settings. For example, an installation path which shouldn't be shared across machines.

I had to try it to ensure I understand it correctly, here is what I did:

- Set the property to some path (i.e. c:\temp) in the User settings
- Open the folder in VSCode Remote (WSL is the one I tried)
- Now the setting is not found under "User" and it shows as empty under "Remote"
image

So this confirms what I understand about the property, where "User" settings do not get used in remote scenarios if the scope is set to "machine", and vice versa. I think the documentation needs to explain it further by using a proper example, the linked issue has more details to it as well.

karolz-ms:
Allright, thanks Hani, what you are saying makes sense. The docs are confusing 😐

package.json
```diff
@@ -1797,7 +1797,8 @@
                 "docker.imageBuildContextPath": {
                     "type": "string",
                     "default": "",
-                    "description": "%vscode-docker.config.docker.imageBuildContextPath%"
+                    "description": "%vscode-docker.config.docker.imageBuildContextPath%",
+                    "scope": "machine"
```
karolz-ms:
```suggestion
                    "scope": "machine-overridable"
```
Don't we want this to be configurable per workspace/folder?

haniamr: Sure, that's fair enough. I don't have specific scenarios in mind to be pro/against so we can do it this way and see the feedback.

PR #1795, opened 2020-03-27
Add more logging and retry logic in test setup

ravipal:
Added more logging and also verification with retry. Hopefully this fixes the transient error.
Fixes: 1788

### DIFF
--- a/test/global.test.ts
+++ b/test/global.test.ts
@@ -58,7 +58,7 @@ export function testInEmptyFolderWithBuildTask(name: string, func?: mocha.AsyncF
     test(name, !func ? undefined : async function (this: mocha.Context) {
         // Ensure build task is created which is required for NetCore scaffolding.
         const workspacefolder = vscode.workspace.workspaceFolders[0];
-        await createBuildTask(workspacefolder);
+        await createBuildTask(name, workspacefolder);
 
         // Delete everything in the root testing folder
         assert(path.basename(testRootFolder) === constants.testOutputName, "Trying to delete wrong folder");;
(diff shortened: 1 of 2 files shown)


test/global.test.ts
```diff
@@ -93,9 +94,30 @@ export async function createBuildTask(folder: vscode.WorkspaceFolder): Promise<v
...
+        verifyBuildTask(testName, folder);
     }
 }
 
+function verifyBuildTask(testName: string, folder: vscode.WorkspaceFolder): void {
+    const maxTry = 5;
+    let i = 0;
+    let buildTaskPresent = isBuildTaskPresent(folder);
+
+    while (!buildTaskPresent && i < maxTry) {
```
bwateratmsft: Once we figure out and fix the real root cause let's remove this.

test/global.test.ts
```diff
@@ -93,9 +94,30 @@ export async function createBuildTask(folder: vscode.WorkspaceFolder): Promise<v
...
+        i++;
+    }
+    assert.equal(buildTaskPresent, true);
+    console.log(`${testName}: Verified the build task is present in the workspace after ${i} tries`);
+}
+
+function isBuildTaskPresent(folder: vscode.WorkspaceFolder): boolean {
+    const workspaceTasks = vscode.workspace.getConfiguration('tasks', folder.uri);
+    const allTasks = workspaceTasks && workspaceTasks.tasks as TaskDefinitionBase[] || [];
+    return allTasks.findIndex(t => t.label === 'build') !== -1
```
bwateratmsft:
```suggestion
    return allTasks.some(t => t.label === 'build');
```
`some` is basically the same as C#'s `Any`

test/global.test.ts
ravipal: Yes. that is the plan.

PR #1796, opened 2020-03-30
Make a few more settings machine-scoped

### LINKED ISSUE #1769: Root password needed sometime when connet remote linux container from mac os
image

If i restart local vscode and the remote vscode, sometimes the root password is not needed.

BTW, I have already add current no-root user to docker user group.

bwateratmsft:
Additional changes for #993. It seems to me that the four settings that control `DOCKER_*` environment variables ought to be machine-scoped. This might also fix #1769.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1807,22 +1807,26 @@
                 "docker.host": {
                     "type": "string",
                     "default": "",
-                    "description": "%vscode-docker.config.docker.host%"
+                    "description": "%vscode-docker.config.docker.host%",
+                    "scope": "machine-overridable"
                 },
                 "docker.certPath": {
                     "type": "string",
                     "default": "",
-                    "description": "%vscode-docker.config.docker.certPath%"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1807,22 +1807,26 @@
                 "docker.host": {
                     "type": "string",
                     "default": "",
-                    "description": "%vscode-docker.config.docker.host%"
+                    "description": "%vscode-docker.config.docker.host%",
+                    "scope": "machine-overridable"
```
ravipal: Can this be applied to other settings (ex: docker.volumes.sortBy). Looks like most of them eligible for machine level settings.

bwateratmsft: Those don't seem to me that they should be machine-scoped. This scoping is specifically to keep remote and local settings apart, i.e. if you have `docker.host` set locally, today it will be picked up in the remote environment; with this setting it will no longer be picked up in the remote environment.

PR #1800, opened 2020-03-30
Copy vsdbg from host machine to container

### LINKED ISSUE #1743: Fail to acquire debugger for single-Dockerfile debugging after compose up debugging
**Repro steps:**

1. Completely remove any existing `~/.vsdbg` folder (to simulate first time run)
1. Create a .NET Core project and add Docker files, including compose
1. Compose up docker-compose**.debug**.yml. This maps `~/.vsdbg` to `/remote_debugger`.
1. Run the "Docker .NET Core Attach (Preview)" launch config. It will prompt to install debugger, allow it to do so.
1. Compose down. The debugger remains installed on the _host_, at `~/.vsdbg`, because it was volume mapped and then installed in the container.
1. Switch to "Docker .NET Core Launch" and try to F5.

Note, I was using an Alpine image when I hit this, but that might not matter.

This fails because when the debugger was installed to `~/.vsdbg`, it has `root` as owner and group, and the local user has no permissions:

```
Acquiring the latest .NET Core debugger...
Unable to acquire the .NET Core debugger.
Error: EACCES: permission denied, open '/home/brandon/.vsdbg/getvsdbg.sh'
```

**Workaround:**

1. `chown` the `~/.vsdbg` directory to the current user
1. Completely remove the `~/.vsdbg` directory

IMO, this is a minor bug with an easy workaround but it isn't obvious to users what happened. I wanted to seed this issue in case people stumble upon it, but I'm OK with won't-fixing this unless we get explicit user feedback.

### LINKED ISSUE #1794: Fail to run 'Docker .Net Core Attach (Preview)' configuration for .NET Core 5.0 project
**OS:** Windows
**Build Version:** [20200326.8](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=16781&view=artifacts&type=publishedArtifacts)
**.NET Core SDK:** 5.0.100-preview.1.20155.7

**Repro Steps:**
1. Create a .NET Core MVC project by "dotnet new mvc --no-https".
2. Open the project in VS Code --> Prompt to add the required assets --> Yes
3. F1 --> Type and select "Docker: Add Docker Files to Workspace" --> ASP.NET Core --> Linux --> Include optional Docker Compose files.
4. Right click the docker-compose.yml --> Select "Compose Up".
5. Open launch.json and click 'Add Configration…', and choose 'Docker .NET Core Attach (Preview)'.
6. Run the above configuration, select the running container from step4 and click yes to allow installing vsdbg on the container.

**Expect:**
Run the configuration successfully without any error.

**Actual:**
An error occurs "The pipe program 'docker' exited unexpectedly with code 126".
image

Docker output:
  _Installing the latest .NET Core debugger..._
  > /bin/sh: 1: curl: not found
  > 
  > debian
  > 
  _Debugger installed_

Debug Console:
_Starting: "docker" exec -i console_console_1 /remote_debugger/vsdbg --interpreter=vscode
OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"/remote_debugger/vsdbg\": stat /remote_debugger/vsdbg: no such file or directory": unknown_

**More Info:**
The debugger installed successfully with alpine image but the debug will be auto-stopped.

ravipal:
net core 5.0 images removed the curl from base image and results into vsdbg install script on container. Now the vsdbg is copied to container from host machine.

Fixes #1794, fixes #1743

### DIFF
--- a/src/debugging/coreclr/CliDockerClient.ts
+++ b/src/debugging/coreclr/CliDockerClient.ts
@@ -266,6 +266,17 @@ export class CliDockerClient implements DockerClient {
         return id.substring(0, 12);
     }
 
+    public async copy(sourcePath: string, destinationPath: string): Promise<string> {
+        const command = CommandLineBuilder
+            .create('docker', 'cp')
+            .withQuotedArg(sourcePath)
+            .withQuotedArg(destinationPath)
+            .build();
+
+        const output = await this.processProvider.exec(command, {});
+        return output.stdout;
+    }
+
     public async exec(containerNameOrId: string, args: string, options?: DockerExecOptions): Promise<string> {
         options = options || {};
 
--- a/src/debugging/netcore/NetCoreDebugHelper.ts
+++ b/src/debugging/netcore/NetCoreDebugHelper.ts
@@ -198,7 +198,7 @@ export class NetCoreDebugHelper implements DebugHelper {
             debuggerPath = `${debuggerDirectory}/vsdbg`;
             const isDebuggerInstalled: boolean = await this.isDebuggerInstalled(containerName, debuggerPath);
             if (!isDebuggerInstalled) {
-                await this.copyDebuggerToContainer(containerName, debuggerDirectory);
+                debuggerPath = await this.copyDebuggerToContainer(containerName, debuggerDirectory);
             }
         }
 
--- a/src/debugging/netcore/NetCoreDebugHelper.ts
+++ b/src/debugging/netcore/NetCoreDebugHelper.ts
@@ -16,7 +16,7 @@ import { pathNormalize } from '../../utils/pathNormalize';
 import { PlatformOS } from '../../utils/platform';
 import { unresolveWorkspaceFolder } from '../../utils/resolveVariables';
 import { ChildProcessProvider } from '../coreclr/ChildProcessProvider';
-import CliDockerClient, { DockerExecOptions } from '../coreclr/CliDockerClient';
+import { CliDockerClient } from '../coreclr/CliDockerClient';
 import { CommandLineDotNetClient } from '../coreclr/CommandLineDotNetClient';
 import { LocalFileSystemProvider } from '../coreclr/fsProvider';
 import { AspNetCoreSslManager, LocalAspNetCoreSslManager } from '../coreclr/LocalAspNetCoreSslManager';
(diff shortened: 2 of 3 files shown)


src/debugging/coreclr/CliDockerClient.ts
```diff
@@ -266,6 +266,17 @@ export class CliDockerClient implements DockerClient {
         return id.substring(0, 12);
     }
 
+    public async copy(sourcePath: string, destinationPath: string): Promise<string> {
+        const command = CommandLineBuilder
+            .create('docker', 'cp')
```
bwateratmsft: Doesn't this fail if the directory does not exist? When I tried it did, anyway.

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -287,40 +287,30 @@ export class NetCoreDebugHelper implements DebugHelper {
         return pathNormalize(result, platformOS);
     }
 
-    private async installDebuggerInContainer(containerName: string, installPath: string): Promise<string> {
+    private async copyDebuggerToContainer(containerName: string, containerDebuggerDirectory: string): Promise<string> {
```
bwateratmsft: Why does this method return a string, but we don't use it?

src/debugging/coreclr/CliDockerClient.ts
ravipal: The copy will not create a parent dir and will fail, that is 'docker cp' behavior. If you copy the a/b/c/.vsdbg to /remote_debugger. It will create the target folder.

---
In reply to: [400954085](https://github.com/microsoft/vscode-docker/pull/1800#discussion_r400954085) [](ancestors = 400954085)

bwateratmsft: So we'll need to do a `docker run mkdir /remote_debugger`, right?

src/debugging/netcore/NetCoreDebugHelper.ts
ravipal: Now it is being used.

---
In reply to: [400957167](https://github.com/microsoft/vscode-docker/pull/1800#discussion_r400957167) [](ancestors = 400957167)

src/debugging/coreclr/CliDockerClient.ts
ravipal: If you need to create a parent directory, yes.

bwateratmsft: Huh, when I tried it last time it didn't work, but it's working now. Oh well.

Issue #333, opened 2018-07-26
Treeview UI for images does not reflect the similarity of Image IDs stored in different registries (add untag?)


PrashanthCorp:
In your existing docker project, build an image twice, with the registry (prefix in the tag) being different. 
Docker behavior: the images have the same ID. 
![image](https://user-images.githubusercontent.com/30420973/43288123-d36a3e64-90db-11e8-8073-810c7947ae50.png)

The side-effect (expected Docker behavior) is that removing any one of these images removes both. Our UI should have some indication of this.

PrashanthCorp:
Moreover, I can't seem to remove a tag. I get the following error:
(HTTP code 409) conflict - conflict: unable to delete 406f227b21f5 (cannot be forced) - image has dependent child images

StephenWeatherford:
Is there an untag?

PrashanthCorp:
This could be helpful: 
https://serverfault.com/questions/703344/how-to-remove-an-image-tag-in-docker-without-removing-the-image-itself

StephenWeatherford:
Have:

  Remove Tag  (new)
  Remove Image  (shows warning if multiple tags for same image)

khaledez:
I faced the same issue. I expected remove docker image in VSCode to behave the same as docker cli, but it doesn't, specially in the case of having more than one tag for an image.
If I try to remove a tag, I get 409 conflict error.

I think the expected behaviour should be the same as docker cli and remove the tag not the image.

PR #1803, opened 2020-03-31
Change to removing images by name

### LINKED ISSUE #1529: Feature to find dependent child images
While I understand the error, it'd be great if there was a feature that helped me determine what child images were causing this error:

image

bwateratmsft:
Fixes #1529, #333 (both indirectly).

If we remove by name instead of by ID, then the 409 conflict error won't happen anymore. If there are child images, then this will simply remove the tag of the parent image; later on if the child is removed it will also remove this parent image (assuming there aren't any other tags).

### DIFF
--- a/src/commands/images/removeImage.ts
+++ b/src/commands/images/removeImage.ts
@@ -23,7 +23,7 @@ export async function removeImage(context: IActionContext, node?: ImageTreeItem,
     if (nodes.length === 1) {
         confirmRemove = localize('vscode-docker.commands.images.remove.confirmSingle', 'Are you sure you want to remove image "{0}"? If there are other tags or child images for this image, only the tag will be removed.', nodes[0].fullTag);
     } else {
-        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove the selected images? If there are other tags or child images for this image, only the tag will be removed.');
+        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove the selected images? If there are other tags or child images for these images, only the tag will be removed.');
     }
 
     // no need to check result - cancel will throw a UserCancelledError
--- a/src/commands/images/removeImage.ts
+++ b/src/commands/images/removeImage.ts
@@ -21,9 +21,9 @@ export async function removeImage(context: IActionContext, node?: ImageTreeItem,
 
     let confirmRemove: string;
     if (nodes.length === 1) {
-        confirmRemove = localize('vscode-docker.commands.images.remove.confirmSingle', 'Are you sure you want to remove image "{0}"? If there are other tags or child images, only the tag will be removed.', nodes[0].fullTag);
+        confirmRemove = localize('vscode-docker.commands.images.remove.confirmSingle', 'Are you sure you want to remove image "{0}"? If there are other tags or child images for this image, only the tag will be removed.', nodes[0].fullTag);
     } else {
-        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove the selected images? If there are other tags or child images, only the tag will be removed.');
+        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove the selected images? If there are other tags or child images for this image, only the tag will be removed.');
     }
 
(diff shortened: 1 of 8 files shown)


src/commands/images/removeImage.ts
```diff
@@ -21,9 +21,9 @@ export async function removeImage(context: IActionContext, node?: ImageTreeItem,
 
     let confirmRemove: string;
     if (nodes.length === 1) {
-        confirmRemove = localize('vscode-docker.commands.images.remove.confirmSingle', 'Are you sure you want to remove image "{0}"? This will remove all matching and child images.', nodes[0].label);
+        confirmRemove = localize('vscode-docker.commands.images.remove.confirmSingle', 'Are you sure you want to remove image "{0}"? If there are other tags or child images, only the tag will be removed.', nodes[0].fullTag);
     } else {
-        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove selected images? This will remove all matching and child images.');
+        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove the selected images? If there are other tags or child images, only the tag will be removed.');
```
karolz-ms:
```suggestion
        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove the selected images? If there are other tags or child images for this image, only the tag will be removed.');
```

karolz-ms: (plus similar change for the confirmation message above)

src/commands/images/removeImage.ts
```diff
@@ -21,9 +21,9 @@ export async function removeImage(context: IActionContext, node?: ImageTreeItem,
 
     let confirmRemove: string;
     if (nodes.length === 1) {
-        confirmRemove = localize('vscode-docker.commands.images.remove.confirmSingle', 'Are you sure you want to remove image "{0}"? This will remove all matching and child images.', nodes[0].label);
+        confirmRemove = localize('vscode-docker.commands.images.remove.confirmSingle', 'Are you sure you want to remove image "{0}"? If there are other tags or child images for this image, only the tag will be removed.', nodes[0].fullTag);
     } else {
-        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove selected images? This will remove all matching and child images.');
+        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove the selected images? If there are other tags or child images for this image, only the tag will be removed.');
```
karolz-ms:
```suggestion
        confirmRemove = localize('vscode-docker.commands.images.remove.confirmMulti', 'Are you sure you want to remove the selected images? If there are other tags or child images for these images, only the selected tags will be removed.');
```
Sorry, should have suggested this first time :-*)

bwateratmsft: English is imprecise...it's on a per-image basis; i.e. if only one of the images has children, we'll still be removing the others (just not this one). Hard to convey that....

karolz-ms: OK, then never mind, no biggie

bwateratmsft: In any case I changed `this image` -> `these images`.

src/tree/images/ImageTreeItem.ts
```diff
@@ -59,25 +57,11 @@ export class ImageTreeItem extends AzExtTreeItem {
     }
 
     public async getImage(): Promise<Image> {
-        return callDockerode(() => ext.dockerode.getImage(this.imageId));
+        return callDockerode(() => ext.dockerode.getImage(this.fullTag));
```
bwateratmsft: Removal fails here for images that are tagged `<none>`, i.e. those where a newer version of the same image has been downloaded.

src/commands/images/runImage.ts
```diff
@@ -26,7 +27,7 @@ async function runImageCore(context: IActionContext, node: ImageTreeItem | undef
     }
 
     const image = await node.getImage();
-    const inspectInfo = await image.inspect();
+    const inspectInfo = await callDockerode(async () => image.inspect());
```
ravipal: Why do we need callDockerrode now?

bwateratmsft: This spot should've had it prior to now. Karol's context changes make flowing through `callDockerode` preferable, because it handles checking for context switches.

PR #1806, opened 2020-04-01
Change to ESLint as recommended extension in CONTRIBUTING with the same in vscode extensions file

nedex:
The extensions.json file has been updated to use the ESLint extension instead of TSLint.
This change removes the inconsistency.


SugahSugah111 (commented):
#1806

PR #1807, opened 2020-04-01
Add cleanUp option to the docker-run task

nedex:
This change allows the user to run a "short-term" container.
By setting this value to `true`, the container will be deleted after its use, using the `--rm` flag of the `docker run` command (default to false, current behavior).
It also helps to make the `dockerRun.containerName` parameter more useful, since a second run would block the task with this error:

> docker: Error response from daemon: Conflict. The container name "/`containerName`" is already in use [...]

Reference:
https://docs.docker.com/engine/reference/run/#clean-up---rm

**Note**: a container with the provided name **_must not exist_** to get this parameter to work (`docker run` command limitation).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1184,9 +1184,9 @@
                                     ]
                                 }
                             },
-                            "cleanUp": {
+                            "remove": {
                                 "type": "boolean",
-                                "description": "%vscode-docker.tasks.docker-run.dockerRun.cleanUp%",
+                                "description": "%vscode-docker.tasks.docker-run.dockerRun.remove%",
                                 "default": false
                             }
                         }
--- a/package.json
+++ b/package.json
@@ -1183,6 +1183,11 @@
                                         "containerPath"
                                     ]
                                 }
+                            },
+                            "cleanUp": {
+                                "type": "boolean",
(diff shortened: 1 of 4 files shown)


bwateratmsft:
We do container cleanup after debugging by default, so this is slightly different. @philliphoff what do you think is the best way to fit this in with that?

nedex:
Just as a note, my use case is something similar to the "_Compile your app inside the Docker container_"  part of the following image:
https://hub.docker.com/_/gcc

While it's not docker containers maintenance (or running something **inside** the container), I don't see any simpler ways to handle this and it's better to using a shell command with "raw" docker commands.

If you have any advices I would appreciate it.
Thanks.

philliphoff:
Looks ok to me.  This could actually be useful for debugging, too, for stacks that only tweak the entrypoint when debugging (like Node.js), as we could use this switch to have the container automatically removed when debugging ends.  Even for .NET Core, we could consider using this as the "remove container after debugging" flag rather than have a completely separate flag.  (In that case, the addition the `--rm` flag to the  `docker run` command is effectively a no-op, as the entrypoint effectively never ends without an explicit stop command.)

bwateratmsft:
Thanks for the contribution @nedex!

PR #1808, opened 2020-04-01
If refresh is manual, reset docker ctx check timer

bwateratmsft:
If a refresh was manually triggered (the command), then reset the timer on `docker context` checks. This will force a hash-check of the Docker config.json file; if needed then `docker context inspect` will be re-called.

As a result, clicking refresh manually should always show data from the immediately current Docker context.

NOTE: It is not needed for `vscode-docker.registries.refresh` because Dockerode is not involved in the registries.

### DIFF
--- a/src/utils/dockerContextManager.ts
+++ b/src/utils/dockerContextManager.ts
@@ -107,7 +107,7 @@ export class DockerContextManager {
         };
     }
 
-    public resetLastCheckedTimestamp(): void {
+    public expediteContextCheck(): void {
         this.lastContextCheckTimestamp = 0;
     }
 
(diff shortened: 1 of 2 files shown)


src/utils/dockerContextManager.ts
```diff
@@ -107,6 +107,10 @@ export class DockerContextManager {
         };
     }
 
+    public resetLastCheckedTimestamp(): void {
```
karolz-ms, line 110:
```suggestion
    public expediteContextCheck(): void {
```
It is a minor thing, but I would prefer to name this method in a way that is less tied to how it is implemented today.

Issue #1766, opened 2020-03-18
The Dockerfiles created by the extension are insecure


itamarst:
1. They shouldn't run as root (https://pythonspeed.com/articles/root-capabilities-docker-security/).

2. They should update system packages to get the latest security updates (`RUN apt-get update && apt-get -y upgrade`).

BigMorty:
Thanks @itamarst for your feedback, we will look into these concerns.

BigMorty:
Thanks for your helpful feedback! Our team discussed your suggestions and I wanted to let you know that our next release of the VS Code Docker extension will include an update to scaffold a Dockerfile that will run the container as a non-root user. In the future, we would ask you to first share security concerns with us privately to prevent exploitations.

In reference to updating system packages in the Dockerfile we scaffold, we do not plan to make that change. The reason is because we want to have reproducible builds. If a project is left untouched for some amount of time, a build that worked with one set of dependencies may become incompatible when every dependency is updated.

Thanks for your support and use of our product,
Mike

itamarst:
Thanks for fixing the root issue.

Re system packages—while reproducibility is a useful goal, not doing security updates doesn't actually improve reproducibility in a particularly meaningful way.

1. The _base_ image isn't going to be identical. Partially this is due to choices you've made: you're using `python:3.8` by default, which means it's going to be 3.8.2, then 3.8.3, etc.. Plus, you're not specifying an operating system, so at some point when next Debian release comes out and user rebuilds from scratch, their base OS will switch from Debian 10 to Debian 11, which is very much _not_ reproducible. So as a starting point, if you want at least some vague level of reproducibility use `python:3.8-buster` as base image.
2. The `python:3.8` image (or any other Python base image) is going to be rebuilt every few weeks, so system packages on the base image are going to change. Not much you can do about it, unless you want to switch to a more stable base image (the Bitnami base images have permanent tags guaranteed to be completely stable).
3. That being said, worries about lack of reproducibility from stable Debian are overblown. Assuming you switch to `python:3.8-buster`, in general Debian will strive for backwards compatibility in system packages; yes, occasionally backwards compatibility will slip through, but usually not.
4. The _real_ problem with lack of reproducibility in practice tends to be Python packages, as listed in `requirements.txt`. If you want to actually have reproducibility you need to transitively pin those using `pip-tools` (or Poetry, or Pipenv though that project appears dead).

[...]

PR #1811, opened 2020-04-02
Add logic to run with a low-rights user in Python

haniamr:
- If there isn't any chosen port that is less than 1024, then scaffold the logic to run as a low-rights user
- Else, show a warning at the top that this will be running as root

Related to #1766

### DIFF
--- a/src/configureWorkspace/configurePython.ts
+++ b/src/configureWorkspace/configurePython.ts
@@ -67,14 +67,14 @@ const flaskRequirements = `flask==1.1.1
 gunicorn==20.0.4`;
 
 const lowRightsUser = `
-# Switching to a non-root user, please refer to https://aka.ms/vscode-docker-python-user
+# Switching to a non-root user, please refer to https://aka.ms/vscode-docker-python-user-rights
 RUN useradd appuser && chown -R appuser /app
 USER appuser
 `;
 
 const rootUserWarning = `
 # Warning: A port below 1024 has been exposed. This requires the image to run as a root user which is not a best practice.
-# For more information, please refer to https://aka.ms/vscode-docker-python-ports`
+# For more information, please refer to https://aka.ms/vscode-docker-python-user-rights`
 
 async function genDockerFile(serviceName: string, target: PythonTarget, projectType: PythonProjectType, ports: number[], outputFolder: string): Promise<string> {
     const exposeStatements = getExposeStatements(ports);
--- a/src/configureWorkspace/configurePython.ts
+++ b/src/configureWorkspace/configurePython.ts
@@ -67,7 +67,7 @@ const flaskRequirements = `flask==1.1.1
 gunicorn==20.0.4`;
 
 const lowRightsUser = `
-# Switching to a low-rights user, please refer to https://aka.ms/vscode-docker-python-user
+# Switching to a non-root user, please refer to https://aka.ms/vscode-docker-python-user
 RUN useradd appuser && chown -R appuser /app
 USER appuser
 `;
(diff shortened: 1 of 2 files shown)


src/configureWorkspace/configurePython.ts
```diff
@@ -66,6 +66,16 @@ gunicorn==20.0.4`;
...
 
+const lowRightsUser = `
+# Switching to a low-rights user, please refer to https://aka.ms/vscode-docker-python-user
+RUN useradd appuser && chown -R appuser /app
+USER appuser
+`;
+
+const rootUserWarning = `
+# Warning: A port below 1024 has been exposed. This requires the image to run as a root user which is not a best practice.
+# For more information, please refer to https://aka.ms/vscode-docker-python-ports`
```
bwateratmsft, line 77: Can't we have just one link?

src/configureWorkspace/configurePython.ts
```diff
@@ -86,8 +96,18 @@ async function genDockerFile(serviceName: string, target: PythonTarget, projectT
         throw new Error(localize('vscode-docker.configurePython.unknownProjectType', 'Unknown project type: {0}', projectType));
     }
 
+    let userSwitchLogic = lowRightsUser;
+    let userWarning = '';
+
+    if (ports?.find(p => p < 1024)) {
```
bwateratmsft, line 102:
```suggestion
    if (ports?.some(p => p < 1024)) {
```
Equivalent of C#'s `Any`

src/configureWorkspace/configurePython.ts
```diff
@@ -86,8 +96,18 @@ async function genDockerFile(serviceName: string, target: PythonTarget, projectT
...
+        userSwitchLogic = '';
+        userWarning = rootUserWarning;
```
bwateratmsft, line 104: The flip-flopping of the local variables is a little confusing. I'd suggest doing something like:
`const rootPorts: boolean = ports?.some(p => p < 1024) ?? false;`

And below, in the replacements:
`.replace(/\$rootUserWarning\$/g, rootPorts ? rootUserWarning : '')`
and similar for `$user$`

src/configureWorkspace/configurePython.ts
```diff
@@ -66,6 +66,16 @@ gunicorn==20.0.4`;
 const flaskRequirements = `flask==1.1.1
 gunicorn==20.0.4`;
 
+const lowRightsUser = `
+# Switching to a low-rights user, please refer to https://aka.ms/vscode-docker-python-user
```
bwateratmsft, line 70: @BigMorty can you look at the text here and on the one below?

karolz-ms: Docker documentation uses term "non-root" user. Let's use it too for consistency

bwateratmsft: Agreed, "low-rights" (as opposed to "high-rights") is more the Windows terminology.

karolz-ms: Forgot the reference: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user

src/configureWorkspace/configurePython.ts
haniamr: I think one link is fair, I'll check back with Uche since he's the one who advised on the wording/links.

src/configureWorkspace/configurePython.ts
haniamr: Changed that to non-root user. Waiting to hear back from @BigMorty and Uche (not sure why I'm unable to tag him) on the wording.

src/configureWorkspace/configurePython.ts
ucheNkadiCode: I would change the link then to 'please refer to https://aka.ms/vscode-docker-python-user-rights`

src/configureWorkspace/configurePython.ts
ucheNkadiCode: Yes I agree, for consistency let's go with "# Switching to a non-root user, please refer to https://aka.ms/vscode-docker-python-user-rights"

PR #1814, opened 2020-04-03
Docker context-related commands

karolz-ms:


src/commands/context/removeDockerContext.ts
```diff
@@ -0,0 +1,29 @@
...
+export async function removeDockerContext(context: IActionContext): Promise<void> {
+    const selectedContext = await selectDockerContext(localize('vscode-docker.commands.context.selectContextToRemove', 'Select Docker context to remove'));
+    if (selectedContext.Current) {
+        context.errorHandling.suppressReportIssue = true;
+        throw new Error(localize('vscode-docker.commands.context.cannotRemoveContextInUse', 'Docker context currently in use cannot be removed'));
+    }
+
+    // no need to check result - cancel will throw a UserCancelledError
+    await ext.ui.showWarningMessage(
+        localize('vscode-docker.commands.context.confirmRemove', 'Are you sure you want to remove Docker context {0}?', selectedContext.Name),
```
bwateratmsft, line 22:
```suggestion
        localize('vscode-docker.commands.context.confirmRemove', 'Are you sure you want to remove Docker context \'{0}\'?', selectedContext.Name),
```

src/commands/context/removeDockerContext.ts
```diff
@@ -0,0 +1,29 @@
...
+        { modal: true },
+        { title: localize('vscode-docker.commands.context.remove', 'Remove') });
+
+    await dockerContextManager.remove(selectedContext.Name);
+    /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
+    vscode.window.showInformationMessage(localize('vscode-docker.commands.context.contextRemoved', 'Docker context {0} has been removed', selectedContext.Name));
```
bwateratmsft, line 28:
```suggestion
    vscode.window.showInformationMessage(localize('vscode-docker.commands.context.contextRemoved', 'Docker context \'{0}\' has been removed', selectedContext.Name));
```

src/commands/context/removeDockerContext.ts
```diff
@@ -0,0 +1,29 @@
...
+import { ext } from '../../extensionVariables';
+import { localize } from '../../localize';
+import { dockerContextManager } from '../../utils/dockerContextManager';
+import { selectDockerContext } from './selectDockerContext';
+
+export async function removeDockerContext(context: IActionContext): Promise<void> {
+    const selectedContext = await selectDockerContext(localize('vscode-docker.commands.context.selectContextToRemove', 'Select Docker context to remove'));
+    if (selectedContext.Current) {
+        context.errorHandling.suppressReportIssue = true;
+        throw new Error(localize('vscode-docker.commands.context.cannotRemoveContextInUse', 'Docker context currently in use cannot be removed'));
```
bwateratmsft, line 17:
```suggestion
        throw new Error(localize('vscode-docker.commands.context.cannotRemoveContextInUse', 'Docker context \'{0}\' is currently in use and cannot be removed', selectedContext.Name));
```

src/commands/context/selectDockerContext.ts
```diff
@@ -0,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
```
bwateratmsft, line 1: I feel like this file should be in `src/utils`, or maybe in `dockerContextManager`, since it isn't itself a command

src/commands/context/selectDockerContext.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { ProgressLocation, window } from 'vscode';
+import { ext } from '../../extensionVariables';
+import { localize } from '../../localize';
+import { dockerContextManager, IDockerContextListItem } from '../../utils/dockerContextManager';
+
+export async function selectDockerContext(prompt: string): Promise<IDockerContextListItem> {
+    const fetchingContexts = localize('vscode-docker.commands.context.fetchingContexts', 'Fetching Docker contexts...');
+    let contexts: IDockerContextListItem[];
+    await window.withProgress(
```
bwateratmsft, line 14: It should work either way but I think you can also do something like
`const contexts = await window.withProgress({ ... }, async () => dockerContextManager.listAll());`

Plus or minus an `async` or `await` somewhere in there :smile:

src/commands/context/selectDockerContext.ts
```diff
@@ -0,0 +1,31 @@
...
+        { location: ProgressLocation.Notification, title: fetchingContexts },
+        async () => {
+            contexts = await dockerContextManager.listAll()
+        });
+
+    const contextItems = contexts.map(ctx => ({
+        label: `${ctx.Name} ${ctx.Current ? '*' : ''}`.trim(),
```
bwateratmsft, line 21: I'd do:
```typescript
    label: ctx.Name,
    detail: ctx.Current ? localize(..., 'Currently selected') : '',
```

You can also try `description` instead of `detail`, I'm not sure exactly how each is rendered--worth looking at both.

src/commands/context/selectDockerContext.ts
```diff
@@ -0,0 +1,31 @@
...
+        context: ctx
```
bwateratmsft, line 22: It's not significantly different but other places use `IAzureQuickPickItem<T>` which is designed for this sort of quickpick, in this case that would mean instead of putting the context in `context`, it would be in `data`.
```typescript
const contextItems: IAzureQuickPickItem<IDockerContextListItem>[] = ...
    {
        label: ...
        data: ctx
    }
```

And below, `return selectedContextItem.data;`

src/commands/context/useDockerContext.ts
```diff
@@ -0,0 +1,19 @@
...
+import { dockerContextManager } from '../../utils/dockerContextManager';
+import { selectDockerContext } from './selectDockerContext';
+
+export async function useDockerContext(_context: IActionContext): Promise<void> {
+    const selectedContext = await selectDockerContext(localize('vscode-docker.commands.context.selectContextToUse', 'Select Docker context to inspect'));
+
+    await dockerContextManager.use(selectedContext.Name);
+
+    /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
+    vscode.window.showInformationMessage(localize('vscode-docker.commands.context.contextInUse', 'Using Docker context {0}', selectedContext.Name));
```
bwateratmsft, line 18:
```suggestion
    vscode.window.showInformationMessage(localize('vscode-docker.commands.context.contextInUse', 'Using Docker context \'{0}\'', selectedContext.Name));
```

src/commands/context/inspectDockerContext.ts
```diff
@@ -0,0 +1,20 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext, openReadOnlyContent } from 'vscode-azureextensionui';
+import { localize } from '../../localize';
+import { dockerContextManager } from '../../utils/dockerContextManager';
+import { selectDockerContext } from './selectDockerContext';
+
+export async function inspectDockerContext(_context: IActionContext): Promise<void> {
```
bwateratmsft, line 11: Here and in the rest of the commands, I'd rename `_context` / `context` to `actionContext`, to avoid confusion.

src/commands/context/selectDockerContext.ts
```diff
@@ -0,0 +1,31 @@
...
+    }));
+
+    const selectedContexItem = await ext.ui.showQuickPick(contextItems, {
```
bwateratmsft, line 25:
```suggestion
    const selectedContextItem = await ext.ui.showQuickPick(contextItems, {
```

src/commands/context/selectDockerContext.ts
```diff
@@ -0,0 +1,31 @@
...
+        placeHolder: prompt,
+        suppressPersistence: true
+    });
+
+    return selectedContexItem.context;
```
bwateratmsft, line 30:
```suggestion
    return selectedContextItem.context;
```

src/utils/dockerContextManager.ts
```diff
@@ -9,13 +9,21 @@ import * as fs from 'fs';
...
-import { execAsync } from './spawnAsync';
+import { execAsync, spawnAsync } from './spawnAsync';
 import { timeUtils } from './timeUtils';
 
-const ContextInspectExecOptions: ExecOptions = { timeout: 5000 }
+// CONSIDER
+// Any of the commands related to Docker context can take a very long time to execute (a minute or longer)
+// if the current context refers to a remote Docker engine that is unreachable (e.g. machine is shut down).
+// Consider having our own timeout for execution of any context-related Docker CLI commands.
+// The following timeout is for _starting_ the command only; in the current implementation there is no timeot
```
bwateratmsft, line 23: Right. For `refreshDockerode` we had to do a `Promise.race` to make a _real_ timeout. It might be helpful for `spawnAsync` to have an option that provides real timeout.

package.json
```diff
@@ -97,6 +97,9 @@
         "onCommand:vscode-docker.volumes.prune",
         "onCommand:vscode-docker.volumes.refresh",
         "onCommand:vscode-docker.volumes.remove",
+        "onCommand:vscode-docker.context.use",
+        "onCommand:vscode-docker.context.remove",
```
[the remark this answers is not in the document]
karolz-ms: That is what we agreed on. We do the same for volumes (there is no "Volume Create" command)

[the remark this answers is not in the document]
bwateratmsft: That makes sense, good point.

src/utils/dockerContextManager.ts
bwateratmsft: <You can resolve this, just putting my thoughts into writing>

src/commands/context/selectDockerContext.ts
karolz-ms: I was struggling with this too. It is not a general-purpose utility, so `src/utils` is not the best place IMO. It involves UI, so I did not want to put it in `dockerContextManager`. That is why the file ended up together with the rest of context command code.

src/commands/context/selectDockerContext.ts
karolz-ms: Nice. Applied

src/commands/context/selectDockerContext.ts
karolz-ms: This mimics the output of `docker context ls`, that is why I did it this way.

bwateratmsft: Can we do the `*` as `detail` or `description`? @BigMorty what do you think?

src/commands/context/selectDockerContext.ts
karolz-ms: Updated

src/commands/context/useDockerContext.ts
karolz-ms: Fixed!

src/commands/context/inspectDockerContext.ts
karolz-ms: Yes, absolutely. Fixed

src/utils/dockerContextManager.ts
karolz-ms: Yep, I saw that.

src/commands/context/selectDockerContext.ts
BigMorty: I like the more "(current)" or "(selected)" better than just an asterisk.

karolz-ms: OK will use `(current)`

karolz-ms:
Thanks for the review, Brandon, much appreciated ☺️

Issue #1785, opened 2020-03-25
Offer to install Docker Desktop when not installed


dbreshears:
In places where error messages display that Docker is not installed,  it should now offer to install Docker Desktop using specific URL.

bwateratmsft:
Installation steps will vary by OS.

PR #1820, opened 2020-04-04
Download and Install Docker

ravipal:
When docker is not installed, earlier the docker view shows 'Install Docker' that will open up a web page. Now clicking the node will start the download and initiate the installation.

Fixes: #1785


src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+import { IActionContext } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { localize } from '../localize';
+import LocalOSProvider from '../utils/LocalOSProvider';
+import { DockerInstallerBase } from './dockerInstallerBase';
+
+class WindowsDockerInstaller extends DockerInstallerBase {
+    protected downloadUrl: string = 'https://aka.ms/download-docker-windows-vscode';
+    protected fileExtension: string = 'exe';
+    protected installationMessage: string = localize('vscode-docker.commands.WindowsDockerInstaller.installationMessage', 'Installer is launched. Please follow the prompt and complet the installation and start the Docker Desktop.');
```
bwateratmsft, line 15:
```suggestion
    protected installationMessage: string = localize('vscode-docker.commands.WindowsDockerInstaller.installationMessage', 'Installer is launched. Please follow the prompts to complete the installation, and then start the Docker Desktop.');
```

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+    protected getInstallCommand(fileName: string): string {
+        // Windows require double quote.
+        return `"${fileName}"`;
+    }
+}
+
+class MacDockerInstaller extends DockerInstallerBase {
+    protected downloadUrl: string = 'https://aka.ms/download-docker-mac-vscode';
+    protected fileExtension: string = 'dmg';
+    protected installationMessage: string = localize('vscode-docker.commands.MacDockerInstaller.installationMessage', 'Installer is launched. Please follow the prompt and complete the installation and start the Docker Desktop.');
```
bwateratmsft, line 25:
```suggestion
    protected installationMessage: string = localize('vscode-docker.commands.MacDockerInstaller.installationMessage', 'Installer is launched. Please follow the prompts to complete the installation, and then start the Docker Desktop.');
```

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+    protected getInstallCommand(fileName: string): string {
+        return `Chmod +x '${fileName}' && open '${fileName}'`;
```
bwateratmsft, line 27:
```suggestion
        return `chmod +x '${fileName}' && open '${fileName}'`;
```

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+class LinuxDockerInstaller extends DockerInstallerBase {
+    protected downloadUrl: string = 'https://aka.ms/download-docker-linux-vscode';
+    protected fileExtension: string = 'sh';
+    protected installationMessage: string = localize('vscode-docker.commands.LinuxDockerInstaller.installationMessage', 'Please follow the prompt in terminal window and complete the installation and start the Docker.');
```
bwateratmsft, line 34:
```suggestion
    protected installationMessage: string = localize('vscode-docker.commands.LinuxDockerInstaller.installationMessage', 'Please follow the prompt in terminal window to complete the installation, and then start Docker.');
```

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+    protected getInstallCommand(fileName: string): string {
+        return `chmod +x '${fileName}' && sh '${fileName}'`;
+    }
+
+    protected async install(fileName: string): Promise<void> {
+        const terminal = ext.terminalProvider.createTerminal(localize('vscode-docker.commands.LinuxDockerInstaller.terminalTitle', 'Docker Install.'));
```
bwateratmsft, line 40:
```suggestion
        const terminal = ext.terminalProvider.createTerminal(localize('vscode-docker.commands.DockerInstaller.terminalTitle', 'Docker Install'));
```

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+        const command = this.getInstallCommand(fileName);
+        terminal.sendText(command);
+        terminal.show();
+        return;
+    }
+}
+
+export async function installDocker(context: IActionContext): Promise<void> {
```
bwateratmsft, line 48: I don't have any specific ideas, but is there any extra telemetry properties we want to attach to `context.telemetry`?

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { localize } from '../localize';
+import LocalOSProvider from '../utils/LocalOSProvider';
+import { DockerInstallerBase } from './dockerInstallerBase';
+
+class WindowsDockerInstaller extends DockerInstallerBase {
```
bwateratmsft, line 12: IMO these three classes make more sense to be in the same file as `DockerInstallerBase`, or otherwise files of their own, rather than in here.

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
+/*---------------------------------------------------------------------------------------------
```
bwateratmsft, line 1: Change file name to `DockerInstallerBase.ts` (classes get capitalized file names, at least in this repo).

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+import LocalOSProvider from '../utils/LocalOSProvider';
+
+export abstract class DockerInstallerBase {
+    protected abstract downloadUrl: string;
+    protected abstract fileExtension: string;
+    protected abstract installationMessage: string;
+    protected abstract getInstallCommand(fileName: string): string;
+
+    public async downloadAndInstallDocker(): Promise<void> {
+        const confirmInstall: string = localize('vscode-docker.commands.DockerInstallerBase.confirm', 'Are you sure you want to install docker on this machine.');
```
bwateratmsft, line 22:
```suggestion
        const confirmInstall: string = localize('vscode-docker.commands.DockerInstallerBase.confirm', 'Are you sure you want to install Docker on this machine?');
```

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+        const installTitle: string = localize('vscode-docker.commands.DockerInstallerBase.install', 'Install');
+        const downloadingMessage: string = localize('vscode-docker.commands.DockerInstallerBase.downloading', 'Downloading Docker installer...');
+        const downloadCompleteMessage: string = localize('vscode-docker.commands.DockerInstallerBase.downloadCompleteMessage', 'Download completed');
+        let downlaodedFileName: string;
```
bwateratmsft, line 26:
```suggestion
        let downloadedFileName: string;
```

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        // no need to check result - cancel will throw a UserCancelledError
+        await ext.ui.showWarningMessage(confirmInstall, { modal: true }, { title: installTitle });
+
+        await vscode.window.withProgress(
+            { location: vscode.ProgressLocation.Notification, title: downloadingMessage },
+            async () => {
+                downlaodedFileName = await this.downloadInstaller();
```
bwateratmsft, line 34:
```suggestion
                downloadedFileName = await this.downloadInstaller();
```

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    public async downloadAndInstallDocker(): Promise<void> {
+        const confirmInstall: string = localize('vscode-docker.commands.DockerInstallerBase.confirm', 'Are you sure you want to install docker on this machine.');
+        const installTitle: string = localize('vscode-docker.commands.DockerInstallerBase.install', 'Install');
+        const downloadingMessage: string = localize('vscode-docker.commands.DockerInstallerBase.downloading', 'Downloading Docker installer...');
+        const downloadCompleteMessage: string = localize('vscode-docker.commands.DockerInstallerBase.downloadCompleteMessage', 'Download completed');
+        let downlaodedFileName: string;
+
+        // no need to check result - cancel will throw a UserCancelledError
+        await ext.ui.showWarningMessage(confirmInstall, { modal: true }, { title: installTitle });
```
bwateratmsft, line 29: Make sure @BigMorty is OK with a modal dialog, they are very annoying.

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        await vscode.window.withProgress(
+            { location: vscode.ProgressLocation.Notification, title: downloadingMessage },
+            async () => {
+                downlaodedFileName = await this.downloadInstaller();
+                // eslint-disable-next-line @typescript-eslint/no-floating-promises
+                vscode.window.showInformationMessage(downloadCompleteMessage);
```
bwateratmsft, line 36: Do we need another toast to say download is complete? IMO when the "Downloading Docker installer..." progress toast disappears it's pretty clear that download is done. Also, immediately after this we show the installation message, so I'd rather just remove this toast entirely.

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+            }
+        );
+
+        const command = this.getInstallCommand(downlaodedFileName);
```
bwateratmsft, line 40:
```suggestion
        const command = this.getInstallCommand(downloadedFileName);
```

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        vscode.window.showInformationMessage(this.installationMessage);
+        await this.install(downlaodedFileName, command);
```
bwateratmsft, line 43:
```suggestion
        await this.install(downloadedFileName, command);
```

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+        await streamToFile(this.downloadUrl, fileName);
+        return fileName;
+    }
+
+    protected async install(fileName: string, cmd: string): Promise<void> {
+        const fsProvider = new LocalFileSystemProvider();
+        try {
+            const processProvider = new ChildProcessProvider();
+            ext.outputChannel.appendLine(localize('vscode-docker.commands.DockerInstallerBase.downloadCompleteMessage', 'Executing command {0}', cmd));
+            await processProvider.exec(cmd, {});
```
bwateratmsft, line 60: Use the terminal for all OSs, don't spawn the process directly.

bwateratmsft: Otherwise, at least use `spawnAsync` directly instead of `ChildProcessProvider`.

src/utils/httpRequest.ts
```diff
@@ -55,3 +57,29 @@ export async function httpsRequestBinary(opts: https.RequestOptions | string): P
         req.on('error', reject);
     });
 }
+
+export async function streamToFile(downloadUrl: string, fileName: string): Promise<void> {
```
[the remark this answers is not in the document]
ravipal: Nope that didn't work. Also that method is not efficient for large files. Streaming is the way to go.

src/commands/dockerInstallerBase.ts
ravipal: Here the modal dialog make sense as we need user input to proceed. @BigMorty reopen this comment, if you don't agree.

src/commands/dockerInstallerBase.ts
ravipal: @BigMorty, Your call.

src/commands/dockerInstallerBase.ts
ravipal: We decided to user Terminal only where we need user input.
Also what is the reasoning behind not using ChildProcessProvider.

bwateratmsft: Eventually the coreclr things are going to go away or get refactored. I don't want to keep adding to the amount of work that's going to take. `ChildProcessProvider` just calls `spawnAsync` underneath it.

src/commands/installDocker.ts
ravipal: This is defined in LinuxDockerInstaller so using that in the ID.

src/commands/installDocker.ts
ravipal: We need OS and success and failure state, which are captured already.

src/commands/dockerInstallerBase.ts
bwateratmsft: Users have repeatedly stated that toasts are annoying. We should minimize them.

src/commands/installDocker.ts
ravipal: I prefer to keep the base class separate. These 3 classes don't have much logic other than providing the data. I think having it here will give you full picture. Having a separate file each one will be an overkill.

bwateratmsft: In that case they should move to DockerInstallerBase.

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+    protected downloadUrl: string = 'https://aka.ms/download-docker-windows-vscode';
+    protected fileExtension: string = 'exe';
+    protected installationMessage: string = localize('vscode-docker.commands.WindowsDockerInstaller.installationMessage', 'Installer is launched. Please follow the prompts to complete the installation, and then start the Docker Desktop.');
```
bwateratmsft, line 15:
```suggestion
    protected installationMessage: string = localize('vscode-docker.commands.WindowsDockerInstaller.installationMessage', 'Installer is launched. Please follow the prompts to complete the installation, and then start Docker Desktop.');
```
Sorry, I missed this in the first pass.

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+    protected getInstallCommand(fileName: string): string {
+        // Windows require double quote.
+        return `"${fileName}"`;
+    }
+}
+
+class MacDockerInstaller extends DockerInstallerBase {
+    protected downloadUrl: string = 'https://aka.ms/download-docker-mac-vscode';
+    protected fileExtension: string = 'dmg';
+    protected installationMessage: string = localize('vscode-docker.commands.MacDockerInstaller.installationMessage', 'Installer is launched. Please follow the prompts to complete the installation, and then start the Docker Desktop.');
```
bwateratmsft, line 25:
```suggestion
    protected installationMessage: string = localize('vscode-docker.commands.MacDockerInstaller.installationMessage', 'Installer is launched. Please follow the prompts to complete the installation, and then start Docker Desktop.');
```
(Also missed this)

src/commands/dockerInstallerBase.ts
```diff
@@ -0,0 +1,67 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        vscode.window.showInformationMessage(this.installationMessage);
+        await this.install(downloadedFileName, command);
+    }
+
+    private async downloadInstaller(): Promise<string> {
+        const osProvider = new LocalOSProvider();
+        const processProvider = new ChildProcessProvider();
+        const tempFileProvider = new OSTempFileProvider(osProvider, processProvider);
+        let fileName = tempFileProvider.getTempFilename('docker', this.fileExtension);
```
haniamr, line 50: `const`?

src/commands/installDocker.ts
haniamr, line 25: There aren't really any prompts in Mac, it's just the drag-and-drop part. Do we want to reword this to make it more clear to the user?

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+    protected getInstallCommand(fileName: string): string {
+        return `chmod +x '${fileName}' && open '${fileName}'`;
```
haniamr, line 27: Is that required in Mac (the chmod part)? I tried with just the "open" command and it works fine.

src/commands/installDocker.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+class LinuxDockerInstaller extends DockerInstallerBase {
+    protected downloadUrl: string = 'https://aka.ms/download-docker-linux-vscode';
+    protected fileExtension: string = 'sh';
+    protected installationMessage: string = localize('vscode-docker.commands.LinuxDockerInstaller.installationMessage', 'Please follow the prompt in terminal window to complete the installation, and then start Docker.');
```
haniamr, line 34: `Docker Desktop`?

src/commands/installDocker.ts
haniamr: I agree, they should either be in separate files or the base class file.

src/commands/dockerInstallerBase.ts
ravipal: I didn't know that we have a plan to get rid of ChildProcessProvider. I will avoid using it then.

src/commands/installDocker.ts
bwateratmsft: Not on Linux. It's "Docker Community Edition" or something like that.

src/commands/installDocker.ts
bwateratmsft: It is something that could change depending on what Docker does with their installer / dmg in the future, so it's hard to say.

src/commands/dockerInstallerBase.ts
bwateratmsft: Yeah. Definitely a lot of it will be reused/refactored, though.

src/commands/dockerInstallerBase.ts
ravipal: Agree, that the 'Downloaded' is unnecessary. Removed it now. Also sent email to get Kaitlin (UX) input, based on her input, I will adjust.

src/commands/dockerInstaller.ts
```diff
@@ -0,0 +1,100 @@
...
+    public async downloadAndInstallDocker(): Promise<void> {
+        const confirmInstall: string = localize('vscode-docker.commands.DockerInstallerBase.confirm', 'Are you sure you want to install Docker on this machine?');
+        const installTitle: string = localize('vscode-docker.commands.DockerInstallerBase.install', 'Install');
+        const downloadingMessage: string = localize('vscode-docker.commands.DockerInstallerBase.downloading', 'Downloading Docker installer...');
+        let downloadedFileName: string;
+
+        // no need to check result - cancel will throw a UserCancelledError
+        await ext.ui.showWarningMessage(confirmInstall, { modal: true }, { title: installTitle });
+
+        await vscode.window.withProgress(
```
bwateratmsft, line 31: It's not imperative but another way you can do this is like this:

```typescript
const downloadedFileName = await vscode.window.withProgress( { ... }, () => this.downloadInstaller() );
```

(There may need to be +/- some `async`/`await` in there, but in general `withProgress` can return the inner value.)

src/commands/installDocker.ts
ravipal: I added it based on what Devin ran into. He got access denied or something like that. I think it is safe to add execute permission.

src/commands/installDocker.ts
ravipal: Not sure what the right thing to call. I used prompt to indicate the GUI shown to user. Any suggestions?

bwateratmsft: I guess I'm fine with the current verbage. @BigMorty?

src/commands/installDocker.ts
bwateratmsft: (Additionally, there's about a billion different ways to set up Docker on Linux and have something that works)

PR #1823, opened 2020-04-06
Call `docker context inspect` only when needed

### LINKED ISSUE #1804: Improve extension activation performance
(Self explanatory)

Here's the data from a webpacked activation:
```json 
{
	"messages": [],
	"activationTimes": {
		"codeLoadingTime": 1537,
		"activateCallTime": 14,
		"activateResolvedTime": 3152,
		"activationReason": {
			"startup": false,
			"extensionId": {
				"value": "ms-azuretools.vscode-docker",
				"_lower": "ms-azuretools.vscode-docker"
			},
			"activationEvent": "onView:dockerContainers"
		}
	},
	"runtimeErrors": []
}
```

From not-webpacked activation:
```json 
{
	"messages": [],
	"activationTimes": {
		"codeLoadingTime": 9422,
		"activateCallTime": 5,
		"activateResolvedTime": 3082,
		"activationReason": {
			"startup": false,
			"extensionId": {
				"value": "ms-azuretools.vscode-docker",
				"_lower": "ms-azuretools.vscode-docker"
			},
			"activationEvent": "onView:dockerContainers"
		}
	},
	"runtimeErrors": []
}
```

So our actual activation takes ~3 seconds, and code loading from ~1.5 seconds (webpacked) to 9.5 seconds (not webpacked).

### LINKED ISSUE #1818: Investigate running `docker context inspect` only if needed
Since `docker context inspect` takes some time (https://github.com/docker/cli/issues/2420), but many people wouldn't be using it, we can save some activation time by only calling it if there's more than just the `default` context. `default` cannot be modified, it is always based on `DOCKER_HOST`, which we cover already with our environment variable logic.

If any folders exist in `~/.docker/contexts/meta`, there are some custom contexts, and we should use `docker context inspect`; but if that folder is empty we shouldn't have to call it.

bwateratmsft:
Fixes #1818, and along with PR #1816 gets us close enough to fix #1804.

If `~/.docker/contexts/meta` is empty then only the default, non-modifiable `DOCKER_HOST`-based context is available. Since we already handle setting `DOCKER_HOST` based on environment + VSCode setting, it is not necessary to call `docker context inspect` in this case.

I expect that for the majority of users, only the default context will be present.

### DIFF
--- a/src/utils/dockerContextManager.ts
+++ b/src/utils/dockerContextManager.ts
@@ -99,14 +99,8 @@ export class DockerContextManager {
         if (Date.now() - this.lastContextCheckTimestamp > this.contextRefreshIntervalMs) {
             try {
                 if (!(await fse.pathExists(DockerContextMetasPath)) || (await fse.readdir(DockerContextMetasPath)).length === 0) {
-                    return {
-                        Changed: false,
-                        Context: undefined,
-                    };
-                } else if (!this.cachedContext) {
-                    // First-time check
-                    this.lastDockerConfigDigest = await this.getDockerConfigDigest();
-                    contextChanged = await this.refreshCachedDockerContext();
+                    this.cachedContext = undefined;
+                    contextChanged = false;
                 } else {
                     const dockerConfigDigest: string = await this.getDockerConfigDigest();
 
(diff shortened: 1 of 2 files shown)


karolz-ms:
Please don't merge yet, I have some questions

PR #1825, opened 2020-04-07
Allow customer docker-compose file names

### LINKED ISSUE #102: Allow custom file names

ravipal:
This change allows user to set the file associations like below so user can define a custom file name as compose file.
`    "files.associations": {
        "*docker-compose*": "yaml"
    }`
Fixes #102


src/utils/quickPickFile.ts
```diff
@@ -34,14 +34,36 @@ function createFileItem(rootFolder: vscode.WorkspaceFolder, uri: vscode.Uri): It
...
+}
+
+function getGlobPatterns(globPatterns: string[], fileTypeRegEx: RegExp): string[] {
+    const result: string[] = globPatterns;
+    try {
+        const config = vscode.workspace.getConfiguration('files').get<{}>('associations');
+        if (config) {
+            for (const globPattern of Object.keys(config)) {
+                const fileType = <string | undefined>config[globPattern];
+                if (fileType && fileTypeRegEx.test(fileType)) {
```
bwateratmsft, line 52: `files.associations` use language IDs, so we can look for exactly `dockerfile` and `yaml`, rather than a regex test.

ravipal: I am following the existing pattern which checks for dockerfile or Dockerfile. I believe the language id can have alias we define Dockerfile as alias for dockerfile.

src/utils/quickPickFile.ts
```diff
@@ -34,14 +34,36 @@ function createFileItem(rootFolder: vscode.WorkspaceFolder, uri: vscode.Uri): It
...
+                }
+            }
+        }
+    } catch {
+        // ignore and use default
     }
+    return result;
+}
 
+async function resolveFilePatterns(rootFolder: vscode.WorkspaceFolder, filePatterns: string[])
```
bwateratmsft, line 63:
```suggestion
async function resolveFilesOfPattern(rootFolder: vscode.WorkspaceFolder, filePatterns: string[])
```
Since we're getting files matching the given pattern(s), I think this is a better name. I realize this was the old name, but this file has always been super confusing and bad names didn't help...

src/utils/quickPickFile.ts
bwateratmsft: I see, fair enough.

PR #1827, opened 2020-04-07
Add an event listener to Docker Tasks

haniamr:
### Problem
For Python debugging, we listen to the prelaunch task's EndEvent in order to determine if the debugger is ready to attach. However, if any task failed in the chain (i.e. docker-build), we won't get notified and the debugger will still try to attach indefinitely.

### Resolution
The resolution in this PR is to listen to our own events (i.e. docker-build/run) and we can then know if any of them failed, which signals us to stop debugging and indicate failure.

This resolution is only partial, since if the user has a non-Docker task in the chain we still won't get notified. Here's a VSCode ask to change the API to indicate if a task succeeded or failed: https://github.com/microsoft/vscode/issues/92978


src/tasks/DockerTaskEndEventListener.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface DockerTaskEvent {
+    name: string,
+    success: boolean,
+    error?: string
+}
+
+type DockerTaskEventListener = { (taskEvent: DockerTaskEvent): void; };
+
+export class DockerTaskEndEventListener {
```
bwateratmsft, line 14: VSCode has an `EventEmitter<T>` class that we be using or deriving from.

haniamr: That's definitely handy, I think we can just use it there's no need to extend it.

bwateratmsft: Yeah. I honestly don't really know how it works, but it seems like it should do what we need. :smile:

bwateratmsft: Additionally: it produces `Disposable`s, which we might need to manually dispose of, or at least attach to `context.subscriptions`.

haniamr: Yeah, it's perfect to what we need here. So I switched to using that, with appropriately disposing the listener after we're done with it.

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -34,8 +35,15 @@ export namespace PythonExtensionHelper {
     export async function ensureDebuggerReady(prelaunchTask: vscode.Task, debuggerSemaphorePath: string, containerName: string, cliDockerClient: CliDockerClient): Promise<void> {
         // tslint:disable-next-line:promise-must-complete
         return new Promise((resolve, reject) => {
+            const dockerTaskListener = dockerTaskEndEventListener.event((taskEvent: DockerTaskEvent) => {
+                if (!taskEvent.success) {
+                    cleanupListeners();
+                    reject(localize('vscode-docker.tasks.pythonExt.failedToAttach', 'Failed to attach the debugger, please see the terminal output for more details.'));
```
bwateratmsft, line 41: Does this end up as output to the task window or as a toast? If toast, I'd suggest `context.errorHandling.suppressReportIssue = true;` or we'll get tons of mistaken bug reports.

src/tasks/DockerTaskProvider.ts
```diff
@@ -49,6 +50,8 @@ export abstract class DockerTaskProvider implements TaskProvider {
                 context.actionContext.telemetry.properties.dockerPlatform = context.platform;
                 context.actionContext.telemetry.properties.orchestration = 'single' as DockerOrchestration; // TODO: docker-compose, when support is added
                 await this.executeTaskInternal(context, task);
+
+                dockerTaskEndEventListener.fire({name: task.name, success: true});
```
bwateratmsft, line 54: It may be OK but I want call out--this means the event listener is shared between `docker-build` tasks and `docker-run` tasks. Does that cause any issues? (I can't think of any off the top of my head)

haniamr: Well, it runs pretty normal, I didn't see any issues in my testing so far. It's a good point, let's keep an eye out for CTI's testing but from my perspective it looks fine.

src/tasks/python/PythonExtensionHelper.ts
haniamr: It would end up as a toast, but unfortunately due to the convoluted nature of Python's debugging logic, to suppress the button here means we'd have to suppress up the chain here "[resolveDebugConfiguration](https://github.com/microsoft/vscode-docker/blob/ab6fdaf7c32dfdae25dcf94308b5babd65578f51/src/debugging/DockerDebugConfigurationProvider.ts#L51)", since that's where this will be called from. Is that something we'd be okay with?

bwateratmsft: Can we pass `IActionContext` down? If we reject here it should throw it all the way up. If we suppress report issue only here then nothing else is affected.

haniamr: Turns out it is passed down in the `DockerDebugContext `object, so I've set that in Python's `resolveDebugConfiguration`.

PR #1833, opened 2020-04-08
Open docker install instruction page for linux

ravipal:
There are several ways to get docker on Linux. So instead of trying to install using a generic script, we let the user install the docker.

### DIFF
--- a/src/commands/dockerInstaller.ts
+++ b/src/commands/dockerInstaller.ts
@@ -16,13 +16,13 @@ import { execAsync } from '../utils/spawnAsync';
 export abstract class DockerInstallerBase {
     protected abstract downloadUrl: string;
     protected abstract fileExtension: string;
-    protected abstract installationMessage: string;
     protected abstract getInstallCommand(fileName: string): string;
 
     public async downloadAndInstallDocker(): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/commands/dockerInstaller.ts
```diff
@@ -72,29 +72,18 @@ export class WindowsDockerInstaller extends DockerInstallerBase {
...
-}
 
-export class LinuxDockerInstaller extends UnixDockerInstaller {
-    protected downloadUrl: string = 'https://aka.ms/download-docker-linux-vscode';
-    protected fileExtension: string = 'sh';
-    protected installationMessage: string = localize('vscode-docker.commands.LinuxDockerInstaller.installationMessage', 'Opening the Docker installer...');
-    protected getInstallCommand(fileName: string): string {
-        return `chmod +x '${fileName}' && sh '${fileName}'`;
+    protected async install(fileName: string): Promise<void> {
+        const terminal = ext.terminalProvider.createTerminal(localize('vscode-docker.commands.UnixDockerInstaller.terminalTitle', 'Docker Install'));
```
bwateratmsft, line 84:
```suggestion
        const terminal = ext.terminalProvider.createTerminal(localize('vscode-docker.commands.MacDockerInstaller.terminalTitle', 'Docker Install'));
```

Since it moved classes

PR #1839, opened 2020-04-09
Fix platform check

### LINKED ISSUE #1838: The "Install Docker" command on Mac opens the Linux install page, instead of installing

bwateratmsft:
Fixes #1838

### DIFF
--- a/src/commands/installDocker.ts
+++ b/src/commands/installDocker.ts
@@ -3,20 +3,17 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as os from 'os';
 import { IActionContext } from 'vscode-azureextensionui';
-import LocalOSProvider from '../utils/LocalOSProvider';
 import { openExternal } from '../utils/openExternal';
 import { MacDockerInstaller, WindowsDockerInstaller } from './dockerInstaller';
 
 export async function installDocker(context: IActionContext): Promise<void> {
-    const osProvider = new LocalOSProvider();
-    if (osProvider.os === 'Linux') {
-        // eslint-disable-next-line @typescript-eslint/tslint/config
-        await openExternal('https://aka.ms/download-docker-linux-vscode');
+    if (os.platform() === 'win32') {
+        await (new WindowsDockerInstaller()).downloadAndInstallDocker();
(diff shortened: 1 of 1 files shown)


src/commands/installDocker.ts
```diff
@@ -3,20 +3,17 @@
...
-import LocalOSProvider from '../utils/LocalOSProvider';
 import { openExternal } from '../utils/openExternal';
 import { MacDockerInstaller, WindowsDockerInstaller } from './dockerInstaller';
 
 export async function installDocker(context: IActionContext): Promise<void> {
-    const osProvider = new LocalOSProvider();
-    if (osProvider.os === 'Linux') {
-        // eslint-disable-next-line @typescript-eslint/tslint/config
-        await openExternal('https://aka.ms/download-docker-linux-vscode');
+    if (os.platform() === 'win32') {
```
ravipal, line 12: The root cause was osProvider.os is returning the wrong value. We care about just 3 OSes. Can the osProvider.os be updated to return one of (Windows, Linux, Mac)

bwateratmsft: Not without spending the time evaluating every single call to `osProvider.os`. Additionally, it's headed for deprecation. `os.platform()` is better.

ravipal: Why not basically port the same logic that you have here to `osProvider.os`, so all caller don't have to come up with the same login in their code. We have about 8 calls using `osProvider.os` and if this returns the right value, then it shouldn't be an issue. Also we can get rid of osProvider.isMac.

bwateratmsft: Again--we would have to evaluate every single call to `osProvider.os` to see how it would affect them. Now isn't the time to do that.

ravipal: os.platform() returns one of 9 values and the caller needs to reduce it to 3. IMO osProvider.os is a good utility function that reduce it to the 3 values that we are interested in.

bwateratmsft: By all means then, update it everywhere, but make sure you aren't regressing any other features by the change. IMO it's the wrong move right now and I'm not willing to do it.

ravipal: Fair enough. I will make that change later.

PR #1843, opened 2020-04-09
Update osProvider.os to include Mac

ravipal:
Earlier osProvider.os was returning either Windows or Linux. This causes confusion because Linux could also indicate the Mac and callers had additional checks to further narrow down. So including the Mac as one of the values.

### DIFF
--- a/src/debugging/coreclr/CommandLineDotNetClient.ts
+++ b/src/debugging/coreclr/CommandLineDotNetClient.ts
@@ -68,7 +68,7 @@ export class CommandLineDotNetClient implements DotNetClient {
     }
 
     public async isCertificateTrusted(): Promise<TrustState> {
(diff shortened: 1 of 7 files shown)


bwateratmsft:
Did we evaluate every single call to `LocalOSProvider.os` and also every single usage of `PlatformOS`?

ravipal:
Yes, I did check and update all the references to LocalOsProvider.os. Let me know if I missed any. Also I don't this is as risky as you say. We still have lot of time to validate.

bwateratmsft:
> Yes, I did check and update all the references to LocalOsProvider.os. Let me know if I missed any. Also I don't this is as risky as you say. We still have lot of time to validate.

It isn't just `LocalOSProvider.os` though, because you added a value to `PlatformOS` it changed the meaning of `'Linux'`. It used to really mean Unix, now it means exclusively Linux, and we have to evaluate every single usage of that too.

PR #1852, opened 2020-04-11
Ravipal/docker install status

### LINKED ISSUE #1826: Move 'Install Docker...' link from the Docker view to notification
Current implementation shows a Install Docker link in Docker View like below. These links will be shown whenever the docker command throws error. It could be either docker is installed but not started or docker is not installed.
image

Identify whether the docker installed or not and show Install Docker link as notification.

ravipal:
Earlier we were showing install docker link 4 times in the docker view. Now this is shown as notification.
@BigMorty Let me know if you want to take this in 1.1.0.
In Mac and Windows:
image

In Linux:
image

Docker view shows
image

Fixes #1826


src/commands/dockerInstaller.ts
```diff
@@ -84,3 +79,26 @@ export class MacDockerInstaller extends DockerInstallerBase {
         terminal.show();
     }
 }
+
+export async function showDockerInstallNotification(): Promise<void> {
+    const dockerInstallMessage = ext.os.platform === 'linux'
+        ? localize('vscode-docker.commands.dockerInstaller.installDockerInfo', 'Docker is not installed. Would you like to open a page with more information?')
+        : localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to Install?');
+
+    const confirmationPrompt: vscode.MessageItem = ext.os.platform === 'linux' ? { title: 'Open' } : { title: 'Install' };
```
bwateratmsft, line 88: Use the `learnMoreLink` option for Linux, instead of 'Open'. Also, use `ext.ui.showWarningMessage(...)`. This will show a 'Learn More' button which is more consistent with our existing style.

bwateratmsft: Probably should change the text on the Linux prompt as well to fit that 'Learn More' button a little better.

src/commands/dockerInstaller.ts
```diff
@@ -84,3 +79,26 @@ export class MacDockerInstaller extends DockerInstallerBase {
         terminal.show();
     }
 }
+
+export async function showDockerInstallNotification(): Promise<void> {
+    const dockerInstallMessage = ext.os.platform === 'linux'
+        ? localize('vscode-docker.commands.dockerInstaller.installDockerInfo', 'Docker is not installed. Would you like to open a page with more information?')
+        : localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to Install?');
```
bwateratmsft, line 86:
```suggestion
        : localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to install?');
```

src/commands/dockerInstaller.ts
bwateratmsft, line 88: Are `'freebsd'` and `'openbsd'` valid options? They would be excluded by this `=== 'linux'`.

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -286,14 +294,13 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
...
-            new OpenUrlTreeItem(this, localize('vscode-docker.tree.additionalTroubleshooting', 'Additional Troubleshooting...'), 'https://aka.ms/AA37qt2'),
-        ];
+    private getDockerErrorTreeItems(context: IActionContext, error: unknown, dockerInstalled: boolean): AzExtTreeItem[] {
+        const result: AzExtTreeItem[] = dockerInstalled
+            ? [
+                new GenericTreeItem(this, { label: localize('vscode-docker.tree.dockerNotRunning', 'Failed to connect. Is Docker running?'), contextValue: 'dockerConnectionError', iconPath: getThemedIconPath('statusWarning') }),
+                new OpenUrlTreeItem(this, localize('vscode-docker.tree.additionalTroubleshooting', 'Additional Troubleshooting...'), 'https://aka.ms/AA37qt2')
+            ]
+            : [new GenericTreeItem(this, { label: localize('vscode-docker.tree.dockerNotInstalled', 'Failed to connect. Is Docker installed?'), contextValue: 'dockerConnectionError', iconPath: getThemedIconPath('statusWarning') })];
```
bwateratmsft, line 303: If we want we should be able to make clicking this node run the `installDocker` command. See what @BigMorty thinks.

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -108,7 +110,13 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
             this._currentItems = undefined;
             this._failedToConnect = true;
             context.telemetry.properties.failedToConnect = 'true';
-            return this.getDockerErrorTreeItems(context, error);
+            const dockerInstalled = await isDockerInstalled();
```
bwateratmsft, line 113: We should change the order of the logic here, or else `docker -v` will be called four times (once for each of Containers, Images, Networks, Volumes). On my system that would take more than 30 seconds.

src/commands/dockerInstaller.ts
ravipal: Not sure I follow your suggestion. Are you suggesting to use `ext.ui.showWarningMessage(...)` which will change the button from 'Open' to 'Learn More'? So effectively changing the label from 'open' to 'learn more' right?

bwateratmsft: In the options object, `ext.ui.showWarningMessage(...)` has a `learnMoreLink` property, which you can use instead of the `Open` button. This would be more consistent with the rest of our UI which uses the `learnMoreLink` for similar things.

src/tree/LocalRootTreeItemBase.ts
ravipal: In my machine it take half second to execute the command. I don't think it is worth optimizing this code considering that it gets exected only when it fails to connect. Also caching the dockerInstalled state will require additional logic on resetting it.

bwateratmsft: It is _absolutely_ worth optimizing this code. The average execution time for any Docker command on Windows is about 2 seconds, and half second on Linux/Mac. You're gonna make people wait 8 seconds for nothing?

ravipal: docker -v is light weight compare to other docker commands that queries the images and containers. The half second that I measured is on windows machine. Even say if it takes average 2 seconds we are not serially waiting for all tree items. It gets loads in parallel, so all of them get the result at the same time (2 seconds) instead of 6+ seconds. Caching will add complexity and user will still wait for 2 seconds or whatever the time it takes to run 'docker -v' command. Are you saying 40% of the users fails to connect to docker?

ravipal: Also note that when the docker is not installed the command 'docker -v' takes 10 milliseconds.

src/commands/dockerInstaller.ts
bwateratmsft: Discussed offline, here we should stick with `window.showInformationMessage(...)`, and just change the text on that Open button (and probably the dialog as well) to `Learn more`. That way, the toast doesn't stay open when you click `Learn more`.

src/tree/LocalRootTreeItemBase.ts
bwateratmsft: I don't think the caching code is actually difficult enough to justify not using it. @karolz-ms what do you think?

src/commands/dockerInstaller.ts
ravipal: Changed the button from 'Open' to 'Learn more'

src/tree/LocalRootTreeItemBase.ts
ravipal: Lets wait for Mike's input in the email thread that I started on this.

src/commands/dockerInstaller.ts
```diff
@@ -84,3 +79,30 @@ export class MacDockerInstaller extends DockerInstallerBase {
...
+    const installMessageNonLinux = localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to install?');
+    const learnMore = localize('vscode-docker.commands.dockerInstaller.learnMore', 'Learn more')
+    const install = localize('vscode-docker.commands.dockerInstaller.install', 'Install')
+    const osProvider: LocalOSProvider = new LocalOSProvider();
+    const isLinux = osProvider.os === 'Linux' && !osProvider.isMac;
+
+    const dockerInstallMessage = isLinux ? installMessageLinux : installMessageNonLinux;
+    const confirmationPrompt: vscode.MessageItem = isLinux ? { title: learnMore } : { title: install };
+    const response = await vscode.window.showInformationMessage(dockerInstallMessage, ...[confirmationPrompt]);
+    if (response) {
```
bwateratmsft, line 94: Just to verify, if the user hits X it results in `response` being `undefined`, hence this block is skipped--correct?

src/tree/LocalRootTreeItemBase.ts
bwateratmsft: I have one suggestion. In the base class, we could test `this.treePrefix === 'containers'` before doing an install check. For the rest of the views, it is maybe OK to show only the "Additional Troubleshooting" node, and not the "Docker not installed" / "Docker not running" warnings.

This way, we only do one `docker -v`, and avoid duplicating errors in each pane.

src/commands/dockerInstaller.ts
ravipal: Yes that is right.

ravipal:
In recent testing I noticed that the refersh of docker view is broken. Need investigation.

src/tree/LocalRootTreeItemBase.ts
bwateratmsft: We can use the `AsyncLazy` I'm working on, with a relatively short timeout of say 500ms. After all...the installation status is not going to change from not installed -> installed in <500ms, but most likely all four views will have attempted to refresh in this time.

src/tree/LocalRootTreeItemBase.ts
ravipal: In stand-up, we agreed on not show the install link here.

PR #1861, opened 2020-04-14
Add an async lazy class and lazy lifetimes

bwateratmsft:
(Not for 1.1)

Adds an `AsyncLazy` class that is smart enough to not re-run the task if it's currently running. Also adds lifetimes to force `Lazy`s to re-evaluate, if the caller so specifies.

One scenario where this would be useful is in places where we know we are going to query something several times in rapid succession--maybe `docker info`, as an example--and we expect no changes to occur in some time span (e.g. 1 second), but want to share the result across each of these queries to avoid re-running code unnecessarily.

### DIFF
--- a/src/utils/lazy.ts
+++ b/src/utils/lazy.ts
@@ -42,19 +42,28 @@ export class AsyncLazy<T> {
     public constructor(private readonly valueFactory: () => Promise<T>, private readonly _valueLifetime?: number) {
     }
 
+    public get isValueCreated(): boolean {
+        return this._isValueCreated;
+    }
+
     public async getValue(): Promise<T> {
         if (this._isValueCreated) {
             return this._value;
         }
 
-        this._isValueCreated = false;
-
         const isPrimaryPromise = this._valuePromise === undefined; // The first caller is "primary"
-        const fnLocalPromiseRef = this._valuePromise = this._valuePromise ?? this.valueFactory(); // Await any currently-running Promise if there is one
-        this._value = await fnLocalPromiseRef;
-        this._valuePromise = undefined;
 
-        this._isValueCreated = true;
+        if (isPrimaryPromise) {
+            this._valuePromise = this.valueFactory();
+        }
+
+        const result = await this._valuePromise;
+
+        if (isPrimaryPromise) {
+            this._value = result;
+            this._valuePromise = undefined;
+            this._isValueCreated = true;
+        }
 
         if (this._valueLifetime && isPrimaryPromise) {
             const reset = setTimeout(() => {
@@ -69,6 +78,6 @@ export class AsyncLazy<T> {
             }, this._valueLifetime);
         }
 
-        return this._value;
+        return result;
     }
 }
--- a/src/utils/lazy.ts
+++ b/src/utils/lazy.ts
@@ -7,7 +7,7 @@ export class Lazy<T> {
     private _isValueCreated: boolean = false;
     private _value: T | undefined;
 
-    public constructor(private readonly valueFactory: () => T) {
+    public constructor(private readonly valueFactory: () => T, private readonly _valueLifetime?: number) {
     }
 
     public get isValueCreated(): boolean {
@@ -15,13 +15,60 @@ export class Lazy<T> {
     }
 
     public get value(): T {
-        if (!this._isValueCreated) {
-            this._value = this.valueFactory();
-            this._isValueCreated = true;
+        if (this._isValueCreated) {
+            return this._value;
+        }
+
+        this._value = this.valueFactory();
+        this._isValueCreated = true;
+
+        if (this._valueLifetime) {
+            const reset = setTimeout(() => {
+                this._isValueCreated = false;
+                this._value = undefined;
+                clearTimeout(reset);
+            }, this._valueLifetime);
         }
 
         return this._value;
     }
 }
 
-export default Lazy;
+export class AsyncLazy<T> {
+    private _isValueCreated: boolean = false;
+    private _value: T | undefined;
+    private _valuePromise: Promise<T> | undefined;
+
+    public constructor(private readonly valueFactory: () => Promise<T>, private readonly _valueLifetime?: number) {
+    }
+
+    public async getValue(): Promise<T> {
+        if (this._isValueCreated) {
+            return this._value;
+        }
+
+        this._isValueCreated = false;
+
+        const isPrimaryPromise = this._valuePromise === undefined; // The first caller is "primary"
+        const fnLocalPromiseRef = this._valuePromise = this._valuePromise ?? this.valueFactory(); // Await any currently-running Promise if there is one
+        this._value = await fnLocalPromiseRef;
+        this._valuePromise = undefined;
+
+        this._isValueCreated = true;
+
+        if (this._valueLifetime && isPrimaryPromise) {
+            const reset = setTimeout(() => {
+                // Will only clear out values if there isn't a currently-running Promise
+                // If there is, this timer will skip, but when that Promise finishes it will go through this code and register a new timer
+                if (this._valuePromise === undefined) {
+                    this._isValueCreated = false;
+                    this._value = undefined;
+                }
+
+                clearTimeout(reset);
+            }, this._valueLifetime);
+        }
+
+        return this._value;
+    }
+}
--- a/test/utils/lazy.test.ts
+++ b/test/utils/lazy.test.ts
@@ -1,0 +1,117 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { Lazy, AsyncLazy } from "../../extension.bundle";
+import { delay } from '../../src/utils/delay';
+
+suite('(unit) Lazy tests', () => {
+    suite('Lazy<T>', () => {
+        test('Normal', async () => {
+            let factoryCallCount = 0;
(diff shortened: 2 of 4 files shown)


src/utils/lazy.ts
```diff
@@ -7,21 +7,68 @@ export class Lazy<T> {
...
+        if (this._isValueCreated) {
+            return this._value;
+        }
+
+        this._isValueCreated = false;
+
+        const isPrimaryPromise = this._valuePromise === undefined; // The first caller is "primary"
+        const fnLocalPromiseRef = this._valuePromise = this._valuePromise ?? this.valueFactory(); // Await any currently-running Promise if there is one
+        this._value = await fnLocalPromiseRef;
+        this._valuePromise = undefined;
```
karolz-ms, line 55: Are you sure you do not have a race condition here? It seems to me that handlers awaiting on older Promises (if more than one was involved) can reset `this._valuePromise` after a new Promise has been created and is not resolved yet.

src/utils/lazy.ts
```diff
@@ -7,21 +7,68 @@ export class Lazy<T> {
...
+
+    public async getValue(): Promise<T> {
+        if (this._isValueCreated) {
+            return this._value;
+        }
+
+        this._isValueCreated = false;
+
+        const isPrimaryPromise = this._valuePromise === undefined; // The first caller is "primary"
+        const fnLocalPromiseRef = this._valuePromise = this._valuePromise ?? this.valueFactory(); // Await any currently-running Promise if there is one
```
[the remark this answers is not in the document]
bwateratmsft: If another execution changes `this._valuePromise` to `undefined` while this one is awaiting, I'm not totally sure what will happen. It might not be necessary.

src/utils/lazy.ts
bwateratmsft: Once the older one finishes, all of the calls that are awaiting it should finish (since it's the same `Promise`)--so all of them would set `this._valuePromise` to undefined. It's overkill, but I'm not sure which of the executions will start working again first--it may not be the "primary" one.

src/utils/lazy.ts
```diff
@@ -7,21 +7,68 @@ export class Lazy<T> {
...
+
+    public constructor(private readonly valueFactory: () => Promise<T>, private readonly _valueLifetime?: number) {
+    }
+
+    public async getValue(): Promise<T> {
+        if (this._isValueCreated) {
+            return this._value;
+        }
+
+        this._isValueCreated = false;
```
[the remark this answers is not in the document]
bwateratmsft: Yes, so that subsequent calls (while this one is `await`ing) don't think there is a value.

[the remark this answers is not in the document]
karolz-ms: My point is on line 46 you check if the value is true and return if it is. So on line 50 it must be false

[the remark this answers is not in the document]
bwateratmsft: 🤦 doh!

I was thinking "I need to set it so that other executions don't think there's a value"...but you're absolutely right, if we got past the above check then it is, by definition, _already_ false.

src/utils/lazy.ts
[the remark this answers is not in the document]
karolz-ms: I am 99% sure `await this._valuePromise` will take the value of `this._valuePromise` at the start of await operation, so subsequent changes to that member will have no effect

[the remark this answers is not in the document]
bwateratmsft: I'm 99% sure I agree. I'll test it out to cover the 1% uncertainty. :smile:

src/utils/lazy.ts
karolz-ms: Consider having only the primary promise owner to change `this._value`, `this._valuePromise`, and `this._isValueCreated`. It will be slightly less efficient, but the correctness will be much easier to determine.

The non-primaries would just check if the value is created, and if not, wait on the promise and return the value.

bwateratmsft: Gotcha--so they'd await `this._valuePromise` and return its value, and only primary would set `this._value` / `this._isValueCreated`?

karolz-ms: Yes, exactly!

src/utils/lazy.ts
```diff
@@ -7,21 +7,68 @@ export class Lazy<T> {
...
+        }
+
+        this._value = this.valueFactory();
+        this._isValueCreated = true;
+
+        if (this._valueLifetime) {
+            const reset = setTimeout(() => {
+                this._isValueCreated = false;
+                this._value = undefined;
+                clearTimeout(reset);
```
ravipal, line 29: What is the need to clear the timeout? The timer will be started only when `this._isValueCreated` is false so any given time only one timer is on.

src/utils/lazy.ts
```diff
@@ -7,21 +7,68 @@ export class Lazy<T> {
...
 
-export default Lazy;
+export class AsyncLazy<T> {
+    private _isValueCreated: boolean = false;
+    private _value: T | undefined;
+    private _valuePromise: Promise<T> | undefined;
+
+    public constructor(private readonly valueFactory: () => Promise<T>, private readonly _valueLifetime?: number) {
+    }
+
```
ravipal, line 44: Do we need isValueCreated similar to non async impl

src/utils/lazy.ts
```diff
@@ -7,21 +7,68 @@ export class Lazy<T> {
...
+        if (this._valueLifetime && isPrimaryPromise) {
+            const reset = setTimeout(() => {
+                // Will only clear out values if there isn't a currently-running Promise
+                // If there is, this timer will skip, but when that Promise finishes it will go through this code and register a new timer
+                if (this._valuePromise === undefined) {
+                    this._isValueCreated = false;
+                    this._value = undefined;
+                }
+
+                clearTimeout(reset);
```
ravipal, line 68: Will this timer ever started twice? I think not, because of the isPrimaryPromise check, so what is the need to clear the timer?

src/utils/lazy.ts
ravipal, line 53: consider one assignment per line.

src/utils/lazy.ts
```diff
@@ -7,21 +7,68 @@ export class Lazy<T> {
...
-export default Lazy;
+export class AsyncLazy<T> {
+    private _isValueCreated: boolean = false;
+    private _value: T | undefined;
+    private _valuePromise: Promise<T> | undefined;
+
+    public constructor(private readonly valueFactory: () => Promise<T>, private readonly _valueLifetime?: number) {
+    }
+
+    public async getValue(): Promise<T> {
```
ravipal, line 45: considering the complexity of this class, please add unit test for this.

src/utils/lazy.ts
```diff
@@ -7,21 +7,68 @@ export class Lazy<T> {
...
+                clearTimeout(reset);
+            }, this._valueLifetime);
         }
 
         return this._value;
     }
 }
 
-export default Lazy;
+export class AsyncLazy<T> {
```
ravipal, line 37: Disposable?

src/utils/lazy.ts
bwateratmsft: The timers are like disposables, and `clearTimeout` is how you "dispose" of them. This is a necessary step.

src/utils/lazy.ts
bwateratmsft: The timers are like disposables, and `clearTimeout` is how you "dispose" of them. This is a necessary step.

src/utils/lazy.ts
bwateratmsft: This class is not disposable.

src/utils/lazy.ts
bwateratmsft: Resolved via the refactoring

src/utils/lazy.ts
bwateratmsft: Absolutely!

bwateratmsft: I added UTs, ran them ~10 times locally without any failures.

test/utils/lazy.test.ts
```diff
@@ -0,0 +1,117 @@
...
+                factoryCallCount++;
+                return true;
+            }, 5);
+
+            lazy.value;
+            lazy.value;
+
+            assert.equal(factoryCallCount, 1, 'Incorrect number of value factory calls.');
+
+            await delay(10);
```
karolz-ms, line 37: It is kind of nice that Node is, to a large extent, deterministic in terms of how callbacks are queued and executed

test/utils/lazy.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        test('With lifetime', async () => {
+            let factoryCallCount = 0;
+            const lazy: Lazy<boolean> = new Lazy(() => {
+                factoryCallCount++;
+                return true;
+            }, 5);
+
+            lazy.value;
+            lazy.value;
```
ravipal, line 33: I assume the second call (line 33) is completed within 5 ms. Otherwise the line 35 assert will fail. Should the lifetime be increased to larger number?

test/utils/lazy.test.ts
bwateratmsft: One thing I'm not sure about is, if you have two calls waiting on the same promise--first A then B--I know A starts then B starts. But, does A finish and then B finish, or the other way around, or random-ish?

test/utils/lazy.test.ts
bwateratmsft: Yes, it will be near-instant since it will go through the first `if` block: `if (this._isValueCreated) { ... }`

src/utils/lazy.ts
ravipal: Great. Thanks!

Issue #1856, opened 2020-04-14
The image in Dockerfile is 2.1 after adding Docker files to a .NET Core 5.0 console app


v-mengwe:
**OS:** Windows
**Build Version:**[20200413.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=17741&view=results)
**.NET Core SDK:** 5.0.100-preview.3.20211.8

**Repro Steps:**

1. Create a .NET Core console app by "dotnet new console".
2. Open this project in VS Code-->Prompt to add the required assets-->yes.
3. F1-->Type and select "Docker: Add Docker Files to Workspace"-->.NET: Core Console-->Linux-->Include optional Docker Compose files
5. Check the image in Dockerfile

**Expect:**
The image in dockerfile should be version 5.0.

**Actual:**
The image in dockerfile shows version 2.1.
![image](https://user-images.githubusercontent.com/62053665/79197247-fc6bb280-7e63-11ea-87ac-0c1cb04b23eb.png)

bwateratmsft:
When we implement this, we should put the determined .NET Core version into the scaffold telemetry.

bwateratmsft:
For the fix, I think the right way is to change our regex. We should look inbetween `<TargetFramework>` (or `<TargetFrameworks>`) tags (also using regex), and look for basically `net(coreapp)?\d\.\d`, i.e. either `netcoreappx.y` or `netx.y`.

Something like this (which is probably missing escape characters somewhere):
`<TargetFrameworks?>.*net(coreapp)?(\d+\.\d+).*</TargetFrameworks?>`

It should _probably_ also be a multi-line inspection, because this is valid XML:
```xml
<TargetFrameworks>
    netcoreapp3.1;
    netcoreapp5.0;
</TargetFrameworks>
```

If there are multiple matches, we should take the highest one.

bwateratmsft:
@ravipal can we add telemetry for the resolved .NET Core version?

PR #1866, opened 2020-04-14
adjust net core version extraction logic to handle net5.0

ravipal:
.Net 5.0 preview 3 generates the console project with below TFM
`<TargetFramework>net5.0</TargetFramework>`

During scaffolding we were looking for only netcoreapp5.0, hence it fails to identify net5.0. Also added the logic to look for the version in TargetFrameworks. Picks the first one defined in the list.

Fixes: #1856

### DIFF
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -191,7 +191,7 @@ $volumes_list$
 
 function extractNetCoreVersion(projFileContent: string): string {
     // Parse version from TargetFramework or TargetFrameworks
-    // Example: netcoreapp1.0 or app5.0
+    // Example: netcoreapp1.0 or net5.0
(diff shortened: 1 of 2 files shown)


src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -189,7 +189,26 @@ $volumes_list$
 `;
 // #endregion
 
-function genDockerFile(serviceNameAndRelativePath: string, platform: Platform, os: PlatformOS | undefined, ports: number[], version: string, artifactName: string, assemblyName: string): string {
+function extractNetCoreVersion(projFileContent: string): string {
+    // Parse version from TargetFramework or TargetFrameworks
+    // Example: netcoreapp1.0 or app5.0
```
haniamr, line 194: `net5.0`?

Issue #1459, opened 2019-12-03
In order to use an SSH DOCKER_HOST on OS X and Linux, you must configure an ssh-agent.


maxsatula:
https://github.com/microsoft/vscode-docker/blob/836a76500c1e3edaadccdea944dec31b3e586c2c/src/utils/refreshDockerode.ts#L63

https://github.com/microsoft/vscode-docker/blob/836a76500c1e3edaadccdea944dec31b3e586c2c/src/utils/refreshDockerode.ts#L81

I doubt setting of `SSH_AUTH_SOCK` variable is necessary. I keep it undefined, but have `IdentityAgent`, `IdentityFile`, `AddKeysToAgent` in my `.ssh/config`. However the code referenced at the top makes my configuration unusable for this extension, while it perfectly works everywhere else outside vscode-docker.

Can this check be removed or replaced with something else? There are other options for SSH to avoid interactive password prompt, such as `BatchMode` or `IdentitiesOnly`.

Thank you!

philliphoff:
Some of the manipulation of environment variables related to SSH is due limitations in the way `dockerode` (used to connect to the Docker host) must be initialized.  That said, there may be more that we could do to support a wider variety of SSH configurations.

maxsatula:
Yep, and dockerode does not seem to respect `.ssh/config` file at all, so the solution might not be that simple as I suggested earlier :(

maxsatula:
Nonetheless, I would appreciate a possibility to customize value of `SSH_AUTH_SOCK`.
I run agent as `ssh-agent -a <my_fixed_path_to_UNIX_socket>`, so the path is known and static.

bwateratmsft:
Dockerode indeed does not respect it; Dockerode is using the [ssh2](https://www.npmjs.com/package/ssh2) node package which does not use `.ssh/config`. I would suggest filing this there because it would be a better solution for it to be implemented there.

maxsatula:
> Nonetheless, I would appreciate a possibility to customize value of `SSH_AUTH_SOCK`.

Do you think it makes sense to implement this part here?

AVSuni:
I get the same error message on three separate laptops (one running Ubuntu 18.04 and two running Windows 10). I don't get it. I have followed all the instructions, and this used to work perfectly well until about October/November.

lcxywfe:
[...]

PR #1868, opened 2020-04-15
Refactor Docker context some

### LINKED ISSUE #1819: Add Docker Desktop installation ID to telemetry

bwateratmsft:
This PR does several things:
- Adds Docker Desktop installation ID to telemetry
- Captures the protocol (local pipe, TCP, SSH, HTTP(S), etc.) in more cases than just Docker context
- Captures the source of the ultimate Dockerode host choice (environment vs. VSCode setting vs. there's-only-one-Docker-context vs. default Docker context vs. custom Docker context)
- Offers a way to do totally bring-your-own-settings to Dockerode (related to #1459 or maybe fixes it)

Resolves #1819.

### DIFF
--- a/src/utils/refreshDockerode.ts
+++ b/src/utils/refreshDockerode.ts
@@ -65,6 +65,10 @@ export async function refreshDockerode(): Promise<void> {
                         /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
                         ext.ui.showWarningMessage(localize('vscode-docker.utils.dockerode.sshAgent', 'In order to use an SSH DOCKER_HOST, you must configure an ssh-agent.'), { learnMoreLink: 'https://aka.ms/AA7assy' });
                     }
+
+                    if (dockerodeOptions) {
+                        dockerodeOptions.sshAuthAgent = newEnv.SSH_AUTH_SOCK;
+                    }
                 }
 
                 try {
@@ -108,7 +112,7 @@ async function getDockerOptionsFromDockerContext(actionContext: IActionContext, 
 
         options.host = host; // Intentionally the full URL (docker-modem can figure out the protocol and hostname from it)
         options.port = parsed.port; // docker-modem can figure out the port if it is not explicit in the URL
-        // TODO dockerodeOptions.username = parsed.username;
+        options.username = parsed.username;
 
         actionContext.telemetry.properties.hostProtocol = parsed.protocol;
     } else {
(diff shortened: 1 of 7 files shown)


src/utils/refreshDockerode.ts
```diff
@@ -22,53 +25,122 @@ const SSH_URL_REGEX = /ssh:\/\//i;
...
-            ext.ui.showWarningMessage(localize('vscode-docker.utils.dockerode.sshAgent', 'In order to use an SSH DOCKER_HOST, you must configure an ssh-agent.'), { learnMoreLink: 'https://aka.ms/AA7assy' });
-        }
+    // Currently the environment variable is the only way to configure this in docker-modem
+    if (dockerContext?.Endpoints?.docker?.SkipTLSVerify) {
+        newEnv.DOCKER_TLS_VERIFY = '';
+    } else {
+        newEnv.DOCKER_TLS_VERIFY = '1';
     }
+
+    // TODO: Attach the TLS material to the options
```
bwateratmsft, line 141: Do? @karolz-ms I don't feel strongly one way or the other on how hard we should try to support this, at least without some user demand. what do you think?

src/utils/refreshDockerode.ts
```diff
@@ -22,53 +25,122 @@ const SSH_URL_REGEX = /ssh:\/\//i;
...
-        // If DOCKER_HOST is an SSH URL, we need to configure / validate SSH_AUTH_SOCK for Dockerode
-        // Other than that, we use default settings, so return undefined
-        if (!await validateSshAuthSock(newEnv)) {
-            // Don't wait
-            /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
-            ext.ui.showWarningMessage(localize('vscode-docker.utils.dockerode.sshAgent', 'In order to use an SSH DOCKER_HOST, you must configure an ssh-agent.'), { learnMoreLink: 'https://aka.ms/AA7assy' });
-        }
+    // Currently the environment variable is the only way to configure this in docker-modem
+    if (dockerContext?.Endpoints?.docker?.SkipTLSVerify) {
+        newEnv.DOCKER_TLS_VERIFY = '';
```
karolz-ms, line 136: Please retain the comment about the fact that to disable TLS (certificate) verification you actually need to **set it to empty value**.

src/utils/refreshDockerode.ts
karolz-ms: Right, I **would not** get into the business of configuring TLS from Docker context unless we hear from customers.

PR #1870, opened 2020-04-15
Capture the net core version in telemetry

ravipal:
Fixes: #1856

### DIFF
--- a/src/configureWorkspace/configureDotNetCore.ts
+++ b/src/configureWorkspace/configureDotNetCore.ts
@@ -9,7 +9,7 @@ import * as path from 'path';
 import * as semver from 'semver';
 import * as vscode from 'vscode';
 import { WorkspaceFolder } from 'vscode';
-import { IActionContext } from 'vscode-azureextensionui';
+import { IActionContext, TelemetryProperties } from 'vscode-azureextensionui';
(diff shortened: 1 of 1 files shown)


src/configureWorkspace/configureDotNetCore.ts
```diff
@@ -416,7 +414,7 @@ export async function scaffoldNetCore(context: ScaffolderContext): Promise<Scaff
     const os = context.os ?? (context.os = await context.promptForOS());
     const isCompose = await context.promptForCompose();
 
-    const telemetryProperties = <ConfigureTelemetryProperties>context.telemetry.properties;
+    const telemetryProperties = <TelemetryProperties & ConfigureTelemetryProperties>context.telemetry.properties;
```
bwateratmsft, line 417: I know it wasn't originally your code, but I'm not sure why we are making a local variable here for this, I don't think there are any other areas where we make a local. If we remove it we can also remove all these extra imports.

ravipal: I added this telemetryProperties to make it easier to read (telemetryProperties instead of context.telemtry.properties). Also this gives intellisense on ConfigureTelemetryProperties.

bwateratmsft: That's fair.

Issue #1845, opened 2020-04-09
Integrate new container tutorial


karolz-ms:
Have a pseudo-node in the container tree if no containers are present. The pseudo-node, when clicked, launches a web page with VS Code-specific container tutorial.

bwateratmsft:
Should we also keep the current entry? 
![image](https://user-images.githubusercontent.com/36966225/78939502-462e5480-7a82-11ea-8859-200f9ffe74ae.png)

karolz-ms:
I think so, yes. Just change what it says and does.

bwateratmsft:
We can also have both nodes show. Maybe mockup some screenshots to compare?

BigMorty:
I was thinking of showing just the new message and not the existing one, but if folks think it is important to call out that no containers were found we could keep. but I think it would look better with just the new message.

The new message should read "Get started with Docker containers". Let's use the whale icon we use when connected to DockerHub. The link to use is https://aka.ms/getstartedwithdocker.

bwateratmsft:
I think it's really important to keep the existing node.

PR #1878, opened 2020-04-16
When no container to show, provider a link to docker tutorial

ravipal:
image

Fixes: #1845


bwateratmsft:
I strongly feel we need to keep the "Successfully connected, but no items found" in all of the views.

ravipal:
> I strongly feel we need to keep the "Successfully connected, but no items found" in all of the views.

My preference is to not show. If we don't show any error to user, they can expect that nothing is wrong with the setup and that implies the connection is successful. In fact, I prefer to not show 'successfully connected ..." in other views as well, instead show something like "Getting Started with Docker..." Lets see what others think.

BigMorty:
I would prefer we don’t show the message either. Could we shorten the message in the other panels to “No items found” ?

karolz-ms:
Not show anything (except the tutorial link) if there is nothing to show.
If there is an error (cannot connect to Docker endpoint), show an error.
This is most in line with the VS Code being lean and to-the-point tool IMO.

bwateratmsft:
@BigMorty and I chatted yesterday and he had an alternative idea that I like way better--do a Help and Feedback tab like what Dapr has. Put it at the top of the explorer. Users can move it or hide it as desired.

Why do we need to put it here? Our average user is not a first time Docker user, they don't need help starting containers.

package.json
```diff
@@ -111,7 +112,8 @@
         "onView:dockerImages",
         "onView:dockerNetworks",
         "onView:dockerRegistries",
-        "onView:dockerVolumes"
+        "onView:dockerVolumes",
+        "vscode-docker.views.help"
```
bwateratmsft, line 116:
```suggestion
        "onCommand:vscode-docker.views.help"
```
I'd also move it up with the other commands

src/commands/registerCommands.ts
```diff
@@ -61,6 +61,7 @@ import { logInToDockerCli } from "./registries/logInToDockerCli";
 import { logOutOfDockerCli } from "./registries/logOutOfDockerCli";
 import { pullImage, pullRepository } from "./registries/pullImages";
 import { reconnectRegistry } from "./registries/reconnectRegistry";
+import { reportIssue } from "./Reportissue";
```
bwateratmsft, line 64:
```suggestion
import { reportIssue } from "./reportIssue";
```

src/tree/help/HelpTreeDataProvider.ts
```diff
@@ -0,0 +1,26 @@
...
+import ReadDocumentationNode from './readDocumentationNode';
+import ReportIssueNode from './ReportIssueNode';
+import ReviewIssuesNode from './ReviewIssuesNode';
+
+export default class HelpTreeDataProvider implements vscode.TreeDataProvider<TreeNode> {
+    public getTreeItem(element: TreeNode): vscode.TreeItem | Thenable<vscode.TreeItem> {
+        return element.getTreeItem();
+    }
+
+    public getChildren(element: TreeNode): vscode.ProviderResult<TreeNode[]> {
```
bwateratmsft, line 18: Why add all these TreeNode classes? It would be much easier to just return four `GenericTreeItem`s or `OpenUrlTreeItem`s.

src/tree/registerTrees.ts
```diff
@@ -76,5 +77,8 @@ export function registerTrees(): void {
         await ext.volumesTree.refresh(node);
     });
 
-    registerCommand('vscode-docker.openUrl', async (_context: IActionContext, node: OpenUrlTreeItem) => node.openUrl());
+    ctx.subscriptions.push(
```
bwateratmsft, line 80: I don't think `ctx` is needed here. The rest of the tree registrations simply use `ext.context.subscriptions.push`, we should do the same here.

src/tree/containers/ContainersTreeItem.ts
```diff
@@ -75,9 +78,22 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<ILocalContainerInf
         }
     }
 
-    public getTreeItemForEmptyList(): AzExtTreeItem[] {
-        const dockerTutorialTreeItem = new OpenUrlTreeItem(this, localize('vscode-docker.tree.conatiner.gettingStarted', 'Get started with Docker containers...'), 'https://aka.ms/getstartedwithdocker');
-        dockerTutorialTreeItem.iconPath = getThemedIconPath('docker')
-        return [dockerTutorialTreeItem];
+    protected getTreeItemForEmptyList(): AzExtTreeItem[] {
+        if (this.isNewContainerUser()) {
+            const dockerTutorialTreeItem = new OpenUrlAzExtTreeItem(this, localize('vscode-docker.tree.conatiner.gettingStarted', 'Get started with Docker containers...'), 'https://aka.ms/getstartedwithdocker');
```
bwateratmsft, line 83:
```suggestion
            const dockerTutorialTreeItem = new OpenUrlAzExtTreeItem(this, localize('vscode-docker.tree.container.gettingStarted', 'Get started with Docker containers...'), 'https://aka.ms/getstartedwithdocker');
```

src/tree/containers/ContainersTreeItem.ts
```diff
@@ -75,9 +78,22 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<ILocalContainerInf
...
+            dockerTutorialTreeItem.iconPath = getThemedIconPath('docker')
+            return [dockerTutorialTreeItem];
+        }
+        return super.getTreeItemForEmptyList()
+    }
+
+    private isNewContainerUser(): boolean {
+        return ext.context.globalState.get<boolean>('vscode-docker.conatiner.newContainerUser', true);
```
bwateratmsft, line 91:
```suggestion
        return ext.context.globalState.get<boolean>('vscode-docker.container.newContainerUser', true);
```

src/tree/containers/ContainersTreeItem.ts
```diff
@@ -75,9 +78,22 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<ILocalContainerInf
...
+    protected getTreeItemForEmptyList(): AzExtTreeItem[] {
+        if (this.isNewContainerUser()) {
+            const dockerTutorialTreeItem = new OpenUrlAzExtTreeItem(this, localize('vscode-docker.tree.conatiner.gettingStarted', 'Get started with Docker containers...'), 'https://aka.ms/getstartedwithdocker');
+            dockerTutorialTreeItem.iconPath = getThemedIconPath('docker')
+            return [dockerTutorialTreeItem];
+        }
+        return super.getTreeItemForEmptyList()
+    }
+
+    private isNewContainerUser(): boolean {
```
bwateratmsft, line 90: Let's use a class `boolean | undefined` here and update it to false in `updateNewContainerUser`, so that we don't have to use the memento every time.

src/tree/containers/ContainersTreeItem.ts
```diff
@@ -75,9 +78,22 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<ILocalContainerInf
...
+        return ext.context.globalState.get<boolean>('vscode-docker.conatiner.newContainerUser', true);
+    }
+
+    private async updateNewContainerUser(items: ILocalItem[]): Promise<void> {
+        if (items && items.length > 0 && this.isNewContainerUser()) {
+            await ext.context.globalState.update('vscode-docker.conatiner.newContainerUser', false);
```
bwateratmsft, line 96:
```suggestion
            await ext.context.globalState.update('vscode-docker.container.newContainerUser', false);
```

src/tree/OpenUrlAzExtTreeItem.ts
```diff
@@ -0,0 +1,29 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzExtParentTreeItem, GenericTreeItem } from "vscode-azureextensionui";
+import { openExternal } from "../utils/openExternal";
+import { getThemedIconPath } from "./IconPath";
+import { OpenUrl } from "./OpenUrlTreeItem";
+
+export class OpenUrlAzExtTreeItem extends GenericTreeItem implements OpenUrl {
```
bwateratmsft, line 11: I'm not sure I understand why this was split apart from `OpenUrlTreeItem` and why two classes are needed.

src/tree/OpenUrlTreeItem.ts
```diff
@@ -3,22 +3,26 @@
...
+import * as vscode from 'vscode';
 import { openExternal } from "../utils/openExternal";
-import { getThemedIconPath } from "./IconPath";
 
-export class OpenUrlTreeItem extends GenericTreeItem {
+export interface OpenUrl {
+    openUrl(): Promise<void>
+}
+
+export class OpenUrlTreeItem extends vscode.TreeItem implements OpenUrl {
```
bwateratmsft, line 13: What's wrong with `GenericTreeItem`?

src/tree/help/HelpTreeDataProvider.ts
```diff
@@ -0,0 +1,26 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import TreeNode from '../TreeItem';
+import GetStartedNode from './GetStartedNode';
+import ReadDocumentationNode from './readDocumentationNode';
+import ReportIssueNode from './ReportIssueNode';
+import ReviewIssuesNode from './ReviewIssuesNode';
+
+export default class HelpTreeDataProvider implements vscode.TreeDataProvider<TreeNode> {
```
bwateratmsft, line 13: Why can't we use `AzExtTreeDataProvider`?

src/tree/OpenUrlAzExtTreeItem.ts
ravipal: I tried to use the existing one but that require to pass a parent tree item and have dependency on AzExtTreeItem. Hence a new one that has no dependency on AzExtTreeItem and also allows using icons from vscode.

src/tree/OpenUrlTreeItem.ts
ravipal: I tried to use the existing one but that require to pass a parent tree item and have dependency on AzExtTreeItem. Hence a new one that has no dependency on AzExtTreeItem and also allows using icons from vscode.

bwateratmsft: You can probably make VSCode icons work, if not I'd submit a PR to vscode-azuretools. In any case, we should use the established pattern since it does everything we need it to without all these extra classes and files.

src/tree/help/HelpTreeDataProvider.ts
ravipal: These are simple nodes with no dependency on Azure and simple to use.

src/tree/help/HelpTreeDataProvider.ts
ravipal: IMO, this is easier to use and understand than the tree items and data provider from AzExtTreeItem

src/tree/OpenUrlTreeItem.ts
ravipal: Why unnecessarily use AzExtTreeItem require implementing AzExtParentTreeItem which require implementing loadMoreChildrenImpl and registering command *.loadMore. These are simple links and commands and doesn't require the functionality provided by the AzExtTreeItem which adds complexity.

ravipal: I will try to use AzExtTreeItem and see how it goes.

src/commands/registerCommands.ts
```diff
@@ -61,6 +61,7 @@ import { logInToDockerCli } from "./registries/logInToDockerCli";
 import { logOutOfDockerCli } from "./registries/logOutOfDockerCli";
 import { pullImage, pullRepository } from "./registries/pullImages";
 import { reconnectRegistry } from "./registries/reconnectRegistry";
+import { reportIssue } from "./reportissue";
```
bwateratmsft, line 64:
```suggestion
import { reportIssue } from "./reportIssue";
```
Linux filesystem is case-sensitive so this might prevent building there.

test/TestTerminalProvider.ts
```diff
@@ -52,6 +52,8 @@ class TestTerminal implements vscode.Terminal {
         this._errFilePath = path.join(root, `.err${this._suffix}`);
         this._semaphorePath = path.join(root, `.sem${this._suffix}`);
     }
+    creationOptions: Readonly<vscode.TerminalOptions> | Readonly<vscode.ExtensionTerminalOptions>;
```
bwateratmsft, line 55: This class will be removed soon in #1888; I think that will cause a merge conflict in this so just a heads up. :smile:

package.json
```diff
@@ -2437,10 +2448,9 @@
         }
     },
     "engines": {
-        "vscode": "^1.41.0"
+        "vscode": "^1.42.0"
```
bwateratmsft, line 2451: Is this for `ThemeIcon` or is there something else?

bwateratmsft: Also, if possible I'd suggest splitting the `vscode` vs `vscode-test` stuff out into another PR; if this depends on that, then we can do that one first and the merge into here.

package-lock.json
```diff
@@ -82,9 +82,9 @@
             }
         },
         "@types/adm-zip": {
```
bwateratmsft, line 84: Merging from `master` branch will probably make most of these changes in `package-lock.json` and `package.json` (for versions) go away; it will get the dependency update from #1889.

src/tree/TreeItem.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
```
bwateratmsft, line 1: Is this interface still used?

src/tree/help/HelpsTreeItem.ts
```diff
@@ -0,0 +1,81 @@
...
+    // private get getStartedTreeItem(): AzExtTreeItem {
+    //     const node = new OpenUrlTreeItem(
+    //         this,
+    //         localize('views.help.getStarted', 'Get Started'),
+    //         'https://aka.ms/helppanel_getstarted',
+    //         new vscode.ThemeIcon('star-empty'));
+    //     node.id = '1';
+
+    //     return node;
+    // }
```
rcjsuen, line 43: Should there be a comment here to explain when this code should be uncommented?

src/tree/TreeItem.ts
ravipal: Not needed. missed deleting this file.

package.json
ravipal: Reverted this change.

src/tree/help/HelpsTreeItem.ts
ravipal: Added a comment.

PR #1882, opened 2020-04-17
Require Blazor manifest to be a file

### LINKED ISSUE #1874: An error "EISDIR: illegal operation on a directory, read" occurs when F5ing a .NET Core 2.1 MVC project
**OS:** Mac
**Build version:** [20200413.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=17741&view=results)
**.NET Core SDK:** 2.1

1. Create a .NET Core MVC project by "dotnet new mvc".
2. Open the project in VS Code-->Prompt to add the required asserts-->Yes.
3. F1-->Type and select "Docker: Add Docker Files to Workspace"-->ASP.NET Core-->Linux-->Include optional Docker Compose files.
4. Go to the Debug tab in VSCode, select the "Docker .NET Core Launch"-->F5

**Expect:**
Debug successfully without error.

**Actual:**
The Terminal has an error: EISDIR: illegal operation on a directory, read.
Screen Shot 2020-04-15 at 7 00 15 PM

**More info:**
This bug does not reproduce with .NET Core 3.1/5.0.

bwateratmsft:
Fixes #1874

### DIFF
--- a/src/tasks/netcore/updateBlazorManifest.ts
+++ b/src/tasks/netcore/updateBlazorManifest.ts
@@ -69,6 +69,7 @@ async function transformBlazorManifest(context: DockerRunTaskContext, inputManif
     if (!inputManifest || // Input manifest can't be empty/undefined
         !outputManifest || // Output manifest can't be empty/undefined
         !(await fse.pathExists(inputManifest)) || // Input manifest must exist
+        !(await fse.stat(inputManifest)).isFile() || // Input manifest must be a file
         !volumes || // Volumes can't be undefined
         volumes.length === 0) { // Volumes can't be empty
         // This isn't considered an error case, we'll just return without doing anything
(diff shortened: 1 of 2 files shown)


src/tasks/netcore/updateBlazorManifest.ts
```diff
@@ -69,6 +69,7 @@ async function transformBlazorManifest(context: DockerRunTaskContext, inputManif
     if (!inputManifest || // Input manifest can't be empty/undefined
         !outputManifest || // Output manifest can't be empty/undefined
         !(await fse.pathExists(inputManifest)) || // Input manifest must exist
+        !(await fse.stat(inputManifest)).isFile() || // Input manifest must be a file
```
bwateratmsft, line 72: We may not need the preceding `pathExists` call; I need to see what `stat` returns if the file does not exist.

bwateratmsft: Nope, `stat` fails if the file doesn't exist. Need to keep `pathExists`.

Issue #1840, opened 2020-04-09
Add semantic highlighting support for Dockerfiles


rcjsuen:
With the finalization of the [semantic highlighting](https://github.com/microsoft/vscode/wiki/Semantic-Highlighting-Overview) API in [Visual Studio Code 1.44](https://code.visualstudio.com/updates/v1_44#_extension-authoring), the Docker extension can now look to resolving long standing bug reports about malformed syntax highlighting in Dockerfiles using code instead of a TextMate grammar. I have tried to do a sweep of the GitHub issues that have been opened in the past to capture the various requests and bug reports about syntax highlighting in Dockerfiles.

- https://github.com/microsoft/vscode-docker/issues/43 (request to add syntax highlighting for `HEALTHCHECK` flags)
- https://github.com/microsoft/vscode-docker/issues/90 (request to add syntax highlighting for `ARG` and `ENV` variables)
- https://github.com/microsoft/vscode-docker/issues/75#issue-216515077 (bug example)
- https://github.com/microsoft/vscode-docker/issues/75#issuecomment-291277804 (bug example)
- https://github.com/microsoft/vscode-docker/issues/75#issuecomment-338812822 (bug example)
- https://github.com/microsoft/vscode-docker/issues/75#issuecomment-401341827 (bug example)
- https://github.com/microsoft/vscode-docker/issues/245#issuecomment-379113269 (bug example)
- https://github.com/microsoft/vscode-docker/issues/292#issue-337954539 (bug example)

**Open question:**
The extension currently depends on Visual Studio Code 1.41 which was released in December 2019.

https://github.com/microsoft/vscode-docker/blob/95fb88977958ea2eb59f81369c399986e094bc2e/package.json#L2439-L2441

Adding semantic highlighting support will change this dependency to 1.44. Not sure to me how many users would be affected by this. Does Microsoft have telemetry information about this to know what percentage of this extension's users are on which release of Visual Studio Code?

bwateratmsft:
I took a look at the last 24 hours, here's the numbers we have for Docker extension users:

| Version | Machine Count |
| --- | --- |
| 1.44.0 | 31205 |
| 1.43.2 | 76885 |
| 1.43.1 | 3809 |
| 1.43.0 | 3359 |
| 1.42.1 | 3757 |
| 1.41.1 | 2607 |

These count the highest version a machine reported in the last 24 hours, i.e. people who have upgraded will only show up in 1.44.0.

[...]

PR #1888, opened 2020-04-20
Reuse terminals when applicable

### LINKED ISSUE #251: suggestion: option to reuse terminals
There is no API to let me see if a terminal is in use, so whenever a command is run we spin up a new terminal.  This leads to a bunch of unused terminals to clean up.  

Suggest to investigate an option that reuses the same terminal, commands are effectively queued up.

bwateratmsft:
Fixes #251 
Unblocks #1840


src/commands/containers/attachShellContainer.ts
```diff
@@ -39,7 +40,5 @@ export async function attachShellContainer(context: IActionContext, node?: Conta
         shellCommand
     );
 
-    const terminal = ext.terminalProvider.createTerminal(`Shell: ${node.containerName}`);
-    terminal.sendText(terminalCommand);
-    terminal.show();
+    await executeAsTask(context, terminalCommand, `Shell: ${node.containerName}`, /* addDockerEnv: */ true);
```
bwateratmsft, line 43: Shell quoting is causing problems here...

bwateratmsft:
Closing this. There are serious problems with shell quoting that make this extremely difficult.

PR #1892, opened 2020-04-21
Pull images from Images pane

### LINKED ISSUE #1313: [Request] Pull image from images list
It would be great to be able to pull a newer version of an image from the list of images, either the latest tag if you select the image, or the specific tag if you select a specific tag in the image's list

bwateratmsft:
Allows pulling images from Images pane, without login. It was different enough from the existing pullImage from the Registries pane that I felt like it needed its own command.

Fixes #1313


package.json
```diff
@@ -352,20 +353,25 @@
                     "group": "images_2_general@1"
                 },
                 {
-                    "command": "vscode-docker.images.push",
```
bwateratmsft, line 355: The diff makes it hard to tell but this is actually inserting `pull` above `push` and re-indexing the others.

src/commands/registerCommands.ts
```diff
@@ -138,7 +140,7 @@ export function registerCommands(): void {
     registerCommand('vscode-docker.context.use', useDockerContext);
     registerCommand('vscode-docker.context.inspect', inspectDockerContext);
     registerCommand('vscode-docker.context.remove', removeDockerContext);
-    registerCommand('vscode-docker.installDocker', installDocker);
+    registerWorkspaceCommand('vscode-docker.installDocker', installDocker);
```
bwateratmsft, line 143: Unrelated, I think this needs to be set up as a workspace command

ravipal: registerWorkspaceCommand ensured the command working on the local host?

bwateratmsft: If I understand it correctly, yes. I want to take some time to investigate and be sure.

src/commands/installDocker.ts
```diff
@@ -5,10 +5,19 @@
 
 import * as os from 'os';
 import { IActionContext } from 'vscode-azureextensionui';
+import { localize } from '../localize';
+import { DockerExtensionKind, getVSCodeRemoteInfo } from '../utils/getVSCodeRemoteInfo';
 import { openExternal } from '../utils/openExternal';
 import { MacDockerInstaller, WindowsDockerInstaller } from './dockerInstaller';
 
 export async function installDocker(context: IActionContext): Promise<void> {
+    const remoteInfo = getVSCodeRemoteInfo(context);
```
bwateratmsft, line 14: @ravipal I looked into `registerWorkspaceCommand`, it's essentially made to warn you if you are in UI mode but try to do something like `docker build`, which must always run in the workspace. So, it doesn't do what we need.

Instead I put this block in; if remoting is in use then we definitely don't want to get into the business of trying to install Docker (plus the target of remoting would almost always be Linux where we don't have support anyway).

Can you take a look at this code and see if it looks OK?

ravipal: Can this check be added during register command? something like `registerLocalCommand` so in future other such commands can reuse the code.

bwateratmsft: Good idea! I'll do that.

PR #1899, opened 2020-04-21
Add semantic highlighting support via the LSP

rcjsuen:
This is a preliminary implementation of semantic highlighting for Dockerfiles. This pull request intends to address #1840 which in a way is itself a collection of many long standing bug reports about malformed syntax highlighting in Dockerfiles.

To test this feature, simply enable semantic highlighting via the settings (or if you're using the JSON settings editor, the key is `"editor.semanticHighlighting.enabled"`).

**Known issues:**
There are a few issues that will be fixed in the next release of the language server. You can copy/paste the Dockerfile below into VS Code to see what I mean.

1. parameter that precedes an escaped string does not get coloured.
1. quoted string values of a variable do not get highlighted as a string
1. variables within quoted strings are not coloured

```Dockerfile
FROM alpine
# second abc is white compared to the first abc that is coloured correctly
RUN echo abc"e"
RUN echo abc\'e\'
# variable value in an enclosed string does not get coloured
ENV abc="def"
# $abc as a variable reference should be coloured differently inside the string
RUN echo "$abc"
```

**Q&A:**
1. How do I enable/disable semantic highlighting only for Dockerfiles?

   From what I understand, you must use the JSON settings editor for this. Use the following JSON snippet to toggle the setting as you please. I'm not sure how you can achieve this with the GUI editor.

   ```JSON
   "[dockerfile]": {
       "editor.semanticHighlighting.enabled": true
   }
   ```

1. Why does your [online Dockerfile editor](https://rcjsuen.github.io/dockerfile-language-service/) seem to colour more things than VS Code?

   I specifically made changes to my own editor so that it would provide custom colours for various situations. Based on the [semantic highlighting wiki page](https://github.com/microsoft/vscode/wiki/Semantic-Highlighting-Overview), there appears to be a way to add custom colour mappings though I may be misunderstanding what is written on the page. I think this will require more investigation and can be addressed in a different pull request so I hope that is not a blocker for this pull request.

1. What is the ETA for the next release of the Dockerfile language server that will address the aforementioned bugs?

   I don't have a timeline to share at the moment.


rcjsuen:
@bwateratmsft Thanks for taking a look.
1. Yeah, I haven't really decided what the best course of action is here as some people just use public images on Docker Hub whereas some others only use private container registries.
1. Yes, that is the same bug that I've pointed out.
1. Good catch! That seems to be a parsing error related to line 4 (`RUN echo abc\'e\'`). Removing that line seems to fix the problem. I'll have to do some poking around to see what the issue is there as other features like "hovering over a keyword" or "highlighting referenced variables" seems to be unaffected. 🤔

rcjsuen:
@bwateratmsft Another interesting thing is that if you type `HEALTH` and then use <kbd>Control</kbd>+<kbd>Space</kbd> to trigger suggestions, it does pop up!

image

Looks like there might be something else that may be amiss...?

bwateratmsft:
I guess also, # 1 is not a new issue, and nobody has complained yet...

rcjsuen:
@bwateratmsft I was able to reproduce the issue with a simple piece of code using the Monaco Editor so I think it may be related to the grammar of file? The system seems convinced that the text cursor is inside a string so it doesn't appear to want to invoke code completion.

1. Open the [Monaco Editor Playground](https://microsoft.github.io/monaco-editor/playground.html).
1. Copy/paste the code below on the left hand side.
1. Click the 'Run' button.
1. Open the debugger and dock it to the side. Open up the console.
1. Follow the instructions typing on the 4th, 6th, and 7th line. Observe the differences in behaviour with regards to when something gets printed to the console or not.


```JavaScript
const content = 
`# type on line 4, it works
# type on line 6, it fails
# type on line 7, it works

RUN \\'e\\'

`;
const language = "dockerfile";
monaco.languages.registerCompletionItemProvider(language, {
    provideCompletionItems: function(_model, position) {
        console.log(`provideCompletionItems called on line ${position.lineNumber} character ${position.column}`)
        return {
            suggestions: []
        };
    }
});

monaco.editor.create(document.getElementById("container"), {
    value: content,
    language: language
});
```

rcjsuen:
> I guess also, # 1 is not a new issue, and nobody has complained yet...

Correct, it's not a new issue and it has been like that since I first implemented Docker Hub linking support.

Regarding the code completion issue, I have opened microsoft/vscode#95924 and microsoft/monaco-editor#1935.

rcjsuen:
@bwateratmsft I decided to just pull the trigger and cut a release so the three aforementioned semantic highlighting bugs should be fixed now.

bwateratmsft:
Awesome, thanks! I verified and all three are fixed.

rcjsuen:
@bwateratmsft [FYI](https://github.com/microsoft/vscode/issues/95924#issuecomment-620613814) regarding the code completion bug you discovered.

bwateratmsft:
@rcjsuen Thanks!

PR #1902, opened 2020-04-22
Install Docker using notification

### LINKED ISSUE #1826: Move 'Install Docker...' link from the Docker view to notification

ravipal:
Earlier we were showing install docker link 4 times in the docker view. Now the install link will be shown as notification like below.
image

Docker view will show
image

Fixes #1826

### DIFF
--- a/src/commands/dockerInstaller.ts
+++ b/src/commands/dockerInstaller.ts
@@ -81,8 +81,8 @@ export class MacDockerInstaller extends DockerInstallerBase {
 }
 
 export async function showDockerInstallNotification(): Promise<void> {
-    const installMessageLinux = localize('vscode-docker.commands.dockerInstaller.installDockerInfo', 'Docker is not installed. Would you like to learn more about installing docker?');
-    const installMessageNonLinux = localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to install?');
+    const installMessageLinux = localize('vscode-docker.commands.dockerInstaller.installDockerInfo', 'Docker is not installed. Would you like to learn more about installing Docker?');
+    const installMessageNonLinux = localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to install it?');
     const learnMore = localize('vscode-docker.commands.dockerInstaller.learnMore', 'Learn more')
     const install = localize('vscode-docker.commands.dockerInstaller.install', 'Install')
     const osProvider: LocalOSProvider = new LocalOSProvider();
--- a/src/commands/dockerInstaller.ts
+++ b/src/commands/dockerInstaller.ts
@@ -38,7 +38,7 @@ export abstract class DockerInstallerBase {
 
     private async preInstallCheck(): Promise<boolean> {
         let proceedInstall = true;
-        if (await dockerInstallStatusProvider.isDockerInstalled()) {
+        if (await dockerInstallStatusProvider.isDockerInstalledRealTimeCheck()) {
             const reinstallMessage = localize('vscode-docker.commands.DockerInstallerBase.reInstall', 'Docker Desktop is already installed. Would you like to reinstall?');
             const install = localize('vscode-docker.commands.DockerInstallerBase.reinstall', 'Reinstall')
             const response = await vscode.window.showInformationMessage(reinstallMessage, ...[install]);
--- a/src/commands/dockerInstaller.ts
+++ b/src/commands/dockerInstaller.ts
@@ -9,6 +9,7 @@ import { LocalFileSystemProvider } from '../debugging/coreclr/fsProvider';
 import { OSTempFileProvider } from '../debugging/coreclr/tempFileProvider';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
+import { dockerInstallStatusProvider } from '../utils/DockerInstallStatusProvider';
 import { streamToFile } from '../utils/httpRequest';
 import LocalOSProvider from '../utils/LocalOSProvider';
 import { execAsync } from '../utils/spawnAsync';
@@ -19,17 +20,32 @@ export abstract class DockerInstallerBase {
     protected abstract getInstallCommand(fileName: string): string;
 
     public async downloadAndInstallDocker(): Promise<void> {
-        const downloadingMessage: string = localize('vscode-docker.commands.DockerInstallerBase.downloading', 'Downloading Docker installer...');
-        const installationMessage: string = localize('vscode-docker.commands.DockerInstallerBase.installationMessage', 'The Docker Desktop installation is started. Complete the installation and then start Docker Desktop.');
+        const shouldInstall: boolean = await this.preInstallCheck();
+        if (shouldInstall) {
+            const downloadingMessage: string = localize('vscode-docker.commands.DockerInstallerBase.downloading', 'Downloading Docker installer...');
+            const installationMessage: string = localize('vscode-docker.commands.DockerInstallerBase.installationMessage', 'The Docker Desktop installation is started. Complete the installation and then start Docker Desktop.');
 
-        const downloadedFileName: string = await vscode.window.withProgress(
-            { location: vscode.ProgressLocation.Notification, title: downloadingMessage },
-            async () => this.downloadInstaller());
+            const downloadedFileName: string = await vscode.window.withProgress(
+                { location: vscode.ProgressLocation.Notification, title: downloadingMessage },
+                async () => this.downloadInstaller());
 
-        const command = this.getInstallCommand(downloadedFileName);
-        // eslint-disable-next-line @typescript-eslint/no-floating-promises
-        vscode.window.showInformationMessage(installationMessage);
-        await this.install(downloadedFileName, command);
+            const command = this.getInstallCommand(downloadedFileName);
+            // eslint-disable-next-line @typescript-eslint/no-floating-promises
+            vscode.window.showInformationMessage(installationMessage);
+            await this.install(downloadedFileName, command);
+        }
+    }
+
+    private async preInstallCheck(): Promise<boolean> {
+        let proceedInstall = true;
+        if (await dockerInstallStatusProvider.isDockerInstalled()) {
+            const reinstallMessage = localize('vscode-docker.commands.DockerInstallerBase.reInstall', 'Docker Desktop is already installed. Would you like to reinstall?');
+            const install = localize('vscode-docker.commands.DockerInstallerBase.reinstall', 'Reinstall')
+            const response = await vscode.window.showInformationMessage(reinstallMessage, ...[install]);
+            proceedInstall = response !== undefined;
+        }
+
+        return proceedInstall;
     }
 
     private async downloadInstaller(): Promise<string> {
--- a/src/commands/dockerInstaller.ts
+++ b/src/commands/dockerInstaller.ts
@@ -19,13 +19,8 @@ export abstract class DockerInstallerBase {
     protected abstract getInstallCommand(fileName: string): string;
 
     public async downloadAndInstallDocker(): Promise<void> {
-        const confirmInstall: string = localize('vscode-docker.commands.DockerInstallerBase.confirm', 'Are you sure you want to install Docker on this machine?');
-        const installTitle: string = localize('vscode-docker.commands.DockerInstallerBase.install', 'Install');
         const downloadingMessage: string = localize('vscode-docker.commands.DockerInstallerBase.downloading', 'Downloading Docker installer...');
         const installationMessage: string = localize('vscode-docker.commands.DockerInstallerBase.installationMessage', 'The Docker Desktop installation is started. Complete the installation and then start Docker Desktop.');
-
-        // no need to check result - cancel will throw a UserCancelledError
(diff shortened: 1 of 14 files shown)


src/commands/dockerInstaller.ts
```diff
@@ -84,3 +79,17 @@ export class MacDockerInstaller extends DockerInstallerBase {
         terminal.show();
     }
 }
+
+export async function showDockerInstallNotification(): Promise<void> {
+    const installMessageLinux = localize('vscode-docker.commands.dockerInstaller.installDockerInfo', 'Docker is not installed. Would you like to learn more about installing docker?');
```
karolz-ms, line 84:
```suggestion
    const installMessageLinux = localize('vscode-docker.commands.dockerInstaller.installDockerInfo', 'Docker is not installed. Would you like to learn more about installing Docker?');
```

src/commands/dockerInstaller.ts
```diff
@@ -84,3 +79,17 @@ export class MacDockerInstaller extends DockerInstallerBase {
...
+    const installMessageNonLinux = localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to install?');
```
karolz-ms, line 85:
```suggestion
    const installMessageNonLinux = localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to install it?');
```

src/utils/DockerInstallStatusProvider.ts
```diff
@@ -0,0 +1,36 @@
...
+            try {
+                await execAsync('docker -v');
+                return true; // As long as the docker command did't throw exception, assume it is installed.
+            } catch (error) {
+                return false; // docker not installed
+            }
+        }, this.maxCacheTime);
+    }
+
+    public static getInstance(): DockerInstallStatusProvider {
```
karolz-ms, line 25: We use a simpler pattern for singletons in the Docker extension codebase, e.g. see https://github.com/microsoft/vscode-docker/blob/master/src/utils/dockerContextManager.ts#L241

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -112,11 +124,18 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
...
             this._itemsFromPolling = undefined;
             this.failedToConnect = false;
+            this._currentDockerStatus = 'Running';
         } catch (error) {
             this._currentItems = undefined;
             this.failedToConnect = true;
             context.telemetry.properties.failedToConnect = 'true';
-            return this.getDockerErrorTreeItems(context, error);
+
+            if (!this._currentDockerStatus) {
```
karolz-ms, line 133: I would not wait inside `loadMoreChildrenImpl` to do a Docker status refresh, even if this is just in the error code path. Can we do the status refresh asynchronously?

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -318,13 +336,21 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
...
+        let pollingDockerStatus: DockerStatus;
+        let isDockerStatusChanged = false;
+
         try {
             this._itemsFromPolling = await this.getSortedItems();
-        } catch {
+            pollingDockerStatus = 'Running';
+        } catch (error) {
             this._itemsFromPolling = undefined;
+            pollingDockerStatus = await this._dockerStatusProvider.isDockerInstalled() ? 'Installed' : 'NotInstalled';
```
karolz-ms, line 347: This is on the UI refresh path. I am worried about performance. Should I be?

src/tree/LocalRootTreeItemBase.ts
ravipal: The status is needed to construct the error nodes. EIther 'docker installed?' or 'docker running?'. This is the same behavior as how the code waits on getSortedItems() in line 124.

src/tree/LocalRootTreeItemBase.ts
ravipal: dockerStatusProvider uses Lazy, so it uses cached value which gets refreshed every 5 seconds. On my machine when the docker is not installed the command takes 10 ms and when docker installed and not running it takes 500 ms. I don't think this will cause any perf issue with the cached value.

src/tree/LocalRootTreeItemBase.ts
karolz-ms: No it isn't quite the same--you are introducing an extra wait to spawn a process and do `docker -v` if initial tree population attempt fails.

Assuming Docker is **not installed**, but the user has Docker view open--how often will we run `docker -v` after this change?

src/tree/LocalRootTreeItemBase.ts
bwateratmsft: To be fair, it's only on the _error_ UI refresh path. If we assume that most people have the extension working most of the time, then this will be hit very little.

src/utils/DockerInstallStatusProvider.ts
```diff
@@ -0,0 +1,33 @@
...
+
+class DockerInstallStatusProvider implements IDockerInstallStatusProvider {
+    private maxCacheTime: number = 5000;
+    private isDockerInstalledLazy: AsyncLazy<boolean>;
+
+    public constructor() {
+        this.isDockerInstalledLazy = new AsyncLazy<boolean>(async () => {
+            try {
+                await execAsync('docker -v');
+                return true; // As long as the docker command did't throw exception, assume it is installed.
```
bwateratmsft, line 21: @ravipal I do have another suggestion; once Docker is installed it's _very_ unlikely that it will become uninstalled during the session. Once we get installed = true, we can skip the `docker -v` calls from then on.

Worst case, if it does get uninstalled, then we'll show the "not running" error instead of the "not installed" error, which I'm totally OK with given the infrequency of the problem.

src/tree/LocalRootTreeItemBase.ts
ravipal: This gets called only when there is a change in dockerinstall status which is updated by hasChanged() function. For user with no docker installed, this will gets called once per pane until the docker status change. hasChanged() gets called for every 2 seconds.

src/utils/DockerInstallStatusProvider.ts
bwateratmsft: This would be beneficial in the probably much-more-common case that Docker is installed but not running, by avoiding repeated `docker -v` calls.

ravipal: If docker is installed, then this method wouldn't get called. For docker not running case, we still should continue to check the status.

src/tree/LocalRootTreeItemBase.ts
ravipal: Also note that line 133 if condition check will make sure that it gets called only when the `this._currentDockerStatus` is undefined which is the first time when user opens the docker view. The next refresh will not await.

src/utils/DockerInstallStatusProvider.ts
bwateratmsft: Ah, I see that is implemented via `this._currentDockerStatus` in `LocalRootTreeItemBase`. Still, I would consider moving this functionality into this class; it would make `LocalRootTreeItemBase` a little cleaner and abstract away install status, leaving it here where it fits better IMO.

karolz-ms: I like @bwateratmsft idea. If we determine during a session that Docker is installed, subsequent failure to refresh any Docker pane (containers, images, etc) is enough to consider Docker as "not running". We do not need to repeat `docker -v` check.

src/utils/DockerInstallStatusProvider.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { AsyncLazy } from "./lazy";
+import { execAsync } from "./spawnAsync";
+
+export interface IDockerInstallStatusProvider {
+    isDockerInstalled(): Promise<boolean>
+}
+
+class DockerInstallStatusProvider implements IDockerInstallStatusProvider {
+    private maxCacheTime: number = 5000;
```
karolz-ms, line 14: How about checking less frequently? Suggest 30 seconds.

src/utils/DockerInstallStatusProvider.ts
bwateratmsft: If done that way we can remove `DockerStatus` / `_currentDockerStatus`, and let the refresh logic be:

1. No error? Yay!
1. Error. Is Docker installed?
    1. Yes: not running
    1. No: not installed

src/tree/LocalRootTreeItemBase.ts
karolz-ms: Just realized this code path will be executed at most once during a session? If that is indeed the case, i am much less worried. Feel free to resolve if my assumption is right

src/utils/DockerInstallStatusProvider.ts
bwateratmsft: An additional thought; installing Docker doesn't (currently) cause it to be automatically started, and I think it would be a great user experience if the UI magically transitioned from "Docker isn't installed" to "Docker isn't running".

src/tree/LocalRootTreeItemBase.ts
karolz-ms: I think we can mitigate the concern by adjusting how `DockerInstallStatusProvider` works

src/utils/DockerInstallStatusProvider.ts
ravipal: I am not sure I follow the suggestion. `DockerInstallStatusProvider` provide the current docker install status either installed or not installed and it caches the value for 5 seconds to optimize for perf. The caller has a timer that periodically tries to refresh the node and if it fails to refresh, then checks for docker status and if it is changed, then it decides to refresh the tree view.
The current implementation will update the view from 'Docker not installed' to 'Docker not running' id docker is installed.

src/tree/LocalRootTreeItemBase.ts
ravipal: That is right. This will gets called once per session.

src/utils/DockerInstallStatusProvider.ts
bwateratmsft: If it only gets called once, then the UI can't change without restarting VSCode.

ravipal: Now docker status is cached forever once it is installed.

src/utils/DockerInstallStatusProvider.ts
```diff
@@ -0,0 +1,36 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { AsyncLazy } from "./lazy";
+import { execAsync } from "./spawnAsync";
+
+export interface IDockerInstallStatusProvider {
+    isDockerInstalled(): Promise<boolean>
+}
+
+class DockerInstallStatusProvider implements IDockerInstallStatusProvider {
```
bwateratmsft, line 13: Although it's unlikely the user would try, perhaps we should block the `installDocker` command by checking if Docker is already installed using this.

ravipal: Added this check. Note that I am not blocking the installation, instead warning the user and it is user's choice.

src/utils/DockerInstallStatusProvider.ts
```diff
@@ -16,21 +16,28 @@ class DockerInstallStatusProvider implements IDockerInstallStatusProvider {
...
             }
+            return dockerInstalled;
         }, this.maxCacheTime);
     }
 
     public async isDockerInstalled(): Promise<boolean> {
         return await this.isDockerInstalledLazy.getValue();
     }
+
+    public async isDockerInstalledRealTimeCheck(): Promise<boolean> {
```
bwateratmsft, line 33: How come we need a realtime check for the install command?

ravipal: We don't want to use cached value which may not be accurate especially for the case where it was installed and then uninstalled.

ravipal: Also we refresh every 30 seconds, so it could have a stale data. For this install flow, we should do real time check.

bwateratmsft: That makes sense; user has it installed and launches VSCode; uninstalls, then uses the command to install.

src/utils/lazy.ts
```diff
@@ -7,13 +7,17 @@ export class Lazy<T> {
...
 
-    public constructor(private readonly valueFactory: () => T, private readonly _valueLifetime?: number) {
+    public constructor(private readonly valueFactory: () => T, private _valueLifetime?: number) {
     }
 
     public get isValueCreated(): boolean {
         return this._isValueCreated;
     }
 
+    public clearLifeTime(): void {
```
karolz-ms, line 17: (minor name change suggestion--represents intent more clearly)
 
```suggestion
    public cacheForever(): void {
```

PR #1906, opened 2020-04-23
Cancel pending Dockerode calls upon context switch

### LINKED ISSUE #1815: The Docker view in the VS Code jumps between local/remote shortly after changing context
**OS:** Windows
**Build Version:** [20200401.21](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=17137&view=results)

**Repro Steps:**
1. Create an Ubuntu 18.04 LTS machine with SSH access--as a remote machine.
2. Install Docker in this Linux machine.
3. In local machine, install Remote-SSH extension in VS Code local instance.
4. Connect to the remote Ubuntu machine via SSH in VS Code.
5. Switch to the VS Code local instance use 'docker context' command to create a context for connecting to the remote machine:
docker context create --docker "host=ssh://username@machineIP" contextname
6. Switch Docker context to the remote machine and verify that it works from command line:
docker context use contextname
docker ps
docker images
7. Check the Docker view in the VS Code local instance.

**Expect:**
Docker view in the VS Code local instance shows Docker assets on the remote machine.

**Actual:**
Docker view in the VS Code local instance can't show the Docker assets on the remote machine.
image

The context output shows as below:
PS C:\Users\AzureTest> docker context inspect
[
    {
        "Name": "test",
        "Metadata": {},
        "Endpoints": {
            "docker": {
                "Host": "ssh://serene@172.18.3.7",
                "SkipTLSVerify": false
            }
        },
        "TLSMaterial": {},
        "Storage": {
            "MetadataPath": "C:\\Users\\AzureTest\\.docker\\contexts\\meta\\9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08",
            "TLSPath": "C:\\Users\\AzureTest\\.docker\\contexts\\tls\\9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08"
        }
    }
]

bwateratmsft:
Fixes #1815


src/utils/promiseUtils.ts
```diff
@@ -24,3 +25,12 @@ export async function delay(ms: number, token?: vscode.CancellationToken): Promi
         }
     });
 }
+
+export async function getCancelPromise(token: vscode.CancellationToken, message?: string): Promise<never> {
+    return new Promise((resolve, reject) => {
+        const disposable = token.onCancellationRequested(() => {
+            disposable.dispose();
+            reject(new UserCancelledError(message));
```
karolz-ms, line 33: I guess if you squint, you can argue it was user action that cancelled the operation in the scenario addressed by this change

src/utils/promiseUtils.ts
```diff
@@ -24,3 +25,12 @@ export async function delay(ms: number, token?: vscode.CancellationToken): Promi
         }
     });
 }
+
+export async function getCancelPromise(token: vscode.CancellationToken, message?: string): Promise<never> {
```
karolz-ms, line 29: What do you think about subclassing `CancellationTokenSource` and adding this method to the new class? This way we can deal with just one thing in the rest of the code

src/utils/promiseUtils.ts
bwateratmsft: My thoughts exactly! Though I could make that a generic.

src/utils/promiseUtils.ts
bwateratmsft: That's an interesting idea, I'll think about it.

bwateratmsft: @philliphoff pointed out that in some cases (for example, `DebugConfigurationProvider.resolveDebugConfiguration`), we do not have `CancellationTokenSource`, only `CancellationToken`. So it seemed best to have both this method, and a derivative class.

Issue #1844, opened 2020-04-09
New panel for Docker context


karolz-ms:
Panel that shows Docker contexts and is a sibling of containers, networks etc.
This panel should be at the bottom and collapsed by default
Current context in use is clearly marked (bolded)
Displays context description on hover.
Hookup the use, delete, and inspect context commands to context menu items named the same.

Later this year this panel will also show all shared contexts a user has in DockerHub, assuming they are signed into DockerHub.

bwateratmsft:
One thing about the hover text--it may have accessibility impact, we would want to validate that.

karolz-ms:
Good point. 

The hover thing is a nice-to-have; if we find spending a lot of time on this, we can just drop it.

bwateratmsft:
This is interesting: https://code.visualstudio.com/api/references/vscode-api#TreeItem

They have a `description` property which is rendered less prominently--probably similar to quick pick items' description/detail--that might be a good place for it. Or, it might be a good place for the "currently selected" label.

BigMorty:
After out design session today, here is what the current thinking is:
- Panel will be named "CONTEXTS"
- Panel will be above the Help and Feedback panel
- Panel will collapsed by default (if VS Code offers this)
- Panel will have Gear, Refresh, and Help icons
- Help icon should take user to new help doc that discusses Contexts
- All local contexts will be displayed in flat list (no parent item)
- Items will show Name & Description by default - Docker endpoint will be the other available field
- Items will not have a icon unless it is the active icon and then the plug icon will be displayed
- Gear icon will show Label, Description, Reset settings
- Context menu items will be Inspect & Use in first section and Remove in the second section

karolz-ms:
Thanks @BigMorty! Two questions:

> Help icon should take user to new help doc that discusses Contexts

Do we really need to have a new documentation topic for contexts? Can we just link to [Docker context doc](https://docs.docker.com/engine/context/working-with-contexts/) instead? We really just expose basic operations around contexts, so having a VSCode docs for them seems like a waste of electrons.

> Gear icon will show Label, Description, Reset settings

I assume this is to the user can list context by Description instead of by Name? (in Docker lingo the context has "Name", not "Label"). Do we really need this Gear icon and associated functionality? Can we just populate the item `description` with context Description, like @bwateratmsft [suggested above](https://github.com/microsoft/vscode-docker/issues/1844#issuecomment-611745695)?

bwateratmsft:
[...]

PR #1916, opened 2020-04-28
Docker context panel

ravipal:
A new panel is added to show the docker context.

Fixes: #1844


src/commands/context/useDockerContext.ts
```diff
@@ -5,15 +5,27 @@
...
+            ...actionContext,
+            noItemFoundErrorMessage: localize('vscode-docker.commands.contexts.use.noContexts', 'No contexts are available to use')
+        });
+        invokedFromCommandPalette = true;
+    }
 
-    await dockerContextManager.use(selectedContext.Name);
+    await node.use(actionContext);
+    await ext.contextsRoot.refresh();
+    dockerContextManager.expediteContextCheck();
```
karolz-ms, line 25: Call to `dockerContextManager.use()` does `expediteContextCheck()` internally, so no need to call it explicitly here.

src/tree/contexts/ContextProperties.ts
```diff
@@ -0,0 +1,14 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { ITreePropertyInfo } from "../settings/ITreeSettingInfo";
+
+export type ContextProperty = 'Name' | 'Description' | 'DockerEndpoint';
+
+export const contextProperties: ITreePropertyInfo<ContextProperty>[] = [
+    { property: 'Name', exampleValue: 'my-context' },
+    { property: 'Description', exampleValue: 'remote linux VM on Azure' },
+    { property: 'DockerEndpoint', exampleValue: 'my-ssh://user1@55.55.5.555' }
```
karolz-ms, line 13:
```suggestion
    { property: 'DockerEndpoint', exampleValue: 'ssh://user1@55.55.5.555' }
```

src/tree/contexts/ContextTreeItem.ts
```diff
@@ -0,0 +1,61 @@
...
+            return getThemedIconPath('connect');
+        }
+    }
+
+    public async deleteTreeItemImpl(context: IActionContext): Promise<void> {
+        await dockerContextManager.remove(this.name);
+    }
+
+    public async inspect(context: IActionContext): Promise<string> {
+        return await dockerContextManager.inspect(this.name)
```
karolz-ms, line 55:
```suggestion
        return dockerContextManager.inspect(this.name)
```

src/tree/contexts/ContextTreeItem.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public async use(context: IActionContext): Promise<void> {
+        return await dockerContextManager.use(this.name);
```
karolz-ms, line 59:
```suggestion
        return dockerContextManager.use(this.name);
```

src/tree/contexts/ContextsTreeItem.ts
```diff
@@ -0,0 +1,84 @@
...
+import { ITreeSettingWizardInfo } from '../settings/ITreeSettingsWizardContext';
+import { ContextGroupTreeItem } from './ContextGroupTreeItem';
+import { contextProperties, ContextProperty } from "./ContextProperties";
+import { ContextTreeItem } from './ContextTreeItem';
+import { LocalContextInfo } from "./LocalContextInfo";
+
+export class ContextsTreeItem extends LocalRootTreeItemBase<LocalContextInfo, ContextProperty> {
+    public treePrefix: string = 'contexts';
+    public label: string = localize('vscode-docker.tree.Contexts.label', 'Contexts');
+    public configureExplorerTitle: string = localize('vscode-docker.tree.Contexts.configure', 'Configure Contexts explorer');
```
karolz-ms, line 21:
```suggestion
    public configureExplorerTitle: string = localize('vscode-docker.tree.Contexts.configure', 'Configure Docker Contexts Explorer');
```

src/tree/contexts/LocalContextInfo.ts
```diff
@@ -0,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IDockerContextListItem } from "../../utils/dockerContextManager";
+import { ILocalItem } from "../LocalRootTreeItemBase";
+
+/**
+ * Wrapper class for Dockerode item, which has inconsistent names/types
```
karolz-ms, line 10: Docker context is not a Dockerode item. Suggest something along the lines of

> Wrapper for IDockerContextListItem that ensures correct comparison and change detection.

src/commands/context/removeDockerContext.ts
```diff
@@ -7,23 +7,40 @@ import * as vscode from 'vscode';
...
+
+    let removeConfirmationMessage: string;
+    if (nodes.length === 1) {
+        removeConfirmationMessage = localize('vscode-docker.commands.context.remove.confirmSingle', 'Are you sure you want to remove context \'{0}\'?', nodes[0].name);
+    } else {
+        const nonCurrentNodes = nodes.filter(n => !n.current)
+        if (nodes.length === nonCurrentNodes.length) {
+            removeConfirmationMessage = localize('vscode-docker.commands.context.remove.confirmMultiple', 'Are you sure you want to remove the selected contexts?');
+        } else {
+            removeConfirmationMessage = localize('vscode-docker.commands.context.remove.confirmSingleFiltered', 'One of the selected context is being used, which can not be removed. Do you want to remove the rest of the contexts?');
```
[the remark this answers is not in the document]
bwateratmsft: Or "other" Docker contexts?

src/commands/registerCommands.ts
```diff
@@ -141,9 +142,13 @@ export function registerCommands(): void {
     registerCommand('vscode-docker.volumes.prune', pruneVolumes);
     registerCommand('vscode-docker.volumes.remove', removeVolume);
 
-    registerCommand('vscode-docker.context.use', useDockerContext);
-    registerCommand('vscode-docker.context.inspect', inspectDockerContext);
-    registerCommand('vscode-docker.context.remove', removeDockerContext);
+    registerCommand('vscode-docker.contexts.configureExplorer', configureDockerContextsExplorer);
+    registerCommand('vscode-docker.contexts.help', dockerContextsHelp);
+    registerCommand('vscode-docker.contexts.inspect', inspectDockerContext);
```
bwateratmsft, line 147: `inspect` \ `remove` \ `use` may need to do `registerWorkspaceCommand` as they result in command line calls...I think. Let's investigate and figure it out for sure.

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -367,6 +367,7 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
         if (!dockerInstallNotificationShownToUser && this._currentDockerStatus === 'NotInstalled') {
             dockerInstallNotificationShownToUser = true;
             // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
bwateratmsft, line 369: I'm not sure that this actually is active; you should be able to remove this top eslint disable.

src/tree/contexts/ContextProperties.ts
bwateratmsft: Not really important, but `.555` isn't allowed in an IPv4 address either, the max is 255.

src/tree/contexts/ContextTreeItem.ts
```diff
@@ -0,0 +1,61 @@
...
+    public get createdTime(): number {
+        return this._item.createdTime;
+    }
+
+    public get label(): string {
+        return ext.contextsRoot.getTreeItemLabel(this._item);
+    }
+
+    // this is the description shown in tree item which can include one or more properites combined.
+    // This is not the description of the context.
```
bwateratmsft, line 31: Can the user configure it to be the actual context description? If so I'd say not _necessarily_ the description.

ravipal: Yes user can still configure to see the contest description. It is in ContexsTreeItem.getPropertyValue

src/tree/registerTrees.ts
```diff
@@ -77,6 +78,15 @@ export function registerTrees(): void {
         await ext.volumesTree.refresh(node);
     });
 
+    ext.contextsRoot = new ContextsTreeItem(undefined);
+    const contextsLoadMore = 'vscode-docker.contexts.loadMore';
+    ext.contextsTree = new AzExtTreeDataProvider(ext.contextsRoot, contextsLoadMore);
+    ext.contextsTreeView = vscode.window.createTreeView('vscode-docker.views.dockerContexts', { treeDataProvider: ext.contextsTree, canSelectMany: true });
+    ext.context.subscriptions.push(ext.contextsTreeView);
+    ext.contextsRoot.registerRefreshEvents(ext.contextsTreeView);
```
bwateratmsft, line 86: As an alternative to automatic refresh, I'd suggest letting Docker Context Manager trigger the refresh here. Additionally, since the refresh rate is default 2 seconds (with a setting), it would be more appropriate to follow the context manager's refresh rate of 20 seconds.

bwateratmsft: Actually, I'm not sure Docker Context Manager would notice if a context is added but not activated, so that's not the best approach. Nevertheless, 2 seconds is definitely way too fast a pace for this, so we should do refresh slower. For example, on my machine, it takes about 5 seconds to do `docker context ls`. Globally the average seems to be about 1 second.

src/commands/registerCommands.ts
ravipal: When I remote ssh into an Linux machine and used the list and inspect (before this change) and it uses the remote machine's context. So I guess this should be fine.

bwateratmsft: That seems fine to me. @karolz-ms, what do you think?

src/tree/registerTrees.ts
ravipal: Good point. I don't think we need 2 second refresh. Now using 20 seconds.

Issue #1913, opened 2020-04-26
Docker installer fails to download


yunanmr:
Does this occur consistently? 
Repro steps:

1.
2.

Action: vscode-docker.installDocker
Error type: ECONNRESET
Error Message: Client network socket disconnected before secure TLS connection was established

Version: 1.1.0
OS: win32
OS Release: 10.0.18363
Product: Visual Studio Code
Product Version: 1.44.2
Language: en

<details>
<summary>Call Stack</summary>

```
connResetException errors.js:559:14
TLSSocket.onConnectEnd _tls_wrap.js:1357:19
Object.onceWrapper events.js:291:20
TLSSocket.emit events.js:208:15
endReadableNT _stream_readable.js:1168:12
processTicksAndRejections task_queues.js:77:11
```

</details>

bwateratmsft:
@ravipal Can you take a look at this one? I suspect that something prevented the installer download.

PR #1923, opened 2020-04-28
Provider better error message for Docker Desktop installer download failure

ravipal:
Docker Desktop installer sometimes failed due to connection issue or proxy settings. Now we provider a better error message and also offer to manually download and install like below
image

Fixes: #1913 , #1926

### DIFF
--- a/src/commands/dockerInstaller.ts
+++ b/src/commands/dockerInstaller.ts
@@ -36,12 +36,7 @@ export abstract class DockerInstallerBase {
             } catch (error) {
                 const message = localize('vscode-docker.commands.DockerInstallerBase.downloadFailed', 'Downloading the Docker Desktop installer failed. Do you want to manually download and install?');
                 const title = localize('vscode-docker.commands.DockerInstallerBase.download', 'Download');
-                const response = await vscode.window.showErrorMessage(message, { title: title });
-                if (response) {
-                    await openExternal(this.downloadUrl);
-                }
-                context.errorHandling.suppressReportIssue = true;
-                context.errorHandling.suppressDisplay = true;
+                this.handleError(context, message, title, this.downloadUrl);
                 throw error;
             }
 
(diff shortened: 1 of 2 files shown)


src/commands/dockerInstaller.ts
```diff
@@ -12,23 +13,39 @@ import { localize } from '../localize';
...
-                async () => this.downloadInstaller());
+            context.telemetry.properties.stage = 'download';
+            try {
+                downloadedFileName = await vscode.window.withProgress(
+                    { location: vscode.ProgressLocation.Notification, title: downloadingMessage },
+                    async () => this.downloadInstaller());
+            } catch (error) {
+                const message = localize('vscode-docker.commands.DockerInstallerBase.downloadFailed', 'Downloading the Docker Desktop installer failed. Do you want to manually download and install?');
+                const title = localize('vscode-docker.commands.DockerInstallerBase.download', 'Download');
+                const response = await vscode.window.showErrorMessage(message, { title: title });
```
bwateratmsft, line 39: Instead of awaiting this (which would delay the telemetry being sent), we can do `vscode.window.showErrorMessage(...).then(response => { if (response === download) { openExternal(...) } } );`

bwateratmsft: There may need to be another argument in addition to `response =>` for handling cancellation, not sure.

bwateratmsft:
Just got #1926 opened, we should capture errors for that as well, and clean up the message plus give a link to the installation docs.

src/commands/dockerInstaller.ts
```diff
@@ -12,27 +13,47 @@ import { localize } from '../localize';
...
+                throw error;
+            }
 
+            context.telemetry.properties.stage = 'install';
             const command = this.getInstallCommand(downloadedFileName);
             // eslint-disable-next-line @typescript-eslint/no-floating-promises
             vscode.window.showInformationMessage(installationMessage);
-            await this.install(downloadedFileName, command);
+            try {
+                await this.install(downloadedFileName, command);
```
bwateratmsft, line 53: We should wrap with an error message that provides context. Otherwise they might get just "Process exited with code X" on Windows, which isn't very helpful out-of-context.

bwateratmsft: We can also do this to avoid duplicating `showErrorMessage`:

```typescript
let message: string;
let button: string;

try {
    message = localize(...);
    title = localize(...);
    // ... download

    message = localize(...);
    button = localize(...);
    // ... install
} catch {
    vscode.window.showErrorMessage(...).then(...);
}
```

ravipal: Added something similar

U31664:
Sorry, I dont understand what I need to do... Modify something? @ravipal

ravipal:
This fix will now provide a link, so user can navigate to a page with instruction on manually installing the Docker Desktop.
In your case, it appears to be something is preventing the installer to run. Did you see the installer getting launched and asking for use permission to proceed with installation? if not, can you please manually download and install and see if that works.

U31664:
Perfect, Thanks @ravipal  , I will try to install manually.

PR #1931, opened 2020-04-29
Scaffold compose-debug for Python apps

haniamr:
Instead of volume mapping the debugger, we have it installed as part of the entry-point itself. The debugger is about 5.5 MB so it doesn't slow-up container start.

### DIFF
--- a/src/configureWorkspace/configurePython.ts
+++ b/src/configureWorkspace/configurePython.ts
@@ -135,7 +135,7 @@ function genDockerComposeDebug(serviceName: string, ports: number[], target: Pyt
     return dockerComposeDebugfile
         .replace(/\$service_name\$/g, getValidImageName(serviceName))
         .replace(/\$ports\$/g, getComposePorts(ports, 5678))
-        .replace(/\$cmd\$/g, cmd)
+        .replace(/\$cmd\$/g, cmd.trimRight())
         .replace(/\$env\$/g, env);
 }
 
--- a/src/configureWorkspace/configurePython.ts
+++ b/src/configureWorkspace/configurePython.ts
@@ -12,7 +12,7 @@ import { dockerDebugScaffoldingProvider, PythonScaffoldingOptions } from '../deb
 import { ext } from "../extensionVariables";
 import { localize } from '../localize';
 import { getValidImageName } from '../utils/getValidImageName';
-import { getPythonProjectType, PythonDefaultPorts, PythonFileExtension, PythonFileTarget, PythonModuleTarget, PythonProjectType, PythonTarget } from "../utils/pythonUtils";
+import { getPythonProjectType, PythonDefaultPorts, PythonFileExtension, PythonFileTarget, PythonModuleTarget, PythonProjectType, PythonTarget, inferPythonArgs } from "../utils/pythonUtils";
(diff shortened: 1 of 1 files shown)


src/configureWorkspace/configurePython.ts
```diff
@@ -111,6 +125,20 @@ function genDockerCompose(serviceName: string, ports: number[]): string {
         .replace(/\$ports\$/g, getComposePorts(ports));
 }
 
+function genDockerComposeDebug(serviceName: string, ports: number[], target: PythonTarget, projectType: PythonProjectType): string {
+    const args: string = (inferPythonArgs(projectType, ports) || []).join(' ');
+    const app: string = 'module' in target ? `-m ${target.module}` : target.file;
```
ravipal, line 130: should this also check if `target.module` has some value?

haniamr: Not really, there will always be a value in either "module" or "file" (we do require non-empty input), the question is which one of them to look at.

haniamr:
> What is the attach story for containers started using docker-compose.yml?

There's a section in the docs about that: https://code.visualstudio.com/docs/containers/debug-python#_how-to-debug-your-app-with-gunicorn

karolz-ms:
> > What is the attach story for containers started using docker-compose.yml?
> 
> There's a section in the docs about that: https://code.visualstudio.com/docs/containers/debug-python#_how-to-debug-your-app-with-gunicorn

And https://code.visualstudio.com/docs/containers/docker-compose#_python

src/configureWorkspace/configurePython.ts
```diff
@@ -60,6 +60,20 @@ services:
...
+const dockerComposeDebugfile = `version: '3.4'
+
+services:
+  $service_name$:
+    image: $service_name$
+    build:
+      context: .
+      dockerfile: Dockerfile
+    entrypoint: /bin/bash
+    command: -c "pip install debugpy -t /tmp && python /tmp/debugpy --wait-for-client --listen 0.0.0.0:5678 $cmd$"
```
bwateratmsft, line 72: How big is `debugpy`? There's an implicit disadvantage that other dependencies are installed via the Dockerfile, and thus benefit from image caching, but this one would not.

bwateratmsft, line 72: Is `--wait-for-client` desirable in this scenario? In contrast, starting a .NET Core app via `docker-compose.debug.yml` will allow the application itself to start and go.

haniamr: Hmm, that's a valid point. I don't really have a strong opinion, I would have thought that the user might want to debug startup logic as well. Did we ever have specific feedback on this?

bwateratmsft: Nothing specific, but our answer for debugging startup logic is "F5", not debug-compose.yml.

haniamr: The problem with F5 is "what if I need more containers to start as well"? Almost every Python example that I saw has at least 1 more container that's usually a database.

bwateratmsft: Same problem exists for .NET :wink:

Issue #1854, opened 2020-04-13
Start Survey to get user feedback and satisfaction


dbreshears:
This survey should not start until June 1st.

bwateratmsft:
Marking for 1.3.0, since it is now checked into 1.2.1.

Issue #1925, opened 2020-04-29
An error occurs when executing Docker related actions


v-ruizh:
**OS:** Windows
**Build Version:** [20200428.22](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=18467&view=results)

**Simple Repro Steps:**
From Docker images view, right click one image and execute "Run" action.
![image](https://user-images.githubusercontent.com/45614674/80583916-7e063780-8a43-11ea-93c9-657506f56464.png)

**Repro Steps:**
1. Create a folder.
2. Open this folder in VS Code Insiders.
3. F1->Docker: Add Docker Files to workspace...->Node.js->Yes to create compose files->Accept default port 3000.

**Expect:**
Add Docker files successfully without any error.

**Actual:**
Add Docker files with an error.
![image](https://user-images.githubusercontent.com/45614674/80581582-efdc8200-8a3f-11ea-9e34-a329c9ab8bcc.png)
Error Text:
_Command 'Docker: Add Docker Files to Workspace...' resulted in an error (Cannot read property 'length' of undefined)_

bwateratmsft:
I can reproduce but I can't get it to give me the error stack...

bwateratmsft:
Got it. This is a regression from `vscode-extension-telemetry`: https://github.com/microsoft/vscode-extension-telemetry/issues/44

ravipal:
I will revert the telemetry change until vscode-extension-telemetry is fixed.

bwateratmsft:
I'll mark this as fixed via #1930. In #1934 we will move back to a higher version.

PR #1934, opened 2020-04-30
Add Experimentation Service client

bwateratmsft:
- Adds Experimentation Service client (and adapter)
- Adds unified activity measurement service
- Removes NPS and Active Use surveys (NPS will return - #1854)
- Guards more things based on user telemetry opt-in, in order to maximize respect of user privacy choice


src/commands/registerCommands.ts
```diff
@@ -71,86 +72,101 @@ import { inspectVolume } from "./volumes/inspectVolume";
...
+
 export function registerCommands(): void {
     registerWorkspaceCommand('vscode-docker.api.configure', configureApi);
     registerWorkspaceCommand('vscode-docker.compose.down', composeDown);
     registerWorkspaceCommand('vscode-docker.compose.restart', composeRestart);
     registerWorkspaceCommand('vscode-docker.compose.up', composeUp);
     registerWorkspaceCommand('vscode-docker.configure', configure);
     registerWorkspaceCommand('vscode-docker.configureCompose', configureCompose);
-    registerCommand('vscode-docker.pruneSystem', pruneSystem);
+    registerCommandWithActivity('vscode-docker.pruneSystem', pruneSystem);
```
bwateratmsft, line 97: I'm still considering using `TelemetryPublisher` along with some sort of categorization, instead of this. `registerCommand` would have to set that categorization to `'command'`, up in `vscode-azureextensionui` (but we'd be blocked on that currently due to #1925).

This way, `TelemetryPublisher` would be the entry point for `recordActivity`. Each of our additional categories, if they ever turn up, would set that categorization.

What do you guys think?

src/telemetry/ActivityMeasurementService.ts
```diff
@@ -0,0 +1,110 @@
...
+    getActivity(type: ActivityType): ActivityMeasurement;
+}
+
+export interface ActivityMeasurement {
+    lastSession: number | undefined;
+    monthlySessions: number;
+    totalSessions: number;
+}
+
+export class ActivityMeasurementService implements IActivityMeasurementService {
```
bwateratmsft, line 28: Need unit tests.

src/commands/registerCommands.ts
bwateratmsft: It would also be blocked on `vscode-azureappservice` adopting version 0.30.0 of `vscode-azureextensionui`.

karolz-ms: @bwateratmsft not sure if I understand the alternative... what would `TelemetryPublisher` report?

bwateratmsft: `TelemetryPublisher` could be the single entrypoint where `recordActivity` would be called, but we would need to know what kind of activity it is--so need something set on the telemetry properties.

src/extensionVariables.ts
```diff
@@ -24,7 +26,12 @@ export namespace ext {
     export let context: ExtensionContext;
     export let outputChannel: IAzExtOutputChannel;
     export let ui: IAzureUserInput;
+
+    export let telemetryOptIn: boolean;
     export let reporter: ITelemetryReporter;
+    export let experimentationService: IExperimentationServiceAdapter | undefined;
+    export let ams: IActivityMeasurementService | undefined;
```
ravipal, line 33: ams is not a well known abbreviation.  Consider the full name.

src/registerListeners.ts
```diff
@@ -20,5 +22,8 @@ function onDidSaveTextDocument(doc: TextDocument): void {
     }
 
     lastUploadTime = Date.now();
-    ext.reporter.sendTelemetryEvent('dockerfilesave', { "lineCount": doc.lineCount.toString() }, {});
+    ext.reporter.sendTelemetryErrorEvent('dockerfilesave', { "lineCount": doc.lineCount.toString() }, {});
```
ravipal, line 25: why line count is considered errorEvent? There is nothing to mask like filepath or user name.

src/tasks/DockerTaskProvider.ts
```diff
@@ -40,6 +41,9 @@ export abstract class DockerTaskProvider implements TaskProvider {
                 actionContext.errorHandling.suppressDisplay = true; // Suppress display. VSCode already has a modal popup and we don't want focus taken away from Terminal window.
                 actionContext.errorHandling.rethrow = true; // Rethrow to hit the try/catch outside this block.
 
+                // eslint-disable-next-line no-unused-expressions, @typescript-eslint/no-floating-promises
+                ext.ams?.recordActivity('overallnoedit');
```
ravipal, line 45: Can the ams internally handle the user opt-in. Otherwise the caller needs to remember it may not be initialized and user ?.

src/registerListeners.ts
bwateratmsft: It's not an error, but as part of this ongoing mess with telemetry and `vscode-extension-telemetry`, `ITelemetryReporter` removed `sendTelemetryEvent`, keeping only `sendTelemetryErrorEvent`. I think that will only be temporary, looks like that interface is going to be removed entirely.

bwateratmsft: I am not sure what we will do for `TelemetryReporterProxy`...

src/commands/registerCommands.ts
bwateratmsft: If `ITelemetryReporter` does end up being removed, `TelemetryPublisher` will no longer be possible.

src/registerListeners.ts
bwateratmsft: Here's the thread for more info on `ITelemetryReporter` maybe leaving: https://github.com/microsoft/vscode-azuretools/issues/693

src/tasks/DockerTaskProvider.ts
bwateratmsft: Changed the implementation to handle opt-in internally for both AMS and experimentation service.

src/telemetry/ActivityMeasurementService.ts
```diff
@@ -0,0 +1,119 @@
...
+                    const now = Date.now();
+
+                    // No need to increment if it's been done already today
+                    if (sameDate(currentValue.lastSession, now, 'day')) {
+                        return;
+                    }
+
+                    const newValue: ActivityMeasurement = {
+                        lastSession: now,
+                        monthlySessions: currentValue.monthlySessions + 1,
```
haniamr, line 59: The name "monthly" implied to me that we'll record 1 instance of ActivityMeasurement per month, which isn't really the case. So I suggest renaming it to something like "currentMonthSessions" or something similar to avoid confusion.

src/telemetry/ExperimentationServiceAdapter.ts
```diff
@@ -0,0 +1,70 @@
...
+        if (!this.flightMap.has(flight)) {
+            this.flightMap[flight] = new AsyncLazy<boolean>(async () => {
+                return await this.wrappedExperimentationService.isCachedFlightEnabled(flight);
+            }, lazyLifetime);
+        }
+
+        return await this.flightMap.get(flight).getValue();
+    }
+
+    private getExtensionVersion(): string {
```
haniamr, line 65: Looks like it might be a useful utility method somewhere.

src/telemetry/TelemetryReporterProxy.ts
```diff
@@ -3,22 +3,41 @@
...
 
-export class TelemetryReporterProxy implements ITelemetryReporter {
     public constructor(
-        private readonly publisher: ITelemetryPublisher,
         private readonly wrappedReporter: ITelemetryReporter) {
     }
 
-    public sendTelemetryEvent(eventName: string, properties?: { [key: string]: string; }, measurements?: { [key: string]: number; }): void {
-        this.wrappedReporter.sendTelemetryEvent(eventName, properties, measurements);
+    public sendTelemetryErrorEvent(eventName: string, properties?: { [key: string]: string; }, measurements?: { [key: string]: number; }, errorProps?: string[]): void {
```
haniamr, line 16: I'm not sure I get this, why is it renamed as an error event? And what do `sharedProperties` represent?

bwateratmsft: It's not an error, but `sendTelemetryEvent` was removed. I'm trying to work around [this issue](https://github.com/microsoft/vscode-azuretools/issues/693). It's probably going to change more.

src/telemetry/TelemetryReporterProxy.ts
```diff
@@ -3,22 +3,39 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ITelemetryReporter } from "vscode-azureextensionui";
-import { ITelemetryPublisher } from "./TelemetryPublisher";
+import { ITelemetryReporter } from 'vscode-azureextensionui';
+import { IExperimentationTelemetry } from 'vscode-tas-client';
+
+export class TelemetryReporterProxy implements ITelemetryReporter, IExperimentationTelemetry {
```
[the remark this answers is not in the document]
bwateratmsft: This class will probably get removed.

src/telemetry/ExperimentationServiceAdapter.ts
bwateratmsft: Implemented it as a `Lazy<string | undefined>` on `constants`.

src/telemetry/TelemetryReporterProxy.ts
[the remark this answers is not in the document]
bwateratmsft: If we end up internalizing `ITelemetryReporter` it will not be possible to validate anything, so the UTs will have to be removed.

src/telemetry/ExperimentationServiceAdapter.ts
```diff
@@ -0,0 +1,53 @@
...
+    public constructor(globalState: vscode.Memento, reporter: tas.IExperimentationTelemetry) {
+        if (!ext.telemetryOptIn) {
+            return;
+        }
+
+        try {
+            const version = extensionVersion.value ?? '1';
+            let targetPopulation: tas.TargetPopulation;
+
+            if (ext.runningTests || process.env.DEBUGTELEMETRY) {
```
karolz-ms, line 28: Again, are we sure about including telemetry from test runs into experiment stats? 🤔

src/telemetry/ExperimentationTelemetry.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * "Handles" telemetry events by adding the shared properties to them
+     * @param context The action context
+     */
+    public async handleTelemetry(context: IActionContext): Promise<void> {
+        // Copy shared properties into event properties
+        for (const key of Object.keys(this.sharedProperties)) {
+            if (context.telemetry.properties[key]) {
+                console.error('Local telemetry property will be overwritten by shared property.');
```
karolz-ms, line 20: Maybe use `console.warn` and also add info which property is being overwritten

src/telemetry/ExperimentationTelemetry.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    /**
+     * Implements `postEvent` for `IExperimentationTelemetry`. This implementation is admittedly strange, but necessary since `ITelemetryReporter` is no longer public.
+     * @param eventName The name of the event
+     * @param props The properties to attach to the event
+     */
+    public postEvent(eventName: string, props: Map<string, string>): void {
+        // This implementation is admittedly strange, but necessary since ITelemetryReporter is no longer public
```
karolz-ms, line 33: 6 months from now no one will remember what `ITelemetryReporter` was. I would just drop the "this is strange" part.

src/tasks/DockerTaskProvider.ts
```diff
@@ -51,7 +55,7 @@ export abstract class DockerTaskProvider implements TaskProvider {
                 context.actionContext.telemetry.properties.orchestration = 'single' as DockerOrchestration; // TODO: docker-compose, when support is added
                 await this.executeTaskInternal(context, task);
 
-                dockerTaskEndEventListener.fire({name: task.name, success: true});
+                dockerTaskEndEventListener.fire({ name: task.name, success: true });
```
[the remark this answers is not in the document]
bwateratmsft: ESLint does this automatically. This was submitted by somebody not using it. I'd rather let the change through.

src/tasks/DockerTaskProvider.ts
```diff
@@ -61,7 +65,7 @@ export abstract class DockerTaskProvider implements TaskProvider {
                 context.terminal.writeErrorLine(error.message);
             }
 
-            dockerTaskEndEventListener.fire({name: task.name, success: false, error: error.message});
+            dockerTaskEndEventListener.fire({ name: task.name, success: false, error: error.message });
```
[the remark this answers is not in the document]
bwateratmsft: (see above)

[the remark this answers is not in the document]
karolz-ms: OK

src/extension.ts
```diff
@@ -49,11 +49,13 @@ const DOCUMENT_SELECTOR: DocumentSelector = [
 ];
 
 function initializeExtensionVariables(ctx: vscode.ExtensionContext): void {
+    ext.context = ctx;
+    ext.telemetryOptIn = vscode.workspace.getConfiguration('telemetry').get('enableTelemetry', false) || ext.runningTests;
```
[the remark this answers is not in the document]
bwateratmsft: I'm going to set `DEBUGTELEMETRY = 1` for our tests instead; otherwise we have to refactor everything around `ext.telemetryOptIn` in order to still be able to test `ActivityMeasurementService`.

src/telemetry/ExperimentationServiceAdapter.ts
bwateratmsft: (see above, will set `DEBUGTELEMETRY = 1` for tests)

src/commands/registerCommands.ts
```diff
@@ -71,6 +72,21 @@ import { inspectVolume } from "./volumes/inspectVolume";
...
 import { removeVolume } from "./volumes/removeVolume";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function registerCommand(commandId: string, callback: (context: IActionContext, ...args: any[]) => any, debounce?: number): void {
+    registerCommandAzUI(
+        commandId,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        async (context, ...args: any[]) => {
+            // eslint-disable-next-line @typescript-eslint/no-floating-promises
+            ext.activityMeasurementService.recordActivity('overallnoedit');
```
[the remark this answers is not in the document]
bwateratmsft: I'd rather leave it as a parameter; if we add other types of activity this is how it will be done, no need to start with an interface we know we'd have to change.

PR #1948, opened 2020-05-06
.Net Core Attach support for windows container

ravipal:
Earlier we were not handling the windows container during attach. Now added support. But `docker cp` command is not supported on hyper-v containers, which is the default in windows 10. So still this doesn't address most of the windows container. At least now this will provide a better error message.
Fixes: #1662

### DIFF
--- a/src/debugging/netcore/NetCoreDebugHelper.ts
+++ b/src/debugging/netcore/NetCoreDebugHelper.ts
@@ -14,7 +14,7 @@ import { NetCoreTaskHelper, NetCoreTaskOptions } from '../../tasks/netcore/NetCo
 import { ContainerTreeItem } from '../../tree/containers/ContainerTreeItem';
 import { callDockerodeWithErrorHandling } from '../../utils/callDockerode';
 import { LocalOSProvider } from '../../utils/LocalOSProvider';
-import { ContainerOSType, getContainerOSType } from '../../utils/osUtils';
+import { DockerOSType, getDockerOSType } from '../../utils/osUtils';
 import { pathNormalize } from '../../utils/pathNormalize';
 import { PlatformOS } from '../../utils/platform';
 import { unresolveWorkspaceFolder } from '../../utils/resolveVariables';
@@ -188,7 +188,6 @@ export class NetCoreDebugHelper implements DebugHelper {
     }
 
     public async resolveAttachDebugConfiguration(context: DockerDebugContext, debugConfiguration: DockerAttachConfiguration): Promise<ResolvedDebugConfiguration | undefined> {
-        // TODO: Validate the target container OS and fail debugging
         // Get Container Name if missing
         const containerName: string = debugConfiguration.containerName ?? await this.getContainerNameToAttach();
 
@@ -196,7 +195,7 @@ export class NetCoreDebugHelper implements DebugHelper {
 
         // If debugger path is not specified, then install the debugger if it doesn't exist in the container
         if (!debuggerPath) {
-            const containerOS = await getContainerOSType(containerName, context.actionContext);
+            const containerOS = await getDockerOSType(context.actionContext);
             const osProvider = new LocalOSProvider();
             const debuggerDirectory = containerOS === 'windows' ? 'C:\\remote_debugger' : '/remote_debugger';
             debuggerPath = containerOS === 'windows'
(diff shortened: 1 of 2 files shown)


src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -287,20 +292,32 @@ export class NetCoreDebugHelper implements DebugHelper {
...
     }
 
-    private async copyDebuggerToContainer(containerName: string, containerDebuggerDirectory: string): Promise<string> {
+    private async copyDebuggerToContainer(context: DockerDebugContext, containerName: string, containerDebuggerDirectory: string, containerOS: ContainerOSType): Promise<void> {
+        const dockerClient = new CliDockerClient(new ChildProcessProvider());
+        if (containerOS === 'windows') {
+            const isolation = await dockerClient.inspectObject(containerName, { format: '{{ .HostConfig.Isolation}}' });
+            if (isolation && isolation === 'hyperv') {
+                context.actionContext.errorHandling.suppressReportIssue = true;
+                throw new Error(localize('vscode-docker.debug.netcore.isolationNotSupported', 'Unable to copy debugger. This operation is not supported on Hyper-V container.'));
```
karolz-ms, line 301: If this message is supposed to be seen by the user, maybe it would be better to refer to what they were trying to do. Something like

    Attaching a debugger to a Hyper-V container is not supported.

src/utils/osUtils.ts
```diff
@@ -88,3 +90,9 @@ export async function getDockerOSType(context: IActionContext): Promise<DockerOS
         return info.OSType;
     }
 }
+
+export async function getConatinerOSType(containerId: string): Promise<ContainerOSType> {
```
bwateratmsft, line 94:
```suggestion
export async function getContainerOSType(containerId: string): Promise<ContainerOSType> {
```

bwateratmsft, line 94: `getDockerOSType` is already doing this, we don't need this method.

bwateratmsft, line 94:
```suggestion
export async function getContainerOSType(containerId: string): Promise<ContainerOSType> {
```

src/utils/osUtils.ts
```diff
@@ -88,3 +90,9 @@ export async function getDockerOSType(context: IActionContext): Promise<DockerOS
...
+    const inspectCmd: string = `docker inspect --format="{{ .Platform}}" ${containerId}`;
+    const execResult = await execAsync(inspectCmd);
```
bwateratmsft, line 96: Should be able to use Dockerode instead which will be much faster.

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -194,11 +195,15 @@ export class NetCoreDebugHelper implements DebugHelper {
 
         // If debugger path is not specified, then install the debugger if it doesn't exist in the container
         if (!debuggerPath) {
-            const debuggerDirectory = '/remote_debugger';
-            debuggerPath = `${debuggerDirectory}/vsdbg`;
-            const isDebuggerInstalled: boolean = await this.isDebuggerInstalled(containerName, debuggerPath);
+            const conatinerOS = await getConatinerOSType(containerName);
```
bwateratmsft, line 198:
```suggestion
            const containerOS = await getConatinerOSType(containerName);

```

src/utils/osUtils.ts
ravipal: Dockerode didn't provide this property and moreover it require to construct the the container object, then query it.

bwateratmsft: "Constructing" the container object does not involve any HTTP requests and only requires the ID, so that's not an issue. There's probably another property we can use to figure it out.

bwateratmsft: The property is there, `.Platform` as expected.

src/utils/osUtils.ts
```diff
@@ -88,3 +90,9 @@ export async function getDockerOSType(context: IActionContext): Promise<DockerOS
         return info.OSType;
     }
 }
+
+export async function getContainerOSType(containerId: string): Promise<ContainerOSType> {
```
bwateratmsft, line 94: On non-Windows, it's always Linux containers so we can short-cirtcuit there.

src/utils/osUtils.ts
ravipal: Are you suggesting to fix the dockerode package?

bwateratmsft: Just `@types/dockerode`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/dockerode

ravipal: Made a PR to update the types/dockerode. Once it is merged and a new version is available, I will consume here.

src/utils/osUtils.ts
```diff
@@ -88,3 +89,15 @@ export async function getDockerOSType(context: IActionContext): Promise<DockerOS
         return info.OSType;
     }
 }
+
+export async function getContainerOSType(containerId: string, context: IActionContext): Promise<ContainerOSType> {
```
bwateratmsft, line 93: @ravipal I just realized; I think `getDockerOSType` is sufficient for this. The Docker server OS must match the container OS.

ravipal: Now using DockerOS instead of container OS and it will break only LOCW.

PR #1949, opened 2020-05-07
Remove "Learn how to contribute..." from pick list

bwateratmsft:
Removing this entry from the registry provider pick list, since we're working on developing a new extensibility model.


src/tree/registries/RegistriesTreeItem.ts
```diff
@@ -87,10 +87,10 @@ export class RegistriesTreeItem extends AzExtParentTreeItem {
             };
         })
         picks = picks.sort((p1, p2) => p1.label.localeCompare(p2.label));
-        picks.push({
+        /* picks.push({
```
karolz-ms, line 90: Just delete. It will be in the file history in case someone needs it

Issue #1954, opened 2020-05-07
compose up command customization with no match produces wrong command


ravipal:
By default the compose up command runs the command
`docker-compose -f "docker-compose.yml" up -d --build`

But when a command customization is provided with no match, it suppose to use the default command same as above. But it runs the below command (note the location of up) which is not valid.
`docker-compose up -f "docker-compose.yml" -d --build`

bwateratmsft:
We talked in water cooler, if we take another fix (e.g. for #1958) we'll include this one as well (so as to avoid needing to branch), otherwise it will wait until 1.3.0.

bwateratmsft:
The fix for this is now released in 1.3.0.

PR #1956, opened 2020-05-08
fix default compose up command

ravipal:
Fixes: #1954

### DIFF
--- a/src/commands/selectCommandTemplate.ts
+++ b/src/commands/selectCommandTemplate.ts
@@ -26,8 +26,8 @@ const defaults: { [key in TemplateCommand]: CommandTemplate } = {
     'runInteractive': { label: 'Docker Run (Interactive)', template: 'docker run --rm -it ${exposedPorts} ${tag}' },
     'attach': { label: 'Docker Attach', template: 'docker exec -it ${containerId} ${shellCommand}' },
     'logs': { label: 'Docker Logs', template: 'docker logs -f ${containerId}' },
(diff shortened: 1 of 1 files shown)


bwateratmsft:
I'd be OK with taking this for 1.2.0.

src/commands/selectCommandTemplate.ts
```diff
@@ -26,8 +26,8 @@ const defaults: { [key in TemplateCommand]: CommandTemplate } = {
     'runInteractive': { label: 'Docker Run (Interactive)', template: 'docker run --rm -it ${exposedPorts} ${tag}' },
     'attach': { label: 'Docker Attach', template: 'docker exec -it ${containerId} ${shellCommand}' },
     'logs': { label: 'Docker Logs', template: 'docker logs -f ${containerId}' },
-    'composeUp': { label: 'Compose Up', template: 'docker-compose up ${configurationFile} ${detached} ${build}' },
-    'composeDown': { label: 'Compose Down', template: 'docker-compose down ${configurationFile}' },
+    'composeUp': { label: 'Compose Up', template: 'docker-compose ${configurationFile} up ${detached} ${build}' },
```
bwateratmsft, line 29: These are already corrected in `package.json` right?

bwateratmsft: Yep, they are.

ravipal:
> I'd be OK with taking this for 1.2.0. Anyone else agree/disagree?

This is really a corner case. I am fine either way.

PR #1969, opened 2020-05-12
Show progress during context switch

### LINKED ISSUE #1965: When doing Context > Use in explorer view, takes 20s for "Connected" icon to move
1. Create a Docker context
1. In Explorer view, right click it, "Use"

Expected:
"Connected" icon should move quickly

Actual:
"Connected" icon takes 20s to move over, long after the containers/images/etc. views are already refreshed with the new context

image

ravipal:
1. Show the loading icon next to the context being switched
image
2. New auto refresh will be blocked while context switch is in progress.

Fixes #1965

### DIFF
--- a/src/extensionVariables.ts
+++ b/src/extensionVariables.ts
@@ -62,7 +62,6 @@ export namespace ext {
     export let contextsTree: AzExtTreeDataProvider;
     export let contextsTreeView: TreeView<AzExtTreeItem>;
     export let contextsRoot: ContextsTreeItem;
-    export let isContextSwitchInProgress: boolean = false;
     /**
      * A test-injectable structure defining the current operating system and version
      */
--- a/src/extensionVariables.ts
+++ b/src/extensionVariables.ts
@@ -62,6 +62,7 @@ export namespace ext {
     export let contextsTree: AzExtTreeDataProvider;
     export let contextsTreeView: TreeView<AzExtTreeItem>;
     export let contextsRoot: ContextsTreeItem;
+    export let isContextSwitchInProgress: boolean = false;
     /**
      * A test-injectable structure defining the current operating system and version
      */
--- a/src/tree/LocalRootTreeItemBase.ts
+++ b/src/tree/LocalRootTreeItemBase.ts
@@ -100,9 +100,6 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
                             // So check again before starting the refresh.
                             if (this.autoRefreshEnabled) {
                                 await this.refresh();
-                            } else {
-                                // clear the cache if the this.autoRefreshEnabled is disabled, so the refresh will not use old cached data.
-                                this._itemsFromPolling = undefined;
                             }
                         }
                     },
@@ -141,7 +138,7 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
             ext.activityMeasurementService.recordActivity('overallnoedit');
 
             this._currentItems = this._itemsFromPolling || await this.getSortedItems();
-            this._itemsFromPolling = undefined;
+            this.clearPollingCache();
             this.failedToConnect = false;
             this._currentDockerStatus = 'Running';
         } catch (error) {
@@ -362,7 +359,7 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
             this._itemsFromPolling = await this.getSortedItems();
             pollingDockerStatus = 'Running';
         } catch (error) {
-            this._itemsFromPolling = undefined;
+            this.clearPollingCache();
             pollingDockerStatus = await dockerInstallStatusProvider.isDockerInstalled() ? 'Installed' : 'NotInstalled';
             isDockerStatusChanged = pollingDockerStatus !== this._currentDockerStatus;
         }
--- a/src/tree/LocalRootTreeItemBase.ts
+++ b/src/tree/LocalRootTreeItemBase.ts
@@ -89,7 +89,7 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
                 const refreshInterval: number = this.getRefreshInterval();
                 intervalId = setInterval(
                     async () => {
-                        if (window.state.focused && await this.hasChanged()) {
+                        if (window.state.focused && ext.isContextSwitchInProgress !== true && await this.hasChanged()) {
                             await this.refresh();
                         }
                     },
@@ -116,6 +116,10 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
             iconPath: getThemedIconPath('info'),
             contextValue: 'dockerNoItems'
         })];
+    }
+
+    public clearPollingCache(): void {
+        this._itemsFromPolling = undefined;
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
@@ -342,7 +346,12 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
         let isDockerStatusChanged = false;
 
         try {
-            this._itemsFromPolling = await this.getSortedItems();
+            const pollingItems = await this.getSortedItems();
+            if (ext.isContextSwitchInProgress === true) {
+                this._itemsFromPolling = undefined;
+                return false;
+            }
+            this._itemsFromPolling = pollingItems;
             pollingDockerStatus = 'Running';
         } catch (error) {
             this._itemsFromPolling = undefined;
--- a/src/tree/LocalRootTreeItemBase.ts
+++ b/src/tree/LocalRootTreeItemBase.ts
@@ -54,6 +54,8 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty 
     public abstract getItems(): Promise<TItem[] | undefined>;
     public abstract getPropertyValue(item: TItem, property: TProperty): string;
 
+    public static autoRefreshViews: boolean = true;
+
     public groupBySetting: TProperty | CommonGroupBy;
     public sortBySetting: CommonSortBy;
     public labelSetting: TProperty;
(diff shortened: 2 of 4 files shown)


src/tree/contexts/ContextTreeItem.ts
```diff
@@ -60,6 +61,18 @@ export class ContextTreeItem extends AzExtTreeItem {
     }
 
     public async use(context: IActionContext): Promise<void> {
-        return dockerContextManager.use(this.name);
+        try {
+            ext.isContextSwitchInProgress = true;
+            return this.runWithTemporaryDescription(
+                localize('vscode-docker.tree.context.using', 'Using...'),
```
bwateratmsft, line 67: I think "changing" or "loading" is maybe a better word, even though the Docker CLI calls it "use".

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -342,7 +346,12 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
         let isDockerStatusChanged = false;
 
         try {
-            this._itemsFromPolling = await this.getSortedItems();
+            const pollingItems = await this.getSortedItems();
```
karolz-ms, line 349: If refresh is suspended (Docker context switch in progress), you do not need to make this call.

src/tree/contexts/ContextTreeItem.ts
karolz-ms: Oh, I see, this shows next to the context name in the tree. Disregard my previous suggestion, but I still think @bwateratmsft is right. "Using" suggests that the context is already being used, which is not the case here. "Changing" or "activating" conveys the state much better.

src/extensionVariables.ts
```diff
@@ -62,6 +62,7 @@ export namespace ext {
     export let contextsTree: AzExtTreeDataProvider;
     export let contextsTreeView: TreeView<AzExtTreeItem>;
     export let contextsRoot: ContextsTreeItem;
+    export let isContextSwitchInProgress: boolean = false;
```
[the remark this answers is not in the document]
bwateratmsft: Or alternatively, something like this (which makes it clear that refresh being enabled depends on several things, namely, window being focused and not in the middle of a context switch):
```typescript
if (this.refreshEnabled && await this.hasChanged()) {
    await this.refresh();
}
```
\+
```typescript
public get refreshEnabled(): boolean {
    return window.state.focused && !ext.isContextSwitchInProgress;
}

Comment (bwateratmsft) [inline src/extensionVariables.ts:65]: Is it possible to make this a public-modifiable static variable on `LocalRootTreeItemBase` instead of an extension variable? It only has relevance to `LocalRootTreeItemBase`.

File: src/tree/contexts/ContextTreeItem.ts [hunk 4]
```diff
@@ -60,6 +61,18 @@ export class ContextTreeItem extends AzExtTreeItem {
     }
 
     public async use(context: IActionContext): Promise<void> {
-        return dockerContextManager.use(this.name);
+        try {
+            ext.isContextSwitchInProgress = true;
+            return this.runWithTemporaryDescription(
```
bwateratmsft, line 66: Does this actually run the `finally` in the expected order since we are not `await`ing it here?

src/tree/contexts/ContextTreeItem.ts
```diff
@@ -60,6 +61,18 @@ export class ContextTreeItem extends AzExtTreeItem {
...
+                localize('vscode-docker.tree.context.using', 'Using...'),
+                async () => {
+                    await dockerContextManager.use(this.name);
+                    // The refresh logic will use the cached contexts retrieved by polling (auto refresh logic)
+                    // Since the context is changed, clear the cached contexts, otherwise the old context will be shown as active
+                    ext.contextsRoot.clearPollingCache();
```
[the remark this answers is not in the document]
bwateratmsft: Agreed, less computation cost.

src/tree/contexts/ContextTreeItem.ts
karolz-ms: Excellent catch @bwateratmsft ! I bet the `finally` clause needs to be inside the lambda passed to `runWithTemporaryDescription`

bwateratmsft: I'm not 100% certain that would have the expected behavior, but I am 100% that `await this.runWithTemporaryDescription(...)` and then `finally` would. If `runWithTemporaryDescription` threw an exception for whatever reason, we would not hit a `finally` inside the lambda.

src/tree/contexts/ContextTreeItem.ts
ravipal: Since the loading icon is shown next to it and also the "...," next to using, it conveys that the using is in progress. Lets see what @BigMorty thinks.

src/tree/contexts/ContextTreeItem.ts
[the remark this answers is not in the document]
ravipal: Without clearing the cache, the auto refresh will use the old cached value, so we must clear the cache. I am reluctant to modifying the view model which require access to the old current and it is better to always reflect the actual data coming from backend.

src/tree/LocalRootTreeItemBase.ts
ravipal: Before `hasChanged()` is called we do check if refresh is suspended. But this is to handle the case where the refresh is suspended while waiting on the result of `this.getSortedItems()`.

src/tree/contexts/ContextTreeItem.ts
ravipal: Good catch!

src/tree/LocalRootTreeItemBase.ts
karolz-ms: Should we then check both before and after the call?

src/tree/contexts/ContextTreeItem.ts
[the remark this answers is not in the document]
karolz-ms: It is fine to clear the cache, although `hasChanged()` will do so automatically anyway next time it is called. My point is that after `dockerContextManager.use()` we **know** that the view model is invalid and what the right view model should be, so no need to wait for the refresh process clean things up, we can make the change immediately. This results in better UI responsiveness.

Also, with this change the `ContextTreeItem` now "knows" about the refresh & data caching logic in the root tree item base class. This increases coupling and makes things more error prone.

src/tree/contexts/ContextTreeItem.ts
BigMorty: Where on the Context node does this show, after the name instead of the description?
Regardless, I think "switching" or "loading" makes sense, would probably go with "switching".

src/tree/contexts/ContextTreeItem.ts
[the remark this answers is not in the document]
ravipal: The cache must be cleared because of the way how the refresh logic works in `LocalRootTreeItemBase`. The specific scenario is
1. Auto refresh caches the context list as part of auto refresh. The next call to `hasChanged()` will happen after 20 seconds.
2. Now user switches context and we invoke refresh.
3. Without clearing the cache, the refresh will use the cached context and will not show the latest current context.

Manipulating the view model require manipulating both current and previous current node and also require access to `RefreshUI` which is not exposed now. Also require copying the loading icon to this extension. In short, we need to duplicate what runWithTemporaryDescription does in the base class and it gets complex and it is not worth.

To make the `ContextTreeItem` not know about the refresh logic, I moved the code to `userDockerContext`

src/tree/contexts/ContextTreeItem.ts
ravipal: Now using 'Switching...'

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -72,6 +74,10 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
         return workspace.getConfiguration(`${configPrefix}.${this.treePrefix}`);
     }
 
+    private get autoRefreshEnabled(): boolean {
+        return window.state.focused && LocalRootTreeItemBase.autoRefreshViews === true;
```
bwateratmsft, line 78: Nit, we rarely do `=== true`
```suggestion
        return window.state.focused && LocalRootTreeItemBase.autoRefreshViews;
```

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -89,8 +95,12 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
                 const refreshInterval: number = this.getRefreshInterval();
                 intervalId = setInterval(
                     async () => {
-                        if (window.state.focused && await this.hasChanged()) {
-                            await this.refresh();
+                        if (this.autoRefreshEnabled && await this.hasChanged()) {
+                            // Auto refresh could be disabled while invoking the hasChaged()
```
bwateratmsft, line 99:
```suggestion
                            // Auto refresh could be disabled while invoking the hasChanged()
```

src/tree/LocalRootTreeItemBase.ts
```diff
@@ -89,8 +95,12 @@ export abstract class LocalRootTreeItemBase<TItem extends ILocalItem, TProperty
                 const refreshInterval: number = this.getRefreshInterval();
                 intervalId = setInterval(
                     async () => {
-                        if (window.state.focused && await this.hasChanged()) {
-                            await this.refresh();
+                        if (this.autoRefreshEnabled && await this.hasChanged()) {
```
bwateratmsft, line 98: Not really important, and it admittedly looks a little odd, but I believe you can do this to avoid another `if` inside:
```suggestion
                        if (this.autoRefreshEnabled && await this.hasChanged() && this.autoRefreshEnabled) {
```
I'd definitely keep the comment explaining why if we do it that way

ravipal: Indeed that looks odd. I think another if is more readable.

src/tree/contexts/ContextTreeItem.ts
```diff
@@ -60,6 +60,6 @@ export class ContextTreeItem extends AzExtTreeItem {
     }
 
     public async use(context: IActionContext): Promise<void> {
-        return dockerContextManager.use(this.name);
+        await dockerContextManager.use(this.name);
```
karolz-ms, line 63: The original code is fine, It is only if you mix async code and exception handling where you might get in trouble without `await`.

src/tree/LocalRootTreeItemBase.ts
karolz-ms: I think what Ravi has here is good enough, no need to be too clever

PR #1972, opened 2020-05-13
Rearrange top of file and add animated gif showing overview of key features

BigMorty:


README.md
```diff
@@ -1,10 +1,13 @@
...
+
+**Check out the [Working with containers](https://aka.ms/AA7arez) topic on the Visual Studio Code documentation site to get started**.
 
-The Docker extension makes it easy to build, manage and deploy containerized applications from Visual Studio Code.
+![Docker extension overview](resources/readme/overview.gif)
 
-**Check out [Working with containers](https://aka.ms/AA7arez) topic on Visual Studio Code documentation site to get started**.
+## Extension info
+
+[![Version](https://vsmarketplacebadge.apphb.com/version/ms-azuretools.vscode-docker.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) [![Installs](https://vsmarketplacebadge.apphb.com/installs-short/ms-azuretools.vscode-docker.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) [![Build Status](https://dev.azure.com/ms-azuretools/AzCode/_apis/build/status/Nightly/vscode-docker-nightly-2?branchName=master)](https://dev.azure.com/ms-azuretools/AzCode/_build/latest?definitionId=22&branchName=master)
```
bwateratmsft, line 10: These badges should stay at the top. Every other project I've ever seen using badges has them at the top (or immediately under a title). Examples:

- https://github.com/microsoft/vscode-azurefunctions/blob/master/README.md
- https://github.com/OmniSharp/omnisharp-vscode/blob/master/README.md
- https://github.com/microsoft/vscode-eslint/blob/master/README.md

README.md
```diff
@@ -1,10 +1,13 @@
-## Docker for Visual Studio Code
```
bwateratmsft, line 1: Why remove the title? Most or all other first-party extensions have the extension name at the top of their README (examples in a comment below)

README.md
```diff
@@ -1,10 +1,13 @@
...
 
-[![Version](https://vsmarketplacebadge.apphb.com/version/ms-azuretools.vscode-docker.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) [![Installs](https://vsmarketplacebadge.apphb.com/installs-short/ms-azuretools.vscode-docker.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) [![Build Status](https://dev.azure.com/ms-azuretools/AzCode/_apis/build/status/Nightly/vscode-docker-nightly-2?branchName=master)](https://dev.azure.com/ms-azuretools/AzCode/_build/latest?definitionId=22&branchName=master)
+The Docker extension makes it easy to build, manage, and deploy containerized applications from Visual Studio Code. It also provides one-click debugging for Node.js, Python, and .NET Core.
+
+**Check out the [Working with containers](https://aka.ms/AA7arez) topic on the Visual Studio Code documentation site to get started**.
 
-The Docker extension makes it easy to build, manage and deploy containerized applications from Visual Studio Code.
+![Docker extension overview](resources/readme/overview.gif)
```
bwateratmsft, line 6: I think the GIF covers the right content, but there's a couple changes I'd make to it--
1. Shrink the window and zoom in a bit
1. Record only the VSCode window, thus hiding your terrifying launch bar :smile:

README.md
karolz-ms: +1 on that e.g. https://github.com/Microsoft/vscode-remote-containers Understand the desire to gain screen real estate, but it looks weird, like a book without front cover

README.md
karolz-ms: The lanuch bar is downright scary 😆

README.md
karolz-ms: I would actually put them at the very bottom of the doc. Only people who contribute code to the extension care about these badges, which is a very small percentage of our users. They take valuable real estate for no good reason.

README.md
```diff
@@ -1,10 +1,13 @@
...
 
-**Check out [Working with containers](https://aka.ms/AA7arez) topic on Visual Studio Code documentation site to get started**.
+## Extension info
```
karolz-ms, line 8: Having extra paragraph for just the badges and one sentence is a waste of space

README.md
```diff
@@ -1,10 +1,13 @@
-## Docker for Visual Studio Code
 
-[![Version](https://vsmarketplacebadge.apphb.com/version/ms-azuretools.vscode-docker.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) [![Installs](https://vsmarketplacebadge.apphb.com/installs-short/ms-azuretools.vscode-docker.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) [![Build Status](https://dev.azure.com/ms-azuretools/AzCode/_apis/build/status/Nightly/vscode-docker-nightly-2?branchName=master)](https://dev.azure.com/ms-azuretools/AzCode/_build/latest?definitionId=22&branchName=master)
+The Docker extension makes it easy to build, manage, and deploy containerized applications from Visual Studio Code. It also provides one-click debugging for Node.js, Python, and .NET Core.
```
karolz-ms, line 2: > It also provides one-click debugging for Node.js, Python, and .NET Core.

It is not very clear from this sentence that the debugging capability is about **debugging inside a container**.

README.md
rcjsuen: > These badges should stay at the top. Every other project I've ever seen using badges has them at the top (or immediately under a title).

I agree with @bwateratmsft and feel they should be at the top for consistency purposes with other projects.

> Only people who contribute code to the extension care about these badges, which is a very small percentage of our users.

For myself, even if I am not looking to contribute to a project, I use the badges to get a quick overview of a project's popularity. I am more inclined to use something that has a million downloads compared to something that has a thousand downloads.

karolz-ms: @rcjsuen good point about popularity, thx for pointing that out

BigMorty: The reason to move them down was to gain real estate when the ReadMe is shown in VS Code, but I see the point about the download counts. Will move them back up.

README.md
BigMorty: What stinks is the ReadMe is used in the repo and in VS Code when looking at the extension. In VS Code it is a waste of space because it duplicates the name from the VS Code header. That said, it does seem lacking when viewed in GitHub. I will add it back.

README.md
BigMorty: LOL, good point. Will re-record.

README.md
```diff
@@ -1,10 +1,13 @@
+
```
bwateratmsft, line 1: Nit: there's a spurious newline here at the beginning

README.md
```diff
@@ -1,10 +1,13 @@
...
 ## Docker for Visual Studio Code
 
 [![Version](https://vsmarketplacebadge.apphb.com/version/ms-azuretools.vscode-docker.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) [![Installs](https://vsmarketplacebadge.apphb.com/installs-short/ms-azuretools.vscode-docker.svg)](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) [![Build Status](https://dev.azure.com/ms-azuretools/AzCode/_apis/build/status/Nightly/vscode-docker-nightly-2?branchName=master)](https://dev.azure.com/ms-azuretools/AzCode/_build/latest?definitionId=22&branchName=master)
```
karolz-ms, line 4: I just realized you can put these badges next to the title e.g. [how Serilog does it](https://github.com/serilog/serilog/blob/dev/README.md) That will save some screen real estate

bwateratmsft: I like it!

README.md
BigMorty: I removed the line, good catch.

README.md
BigMorty: Agree, looks good

README.md
BigMorty: Updated to mentnion "inside a container"

PR #1983, opened 2020-05-15
Add a timeout to Dockerode requests

### LINKED ISSUE #1947: Explorer is not responsive for newly created remote SSH Docker Context
I then tried to use the command line to SSH into the machine directly, that is when I got this prompt:

image

I answered 'yes' and the SSH connection was made. I was then able to use the Context in VS Code and the panels refreshed as expected. I found the following StackOverflow item that discuss this, it has to do with the ~/.ssh/known_hosts file.
https://stackoverflow.com/questions/3663895/ssh-the-authenticity-of-host-hostname-cant-be-established

Is there a way we can alert the user to this issue?

bwateratmsft:
Fixes #1947


src/utils/callDockerode.ts
```diff
@@ -24,32 +27,48 @@ export async function callDockerode<T>(dockerodeCallback: () => T): Promise<T> {
...
-            /* eslint-disable no-unused-expressions */
-            // This will cause any still-awaiting promises to reject with UserCancelledError, and then cps will be recreated
-            cps?.cancel();
-            cps?.dispose();
-            cps = undefined;
-            /* eslint-enable no-unused-expressions */
-
-            await refreshDockerode();
+export async function callDockerodeAsync<T>(dockerodeAsyncCallback: () => Promise<T>, context?: IActionContext): Promise<T> {
+    const timeoutPromiseSource = new CancellationPromiseSource(Error, localize('vscode-docker.utils.dockerode.timeout', 'Request timed out.'));
```
karolz-ms, line 34: I think we need another utility class for this, e.g. `TimeoutPromiseSource`, in `promiseUtils`. The timer handling logic could go there. This utility would take a timeout and a lambda that would be called if the timeout elapses. Disposing would stop the internal timer automatically.

Having such utility class would simplify the code in `callDockerode`

src/utils/promiseUtils.ts
```diff
@@ -45,8 +45,37 @@ export class CancellationPromiseSource extends vscode.CancellationTokenSource {
...
+    private timeoutTimer: NodeJS.Timeout | undefined;
+    private readonly cps: CancellationPromiseSource;
+
+    public constructor(private readonly timeoutMs: number, private readonly context?: IActionContext) {
+        this.cps = new CancellationPromiseSource(Error, localize('vscode-docker.utils.promiseUtils.timeout', 'Request timed out.'));
+    }
+
+    public get promise(): Promise<never> {
+        this.timeoutTimer = setTimeout(() => {
+            if (this.context) {
```
karolz-ms, line 64: Consider having part about suppressing issue reporting in the `callDockerode` code and passing a lambda to the `TimeoutPromiseSource` constructor (instead of `IActionContext`). This will make `TimeoutPromiseSource` more reusable

bwateratmsft: That's fair. Should we generalize and call it just "callback", something that gets called before the cancel?

karolz-ms: Yes, or "onTimeout", or something like that

src/utils/callDockerode.ts
```diff
@@ -9,12 +9,15 @@ import { IActionContext, parseError, UserCancelledError } from 'vscode-azureexte
...
+import { CancellationPromiseSource, TimeoutPromiseSource } from './promiseUtils';
 import { refreshDockerode } from './refreshDockerode';
 
+// 20 s timeout for all calls (enough time for a possible Dockerode refresh + the call, but short enough to be UX-reasonable)
+const dockerodeCallTimeout = 20 * 1000;
+
 let cps: CancellationPromiseSource | undefined;
 
-export async function callDockerode<T>(dockerodeCallback: () => T): Promise<T> {
+export async function callDockerode<T>(dockerodeCallback: () => T, context?: IActionContext): Promise<T> {
```
ravipal, line 20: The caller should have an ability to pass the timeout. Not all calls require 20 seconds timeout

bwateratmsft: All calls can result in a Dockerode refresh, so it makes sense to use the same timeout.

ravipal: I don't understand your comment. What is the relation between any dockerode call to dockerode refresh? Say if I want to issue an inspect command with 5 seconds time out, how do I do that?

bwateratmsft: The point of wrapping everything in `callDockerode` is so that we can detect context changes and re-initialize the Dockerode object if a change is detected. Ideally, Dockerode would do this internally, but that isn't the case.

Any Dockerode call (including inspect) _could_ be the first one to detect a context switch, in which case we cancel all outstanding Dockerode calls, refresh Dockerode, then return. Detecting requires about 25 ms (hash check of the config file), then reinitializing Dockerode takes ~1-6 seconds depending on how long `docker context inspect` feels like taking.

ravipal: Yes the context refresh happens when the context is changed. My suggestion is to take timeout as optional parameters and use it when provided.

bwateratmsft: That's reasonable, but there isn't anywhere I'd actually use it right now. Can we just defer adding that option until we actually find that we need it?

ravipal: Since now this supports timeout, it is reasonable to expose the timeout to the caller. I don't think it is much work to add an optional parameter. In any case, it is your call.

bwateratmsft: I agree that it would be very easy to add an optional parameter to override the default. However, since there isn't currently a use case, I'm OK adding that parameter if/when we need it.

PR #1993, opened 2020-05-18
[Python] Use debugpy launch

haniamr:
Here is how the flow of the launch request works:
_(host) [VSCode] -> (host) [Debug Adapter] -> (host) [Custom launcher] -> (container) [Debugpy launcher] -> (container) [debugpy]_

We added our own custom launcher which doesn't do much but pass the call to the real launcher in the container using a `docker exec` command.

When debugpy gets the request, it connects back to VSCode on the specified host:port. For Windows/Mac, we use `host.docker.internal:{randomPort}`, however this doesn't work for Linux. So we use the ip address from Docker's predefined network "bridge" and it works in all of the cases I tried (we might need to check with Docker so that we're not basing off wrong assumptions).

The rest of the changes get rid of the attach logic we have.

### DIFF
--- a/resources/python/launcher.py
+++ b/resources/python/launcher.py
@@ -1,0 +1,8 @@
+import os
+import sys
+
+containerId = sys.argv[-1]
+args = sys.argv[1:-1]
+
+print("docker exec -d " + containerId + " python /pydbg/debugpy/launcher " + ' '.join(args))
+os.execvp('docker', ['docker', 'exec', '-d', containerId, 'python', '/pydbg/debugpy/launcher'] + args)
--- a/resources/python/launcher.py
+++ b/resources/python/launcher.py
@@ -1,8 +1,13 @@
-import os
-import sys
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE.md in the project root for license information.
 
+# This acts as a simple launcher for debugpy that only redirects the args to the actual launcher inside the container
+import os, sys
+
+# Container id is the last arg
 containerId = sys.argv[-1]
 args = sys.argv[1:-1]
 
 print("docker exec -d " + containerId + " python /pydbg/debugpy/launcher " + ' '.join(args))
+
 os.execvp('docker', ['docker', 'exec', '-d', containerId, 'python', '/pydbg/debugpy/launcher'] + args)
--- a/resources/python/launcher.py
+++ b/resources/python/launcher.py
@@ -8,6 +8,8 @@ import os, sys
 containerId = sys.argv[-1]
 args = sys.argv[1:-1]
 
-print("docker exec -d " + containerId + " python /pydbg/debugpy/launcher " + ' '.join(args))
+launcherPort = args[0]
+args[0] = 'host.docker.internal:' + launcherPort
 
+print("docker exec -d " + containerId + " python /pydbg/debugpy/launcher " + ' '.join(args))
 os.execvp('docker', ['docker', 'exec', '-d', containerId, 'python', '/pydbg/debugpy/launcher'] + args)
--- a/resources/python/launcher.py
+++ b/resources/python/launcher.py
@@ -8,8 +8,10 @@ import os, sys
 containerId = sys.argv[-1]
 args = sys.argv[1:-1]
 
-launcherPort = args[0]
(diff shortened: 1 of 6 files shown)


bwateratmsft:
Ahh, I love it: 
image

haniamr:
> Ahh, I love it:
> image
> Removing code feels like spring cleaning 😄

Yeah, it was very relieving 😄

rcjsuen:
> > Ahh, I love it:
> > image
> > Removing code feels like spring cleaning 😄
> 
> Yeah, it was very relieving 😄

Deleted code is the best code.

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -114,21 +72,14 @@ export class PythonDebugHelper implements DebugHelper {
...
-        const debugOptions = ['FixFilePathCase', 'RedirectOutput', 'ShowReturnValue'];
-
-        if (os.platform() === 'win32') {
-            debugOptions.push('WindowsClient');
-        }
+        const args = [...debugConfiguration.args || pythonRunTaskOptions.args || [], containerName];
+        const launcherPath = path.join(ext.context.asAbsolutePath('resources'), 'python', 'launcher.py');
 
         return {
-            ...debugConfiguration,
```
bwateratmsft, line 126: Any reason it needs to be, "copy in a bunch of individual things from `debugConfiguration`", instead of the original `...debugConfiguration`?

resources/python/launcher.py
```diff
@@ -0,0 +1,20 @@
...
+# If the adapterHost is only a port number, then append the default DNS name 'host.docker.internal'
+adapterHost = args[0]
+
+if adapterHost.isnumeric():
+    args[0] = 'host.docker.internal:' + adapterHost
+
+dockerExecArgs = ['docker', 'exec', '-d', containerId, 'python', '/pydbg/debugpy/launcher'] + args
+
+print(' '.join(dockerExecArgs))
+os.execvp('docker', dockerExecArgs)
```
bwateratmsft, line 20: Just curious, any reason `execvp` in particular? `p` I understand so that we can make use of `$PATH` to execute `docker`.

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -128,18 +35,19 @@ export namespace PythonExtensionHelper {
             return undefined;
         }
 
-        const debuggerPath = path.join(pyExt.extensionPath, 'pythonFiles', 'lib', 'python');
-        const oldDebugger = path.join(debuggerPath, 'old_ptvsd');
-        const newDebugger = path.join(debuggerPath, 'new_ptvsd');
+        const version = new semver.SemVer(pyExt.packageJSON.version);
+
+        if (version.compare(minPyExtensionVersion) == -1) {
```
bwateratmsft, line 40:
```suggestion
        if (version.compare(minPyExtensionVersion) < 0) {
```
Or otherwise, `===`.

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -36,10 +35,6 @@ export interface PythonDockerDebugConfiguration extends DockerDebugConfiguration
 }
 
 export class PythonDebugHelper implements DebugHelper {
```
bwateratmsft, line 37: Is there any issue from bringing existing debug config / tasks (old scaffolding) into this new model?

src/debugging/python/PythonDebugHelper.ts
haniamr: Yeah, the launch configuration for Python has a property called "python", which conflicts with the one we have in our own configuration so had to go this way.

resources/python/launcher.py
haniamr: I chose the `p` variant for the PATH, and `execv` to ensure arguments are passed as a list so we don't have to worry about concatenating the args and escaping.

src/debugging/python/PythonDebugHelper.ts
haniamr: Not that I can think of, we previously had port and volume mappings specific for the debugger but now are not needed. The only compatibility issue I can think of is related to the version of the Python extension, so I do an explicit check there.

PR #1952, opened 2020-05-07
Deprecation warning for `docker-coreclr` debug

bwateratmsft:

PR #2007, opened 2020-05-21
Fix issues with deprecation warning

bwateratmsft:
Fixes an issue with #1952 and also deals with the new C# extension publisher.


src/debugging/coreclr/prereqManager.ts
```diff
@@ -52,7 +52,7 @@ export class DotNetExtensionInstalledPrerequisite implements Prerequisite {
         //       As this extension targets Docker in general and not .NET Core in particular, we don't want the
         //       extension as a whole to depend on it.  Hence, we only check for its existence if/when asked to
         //       debug .NET Core in Docker containers.
-        const dependenciesSatisfied = this.getExtension('ms-vscode.csharp') !== undefined;
+        const dependenciesSatisfied = this.getExtension('ms-vscode.csharp') !== undefined || this.getExtension('ms-dotnettools.csharp') !== undefined;
```
bwateratmsft, line 55: The C# extension changed their publisher (and thus, extension ID) without bothering to see who was using it, I noticed today when testing the old-style debugging for .NET Core.

PR #2073, opened 2020-06-17
Check for and flag outdated images

### LINKED ISSUE #1493: Mark outdated images
The tags of some of the images in the "Images" tree view are likely to get outdated when maintainers push new versions to the corresponding registries. Similarly to how modifications are marked in git repos, I think it would be useful to have a symbol added to outdated images.

Related to #1313, this feature would help users decide when they need to explicitly pull the image.

bwateratmsft:
Fixes #1493. Looks like this:
image


bwateratmsft:
@BigMorty:

Couple questions for you that @karolz-ms and I were discussing.

1. There's two ways to decide if an image is out-of-date:
    1. If it's anything other than the latest image for this tag (i.e. compare their IDs), it's out-of-date.
    1. If it's more than X amount of time older than the latest image for this tag, it's out-of-date. If we want to do this instead, what is X?
1. Refreshing--should a manual refresh trigger a new outdated check or is once per session good enough? Alternatively, should we offer a command to trigger a new outdated check?
    1. NOTE: if you pull a new version of an image, it will automatically no longer show as "out of date"--so an outdated check is _not_ needed just to make the yellow icon go away on an image that is now up-to-date.

BigMorty:
1. ImageID
2. Start with once per session and see the feedback
3. On by default, can always turn this off if we get feedback

package.json
```diff
@@ -2057,6 +2057,11 @@
                     "type": "boolean",
                     "default": true,
                     "description": "%vscode-docker.config.docker.showRemoteWorkspaceWarning%"
+                },
+                "docker.checkForOutdatedImages": {
```
bwateratmsft, line 2061: Karol suggested moving this up among the other view-related settings, I agree.

src/tree/images/ImagesTreeItem.ts
```diff
@@ -13,14 +13,19 @@ import { ITreeArraySettingInfo, ITreeSettingInfo } from "../settings/ITreeSettin
 import { ImageGroupTreeItem } from './ImageGroupTreeItem';
 import { getImagePropertyValue, imageProperties, ImageProperty } from "./ImageProperties";
 import { ImageTreeItem } from "./ImageTreeItem";
+import { OutdatedImageChecker } from "./OutdatedImageChecker";
+
+export type DatedDockerImage = DockerImage & { Outdated?: boolean };
```
bwateratmsft, line 18: Just make this an interface

karolz-ms:
@bwateratmsft is this ready for review?

bwateratmsft:
> @bwateratmsft is this ready for review?

@karolz-ms Yes, I think so

package.nls.json
```diff
@@ -136,6 +136,7 @@
     "vscode-docker.config.docker.images.description": "Any secondary properties to display for a image (an array). Possible elements include: CreatedTime, FullTag, ImageId, Registry, Repository, RepositoryName, RepositoryNameAndTag, and Tag",
     "vscode-docker.config.docker.images.label": "The primary property to display for a image: CreatedTime, FullTag, ImageId, Registry, Repository, RepositoryName, RepositoryNameAndTag, or Tag",
     "vscode-docker.config.docker.images.sortBy": "The property to use to sort images in Docker view: CreatedTime or Label",
+    "vscode-docker.config.docker.images.checkForOutdatedImages": "Check for outdated base images once per session",
```
karolz-ms, line 139:
```suggestion
    "vscode-docker.config.docker.images.checkForOutdatedImages": "Check for outdated base images (once per VS Code session)",
```

PR #2074, opened 2020-06-18
[Python] Use the Python Extension API to get the debugger location

haniamr:
There is a recent breaking change since the location of "debugpy" changed in the Python extension. So the resolution is to use an API provided by the Python Extension to resolve the location of the debugger to be used.

### DIFF
--- a/src/tasks/python/PythonExtensionHelper.ts
+++ b/src/tasks/python/PythonExtensionHelper.ts
@@ -5,7 +5,6 @@
 
 // This will eventually be replaced by an API in the Python extension. See https://github.com/microsoft/vscode-python/issues/7282
 
-import * as fse from 'fs-extra';
 import * as path from 'path';
 import * as semver from 'semver';
 import * as vscode from "vscode";
@@ -19,6 +18,20 @@ export namespace PythonExtensionHelper {
     }
 
     export async function getLauncherFolderPath(): Promise<string> {
+        const pyExt = await getPythonExtension();
+
+        if (pyExt?.exports?.debug) {
+            const debuggerPath = await pyExt.exports.debug.getDebuggerPackagePath();
+
+            if (debuggerPath) {
+                return debuggerPath;
+            }
+        }
+
+        throw new Error(localize('vscode-docker.tasks.pythonExt.noDebugger', 'Unable to find the debugger in the Python extension.'));
+    }
+
+    export async function getPythonExtension() : Promise<vscode.Extension<any>> | undefined {
         const pyExtensionId = 'ms-python.python';
         const minPyExtensionVersion = new semver.SemVer('2020.5.78807');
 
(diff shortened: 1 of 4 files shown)


src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -19,6 +17,23 @@ export namespace PythonExtensionHelper {
     }
 
     export async function getLauncherFolderPath(): Promise<string> {
+        const pyExt = await getPythonExtension();
+
+        // tslint:disable:no-unsafe-any no-any
```
bwateratmsft, line 22: Can you install eslint instead of tslint? The suppressions are different for that as well.

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -19,6 +17,23 @@ export namespace PythonExtensionHelper {
...
+        }
+        /* eslint-enable @typescript-eslint/tslint/config */
+
+        throw new Error(localize('vscode-docker.tasks.pythonExt.noDebugger', 'Unable to find the debugger in the Python extension.'));
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    export async function getPythonExtension(): Promise<vscode.Extension<any>> | undefined {
         const pyExtensionId = 'ms-python.python';
         const minPyExtensionVersion = new semver.SemVer('2020.5.78807');
```
bwateratmsft, line 38: Shouldn't there be a new minimum version?

src/tasks/python/PythonExtensionHelper.ts
```diff
@@ -19,6 +17,23 @@ export namespace PythonExtensionHelper {
...
+                return debuggerPath;
+            }
+        }
+        /* eslint-enable @typescript-eslint/tslint/config */
+
+        throw new Error(localize('vscode-docker.tasks.pythonExt.noDebugger', 'Unable to find the debugger in the Python extension.'));
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    export async function getPythonExtension(): Promise<vscode.Extension<any>> | undefined {
```
bwateratmsft, line 36: We should make an interface file containing their extension API. It won't be any more susceptible to out-of-sync issues than we already are, but would let us get rid of suppressions and make it clearer what the Python extension is giving us.

bwateratmsft: It's fine to include only the parts of the API we are using. We should also have a link in it to their source file.

src/tasks/python/PythonExtensionHelper.ts
haniamr: No, the API already exists in the current minimum version (based on check-in dates at least). I can quickly verify that with Karthik.

PR #2084, opened 2020-06-18
Update test framework to vscode-test

### LINKED ISSUE #1737: Revisit test infrastructure
Our tests haven't gotten much TLC lately and need some attention. The infrastructure is dated and many of the tests are noisy. Let's see what we can do.

Additionally, the [vscode](https://www.npmjs.com/package/vscode) NPM package for testing is deprecated, and replaced by [vscode-test](https://www.npmjs.com/package/vscode-test). We also need a newer version of mocha.

bwateratmsft:
Fixes #1737


.vscode/launch.json
```diff
@@ -30,14 +30,15 @@
 				"--extensionDevelopmentPath=${workspaceFolder}"
 			],
 			"env": {
+				"AZCODE_DOCKER_IGNORE_BUNDLE": "",
```
bwateratmsft, line 33: It may have been this bug: https://github.com/microsoft/vscode/issues/100265

But it seems like env vars that were set in a previous launch config persisted into the next launch, which was messing a lot of things up.

philliphoff: Yes, I've seen this happen from time to time.

PR #2089, opened 2020-06-20
Deliver signals to the app

wojpawlik:


karolz-ms:
We should use [the "exec form" of the CMD command](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#cmd) This will ensure the signals are properly delivered and eliminate shell dependency.

@wojpawlik would you like to make that adjustment to your PR?

wojpawlik:
No, I'm too unfamiliar with the code. This would require changing the type of https://github.com/microsoft/vscode-docker/blob/05d585ed9c98533d506ed70f0236968faf278129/src/configureWorkspace/configure.ts#L28.

karolz-ms:
Fair enough. I have opened https://github.com/microsoft/vscode-docker/issues/2090 to have this problem fixed in our whole codebase.

Thank you for helping us make the Docker extension better--much appreciated! ☺️

PR #2100, opened 2020-06-24
Better UX on context removal

### LINKED ISSUE #2098: The error is inconsistent when single remove default context and remove it with other contexts together
**OS:** Win 10
**Build Version:** 20200623.1

**Repro Steps:**
1. Make sure there are several contexts in Docker context view and use default context.
2. F1 --> Docker Contexts: Remove... --> Select all context to remove.
3. Check the error message.

**Expect:**
An error indicating that the in use context will not be removed.
image

**Actual:**
The error is "Process 'docker context rm default' exited with code 1".
image

**More Info:**
1. An error indicating that the in use context will not be removed if there is only one default context.
2. The error is "Process 'docker context rm default' exited with code 1" when using other context and remove the default context.

bwateratmsft:
Fixes #2098

1. Context tree is no longer multi-select enabled (removing multiple contexts at once is not a realistic use case, and inspect and use aren't enabled anyway [use isn't even possible for multi-select nor does it make sense])
1. The default and current context cannot be removed--the "Remove" option won't even appear in the context menu, nor will the default or current context show when using "Remove" from the command palette
1. Help tree is no longer multi-select enabled (it never needed to be)

Can't remove in-use custom context, or default context, from tree:
image image

Or from palette:
image

Upon switching back to default:
image image

### DIFF
--- a/package.json
+++ b/package.json
@@ -541,17 +541,17 @@
                 },
                 {
                     "command": "vscode-docker.contexts.inspect",
-                    "when": "view == vscode-docker.views.dockerContexts && viewItem == context",
+                    "when": "view == vscode-docker.views.dockerContexts && viewItem =~ /Context$/i",
                     "group": "contexts_1_general@1"
                 },
                 {
                     "command": "vscode-docker.contexts.use",
-                    "when": "view == vscode-docker.views.dockerContexts && viewItem == context",
+                    "when": "view == vscode-docker.views.dockerContexts && viewItem =~ /Context$/i",
                     "group": "contexts_1_general@2"
                 },
                 {
                     "command": "vscode-docker.contexts.remove",
-                    "when": "view == vscode-docker.views.dockerContexts && viewItem == context",
+                    "when": "view == vscode-docker.views.dockerContexts && viewItem =~ /^customContext$/i",
                     "group": "contexts_2_destructive@1"
                 }
             ]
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -541,17 +541,17 @@
                 },
                 {
                     "command": "vscode-docker.contexts.inspect",
-                    "when": "view == vscode-docker.views.dockerContexts && viewItem == context",
+                    "when": "view == vscode-docker.views.dockerContexts && viewItem =~ /Context$/i",
```
bwateratmsft, line 544: TODO: Add some screenshots

src/tree/contexts/ContextTreeItem.ts
```diff
@@ -10,15 +10,26 @@ import { getThemedIconPath, IconPath } from '../IconPath';
...
+
     private readonly _item: LocalContextInfo;
 
     public constructor(parent: AzExtParentTreeItem, item: LocalContextInfo) {
         super(parent);
         this._item = item;
     }
 
+    public get contextValue(): string {
+        if (/^default$/i.test(this.name)) {
```
karolz-ms, line 24: Turns out it is completely possible to create an additional Docker context named `DEfault` or something like this, even on Windows. Should we check `this.name === 'default'` instead?

bwateratmsft: That should be fine. We also do this same comparison in `refreshDockerode`. I'll change both.

Issue #2102, opened 2020-06-25
Support ACI contexts in Explorer Panel


dbreshears:
When switching to ACI context, the containers view should reflect the containers running in the ACI context. Images, networks, and volumes views should show "This view is not supported in the current Docker context."

Additionally, as many of the context commands as possible should be enabled.

bwateratmsft:
We have released Docker 1.4.0 which contains this feature.

PR #2119, opened 2020-06-30
Create a client API for wrapping SDK calls

bwateratmsft:
Related to #2102.

This is a pretty substantial PR but fundamentally it is wrapping Dockerode + the new Docker SDK so that we can intelligently use the right one based on context. It also refactors the Docker context manager somewhat to be event-based.


src/docker/ContextManager.ts
```diff
@@ -0,0 +1,201 @@
...
+    readonly onContextChanged: Event<DockerContext>;
+    refresh(): Promise<void>;
+    getContexts(): Promise<DockerContext[]>;
+
+    inspect(actionContext: IActionContext, contextName: string): Promise<DockerContextInspection>;
+    use(actionContext: IActionContext, contextName: string): Promise<void>;
+    remove(actionContext: IActionContext, contextName: string): Promise<void>;
+}
+
+// TODO: consider a periodic refresh as a catch-all; but make sure it compares old data to new before firing a change event
```
bwateratmsft, line 53: Consider this

src/docker/ContextChangeCancelClient.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    protected async withTimeoutAndCancellations<T>(context: IActionContext, callPromise: () => Promise<T>, timeout: number, token?: CancellationToken): Promise<T> {
+        const tps = new TimeoutPromiseSource(timeout);
+        const evt = tps.onTimeout(() => {
+            context.errorHandling.suppressReportIssue = true;
+        });
+
+        try {
+            const promises: Promise<T>[] = [tps.promise, this.contextChangeCps.promise, callPromise()];
```
karolz-ms, line 41: Isn't it nice that `Promise<never>` can act as `Promise<T>`? ☺️

src/docker/ContextManager.ts
```diff
@@ -0,0 +1,201 @@
...
+const dockerConfigFile = path.join(os.homedir(), '.docker', 'config.json');
+const dockerContextsFolder = path.join(os.homedir(), '.docker', 'contexts', 'meta');
+
+const WindowsLocalPipe = 'npipe:////./pipe/docker_engine';
+const UnixLocalPipe = 'unix:///var/run/docker.sock';
+
+const defaultContext: Partial<DockerContext> = {
+    Name: 'default',
+    Description: 'Current DOCKER_HOST based configuration',
+    CreatedTime: undefined, // Not defined for contexts
```
karolz-ms, line 40: If the type of the variable is `Partial<DockerContext>` do you need to set the `CreatedTime` to undefined?

src/docker/ContextManager.ts
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        try {
+            this.refreshing = true;
+
+            this.contextsCache.clear();
+            const contexts = await this.contextsCache.getValue();
+            const currentContext = contexts.find(c => c.Current);
+
+            if (currentContext.DockerEndpoint === '') { // TODO: check based on type
```
karolz-ms, line 93: This is a little awkward. Maybe use a map (endpoint type) --> (Docker client factory method)?

I am also wondering if we should always create a new client when refreshing, even if the current client is of the expected kind.

src/docker/DockerApiClient.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Disposable } from 'vscode';
+import { IActionContext } from 'vscode-azureextensionui';
+import { CancellationToken } from 'vscode-languageclient';
+import { DockerInfo, PruneResult } from './Common';
+import { DockerContainer, DockerContainerInspection } from './Containers';
+import { DockerImage, DockerImageInspection } from './Images';
+import { DockerNetwork, DockerNetworkInspection, DriverType } from './Networks';
+import { DockerVolume, DockerVolumeInspection } from './Volumes';
+
+export interface DockerApiClient extends Disposable {
```
karolz-ms, line 15: Nice and clean, love it

src/tree/contexts/ContextsTreeItem.ts
```diff
@@ -42,21 +41,20 @@ export class ContextsTreeItem extends LocalRootTreeItemBase<LocalContextInfo, Co
...
+                return item.Name;
             case 'Description':
-                return item.data.Description;
+                return item.Description;
             case 'DockerEndpoint':
-                return item.data.DockerEndpoint;
+                return item.DockerEndpoint;
             default:
-                return getCommonPropertyValue(item, property);
+                return '';
```
karolz-ms, line 57: A comment why this is OK would be nice

src/docker/Contexts.ts
```diff
@@ -0,0 +1,19 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { DockerObject } from './Common';
+
+export interface DockerContext extends DockerObject {
+    readonly Description?: string;
+    readonly DockerEndpoint: string;
+    readonly Current: boolean;
+
+    readonly Id: undefined; // Not defined for contexts
```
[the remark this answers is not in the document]
karolz-ms: I can see how that might complicate the tree code though... 🤔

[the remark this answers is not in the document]
bwateratmsft: Yeah 😞 

I'll look again at it and see if I can figure out something better.

src/docker/Containers.ts
```diff
@@ -0,0 +1,41 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DockerObject } from './Common';
+
+export interface DockerPort {
+    readonly IP?: string;
+    readonly PrivatePort?: number;
```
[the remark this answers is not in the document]
bwateratmsft: That's correct, but I wanted to match the Docker Engine API's shape: https://docs.docker.com/engine/api/v1.40/#operation/ContainerList

This way there is less translation needed on the objects Dockerode gets.

src/docker/ContextManager.ts
bwateratmsft: For ACI at least, we shouldn't need to recreate on context switch, because the SDK + daemon should be context aware.

src/docker/ContextChangeCancelClient.ts
bwateratmsft: I'm glad, since it was the only way to implement this that the transpiler was happy with!

src/docker/Contexts.ts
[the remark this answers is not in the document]
bwateratmsft: I pushed that, take a look and let me know what you think.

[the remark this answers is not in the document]
karolz-ms: Looks good!

PR #2130, opened 2020-07-06
Use python3 explicitly

### LINKED ISSUE #2123: How do you set the debugpy launcher to use python3?
https://github.com/microsoft/vscode-docker/blob/master/resources/python/launcher.py

This file has a section of args that are passed to docker exec.

```
dockerExecArgs = ['docker', 'exec', '-d', containerId, 'python', '/debugpy/launcher'] + args
```

My debug fails, because the container receives `python` as python2.7, but my program needs python3, which is available in the container. If i change this launcher.py file to python3 it works. Can i pass this somehow to the launcher?

bwateratmsft:
Fixes #2123

### DIFF
--- a/resources/python/launcher.py
+++ b/resources/python/launcher.py
@@ -14,7 +14,7 @@ adapterHost = args[0]
 if adapterHost.isnumeric():
     args[0] = 'host.docker.internal:' + adapterHost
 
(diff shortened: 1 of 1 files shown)


bwateratmsft:
@int19h do you have any concerns with this change?

int19h:
Should be good per https://www.python.org/dev/peps/pep-0394/

PR #2131, opened 2020-07-06
Use latest TAS client

bwateratmsft:
This will give us the language traffic filter

### DIFF
--- a/src/telemetry/surveys/awareness.ts
+++ b/src/telemetry/surveys/awareness.ts
@@ -3,7 +3,6 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as vscode from 'vscode';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
 import { Survey } from './SurveyManager';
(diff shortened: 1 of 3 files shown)


src/telemetry/surveys/awareness.ts
```diff
@@ -24,7 +23,6 @@ async function isEligible(): Promise<boolean> {
     const overallActivity = ext.activityMeasurementService.getActivityMeasurement('overall');
     const noEditActivity = ext.activityMeasurementService.getActivityMeasurement('overallnoedit');
 
-    return (vscode.env.language === 'en' || vscode.env.language.startsWith('en-')) &&
```
bwateratmsft, line 27: TODO: this change essentially means we only show to Docker extension 1.4.0, assuming we also turn on the language filter on the awareness flight. Is that OK?

karolz-ms: I'd think so, given how quickly people have been upgrading the extension in the past

bwateratmsft: Talked to @BigMorty, he's OK with locking this experiment to 1.4.0+, thus doing this change is OK.

PR #2134, opened 2020-07-07
Add new Docker SDK-based client

### LINKED ISSUE #2102: Support ACI contexts in Explorer Panel

bwateratmsft:
Fixes #2102 

Add the new Docker SDK-based client under the "DockerServe" client implementation. Supported context-menu commands on containers include:
- View Logs
- Attach Shell
- Inspect
- Open in Browser
- Remove

These are _not_* implemented in the Docker SDK yet and will show an error "This action is not supported in the current Docker context.":
- Restart
- Start
- Stop

*Stop is but the behavior is not exactly the same as `docker stop`, due to the limitations of ACI. We'll leave it unimplemented for now.


src/docker/DockerServeClient/DockerServeUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    return {
+        Id: container.id,
+        Image: container.image,
+        Name: container.id, // TODO ?
+        State: container.status,
+        Status: container.status,
+        ImageID: undefined, // TODO ?
+        CreatedTime: undefined, // TODO ?
+        Labels: labels, // TODO--not working
```
karolz-ms, line 37: Presumably it will work once the SDK supports local Docker engine. It does not work with ACI because ACI does not support labels.

What I am saying let's put a better comment here ☺️

gtardif:
landed here following issues from https://github.com/docker/node-sdk ; just noticed "remove" is not planned to be implemented as "not supported in sdk". 
You should be able to implement the functionality with the Delete Method. 
With the warning that there is a slight semantic difference between ACI and local containers  (at least in the command line) : `docker rm xxx` requires the container to be stopped for local containers. With ACI, `docker rm xxx` will kill the container and remove it (equivalent of `docker rm -f xxx`).

bwateratmsft:
> landed here following issues from https://github.com/docker/node-sdk ; just noticed "remove" is not planned to be implemented as "not supported in sdk".
> You should be able to implement the functionality with the Delete Method.
> With the warning that there is a slight semantic difference between ACI and local containers (at least in the command line) : `docker rm xxx` requires the container to be stopped for local containers. With ACI, `docker rm xxx` will kill the container and remove it (equivalent of `docker rm -f xxx`).

That was a mis-type by me, Remove _will_ be present. Just stop/start/restart are being turned off (for now).

src/docker/DockerServeClient/DockerServeClient.ts
```diff
@@ -110,4 +153,28 @@ export class DockerServeClient extends ContextChangeCancelClient implements Dock
         throw new NotSupportedError(context);
     }
     // #endregion Not supported by the Docker SDK yet
+
+    private async promisify<TRequest, TResponse>(
+        context: IActionContext,
+        thisArg: unknown,
```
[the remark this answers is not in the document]
bwateratmsft: When we eventually add more things (images, networks, etc.) it won't always be `this.containersClient`. There are other ways to accomplish this same thing but all of them involve either binding `this.containersClient` to the callback or passing it in as a `thisArg`.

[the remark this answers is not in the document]
karolz-ms: makes sense

src/docker/DockerServeClient/DockerServeClient.ts
```diff
@@ -13,22 +16,57 @@ import { DockerImage, DockerImageInspection } from '../Images';
...
+    }
+
     public async info(context: IActionContext, token?: CancellationToken): Promise<DockerInfo> {
         throw new NotSupportedError(context);
     }
 
     public async getContainers(context: IActionContext, token?: CancellationToken): Promise<DockerContainer[]> {
-        throw new NotSupportedError(context);
+        const response: ListResponse = await this.promisify(context, this.containersClient, this.containersClient.list, new ListRequest(), token);
+        const result = response.getContainersList();
```
bwateratmsft, line 43: Does this handle compose-started groups?

bwateratmsft: Yup

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -69,6 +69,11 @@ export class PythonDebugHelper implements DebugHelper {
         const projectType = debugConfiguration.python.projectType;
         const pythonRunTaskOptions = (context.runDefinition as PythonRunTaskDefinition)?.python || {};
 
+        ext.outputChannel.appendLine('Run task options:');
```
bwateratmsft, line 72: This change shouldn't be in here

src/commands/containers/attachShellContainer.ts
```diff
@@ -20,7 +21,14 @@ export async function attachShellContainer(context: IActionContext, node?: Conta
         });
     }
 
-    let osType = await getDockerOSType(context);
+    let osType: DockerOSType;
+    try {
+        osType = await getDockerOSType(context);
```
bwateratmsft, line 26: Get OS type from container instead of system? That would allow for Windows containers in ACI to work.

* Need the SDK to provide platform

src/tasks/TaskHelper.ts
```diff
@@ -125,7 +125,15 @@ export async function getAssociatedDockerRunTask(debugConfiguration: DockerDebug
     const workspaceTasks = workspace.getConfiguration('tasks');
     const allTasks: TaskDefinitionBase[] = workspaceTasks && workspaceTasks.tasks as TaskDefinitionBase[] || [];
 
-    return await recursiveFindTaskByType(allTasks, 'docker-run', debugConfiguration) as DockerRunTaskDefinition;
+    ext.outputChannel.appendLine('Getting run tasks. VSCode returned this for all tasks:');
```
bwateratmsft, line 128: This should not be in here

package-lock.json
```diff
@@ -81,6 +115,127 @@
                 }
             }
         },
+        "@octokit/auth-token": {
```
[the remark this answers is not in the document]
bwateratmsft: Webpack's tree shaking is removing the Octokit stuff (as expected), so I think it's OK to merge these changes. We'll pick up a new version of the Docker SDK as soon as one is available.

PR #2135, opened 2020-07-07
Support creating an ACI context from palette / contexts view

### LINKED ISSUE #2114: Support creating a new ACI context from the Contexts panel
Similar to the + sign in the Networks panel we should support creating an ACI context. Once we support ACI and SSH context creation the + sign icon should prompt user for type of context. For ACI we should then sign into Azure if needed, then show list of subscriptions, then list of resource groups or ability to create a new resource group.

bwateratmsft:
Fixes (mostly) #2114. Adds support for creating an ACI context in the command palette and contexts view. The command will _not_ show up unless we can positively confirm that the new CLI is present, according to the following rules--
1. If another ACI context already exists, OR
1. If the output of `docker serve --help` starts with "Start an api server"

TODO: add screenshots

### DIFF
--- a/src/commands/context/aci/createAciContext.ts
+++ b/src/commands/context/aci/createAciContext.ts
@@ -76,7 +76,6 @@ class AciContextCreateStep extends AzureWizardExecuteStep<IAciWizardContext> {
         } catch (err) {
             const error = parseError(err);
 
-            // TODO: get the real error code
             if (error.errorType === '5' || /not logged in/i.test(error.message)) {
                 // If error is due to being not logged in, we'll go through login and try again
                 await execAsync('docker login azure');
--- a/src/commands/context/aci/createAciContext.ts
+++ b/src/commands/context/aci/createAciContext.ts
@@ -77,7 +77,7 @@ class AciContextCreateStep extends AzureWizardExecuteStep<IAciWizardContext> {
             const error = parseError(err);
 
             // TODO: get the real error code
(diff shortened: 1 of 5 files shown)


src/docker/ContextManager.ts
```diff
@@ -96,8 +103,12 @@ export class DockerContextManager implements ContextManager, Disposable {
 
             // Create a new client
             if (currentContext.Type === 'aci') {
+                await this.setVsCodeContext('vscode-docker:aciContext', true);
```
bwateratmsft, line 106: Right now, `vscode-docker:aciContext` and `vscode-docker:newSdkContext` mean the same thing, but that probably will not be true in the future, so I set it up as two different contexts.

karolz-ms: Put this as a comment in the code please

src/commands/context/aci/createAciContext.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        const command = `docker context create aci ${wizardContext.contextName} --subscription-id ${wizardContext.subscriptionId} --location ${wizardContext.location.name} --resource-group ${wizardContext.resourceGroup.name}`;
+
+        try {
+            await execAsync(command);
+        } catch (err) {
+            const error = parseError(err);
+
+            // TODO: get the real error code
+            if (error.errorType === '1234') {
```
bwateratmsft, line 80: Do

Issue #2055, opened 2020-06-10
Request textDocument/semanticTokens failed


csakaszamok:
during writing a Dockerfile i get error messages:  

[Error - 10:47:31 AM] Request textDocument/semanticTokens failed.
  Message: Request textDocument/semanticTokens failed with message: Maximum call stack size exceeded
  Code: -32603

rcjsuen:
@csakaszamok Thank you for your bug report. I presume this bug is not reproducible?

Can you share your Dockerfile? Do you remember what you were changing at the time?

csakaszamok:
yes, you can reproduce it with a litte sample Dockerfile:
FROM ubuntu
ENV \
# common variables:
  ENV1=something \
  ENV2=999 

if you remove or move the comment, everything is ok
so it is good:
FROM ubuntu
ENV \
  ENV1=something \
# common variables:
  ENV2=999 


thank you

rcjsuen:
@csakaszamok Thank you for sharing your Dockerfile. I have opened https://github.com/rcjsuen/dockerfile-language-service/issues/69 to track this.

csakaszamok:
Thx

JoseFMP:
Still happening over here.

rcjsuen:
> Still happening over here.

Hi, @JoseFMP. You are correct. The bug fix is not in a release yet so you will continue to encounter this issue until then. Thank you for your understanding.

rcjsuen:
@JoseFMP Would you mind sharing your Dockerfile here so we can ensure that your bug is covered?

rcjsuen:
Thank you @csakaszamok and @JoseFMP for your bug report. I have opened #2146 to address this.

bwateratmsft:
We have released Docker 1.4.0 which contains a fix for this.

khs1994:
```docker
FROM alpine

ENV PHP_EXTRA_CONFIGURE_ARGS \
        kkk \
        # comment
        # comment
        # comment
        # comment
        kkk2
```

with one more comment line,show this error message

rcjsuen:
@khs1994 Thank you for your bug report. I have reproduced the problem and have opened https://github.com/rcjsuen/dockerfile-language-service/issues/74 to track this.

Issue #2043, opened 2020-06-05
Ctrl Click on FROM leads to a DockerHub link, instead of a stage in a multi-stage config


samhwang:
Hey VSCode Docker team,

I just noticed that the recent update has made it possible to Ctrl+Click on the image name, and then open an appropriate one in DockerHub, which is *fantastic!* . So I just encountered this little issue, and absolutely not a deal breaker at all, but it would be nice to have it fixed.

## The situation:
I want all my stages to have the same `WORKDIR` and `LABEL`, and `EXPOSE`-ing the same port, so I put an official image at the top, and made it a `base` stage. A snippet of what I'm working is like this.

```
FROM node:12-alpine as base
LABEL MAINTAINER="Sam Huynh"
WORKDIR /srv/app
EXPOSE 1337

# Dev dependencies
FROM base as dev_dependencies
COPY package.json yarn.lock ./
RUN yarn install && \
    yarn cache clean

# Dev stage
FROM base as development

COPY . .
COPY --from=dev_dependencies /srv/app/node_modules ./node_modules
CMD ["yarn", "develop"]

# Other stages
## Build something else...
```

## Working:
- Ctrl+Clicking on `node` in `FROM node:12-alpine` prompts me to open [The official Node docker image](https://hub.docker.com/_/node), which is expected.
![Screenshot from 2020-06-05 10-24-11](https://user-images.githubusercontent.com/4188758/83823212-b79f2200-a716-11ea-947c-db338f67f4d5.png)

- Ctrl+clicking on `dev_dependencies` in `--from=dev_dependencies` brings me back to the that stage, which is also correct.
![Screenshot from 2020-06-05 10-27-08](https://user-images.githubusercontent.com/4188758/83823339-22e8f400-a717-11ea-888d-1f4e2215ee48.png)

## Expected:
- Ctrl+Clicking at `base` in `FROM base` goes back to the `base` stage.

## What happened:
- Ctrl+Clicking at `base` in `FROM base` prompts me to open [The "offical" base docker image](https://hub.docker.com/_/base), which doesn't exist.
![Screenshot from 2020-06-05 10-30-10](https://user-images.githubusercontent.com/4188758/83823526-92f77a00-a717-11ea-9056-17e53a835bb6.png)

Please let me know if there are any other info I can put in here.

rcjsuen:
[...]

PR #2146, opened 2020-07-11
Upgrade language server to 0.1.0

rcjsuen:
This update provides the following changes:
- continued improvements and fixes to the semantic tokens parser (#2055)
- editor visuals for the deprecated `MAINTAINER` instruction
- linting fixes due to incorrect calculations of ranges
- navigation of build stages (#2043)

image

Please use the Dockerfile to test the new changes and bug fixes.
```Dockerfile
FROM alpine AS base
# this should not be a link anymore, issue 2043
# instead, clicking on it now should jump to the stage definition above
# https://github.com/microsoft/vscode-docker/issues/2043
FROM base
# this will now have a strikeout decoration to indicate that it is deprecated
# not sure if this is only for VS Code 1.47 or not though...?
MAINTAINER user

# the A should be coloured like X and Y
RUN --mount=X=Y,A echo

FROM scratch
# this should create an error because the ENV has no arguments
ENV \
    # comment
```
The following Dockerfiles have very specific whitespace characters to reproduce the error in question. Simply copy and paste the content into a Dockerfile and and open the Output panel and select the Dockerfile Language Server to verify that there are no errors being logged. Do not save the file as the formatter will change the integrity of the file.
```Dockerfile
# issue 2055
# https://github.com/microsoft/vscode-docker/issues/2055
FROM alpine
ENV \
    # common variables:
    ENV1=something
```
```Dockerfile
FROM alpine
COPY a b \
\ 
/dir
```
```Dockerfile
# there are two problems here
# 1 - the semantic tokens is buggy and there is an infinite loop here
# 2 - the validation is wrong, the red squiggles should be over the whole argument
# which spans across all 3 lines instead of simply the `/di\` in the beginning
FROM alpine
COPY a b /di\
\
\r
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2961,73 +2961,42 @@
             }
         },
         "dockerfile-language-server-nodejs": {
-            "version": "0.0.24",
-            "resolved": "https://registry.npmjs.org/dockerfile-language-server-nodejs/-/dockerfile-language-server-nodejs-0.0.24.tgz",
-            "integrity": "sha512-tG0cE6yIyj80eKHMbDBMNPHc1l40GW8v7b8jeDRs0Jv42boG0s6DEsM/eDFQw92mwAtBry07mB1tM+jSwqGzJw==",
+            "version": "0.1.0",
+            "resolved": "https://registry.npmjs.org/dockerfile-language-server-nodejs/-/dockerfile-language-server-nodejs-0.1.0.tgz",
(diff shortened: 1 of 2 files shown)


rcjsuen:
Looks like the macOS build failed but I have no idea why.

bwateratmsft:
The macOS pipeline is really temperamental for whatever reason...

rcjsuen:
@bwateratmsft Thanks for reviewing. See you again in a month or two I guess...? :P

PR #2151, opened 2020-07-13
Disable ACI create command

bwateratmsft:
The fix to stop prompting for subscription ID when it is already passed hasn't been implemented yet, so disabling the ACI create button/command. Related to #2114.


bwateratmsft:
I'll hold off on merging this until we know for sure if it is needed.

bwateratmsft:
I have verified that the fix is working in the latest build, so I will close this.

PR #2152, opened 2020-07-13
Change to hashed Docker Desktop ID

bwateratmsft:
The unhashed ID is automatically discarded by the telemetry system, so change to a hashed ID. We only need it for uniqueness anyway.


src/extension.ts
```diff
@@ -81,7 +82,7 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
     await callWithTelemetryAndErrorHandling('docker.activate', async (activateContext: IActionContext) => {
         activateContext.telemetry.properties.isActivationEvent = 'true';
         activateContext.telemetry.measurements.mainFileLoad = (perfStats.loadEndTime - perfStats.loadStartTime) / 1000;
-        activateContext.telemetry.properties.dockerInstallationID = await getDockerInstallationID();
+        activateContext.telemetry.properties.dockerInstallationIDHash = await getDockerInstallationIDHash();
```
bwateratmsft, line 85: Changing the name is needed because the old data's classification cannot be changed.

Issue #1959, opened 2020-05-08
Getting issue when i deploy to docker registry


kalanchiyamtutorial:
Does this occur consistently? 
Repro steps:

1.
2.

Action: vscode-docker.registries.logInToDockerCli
Error type: Error
Error Message: In order to log in to the Docker CLI using tokens, you currently need to go to 
Open your Docker config file and remove "credsStore": "wincred" from the config.json file, then try again. 
Doing this will disable wincred and cause Docker to store credentials directly in the .docker/config.json file. All registries that are currently logged in will be effectly logged out.

Version: 1.1.0
OS: win32
OS Release: 10.0.17763
Product: Visual Studio Code
Product Version: 1.44.2
Language: en

<details>
<summary>Call Stack</summary>

```
extension.bundle.js:164:252001extension.bundle.js:164:252001
ChildProcess.exithandler child_process.js:315:5
ChildProcess.emit events.js:203:13
maybeClose child_process.js:1021:16
Process.ChildProcess._handle.onexit child_process.js:283:5
```

</details>

bwateratmsft:
Have you tried removing `"credsStore": "wincred"` from the Docker config.json?

Naninani:
> Have you tried removing `"credsStore": "wincred"` from the Docker config.json?

Removing the `"credsStore": "wincred"` is exposing the auth token in the config file. Sure that works, but how is that considered secure and compliant approach?

bwateratmsft:
@Naninani I looked at our code and it appears this warning message was created to work around https://github.com/Azure/azure-cli/issues/4843 -- now fixed for years. I'll take a look to see if it is needed anymore.

bwateratmsft:
@Naninani the problem still exists and is outside of our control--wincred has a limit of 2.5K and the refresh tokens used for CLI login far exceed that. In an [old comment](https://github.com/Azure/azure-cli/issues/4843#issuecomment-342712079), @djyou implied that there would be a server-side change to limit the token length; @djyou, has that actually happened? A refresh token I got just now is 2717 characters long, so the error reproduces.

In the meantime, I'm going to clean up the messy grammar in the message. You are absolutely correct that turning off wincred is less secure, but there's nothing the Docker extension can do about the limitations of it.

bwateratmsft:
We think this should be fixed by #2153. Users who want to remove the `"credsStore": "desktop"` entry can do so, but may need to log out and back in.

bwateratmsft:
[...]

PR #2153, opened 2020-07-13
Use access_token grant instead of refresh_token

bwateratmsft:
This substantially mitigates #1959 or possibly eliminates the issue entirely. Changing the ADAL token <=> ACR token exchange's `grant_type` to `access_token`, and not passing in the ADAL `refresh_token`, dramatically reduces the size of the obtained ACR `refresh_token`. My token's size was pretty consistently about 950 bytes, compared to > 2600 before.

It may be functionally impossible to get a token using this different mechanism that is too long for Docker CLI. We won't update people's existing Docker configs (at least not now...), but we hopefully will never need to suggest disabling Wincred again.

The disadvantage to this grant type is that if the ADAL `access_token` is near expiry, they (ACR) will not be able to use the ADAL `refresh_token` to get a new one--however, that shouldn't be an issue for us since we get a fresh ADAL token from the Azure Account extension every time.

### DIFF
--- a/src/commands/context/aci/createAciContext.ts
+++ b/src/commands/context/aci/createAciContext.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Progress } from 'vscode';
-import { AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, IActionContext, IResourceGroupWizardContext, LocationListStep, parseError, ResourceGroupListStep } from 'vscode-azureextensionui';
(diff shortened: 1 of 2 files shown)


src/commands/context/aci/createAciContext.ts
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Progress } from 'vscode';
-import { AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, IActionContext, IResourceGroupWizardContext, LocationListStep, parseError, ResourceGroupListStep } from 'vscode-azureextensionui';
```
karolz-ms, line 7: This file has unrelated changes--please exclude

bwateratmsft: I was gonna put them together (me lazy) but I can separate.

PR #2169, opened 2020-07-16
Rewrite compose commands for ACI contexts

### LINKED ISSUE #2168: Compose up / down commands will not work in ACI context
The command needs to be altered so that `docker-compose` becomes `docker compose`. Additionally, `--build` needs to be removed.

bwateratmsft:
1.4.0 temporary fix for #2168


bwateratmsft:
> 😑 This is fine for the upcoming release, but we should think about the implications on command customization. I think sooner or later, we will have to have something that allows users to customize commands depending on the Docker context type or associated VS Code context value.
> 
> Maybe a "CONSIDER" comment would be useful

I like what we talked about offline--add a `contextType` sub-property for command customizations, and have new default values that are arrays (where needed) instead of single values. I'll look at how much implementation work that would be.

PR #2171, opened 2020-07-17
Add image size as label and sortable (but not groupable)

### LINKED ISSUE #2047: Add image size on description
Issue Type: <b>Feature Request</b>

In the settings of the list of images there is not enough opportunity to display image size. The description configuration does not have this option.

image

Extension version: 1.2.1
VS Code version: Code 1.45.1 (5763d909d5f12fe19f215cbfdd29a91c0fa9208a, 2020-05-14T08:27:35.169Z)
OS version: Windows_NT x64 10.0.18363
Remote OS version: Linux x64 4.15.0-101-generic

bwateratmsft:
Fixes #2047. This will always show the size in MB--there are very few images <1MB and very few >1GB, so it's an easy way to cover as many bases as possible.

With size in the label, sorted alphabetically:
image

With size in label, sorted by size:
image

Can sort by but not group by size:
image
image

Containers/etc. do NOT show size in sorting:
image


test/tree/imagesTree.test.ts
```diff
@@ -11,77 +11,92 @@ const testImages: DockerImage[] = [
     {
         Name: 'a',
         Id: 'sha256:b0648d86f18e6141a8bfa98d4d17d5180aa2699af7f27eac5491fd1f950f6f05',
-        CreatedTime: generateCreatedTimeInSec(2)
+        CreatedTime: generateCreatedTimeInSec(2),
+        Size: 2 * 1024 * 1024,
```
karolz-ms, line 15: Consider changing the test data so that the image sizes are not increasing throughout (and the data is not already sorted) 😁

PR #2172, opened 2020-07-17
Use recommended EXEC form of commands

### LINKED ISSUE #2090: Use exec form of the CMD Dockerfile command in files we scaffold
This is Docker best practice https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#cmd that ensures signals are properly delivered to the main container process and eliminates shell dependency

https://github.com/microsoft/vscode-docker/pull/2089 highlighted a problem in the Node.js Dockerfiles; there might be other places that need changing

bwateratmsft:
Fixes #2090

NOTE: the "Other" scaffolding still uses the other form (`CMD /usr/games/fortune -a | cowsay`). I don't think there's value in changing that for two reasons--
1. The "Other" Dockerfile is little more than a demonstration of how Dockerfiles look.
1. The intended behavior of the image does not work when using the EXEC form + `sh`, `-c`. It is supposed to output a random bit of text (the output from `/usr/games/fortune`) and put it into a text box being spoken by the Docker whale (the pipe into `cowsay`). If the EXEC form is used, nothing is output to the terminal doing the `docker run`; instead the output is just into the container itself.

### DIFF
--- a/src/configureWorkspace/configure.ts
+++ b/src/configureWorkspace/configure.ts
@@ -28,6 +28,7 @@ export interface PackageInfo {
     author: string;
     version: string;
     artifactName: string;
+    main?: string;
 }
 
 interface JsonPackageContents {
@@ -156,7 +157,8 @@ function getDefaultPackageInfo(): PackageInfo {
         cmd: ['npm', 'start'],
         author: 'author',
         version: '0.0.1',
-        artifactName: ''
+        artifactName: '',
+        main: 'index.js',
     };
 }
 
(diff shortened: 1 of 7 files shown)


karolz-ms:
Regarding the "Other" Dockerfile, the EXEC form works for me:

```dockerfile
FROM docker/whalesay:latest
LABEL Name=scaffold Version=0.0.1
RUN apt-get -y update && apt-get install -y fortunes
CMD ["sh", "-c", "/usr/games/fortune -a | cowsay"]
```

```shell
$ docker run --rm -it  scaffold:latest
 ______________________________________ 
/ "Intelligence without character is a \
| dangerous thing."                    |
|                                      |
\ -- G. Steinem                        /
 -------------------------------------- 
    \
     \
      \     
                    ##        .            
              ## ## ##       ==            
           ## ## ## ##      ===            
       /""""""""""""""""___/ ===        
  ~~~ {~~ ~~~~ ~~~ ~~~~ ~~ ~ /  ===- ~~~   
       \______ o          __/            
        \    \        __/             
          \____\______/   
```

bwateratmsft:
> Regarding the "Other" Dockerfile, the EXEC form works for me:

Weird, I did exactly what you're doing and it did not work for me. I'll try again.

bwateratmsft:
> > Regarding the "Other" Dockerfile, the EXEC form works for me:
> 
> Weird, I did exactly what you're doing and it did not work for me. I'll try again.

Yup, this time it's working. *shrug*

PR #2173, opened 2020-07-17
Use tasks to execute things instead of terminal windows

### LINKED ISSUE #251: suggestion: option to reuse terminals

bwateratmsft:
Fixes #251. This also requires doc updates for command customization.


src/utils/executeAsTask.ts
```diff
@@ -0,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { IActionContext } from 'vscode-azureextensionui';
+import { addDockerSettingsToEnv } from './addDockerSettingsToEnv';
+
+export async function executeAsTask(context: IActionContext, command: string, name: string, addDockerEnv: boolean, workspaceFolder?: vscode.WorkspaceFolder, cwd?: string): Promise<vscode.TaskExecution> {
```
karolz-ms, line 10: Nit: just use an Enum, or string literal type for `addDockerEnv` parameter. Then you don't need to put a comment everywhere to remind ourselves what the `true` or `false` value stands for.

src/utils/spawnAsync.ts
```diff
@@ -121,7 +121,7 @@ export async function execAsync(command: string, options?: cp.ExecOptions, progr
 }
 
 export function bufferToString(buffer: Buffer): string {
-    // Node.js treats null bytes as part of the length, which makes everything mad
-    // There's also a trailing newline everything hates, so we'll remove
-    return buffer.toString().replace(/\0/g, '').replace(/\r?\n$/g, '');
+    // Remove non-printing control characters and trailing newlines
```
karolz-ms, line 124: nice

PR #2176, opened 2020-07-21
Make command templates context type aware

### LINKED ISSUE #2168: Compose up / down commands will not work in ACI context

bwateratmsft:
Fixes #2168. Template commands are made context-aware, so that different commands can match different contexts.

With the old behavior, the order is as follows:
1. Look for any template with a match property that matches
1. Look for any template without a match property
1. Use the default template

With this new behavior, the order is as follows (more or less the same as above):
1. Look for any template whose context type matches AND with a match property that matches
1. Look for any template whose context type matches AND without a match property
1. Use the default template whose context type matches

If no `contextType` property is explicitly specified on any template, it is treated as `'all'` and will apply in any context. `'legacy'` currently matches only `'moby'` contexts, and `'new'` currently matches only `'aci'` contexts (though it is likely there will be more in the future).

TODO: create an issue or PR to update the docs.


bwateratmsft:
@karolz-ms and I discussed an alternative approach to `new` / `aci` / `legacy` / `all`:

1. Only `aci` and `moby` would be defined, right now (same as what is defined for `ContextType`).
1. An undefined `contextType` value would be equivalent to "all". Today, it would mean `moby` or `aci`, but if a new context type `abc` appears, it would still match this and work for it.
1. The existing commands except compose up/down would have `contextType` undefined.
1. The _new_ compose up/down would have `contextType` undefined, and the _old_ ones would have `contextType == moby`.
1. Context type could be defined as an "or", i.e. `moby|aci` would mean either `moby` or `aci` (today, this is all possible values, but that may not be true in the future). For example, if a new context type `abc` appears, then a command with `contextType == moby|aci` _would not_ match it, but one with `contextType == undefined` _would_ match it.
1. Template preference would go similar to the existing `match` property, preferring, in order:
    1. Settings-defined templates with either `match` or `contextType`, where all specified properties match.
    1. Settings-defined templates with neither `match` nor `contextType`.
    1. Hard-coded default templates (only one should ultimately match, to avoid prompting).

src/commands/selectCommandTemplate.ts
```diff
@@ -88,53 +98,68 @@ export async function selectComposeCommand(context: IActionContext, folder: vsco
...
+    const hardcodedTemplates = defaultCommandTemplates[command];
 
-        return false;
-    });
+    // Build the template selection matrix. Settings-defined values are preferred over hardcoded, and constrained over unconstrained.
+    // Constrained templates have either `match` or `contextTypes`, and must match the constraints.
+    // Unconstrained templates have neither `match` nor `contextTypes`.
+    const templateMatrix: CommandTemplate[][] = [];
+
+    // 0. Settings-defined templates with either `match` or `contextTypes`, that satisfy the constraints
```
karolz-ms, line 129: Kudos for explaining the design/intention in the comments!! ❤️

src/commands/selectCommandTemplate.ts
```diff
@@ -159,3 +184,47 @@ async function quickPickTemplate(context: IActionContext, templates: CommandTemp
 
     return selection.data;
 }
+
+function getConstrainedTemplates(templates: CommandTemplate[], matchContext: string[], currentContextType: ContextType): CommandTemplate[] {
+    return templates.filter(template => {
+        if (!template.contextTypes && !template.match) {
+            // If neither contextTypes nor match is defined, this is an unconstrained template
```
karolz-ms, line 191: Add `isConstrainedTemplate` property to `CommandTemplate`? You could use it here and in `getUnconstrainedTemplates` method below

bwateratmsft: I'm not sure I follow. If I added such a property I would basically have to apply the same logic in these methods in order to set it; wouldn't it be about the same amount of code?

bwateratmsft: (If `CommandTemplate` was a class I could easily use a `get` method but it's just an interface...)

karolz-ms: OK never mind then

PR #2193, opened 2020-07-24
Fix context type issue

### LINKED ISSUE #2190: Fail to show the Docker assets with error "This view is not supported in the current Docker context"
**OS:** Win 10/Mac
**Build Version:** 20200723.12
**Regression:** Previous Build 

**Repro Steps:**
1. Install the Docker extension.
2. Navigate to "Docker View" and check the Docker assets.

**Expect:**
All Docker assets show well.

**Actual:**
Fail to show the Docker assets with an error "This view is not supported in the current Docker context".
image

**More Info:**
1. This issue reproduces if there is only one default Docker context.
2. This issue does not reproduce if there are several Docker contexts.
3. This issue does not reproduce on the previous build 20200723.3.

bwateratmsft:
Fixes #2190

### DIFF
--- a/src/docker/ContextManager.ts
+++ b/src/docker/ContextManager.ts
@@ -38,6 +38,7 @@ const defaultContext: Partial<DockerContext> = {
     Id: 'default',
     Name: 'default',
     Description: 'Current DOCKER_HOST based configuration',
+    Type: 'moby',
 };
 
 // These contexts are used by external consumers (e.g. the "Remote - Containers" extension), and should NOT be changed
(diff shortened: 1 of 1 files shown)


src/docker/ContextManager.ts
```diff
@@ -38,6 +38,7 @@ const defaultContext: Partial<DockerContext> = {
     Id: 'default',
     Name: 'default',
     Description: 'Current DOCKER_HOST based configuration',
+    Type: 'moby',
```
bwateratmsft, line 41: The real bug was at line 202, the lack of setting `Type: 'moby'`. Setting it in the default context here makes sense since a `DOCKER_HOST`-based configuration is always `'moby'`, and it removes the need for the same line at line 250 below.

Issue #722, opened 2018-12-30
Allow usage without gnome-keyring installed


fedoraJ:
Repro steps:
When I attempt to add a private registry that is hosted on my local (home lab) network I get the following error. The private registry has no authentication and I can see it is giving a 200 OK response to VSCode. 

Action: vscode-docker.connectCustomRegistry
Error type: Error
Error Message: The name org.freedesktop.secrets was not provided by any .service files

Version: 0.4.0
OS: linux (Fedora 29 KDE spin)

This issue is related to #470 in that you can work around it by installing gnome-keyring. However, I have been purging everything gnome off of my system. Gnome-keyring was removed to force application to use KWallet instead. So, no I will **not** install it. Nor am I going to badger anyone about supporting KWallet. Rather I am asking that instead of throwing an error when it can't connect to libsecret, it prompts the user for a username and password when one is required (my private registry has none, it's not accessible outside my network.)

StephenWeatherford:
Currently on vacation.

StephenWeatherford:
We'll take a look, thanks.

StephenWeatherford:
Also see PR comments in https://github.com/Microsoft/vscode-docker/pull/819#discussion_r263573344

PR #2220, opened 2020-08-06
Allow authentication to registries without gnome-keyring

karolz-ms:
Fixes https://github.com/microsoft/vscode-docker/issues/722

It took me some time to convince myself that this (ignoring keytar errors) is both sufficient and safe to fix the issue above, but at the end I think it is.


karolz-ms:
Thanks, Ravi! 😄

Issue #2231, opened 2020-08-11
Fail to execute command "Compose Restart" after executing command "Compose Up"


v-mengwe:
**OS:** Win10/Mac
**Build Version:** 20200807.6

**Repro Steps:**

1. Create an Express App project and open it in VS Code.
2. Command Palette -> Add Docker files to workspace -> Node.js -> Enter port.
3. Right click docker-compose.yml/docker-compose.debug.yml -> Select "Compose Up".
4. A new running container appears in the Containers tree view.
5. Right click docker-compose.yml/docker-compose.debug.yml -> Select "Compose Restart".
6. Check the result.

**Expect:**
Execute command "Compose Restart" successfully without any error.

**Actual:**
Execute command "Compose Restart" failed with an error.
![image](https://user-images.githubusercontent.com/62053665/89871206-9ad90c00-dbe9-11ea-8f1c-0fc447580a57.png)

**More Info:**
This issue also reproduces for a .NET Core MVC project.

karolz-ms:
I can reproduce. Investigating...

karolz-ms:
What we are *trying* to do in our code is to do `compose down`, wait for that to finish, then do `compose up`. However, as we now run commands as tasks, what we are doing to wait for the task execution to finish, actually does not work. There is simply no easy way to wait for a task to finish.  https://github.com/microsoft/vscode/issues/98483 Investigating alternatives...

karolz-ms:
Fix available. This is good bug; affects more than just this scenario and could result in unpredictable errors/race conditions. Good job, @v-mengwe !

PR #2234, opened 2020-08-11
Wait for task execution properly

karolz-ms:
Fixes https://github.com/microsoft/vscode-docker/issues/2231

### DIFF
--- a/src/utils/executeAsTask.ts
+++ b/src/utils/executeAsTask.ts
@@ -37,8 +37,9 @@ export async function executeAsTask(context: IActionContext, command: string, na
     const taskExecution = await vscode.tasks.executeTask(task);
 
     const taskEndPromise = new Promise<void>((resolve) => {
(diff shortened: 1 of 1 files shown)


src/utils/executeAsTask.ts
```diff
@@ -34,5 +34,15 @@ export async function executeAsTask(context: IActionContext, command: string, na
         task.definition.idRandomizer = Math.random();
     }
 
-    return vscode.tasks.executeTask(task);
+    const taskExecution = await vscode.tasks.executeTask(task);
+
+    const taskEndPromise = new Promise<void>((resolve) => {
+        vscode.tasks.onDidEndTaskProcess(e => {
```
philliphoff, line 40: The event subscription returns a `Disposable` that should be invoked prior to resolving, else this may get continued to get called after the task is done.

karolz-ms: Good catch, thank you!

PR #2240, opened 2020-08-17
Update version and dependencies

karolz-ms:
(starting 1.6.0 cycle)


karolz-ms:
Thanks, @ravipal ! ☺️

PR #2255, opened 2020-08-26
Disable NPS survey for now

bwateratmsft:
For slight performance gain we can disable the NPS survey since we do not have the experiment enabled at this time.

@BigMorty Are you OK removing this? Here's the implications--
1. Slight performance improvement
1. We can't re-enable this NPS survey from server side

NOTE: For the next NPS survey, we will need to make a code change anyway--to advance the survey ID from `nps2` to `nps4` (I skipped 3 because we called this one "NPS 3" in the experimentation service).


BigMorty:
I am fine removing this.

PR #2265, opened 2020-08-31
Add ability to deploy from ACR / Docker Hub to ACI

### LINKED ISSUE #1718: Add ability to deploy an image in ACR (or other registry?) to Azure Container Instances (ACI)
Similar to how I can click on an image in ACR and deploy it to Azure App Service, I would like to be able to deploy it to ACI as well.

bwateratmsft:
Fixes #1718. Also implements a few of the now-available APIs in the new Docker Serve client.

### DIFF
--- a/src/commands/containers/aciContainerActionWarning.ts
+++ b/src/commands/containers/aciContainerActionWarning.ts
@@ -1,0 +1,23 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DialogResponses, IActionContext } from 'vscode-azureextensionui';
+import { ext } from '../../extensionVariables';
+import { localize } from '../../localize';
+import { ContainerTreeItem } from '../../tree/containers/ContainerTreeItem';
+
+export async function aciContainerActionWarning(context: IActionContext, nodes: ContainerTreeItem[]): Promise<string[]> {
+    if ((await ext.dockerContextManager.getCurrentContext()).Type !== 'aci') {
+        return nodes.map(n => n.containerId);
+    }
+
+    const confirm = localize('vscode-docker.commands.containers.aciContainerActionWarning.confirm', 'ACI containers can only be started, stopped, or removed in a group. This action will apply to all containers in \'{0}\'. Do you want to proceed?');
+
+    // No need to check result - cancel will throw a UserCancelledError
+    await ext.ui.showWarningMessage(confirm, { modal: true }, DialogResponses.yes);
+
+    // TODO
+    return [foobarbaz];
+}
--- a/src/commands/containers/aciContainerActionWarning.ts
+++ b/src/commands/containers/aciContainerActionWarning.ts
@@ -6,18 +6,31 @@
 import { DialogResponses, IActionContext } from 'vscode-azureextensionui';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
+import { getComposeProjectName, NonComposeGroupName } from '../../tree/containers/ContainersTreeItem';
 import { ContainerTreeItem } from '../../tree/containers/ContainerTreeItem';
 
 export async function aciContainerActionWarning(context: IActionContext, nodes: ContainerTreeItem[]): Promise<string[]> {
-    if ((await ext.dockerContextManager.getCurrentContext()).Type !== 'aci') {
+    if ((await ext.dockerContextManager.getCurrentContext()).Type !== 'aci' ||
+        nodes.every(n => getComposeProjectName(n.containerItem) === NonComposeGroupName)) {
+        // If we're not in an ACI context, or every node in the list is not part of any ACI container group, return unchanged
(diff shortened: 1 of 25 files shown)


src/commands/containers/aciContainerActionWarning.ts
```diff
@@ -0,0 +1,36 @@
...
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DialogResponses, IActionContext } from 'vscode-azureextensionui';
+import { ext } from '../../extensionVariables';
+import { localize } from '../../localize';
+import { getComposeProjectName, NonComposeGroupName } from '../../tree/containers/ContainersTreeItem';
+import { ContainerTreeItem } from '../../tree/containers/ContainerTreeItem';
+
+export async function aciContainerActionWarning(context: IActionContext, nodes: ContainerTreeItem[]): Promise<string[]> {
```
karolz-ms, line 12: Nit: the name of this function could be better, as it is not just about showing a warning to the user; it also returns all containers affected by the operation. Consider `confirmAllAffectedContainers()` or something like that

src/commands/registries/azure/deployImageToAci.ts
```diff
@@ -0,0 +1,83 @@
...
+        await executeAsTask(context, command, title, options);
+    }
+}
+
+async function getImagePorts(fullTag: string): Promise<number[]> {
+    try {
+        const result: number[] = [];
+
+        // 1. Pull the image to the default context
+        await execAsync(`docker --context default pull ${fullTag}`);
```
karolz-ms, line 64: If the user did not have the image in the local image store, the act of deploying it to ACI will have a side effect to putting it also in the local image store. Is that acceptable, or should we try to clean it up?

src/commands/registries/azure/deployImageToAci.ts
```diff
@@ -0,0 +1,83 @@
...
+export async function deployImageToAci(context: IActionContext, node?: RemoteTagTreeItem): Promise<void> {
+    if (!node) {
+        node = await ext.registriesTree.showTreeItemPicker<RemoteTagTreeItem>([registryExpectedContextValues.dockerHub.tag, registryExpectedContextValues.dockerV2.tag], context);
+    }
+
+    const aciContext = await ext.contextsTree.showTreeItemPicker<ContextTreeItem>([/aciContext;/i], context);
+
+    // Switch to the other context if needed
+    if (!aciContext.current) {
+        await vscode.commands.executeCommand('vscode-docker.contexts.use', aciContext);
```
karolz-ms, line 26: Since we specify the ACI context explicitly for `docker run`, this is not strictly necessary. I assume this is done so that the user gets visual feedback that their deployment action succeeded.

src/commands/registries/azure/deployImageToAci.ts
```diff
@@ -0,0 +1,83 @@
...
+        title: localize('vscode-docker.commands.registries.deployImageToAci.gettingPorts', 'Determining ports from image...'),
+    };
+    const ports = await vscode.window.withProgress(progressOptions, async () => {
+        return getImagePorts(node.fullTag);
+    });
+    const portsArg = ports.map(port => `-p ${port}:${port}`).join(' ');
+
+    addImageTaggingTelemetry(context, node.fullTag, '');
+
+    const command = `docker --context ${aciContext.name} run -d ${portsArg} ${node.fullTag}`;
```
karolz-ms, line 44: Since we set the chosen ACI context as the current one, is `--context` option necessary here?

karolz-ms (approved):
Looks great, just a few clarifying questions

src/commands/containers/aciContainerActionWarning.ts
bwateratmsft: Agree, I never liked the name of this method. Good suggestion.

src/commands/registries/azure/deployImageToAci.ts
bwateratmsft: I don't think there's any particularly good way to find out if it was already there (thus we should not remove it), or was not (and maybe we should remove it). It could already be there but out of date, in which case we should probably update it before inspecting ports--it's a whole can of worms. I'd be inclined to say it's fine to just leave it behind.

src/commands/registries/azure/deployImageToAci.ts
bwateratmsft: Exactly. @BigMorty asked specifically to have it switch to the ACI context. Technically the `--context` is no longer necessary on the `docker run` command, but, since we're doing things in multiple contexts already I thought it was safe to just leave it there.

src/commands/registries/azure/deployImageToAci.ts
bwateratmsft: (See also above). It's not technically necessary but also not harmful. We're doing things in multiple contexts already so I'm inclined to leave it, but I don't feel strongly.

karolz-ms: I am fine with that

src/commands/registries/azure/deployImageToAci.ts
karolz-ms: Sounds good

PR #2331, opened 2020-09-17
Remove coreclr debug

### LINKED ISSUE #2197: Remove CoreCLR debug
The legacy .NET debug has been deprecated for a while now with warnings to anyone still using it; I think it's fine to remove at this point. Should look at refactoring what is still in use (e.g. OS provider, vsdbg downloader, etc.) and removing what is not.

bwateratmsft:
Fixes #2197

Tested so far:
- [x] .NET Core debugging (single) (Mac & Windows)
- [x] .NET attach + compose up (Mac & Windows)
- [x] .NET attach + compose debug up (Mac & Windows)
- [x] VsDbg not present
- [x] SSL cert not present
- [x] Python debug (sanity test of browser launch)
- [x] Node debug (sanity test of browser launch)
- [x] Windows debug


src/commands/dockerInstaller.ts
```diff
@@ -124,14 +119,17 @@ export class MacDockerInstaller extends DockerInstallerBase {
...
 export async function showDockerInstallNotification(): Promise<void> {
-    const installMessageLinux = localize('vscode-docker.commands.dockerInstaller.installDockerInfo', 'Docker is not installed. Would you like to learn more about installing Docker?');
-    const installMessageNonLinux = localize('vscode-docker.commands.dockerInstaller.installDocker', 'Docker Desktop is not installed. Would you like to install it?');
-    const learnMore = localize('vscode-docker.commands.dockerInstaller.learnMore', 'Learn more')
-    const install = localize('vscode-docker.commands.dockerInstaller.install', 'Install')
-    const osProvider: LocalOSProvider = new LocalOSProvider();
-    const dockerInstallMessage = osProvider.os === 'Linux' ? installMessageLinux : installMessageNonLinux;
-    const confirmationPrompt: vscode.MessageItem = osProvider.os === 'Linux' ? { title: learnMore } : { title: install };
-    const response = await vscode.window.showInformationMessage(dockerInstallMessage, ...[confirmationPrompt]);
+    const isLinux = os.platform() !== 'win32' && os.platform() !== 'darwin';
```
karolz-ms, line 122: `osUtils.isLinux()` ?

src/commands/installDocker.ts
```diff
@@ -3,17 +3,15 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as os from 'os';
```
karolz-ms, line 6: Why not osUtils?

karolz-ms (approved):

src/commands/dockerInstaller.ts
bwateratmsft: That'd be fine. Need to ensure `osUtils` uses `os` underneath; anything else could have problems with VSCode remoting.

karolz-ms: I just like the idea of having the logic of "are we on Linux/Unix?" or "are we on Windows ver X or newer" to be in one place. That way it is easy to fix it as necessary.

bwateratmsft: I cleaned it up here and elsewhere.

Issue #2313, opened 2020-09-14
Django project debug in docker container doesn't start


jmejia32:
Redirected from https://github.com/microsoft/vscode/issues/106561#issue-700425657

In vscode, when I start the debug in my local computed (no docker), debug starts completely but when I start debug with docker, the container builds and starts while vscode ui waiting for connection and never "starts" debugging process.

bwateratmsft:
@jmejia32 Can you share the project you are encountering this issue in? Or at least the Docker and docker-compose files, tasks.json, and launch.json?

Also, can you share the logs from the following places:
- Output
  - Docker
  - Python
- Terminal
  - `docker-build` task
  - `docker-run` task
  - Python Debug Console
- Debug Console

Debugging in Docker is done differently; the app is run by running Debugpy in the container, which is supposed to connect back to the host. It sounds like Debugpy may not be able to connect back to the host and so VSCode sits there waiting for the connection. If this is the case the "Python Debug Console" output probably has the most important info.

jmejia32:
@bwateratmsft ok, here we go:

* Output
  * Docker: Nothing
  * Python:
```
Python interpreter path: ~/.local/pyenvs/django_env/bin/python
Starting Pylance language server.
> ~/.local/pyenvs/django_env/bin/python ~/.vscode/extensions/ms-python.python-2020.8.108011/pythonFiles/pyvsc-run-isolated.py -c "import notebook"
> ~/.local/pyenvs/django_env/bin/python ~/.vscode/extensions/ms-python.python-2020.8.108011/pythonFiles/pyvsc-run-isolated.py -c "import jupyter"
> conda --version
> pyenv root
> pyenv root
> pyenv root
> pyenv root
> pyenv root
> pyenv root
> pyenv root
> pyenv root
> pyenv root
```
* Terminal
  * `docker-build` task
  ```json
  {
      "type": "docker-build",
      "label": "docker-build",
      "platform": "python",
      "dockerBuild": {
        "tag": "talentumplus:latest",
        "dockerfile": "${workspaceFolder}/Dockerfile",
        "context": "${workspaceFolder}",
        "pull": true
      }
    },
  ```
  * `docker-run` task
  ```json
  {
      "type": "docker-run",
      "label": "docker-run: debug",
      "dependsOn": [
        "docker-build"
      ],
      "dockerRun": {
        "volumes": [
          {
            "containerPath": "/app", "localPath": "${workspaceFolder}"
          }
        ],
[...]

PR #2341, opened 2020-09-22
Fix tests silently failing

### LINKED ISSUE #2302: Tests are failing silently
Tests have been failing but are doing so silently. e.g. https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=25608

bwateratmsft:
Fixes #2302. Also unrelatedly, slightly improves the fix for #2313.

### DIFF
--- a/src/debugging/python/PythonDebugHelper.ts
+++ b/src/debugging/python/PythonDebugHelper.ts
@@ -94,7 +94,7 @@ export class PythonDebugHelper implements DebugHelper {
                 removeContainerAfterDebug: debugConfiguration.removeContainerAfterDebug
             },
             debugLauncherPath: debugConfiguration.debugLauncherPath || launcherPath,
-            debugAdapterHost: debugConfiguration.debugAdapterHost || await this.getDebugAdapterHost(context),
+            debugAdapterHost: debugConfiguration.debugAdapterHost || await this.getDebugAdapterHost(context) || '172.17.0.1',
             console: debugConfiguration.console || "integratedTerminal",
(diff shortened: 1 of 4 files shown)


src/debugging/python/PythonDebugHelper.ts
```diff
@@ -94,7 +94,7 @@ export class PythonDebugHelper implements DebugHelper {
                 removeContainerAfterDebug: debugConfiguration.removeContainerAfterDebug
             },
             debugLauncherPath: debugConfiguration.debugLauncherPath || launcherPath,
-            debugAdapterHost: debugConfiguration.debugAdapterHost || await this.getDebugAdapterHost(context),
+            debugAdapterHost: debugConfiguration.debugAdapterHost || await this.getDebugAdapterHost(context) || '172.17.0.1',
```
karolz-ms, line 97: What is this IP address, why is this one a good default?

bwateratmsft: This is the IP address that represents the host within the container, on the `bridge` network. So far this is the only one I've ever seen--same address on my Windows desktop, Macbook, Linux VM, and multiple customers' Linux machines as well.

Their [docs](https://docs.docker.com/network/network-tutorial-standalone/) explicitly mention that address as well (for `bridge`), but it may be more as an example than a prescriptive statement.

karolz-ms: Roger. Internet poking suggests 172.17.0.0/16 is the default bridge network address space. I would just put a comment on that in the code then

karolz-ms (approved):

PR #2344, opened 2020-09-23
Use ENV key=value in Dockerfile templates

StefanScherer:
Hi,
In our docs.docker.com we've removed the examples that used the old syntax `ENV key value` and replaced it with the better `ENV key=value` syntax (related https://github.com/docker/docker.github.io/pull/11414). 
The old style had its problems when using multiple environment variables `ENV ONE TWO= THREE=world`

Just yesterday I used the Docker extension in preparation for a Docker beginners video course and saw that the Docker extension uses the old style.

This PR changes the templates to the new style and helps users to use the good practice.

### DIFF
--- a/resources/templates/node/Dockerfile.template
+++ b/resources/templates/node/Dockerfile.template
@@ -1,5 +1,5 @@
 FROM node:12.18-alpine
-ENV NODE_ENV production
+ENV NODE_ENV=production
 WORKDIR /usr/src/app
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):
Looks good to me, thanks @StefanScherer!

thaJeztah:
Thanks! FWIW (for people arriving here); both syntaxes will continue to be supported, but the syntax without `=` can be somewhat ambiguous, and it's easy to make mistakes with that format, so we're trying to discourage its use.

PR #2348, opened 2020-09-25
Upgrade to TypeScript 4

### LINKED ISSUE #2346: Upgrade to TypeScript 4
Currently we are on TypeScript 3.9.7 but the latest available at the time of writing is 4.0.3. Unfortunately we are blocked from upgrading due to [this change](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#properties-overriding-accessors-and-vice-versa-is-an-error), which makes many of the accessors in the tree view code errors. ESLint already complains about this within VSCode although the linting task in AzPipelines does not.

~~Need to start by fixing the issue upstream in `vscode-azureextensionui`, and then we can fix it here if need be. Blocked on https://github.com/microsoft/vscode-azuretools/issues/805.~~

bwateratmsft:
Fixes #2346. I found a somewhat less-intrusive way to satisfy (trick) the TypeScript compiler; it depends largely on having an intermediate abstract class that redefines the inherited parent property as an abstract, allowing inheriting classes to choose for it to be either a property or an accessor.

This also deals with some linter issues that came up.


karolz-ms (approved):
Great job, Brandon, this workaround is totally manageable

PR #2353, opened 2020-09-28
Add compose grouping as default, add compose down/restart commands to group

### LINKED ISSUE #2304: Allow compose down / compose restart on grouped-by-compose nodes
We can use the `com.docker.compose.project.config_files` and `com.docker.compose.project.working_dir` labels to locate the CWD to use and the files to add.

### LINKED ISSUE #2324: Group containers by Compose invocation "project" by default
Docker Desktop is doing that already. Given how widely used Compose is it probably make sense to have Compose be the default grouping for containers pane

bwateratmsft:
Fixes #2304, fixes #2324. The default grouping for containers is now compose project instead of "None". Commands for compose down / restart are added.

NOTE: the command template version of `docker-compose down` is _not_ used here. The difficulty of doing so would have been very high.

Nodes are expanded by default, and remain in the same state as they were before if refresh is clicked (so manually collapsed stay collapsed):
image

Commands:
image
image


package.nls.json
```diff
@@ -188,6 +188,8 @@
     "vscode-docker.commands.containers.start": "Start",
     "vscode-docker.commands.containers.stop": "Stop",
     "vscode-docker.commands.containers.viewLogs": "View Logs",
+    "vscode-docker.commands.containers.composeGroup.restart": "Compose Restart",
```
bwateratmsft, line 191: TODO: these strings good?

src/tree/containers/ContainersTreeItem.ts
```diff
@@ -124,6 +124,10 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<DockerContainer, C
 export const NonComposeGroupName = localize('vscode-docker.tree.containers.otherContainers', 'Other Containers');
```
bwateratmsft, line 124: TODO: should this change?

package.json
```diff
@@ -374,6 +384,16 @@
                     "when": "view == dockerContainers && viewItem =~ /container$/i",
                     "group": "containers_2_destructive@1"
                 },
+                {
+                    "command": "vscode-docker.containers.composeGroup.restart",
+                    "when": "view == dockerContainers && viewItem =~ /composeGroup$/i && !vscode-docker:aciContext",
```
bwateratmsft, line 389: These commands show for the "Other containers" group which doesn't make sense

package.nls.json
karolz-ms: Good enough IMO ☺️

src/commands/containers/composeGroup.ts
```diff
@@ -0,0 +1,45 @@
...
+    return composeGroup(context, 'restart', node);
+}
+
+export async function composeGroupDown(context: IActionContext, node: ContainerGroupTreeItem): Promise<void> {
+    return composeGroup(context, 'down', node);
+}
+
+async function composeGroup(context: IActionContext, composeCommand: 'restart' | 'down', node: ContainerGroupTreeItem): Promise<void> {
+    const workingDirectory = getComposeWorkingDirectory(node);
+    const filesArgument = getComposeFiles(node)?.map(f => `-f ${f}`)?.join(' ');
```
karolz-ms, line 23: These file names should probably be quoted, to account for cases like spaces in the name, or the file name beginning with a dash 🙄

src/tree/containers/ContainersTreeItem.ts
karolz-ms: Standalone Containers?

karolz-ms (approved):

src/commands/containers/composeGroup.ts
bwateratmsft: Hm, is `'` reliable in all shells or should we use `"`? Or is that also bust?

karolz-ms: I believe the rule of thumb is single quote on Mac/Linux; double quote on Windows https://github.com/microsoft/vscode-docker/blob/master/src/commands/dockerInstaller.ts#L85

package.nls.json
bwateratmsft: @BigMorty What do you think? These strings are what show in the context menu.

src/tree/containers/ContainersTreeItem.ts
bwateratmsft: @BigMorty What do you think? This string is the header for the grouping of non-compose containers.

bwateratmsft: We talked and decided on "Individual Containers" at least for now.

PR #2354, opened 2020-09-28
Include platform in container list result

bwateratmsft:


bwateratmsft:
Actually Dockerode doesn't have platform on the list API so we don't need this at this point.

PR #2355, opened 2020-09-28
Add `docker.context` setting and support DOCKER_CONTEXT

### LINKED ISSUE #2264: Add a docker.context setting for setting DOCKER_CONTEXT similar to docker.host
Add a `docker.context` setting for setting `DOCKER_CONTEXT` similar to `docker.host` for setting `DOCKER_HOST`. This will allow using docker contexts per window or project rather than globally.

This should probably also be supported by the Remote - Containers extension.

bwateratmsft:
Fixes #2264

Tested:
- [x] Multiple ACI contexts open (using `DOCKER_CONTEXT`)
- [x] ACI and regular contexts open (using `DOCKER_CONTEXT`)
- [x] Multiple ACI contexts open (using `docker.context`)
- [x] ACI and regular contexts open (using `docker.context`)
- [x] Multiple contexts open (mixing `DOCKER_CONTEXT` and `docker.context`)
- [x] Verify `DOCKER_HOST` takes precedence over any context
- [x] Verify `docker.host` takes precedence over any context
- [x] Verify `docker.context` takes precedence over `DOCKER_CONTEXT`
- [x] Check behavior if `docker.context` / `DOCKER_CONTEXT` references an invalid context (compare behavior to CLI)
- [x] Check behavior if only one context exists
- [x] No changing context if `DOCKER_HOST` / `docker.host` / `DOCKER_CONTEXT` / `docker.context` are set


src/docker/ContextManager.ts
```diff
@@ -180,34 +180,40 @@ export class DockerContextManager implements ContextManager, Disposable {
...
                     Current: true,
-                    DockerEndpoint: dockerHost,
+                    DockerEndpoint: dockerHostEnv,
                 } as DockerContext];
             }
 
             // No value for DOCKER_HOST, and multiple contexts exist, so check them
             const result: DockerContext[] = [];
-            const { stdout } = await execAsync('docker context ls --format="{{json .}}"', ContextCmdExecOptions);
+            const { stdout } = await execAsync('docker context ls --format="{{json .}}"', { ...ContextCmdExecOptions, env: { ...process.env, DOCKER_CONTEXT: dockerContextEnv } });
```
karolz-ms, line 216: This code relies on the fact that `docker context ls --context foo` will report "foo" as current context regardless what the ambient context is. This is exactly what we want, but it is not super obvious. Consider adding a comment.

karolz-ms (approved):

src/docker/ContextManager.ts
bwateratmsft: Good idea. Originally I had the logic where it would list them and if `dockerContextEnv` was set, compare names--but I realized that was redundant because the CLI already has the right behavior from the get go. A comment is definitely a good idea here.

Issue #2358, opened 2020-09-29
Context panel shows "Switching..." UI forever


karolz-ms:
VS Code 1.49.2 on Mac Catalina 10.15.6
Docker Extension 1.6.0

Repro steps
1. Create an ACI context
2. Right-click on the ACI context in the Context pane and choose "Use".

Expected: ACI context becomes the context-in-use and the UI updates accordingly

Actual: the UI remains in "Switching..." state forever

![image](https://user-images.githubusercontent.com/15271049/94621652-91713500-0265-11eb-921a-61cf0ef909f3.png)

This does not reproduce every time, but I have been seeing it at least once a day recently. As far as I can tell the ACI context is active, it is just that the UI somehow does not notice that the switch is complete.

karolz-ms:
Some more observations:
- Refreshing the Contexts panel will "fix" it
- In the "Switching.." state the Container panel remains empty. So it is not just the Contexts panel that is not working--entire Docker view is in an intermediate, non-usable state

bwateratmsft:
I'm guessing something isn't calling back under some circumstances, e.g. after the `docker context use` we aren't getting the callback that it finished.

PR #2360, opened 2020-09-30
Fix possible cause of UI getting stuck saying "Switching"

bwateratmsft:
Might fix the root cause of #2358

I wondered if these events were happening out of order. This way if the `onContextChanged` event fires before control returns to this method, it won't get stuck.


karolz-ms (approved):
This is probably it--`onContextChanged` being raised before we install the handler

PR #2361, opened 2020-09-30
Enable debugging node apps in subfolders

### LINKED ISSUE #2057: Adding Node.js files to workspace breaks when package.json in subfolder
I was following the [Deploy images to ACI Tutorial](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-tutorial-prepare-app) and noticed that containerizing the app as-is was failing. I suspect it is because the package.json is found in a folder named `app`. Even though the docker-build and docker-run tasks don't reference this file, I am still blocked from hitting F5

Repro Steps:
From Terminal
1. `git clone https://github.com/Azure-Samples/aci-helloworld.git` into a folder of choice
1. `code .` to open editor
Within VS Code
1. Delete or Rename current Dockerfile to Dockerfile_original
1. `Ctrl + Shift + P` Add Dockerfile to workspace and select Node.js
1. I am met with this error: 
image
1. After moving the package.json to the root folder, the container now builds, but during run there is the error:
image
1. This can be fixed by changing the `main` property in `package.json` to "./app/index.js" but I'm unsure if this problem would have persisted in a fresh node project scenario

bwateratmsft:
Fixes #2057. Also makes locating files a little more consistent and puts the platform-specific code for that with the platforms themselves.

Tested:
- [x] Node in subfolder (F5, compose up, debug compose up)
- [x] Node not in subfolder (F5, compose up, debug compose up)
- [x] .NET in subfolder (F5, compose up, debug compose up)
- [x] .NET not in subfolder (F5, compose up, debug compose up)
- [x] Python (F5, compose up, debug compose up)
- [x] Some other platform

### DIFF
--- a/src/scaffolding/wizard/netCore/NetCoreGatherInformationStep.ts
+++ b/src/scaffolding/wizard/netCore/NetCoreGatherInformationStep.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as path from 'path';
 import { SemVer } from 'semver';
 import { localize } from '../../../localize';
 import { hasTask } from '../../../tasks/TaskHelper';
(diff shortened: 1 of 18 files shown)


src/scaffolding/wizard/netCore/NetCoreGatherInformationStep.ts
```diff
@@ -70,11 +71,26 @@ export class NetCoreGatherInformationStep extends GatherInformationStep<NetCoreS
             wizardContext.serviceName = getValidImageNameFromPath(wizardContext.artifact);
         }
 
+        if (!wizardContext.dockerBuildContext) {
```
bwateratmsft, line 74: This is redundant since the `super` call would set this. Can probably get rid of this and leave a comment.

karolz-ms (approved):
That looks about right ☺️

PR #2368, opened 2020-10-02
Lazy load as many `@azure` dependencies (and some more) as possible

bwateratmsft:
As it stands right now, Docker is probably the slowest activating of the top ~15 extensions in the marketplace, which I'm not excited about.

This change causes most `@azure` dependencies to be lazy-loaded, along with a few others. These tend to be the biggest dependencies and thus take the most time to load. Overall this results in a dramatic reduction in code loading time during extension activation, especially since most users are not using Azure features.

In addition to `@azure` dependencies, this also lazy-loads:
- Dockerode and the new Docker SDK (though on activation, _one_ of those two is going to get loaded--usually Dockerode--but there's still gains since it's no longer both of them)
- Handlebars

I'd like to also lazy-load `moment` but it is used in several synchronous methods in a way that would be very difficult to make asynchronous.

With this change, the dependency chart in the bundle looks like this:
image

I believe the biggest thing still being loaded non-lazily is `moment`.

@EricJizbaMSFT FYI

Validation testing (which includes validating that the dependencies are actually lazy-loaded):
- [x] Create an ACR
- [x] Deploy to Azure
- [x] Run Azure task
- [x] View Azure task logs
- [x] Scaffold files
- [x] Start w/ ACI context active
- [x] Start w/ regular context active
- [x] Deploy to ACI (not that it should matter)


src/utils/lazyLoad.ts
```diff
@@ -1,45 +0,0 @@
-/*---------------------------------------------------------------------------------------------
```
bwateratmsft, line 1: It seems we never needed this complicated pattern for lazy loading, it Just Works (TM) without all the complexity.

src/docker/ContextManager.ts
```diff
@@ -122,11 +120,15 @@ export class DockerContextManager implements ContextManager, Disposable {
                 // But that probably won't be true in the future, so define both as separate concepts now
                 await this.setVsCodeContext('vscode-docker:aciContext', true);
                 await this.setVsCodeContext('vscode-docker:newSdkContext', true);
-                ext.dockerClient = new DockerServeClient(currentContext);
+
+                const dsc = await import('./DockerServeClient/DockerServeClient');
```
karolz-ms, line 124: Would it make sense to cache the result of this import (and the dockerode one)?

karolz-ms: I think the module system does internal caching... ? 🤔

karolz-ms (approved):
Async imports rock!

src/docker/ContextManager.ts
bwateratmsft: Yeah, it does it internally. I set some logpoints before and after this; the first time it was about 200ms (unpacked), every subsequent time it was 0ms. So no need to cache ourselves.

PR #2370, opened 2020-10-05
Switch off `moment` to much lighter-weight `dayjs`

bwateratmsft:
This moves us away from `moment` (about 700 KB packed) to `dayjs` (about 7 KB packed). The result is a bit smaller final VSIX and improved load times.

The calling convention for `dayjs` is a bit cumbersome but IMO the benefits make it worthwhile.

With `moment` gone, there's very little we could do to further optimize code loading times. Both `dockerode` and `@docker/sdk` have fairly large (and mostly non-overlapping) dependency trees, but at least one of them needs to be loaded to accomplish anything.


karolz-ms:
Moment is essentially deprecated: https://momentjs.com/docs/#/-project-status/

karolz-ms (approved):

bwateratmsft:
> Moment is essentially deprecated: https://momentjs.com/docs/#/-project-status/

Yep, one more reason.

PR #2388, opened 2020-10-12
Use command line builder for Docker login

bwateratmsft:


karolz-ms (approved):
Just FYI I looked at other spots in our code where command execution might need to use the builder and I could not find anything

PR #2394, opened 2020-10-13
Resource group name sugggestion

### LINKED ISSUE #2392: There is no default resource group name if select "+Create new resource group" during new ACI context creation
**OS:** Windows 10
**Build Version:** 20201012.4

**Repro Steps:**
1. Right click on one image in the registries view -> Deploy Image to Azure Container Instances... -> Create new ACI context -> Enter a name for ACI context -> Click "+Create new resource group".
2. Check the default resource group name.

**Expect:**
There is a default resource group name which is the same with ACI context name.

**Actual:**
There is no default resource group name.
image

**More Info:**
This issue does not reproduce when creating a new ACR.

bwateratmsft:
Fixes #2392

### DIFF
--- a/src/tree/contexts/aci/ContextNameStep.ts
+++ b/src/tree/contexts/aci/ContextNameStep.ts
@@ -3,14 +3,20 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


karolz-ms (approved):
Nice, didn't know that the UI library already has all this stuff for RG name creation and validation

PR #2438, opened 2020-10-26
Force refresh of Docker context after 10 seconds

### LINKED ISSUE #2383: [Codespaces] The context keeps "Switching" state after using a different one
**OS:** Windows 10
**Build Version:** 20201009.1

**Repro Steps:**
1. Create a Linux Codespace and connect it.
2. Install Docker extension in this Codespace.
3. Create a new context.
4. Use this new context and check the result.

**Expect:**
Switch to the new context successfully.

**Actual:**
The context keeps "Switching" state after using a different one.
image

bwateratmsft:
Fixes #2383

### DIFF
--- a/src/commands/context/useDockerContext.ts
+++ b/src/commands/context/useDockerContext.ts
@@ -39,7 +39,6 @@ export async function useDockerContext(actionContext: IActionContext, node?: Con
                     // This covers the scenario where the filesystem watchers aren't working (e.g. Codespaces)
                     const timer = setTimeout(() => {
                         disposable.dispose();
-                        clearTimeout(timer);
 
                         void ext.dockerContextManager.refresh();
                         resolve();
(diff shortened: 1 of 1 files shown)


src/commands/context/useDockerContext.ts
```diff
@@ -30,8 +30,20 @@ export async function useDockerContext(actionContext: IActionContext, node?: Con
...
+                        clearTimeout(timer);
+
                         resolve();
                     });
+
+                    // If 10 seconds pass and it hasn't been updated, force a refresh
+                    // This covers the scenario where the filesystem watchers aren't working (e.g. Codespaces)
+                    const timer = setTimeout(() => {
+                        disposable.dispose();
+                        clearTimeout(timer);
```
karolz-ms, line 42: Do we need the call to `clearTimeout` here (when we are already executing the timeout function)?

karolz-ms (approved):
Just one small question.

BTW, I noticed that we call this command in the "deploy to ACI" scenario, so this change takes care of that too, nice

src/commands/context/useDockerContext.ts
bwateratmsft: I suppose not; I was conflating it with `setInterval` which loops by default.

PR #2439, opened 2020-10-26
Improve extension activation time

### LINKED ISSUE #2371: Investigate ways to improve activation time
We've just done a significant amount of improvement to code loading time. We should also see what can be done about activation time.

At the time of writing, for 1.6.0 the avg activation times are 278 ms for Linux, 388 ms for Mac, 541 ms for Windows. I'd like to get Windows (and the rest) below 200 ms if at all possible.

bwateratmsft:
Fixes #2371.

This covers:
- [x] Get Docker ID (this is already a one-time penalty, it's cached from then on)
- [x] Filesystem checks for Docker context paths
- [x] Setting the VSCode contexts
- [ ] Don't wait on Docker context loading at all?? Can that be done without substantial refactoring?

In the most common case (no Docker contexts, therefore don't need to run CLI), the remaining time spent on activation is mostly dominated (~75%) by Dockerode loading. Nevertheless that will stay as-is; it was an improvement to lazy load the Docker client provider.

### DIFF
--- a/src/docker/ContextManager.ts
+++ b/src/docker/ContextManager.ts
@@ -118,14 +118,14 @@ export class DockerContextManager implements ContextManager, Disposable {
             if (isNewContextType(currentContext.Type)) {
                 // Currently vscode-docker:aciContext vscode-docker:newSdkContext mean the same thing
                 // But that probably won't be true in the future, so define both as separate concepts now
-                await this.setVsCodeContext('vscode-docker:aciContext', true);
-                await this.setVsCodeContext('vscode-docker:newSdkContext', true);
+                this.setVsCodeContext('vscode-docker:aciContext', true);
+                this.setVsCodeContext('vscode-docker:newSdkContext', true);
 
                 const dsc = await import('./DockerServeClient/DockerServeClient');
                 ext.dockerClient = new dsc.DockerServeClient(currentContext);
             } else {
-                await this.setVsCodeContext('vscode-docker:aciContext', false);
-                await this.setVsCodeContext('vscode-docker:newSdkContext', false);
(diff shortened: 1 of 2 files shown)


bwateratmsft:
> Don't wait on Docker context loading at all?? Can that be done without substantial refactoring?

I've looked into it, and I think the answer is no. The real need is for `ext.dockerClient` to be set, since the activation reason might be any of the commands that use it. The simplest way would be to make `ext.dockerClient` an `AsyncLazy` but it has dozens of callers so it would be a significant amount of refactoring.

src/docker/ContextManager.ts
```diff
@@ -193,7 +193,7 @@ export class DockerContextManager implements ContextManager, Disposable {
                 actionContext.telemetry.properties.hostSource = 'docker.context';
             } else if ((dockerContextEnv = process.env.DOCKER_CONTEXT)) { // Assignment + check is intentional
                 actionContext.telemetry.properties.hostSource = 'envContext';
-            } else if (!(await fse.pathExists(dockerContextsFolder)) || (await fse.readdir(dockerContextsFolder)).length === 0) {
+            } else if (!fse.pathExistsSync(dockerContextsFolder) || fse.readdirSync(dockerContextsFolder).length === 0) { // Sync is intentionally used for performance
```
karolz-ms, line 196: Kudos for adding this comment! Consider mentioning "activation code path" or something like that for extra clarity.

karolz-ms (approved):

src/docker/ContextManager.ts
bwateratmsft: I had the same thought but didn't end up doing it. I'll add it.

PR #2441, opened 2020-10-26
Open a "Getting Started" page after install/upgrade

### LINKED ISSUE #1895: Show "Getting started" page after installation of extension
This getting started page will help users understand what features/functionality the extension has and should offer next steps and/or shortcuts to features. This should work similar to what the Python extension has done.

### LINKED ISSUE #2343: Link to install Docker in the explorer window
1. Create a Docker context (or do anything that will result in using the CLI during context init)
1. Uninstall Docker
1. Launch VSCode and open a Dockerfile

Expect: No error
Actual:
image
image

### LINKED ISSUE #2440: Disable awareness toast
As part of #1895, we'll disable the awareness toast

bwateratmsft:
Fixes #1895, fixes #2343, fixes #2440. This adds a fancy new start page experience similar to what the Python extension has.

The page is theme-smart (it will follow theme colors). All the icons and headings will execute relevant commands.

The code is intended to show the start page once per minor version update (i.e. it will not show at patch versions). Of course, if they uncheck "Show this page...", it will not show automatically anymore, but the "Docker : Open Start Page" command will _always_ open it.

image

image
(Light theme makes my eyes bleed)


src/commands/startPage/openStartPage.ts
```diff
@@ -0,0 +1,42 @@
...
+    const thisVersion = new semver.SemVer(extensionVersion.value);
+    const diff = semver.diff(thisVersion, lastVersion);
+
+    if (!vscode.workspace.getConfiguration('docker').get('showStartPage', false)) {
+        // Don't show: disabled by settings
+        return;
+    } else if (!/^en(-us)?$/i.test(vscode.env.language)) {
+        // Don't show: this page is English only
+        return;
+    } else if (semver.lte(thisVersion, lastVersion) || diff === 'prepatch' || diff === 'patch' || diff === 'prerelease') {
```
bwateratmsft, line 31: I'd like to see this code factored out into a method and unit tests written for it.

resources/startPage.html.template
```diff
@@ -0,0 +1,177 @@
...
+
+<body>
+    <div class="container">
+        <div class="row" id="titleBar">
+            <div class="cell">
+                <a href="https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker"><img src="{{ dockerIconUri }}" /></a>
+            </div>
+            <div class="cell">
+                <a href="https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker" class="linkheading">
+                    <h1>Docker</h1>
```
danegsta, line 64: Purely subjective, but the default h1 font size feels a little small next to the large Docker icon in the header

src/commands/startPage/StartPage.ts
```diff
@@ -0,0 +1,83 @@
...
+
+            const listener = this.activePanel.webview.onDidReceiveMessage(async (message: WebviewMessage) => this.handleMessage(message));
+
+            this.activePanel.onDidDispose(() => {
+                this.activePanel = undefined;
+                listener.dispose();
+            });
+        }
+
+        this.activePanel.webview.html = await this.getWebviewHtml(resourcesRoot, codiconsRoot);
```
philliphoff, line 51: Does this need to be set each time, or only once post creation?

resources/startPage.html.template
```diff
@@ -0,0 +1,177 @@
...
+            color: var(--vscode-editor-foreground);
+        }
+    </style>
+</head>
+
+<body>
+    <div class="container">
+        <div class="row" id="titleBar">
+            <div class="cell">
+                <a href="https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker"><img src="{{ dockerIconUri }}" /></a>
```
philliphoff, line 60: Should all of the icon-based links have `alt` or `a11y` tags for screen readers?

src/commands/startPage/StartPage.ts
bwateratmsft: Yep, every time.

resources/startPage.html.template
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        div .row {
+            display: flex;
+            flex-direction: row;
+            justify-content: flex-start;
+            align-items: flex-start;
+        }
+
+        div .cell {
```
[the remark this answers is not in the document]
bwateratmsft: So replace `div .cell` with `.row > *`? (And remove the `cell` class)

resources/startPage.html.template
bwateratmsft: Yes, for sure!

resources/startPage.html.template
```diff
@@ -0,0 +1,177 @@
...
+
+        a.icon:hover {
+            color: var(--vscode-editor-foreground);
+        }
+
+        div .container {
+            display: flex;
+        }
+
+        div .row {
```
danegsta, line 38: A grid might make more sense for the list of feature icons/descriptions (just in case there's any width differences between the different icons, the columns would always line up):

```
.container {
  display: grid;
  grid-gap: 10px;
  grid: auto-flow / auto 1fr;
}
```

resources/startPage.html.template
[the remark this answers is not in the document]
danegsta: Yeah, it'd have the same effect.

resources/startPage.html.template
danegsta: Each direct child of the container element would alternate between the first column and the second column position on increasing rows.

bwateratmsft: I was able to do it with Flexbox by just setting the width on the Docker logo the same as the icons get.

resources/startPage.html.template
bwateratmsft: I made the Docker icon smaller, which both helps with this and aligns the column better.

resources/startPage.html.template
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        .row {
+            display: flex;
+            flex-direction: row;
+            justify-content: flex-start;
+            align-items: flex-start;
+        }
+
+        .row>* {
```
karolz-ms, line 46: `.row > *` for extra readability

resources/startPage.html.template
```diff
@@ -0,0 +1,180 @@
...
+            <div>
+                <a href="https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker" class="linkheading">
+                    <h1>Docker Extension</h1>
+                    <span>by Microsoft</span>
+                </a>
+            </div>
+        </div>
+        <div class="row">
+            <div>
+                <a href="command:vscode-docker.configure" class="linkheading icon" tabindex="-1">
```
karolz-ms, line 72: Will telemetry capture the fact that this command was invoked from the start page?

src/commands/startPage/StartPage.ts
```diff
@@ -0,0 +1,83 @@
...
+        };
+
+        const template = Handlebars.compile(await fse.readFile(templatePath.fsPath, 'utf-8'));
+        return template(startPageContext);
+    }
+
+    private async handleMessage(message: WebviewMessage): Promise<void> {
+        switch (message.command) {
+            case 'showStartPageClicked':
+                await vscode.workspace.getConfiguration('docker').update('showStartPage', Boolean(message.showStartPage), vscode.ConfigurationTarget.Global);
```
karolz-ms, line 76: Should we add some telemetry that indicates the user clicked the checkbox (and what was the value they selected)?

src/utils/openExternal.ts
```diff
@@ -3,14 +3,8 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { commands, Uri } from 'vscode';
-import { localize } from '../localize';
+import * as vscode from 'vscode';
 
 export async function openExternal(path: string, throwErrorOnFailure: boolean = false): Promise<void> {
```
karolz-ms, line 8: We probably should get rid of this wrapper altogether. Also, `throwErrorOnFailure` parameter is now not used at all.

resources/startPage.html.template
bwateratmsft: I tried but VSCode immediately reformatted it back 😭 

I'm sure there's a setting but for this I don't know it's worth disabling, other CSS formatting features are good.

resources/startPage.html.template
bwateratmsft: No, unfortunately. It's not possible to pass any arguments to the command so it will be completely indistinguishable from one launched by the palette.

resources/startPage.html.template
karolz-ms: Of course 🙄

resources/startPage.html.template
karolz-ms: That's a bummer. Maybe it is worth opening an issue about it against VS Code core?

karolz-ms (approved):

resources/startPage.html.template
bwateratmsft: Good news, I'm wrong! You can pass arguments in the URI: https://code.visualstudio.com/api/extension-guides/command#using-commands

karolz-ms: Ooh, that is fantastic! So we can track what users click on the start page, that's awesome.

src/commands/startPage/StartPage.ts
```diff
@@ -0,0 +1,91 @@
...
+
+        const startPageContext: StartPageContext = {
+            cspSource: webview.cspSource,
+            nonce: cryptoUtils.getRandomHexString(8),
+            codiconsFontUri: webview.asWebviewUri(vscode.Uri.joinPath(codiconsRoot, 'codicon.ttf')).toString(),
+            codiconsStyleUri: webview.asWebviewUri(vscode.Uri.joinPath(codiconsRoot, 'codicon.css')).toString(),
+            dockerIconUri: webview.asWebviewUri(vscode.Uri.joinPath(resourcesRoot, 'docker.png')).toString(),
+            showStartPageChecked: vscode.workspace.getConfiguration('docker').get('showStartPage', false) ? 'checked' : '',
+            isMac: isMac(),
+            showAzure: true, // TODO?
```
bwateratmsft, line 71: Do?

PR #2460, opened 2020-11-03
Don't use red text for stderr in DockerPseudoTerminal

### LINKED ISSUE #2451: Buildkit builds have very ugly output
image

These aren't errors but they certainly look intimidating.

bwateratmsft:
Fixes #2451


src/tasks/DockerPseudoterminal.ts
```diff
@@ -56,33 +63,52 @@ export class DockerPseudoterminal implements Pseudoterminal {
...
         command: CommandLineBuilder,
         folder: WorkspaceFolder,
-        rejectOnStderr?: boolean,
-        stdoutBuffer?: Buffer,
-        stderrBuffer?: Buffer,
         token?: CancellationToken): Promise<void> {
         const commandLine = resolveVariables(command.build(), folder);
 
         // Output what we're doing, same style as VSCode does for ShellExecution/ProcessExecution
         this.write(`> ${commandLine} <\r\n\r\n`, DEFAULTBOLD);
```
bwateratmsft, line 70: This is being output correctly but is pretty quickly overwritten when the build task starts; it completely clears the terminal window. The same does not happen for shell commands. Should try to fix it.

src/tasks/DockerPseudoterminal.ts
```diff
@@ -3,10 +3,13 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as nodeptytypes from 'node-pty';
```
bwateratmsft, line 6: Build is failing because I don't have `node-pty.d.ts` available. Need to find a way to download it to `node_modules`.

src/tasks/DockerPseudoterminal.ts
```diff
@@ -56,33 +62,52 @@ export class DockerPseudoterminal implements Pseudoterminal {
...
+
+                    resolve();
                 }
-            },
-            stderrBuffer,
-            token
-        );
+            );
+
+            const cancelDisposable = token?.onCancellationRequested(() => {
```
philliphoff, line 99: Should `cancelDisposable` be disposed in the exit event handler above as well?

src/tasks/DockerPseudoterminal.ts
```diff
@@ -56,33 +62,52 @@ export class DockerPseudoterminal implements Pseudoterminal {
...
-            },
-            stdoutBuffer,
-            (stderr: string) => {
-                this.writeError(stderr);
-
-                if (rejectOnStderr) {
-                    throw new Error(stderr);
+        const pty = getCoreNodeModule<nodepty>('node-pty');
+
+        this.activePty = pty.spawn(isWindows() ? 'powershell.exe' : 'sh', ['-c', commandLine], {
```
philliphoff, line 73: Should we be using the shell (default or otherwise) set by the user?  (I prefer and setup `CMD` on Windows, for example, and `zsh` on Mac.). Is PowerShell now a core component of Windows or still an optional feature?

src/tasks/DockerPseudoterminal.ts
```diff
@@ -114,3 +139,21 @@ export class DockerPseudoterminal implements Pseudoterminal {
         this.writeEmitter.fire(`\x1b[${color}${message}\x1b[0m`);
     }
 }
+
+/**
+ * Loosely based on https://github.com/microsoft/node-pty/blob/master/typings/node-pty.d.ts, but downloading that is difficult and actually installing
+ * the node-pty package is even more difficult (lots of native code)
+ */
+interface nodepty {
```
philliphoff, line 147: Any reason not to keep this in a `src\types\node-pty.d.ts` file?

src/tasks/DockerPseudoterminal.ts
bwateratmsft: Probably should use `cmd` and `sh`, rather than `powershell` and `sh`. We *could* try to adopt the user's shell choice but that becomes *really* complicated *really* quickly. In defense of just using `cmd`/`sh`, that's what Node does if you do `child_process.spawn`, and since `spawn` is what we were doing previously--that's what we've always had.

src/tasks/DockerPseudoterminal.ts
bwateratmsft: Not particularly, was just thinking this was probably the only place it would be needed. But I don't feel strongly at all.

src/tasks/DockerPseudoterminal.ts
philliphoff: I'm fine with `cmd` and `sh`; I think that should work reasonably well in the vast majority of cases.

src/tasks/DockerPseudoterminal.ts
bwateratmsft: Alex suggested some control character changes that might help.

src/tasks/DockerPseudoterminal.ts
bwateratmsft: Ugh. Obnoxiously, `node-pty` does not handle arguments correctly with `cmd`. Something with quoting is not working.

bwateratmsft: So it's just not plausible to use `node-pty`. There is so much complexity in handling the terminal control characters that it's not worth the effort. Instead I'll just make stderr text be default color. Thanks for nothing, BuildKit!

philliphoff (approved):

PR #2469, opened 2020-11-05
Fix a merge build break

bwateratmsft:


src/tree/containers/ContainersTreeItem.ts
```diff
@@ -59,7 +59,7 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<DockerContainerInf
         const rawResults = await ext.dockerClient.getContainers(context);
 
         // NOTE: We *know* that ACI doesn't currently support showing files, but we'll give the benefit of the doubt to any other context type.
-        const contextType = (await ext.dockerContextManager.getCurrentContext())?.Type;
+        const contextType = (await ext.dockerContextManager.getCurrentContext())?.ContextType;
```
philliphoff, line 62: How were we allowed to merge our branches in with that break?  The PR builds (for mine, at least) showed a clean build.  Weird.

philliphoff (approved):

src/tree/containers/ContainersTreeItem.ts
bwateratmsft: I think your clean build happened before my PR was submitted that changed the field name. We don't force PR rebuilds when `master` changes. Naturally this can result sometimes but I don't think it's a serious problem.

PR #2477, opened 2020-11-10
Replace ADD with COPY

mbelsky:
Hi,

Docker docs [suggests](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy) to use COPY for files and directories that do not require ADD’s tar auto-extraction capability. So there is a little change.


bwateratmsft:
Thank you for your contribution @mbelsky! We're just about to cut a release (1.8.0) so I'd like to hold on to this PR until 1.9.0. We release roughly monthly.

bwateratmsft (approved):
Tested it out and everything seems good to me!

bwateratmsft:
This is now released in Docker extension version 1.9.0.

PR #2514, opened 2020-11-20
Add ability to launch a subset of Compose services

### LINKED ISSUE #2445: Support launching a subset of services from a compose file
Many customers only want to launch some of the services in a compose file at any given time, let's support that scenario.

bwateratmsft:
Fixes #2445.

Screenshot of the tasks:
image

Screenshot of command customization token:
image

Screenshot of service prompt when using command customization token:
image

Documentation PR: https://github.com/microsoft/vscode-docs/pull/4152


src/commands/compose/compose.ts
```diff
@@ -38,26 +39,41 @@ async function compose(context: IActionContext, commands: ('up' | 'down')[], mes
 
     for (const command of commands) {
         if (selectedItems.length === 0) {
-            const terminalCommand = await selectComposeCommand(
+            // Get the initial command
```
philliphoff, line 42: Given these two blocks are nearly identical, is there a way to consolidate them into a single block?  For example, wrap the no-selected-items case into a "dummy" selected item so that you can just use the selected items branch?

src/commands/compose/getComposeServiceList.ts
```diff
@@ -0,0 +1,56 @@
...
+const composeCommandReplaceRegex = /(up|down).*$/i;
+
+export async function getComposeServiceList(context: IActionContext, workspaceFolder: vscode.WorkspaceFolder, composeCommand: string): Promise<string> {
+    const services = await getServices(workspaceFolder, composeCommand);
+
+    // Fetch the previously chosen services list. By default, all will be selected.
+    const workspaceServiceListKey = `vscode-docker.composeServices.${workspaceFolder.name}`;
+    const previousChoices = ext.context.workspaceState.get<string[]>(workspaceServiceListKey, services);
+
+    const pickChoices: IAzureQuickPickItem<string>[] = services.map(s => {
```
philliphoff, line 21: Nit: when returning objects, the `return` and braces can be omitted by using parentheses.
```javascript
s => ({
    label: s,
    data: s,
    picked: ...
}));
```

src/commands/compose/getComposeServiceList.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // Update the cache
+    await ext.context.workspaceState.update(workspaceServiceListKey, subsetChoices.map(c => c.data));
+
+    return subsetChoices.map(c => c.data).join(' ');
+}
+
+async function getServices(workspaceFolder: vscode.WorkspaceFolder, composeCommand: string): Promise<string[]> {
+    // Start by getting a new command with the exact same files list (replaces the "up ..." or "down ..." with "config --services")
+    const configCommand = composeCommand.replace(composeCommandReplaceRegex, 'config --services');
```
philliphoff, line 49: Will the regex work if `up` or `down` happen to show up elsewhere in the command line, such as in the compose filenames?

src/tasks/DockerComposeTaskDefinitionBase.ts
```diff
@@ -0,0 +1,30 @@
...
+    customOptions?: string;
+}
+
+export interface DockerComposeDownOptions {
+    removeImages?: 'all' | 'local';
+    removeVolumes?: boolean;
+    customOptions?: string;
+}
+
+export interface DockerComposeOptions {
```
philliphoff, line 22: I imagine there's a way in TypeScript to describe the fact that `up` or `down` should be included, but not both, and not neither.

src/tasks/DockerComposeTaskProvider.ts
```diff
@@ -0,0 +1,85 @@
...
+            false, // rejectOnStderr
+            undefined, // stdoutBuffer
+            Buffer.alloc(10 * 1024), // stderrBuffer
+            context.cancellationToken
+        );
+        throwIfCancellationRequested(context);
+    }
+
+    private async validateResolvedDefinition(context: DockerComposeTaskContext, dockerCompose: DockerComposeOptions): Promise<void> {
+        if (!dockerCompose.up && !dockerCompose.down) {
```
philliphoff, line 51: What if both are present?

src/tasks/DockerComposeTaskProvider.ts
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+    }
+
+    private async resolveCommandLine(options: DockerComposeOptions): Promise<CommandLineBuilder> {
+        const isNewContext = isNewContextType((await ext.dockerContextManager.getCurrentContext()).ContextType);
+
+        if (options.up) {
+            return CommandLineBuilder
+                .create(isNewContext ? 'docker compose' : 'docker-compose')
```
philliphoff, line 67: I wonder if we should move this to our Docker client API to consolidate it in one place.

package.json
```diff
@@ -1232,6 +1232,109 @@
                         ]
                     }
                 }
+            },
+            {
+                "type": "docker-compose",
+                "properties": {
+                    "dockerCompose": {
```
philliphoff, line 1239: I kind of like the use of nested properties to determine the sub-command for the task (i.e. `up` vs. `down`).  I may steal that for other extensions.  That might have also worked for our Docker `build` and `run` tasks, but I suppose it's too late to for that.  :-)

src/commands/compose/getComposeServiceList.ts
bwateratmsft: So THAT's how to do it!

src/commands/compose/getComposeServiceList.ts
bwateratmsft: No, great catch! I'll enforce word boundaries around those words.

src/tasks/DockerComposeTaskDefinitionBase.ts
bwateratmsft: Looks like it, but it involves creating intermediate types where one is `never` and the other is defined; and then a union of those two is the "final" type: https://stackoverflow.com/a/61281828

I think it's sort of an awkward syntax...

src/tasks/DockerComposeTaskProvider.ts
bwateratmsft: I can add extra validation here. The tasks.json will already have some annoying yellow squiggles if you add both.

src/tasks/DockerComposeTaskProvider.ts
bwateratmsft: I agree. Since it depends on context, we should have it somewhere in there.

package.json
bwateratmsft: Yeah. I was thinking to myself--`docker-compose down` is barely complicated enough to justify a task of its own, how can I do this in the laziest way possible? :smile:

src/tasks/DockerComposeTaskDefinitionBase.ts
bwateratmsft: I found what I think is a reasonable way to do it. It doesn't offer a whole lot, only prevents you from doing this:

```typescript
const foo: DockerComposeOptions = {
    up: { ... },
    down: { ... },
    files: [ ... ],
};
```

In the above example, it forces you to define either `up` or `down` but not both. Having neither is allowed by this syntax though. I guess it's still an improvement, plus I learned something new about TypeScript!

philliphoff (approved):

Issue #1106, opened 2019-07-04
[Linux] Miss a icon for the 'Connected Registries' node


v-xuanzh:
**OS Platform**: Linux Ubuntu
**Extension build:** [20190703.2](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=7464)

**Steps to reproduce:**
1. Start up a private local registry using the command `docker run -d --rm --name noAuthRegistry -p 5100:5000 registry`
2. Go to REGISTRIES -> Click 'Connect Registry...' -> Select 'Generic Docker Registry(Preview)'
3. Input the URL: `http://localhost:5100`
Username: <`Press Enter`>
4. Observe the icon for the 'Connected Registries' node.

**Expect Experience:**

The icon of 'Connected Registries' node shows well.

**Actual Experience:**

Miss a icon for the 'Connected Registries' node.
![image](https://user-images.githubusercontent.com/34729022/60660320-16667680-9e8a-11e9-8dd0-9979fc79a507.png)

**More info:**

This issue doesn't reproduce on Windows/Mac.

v-dantch:
More Info:
There is also no icon for the "GitLab" node.
![image](https://user-images.githubusercontent.com/34729092/64685935-6a10c400-d4ba-11e9-8d82-a090bc3aa34b.png)

bwateratmsft:
@v-ruizh does this still reproduce?

v-ruizh:
@bwateratmsft Sorry for the late reply. This bug also reproduces on the newer build [20200413.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=17741&view=results).

bwateratmsft:
Thanks!

v-mengwe:
Verified on the latest build 20200818.1, this issue still reproduces for now. I will continue to track this issue.

v-ruizh:
@bwateratmsft This issue does not reproduce for "Connected Registries" node, but still reproduces for "GitLab" node. Could you also help to fix this issue for "GitLab" node?

**Expect:**
![image](https://user-images.githubusercontent.com/45614674/101139929-36850100-364d-11eb-8d79-ef705c87cce2.png)

**Actual:**
![image](https://user-images.githubusercontent.com/45614674/101140051-5e746480-364d-11eb-8c09-d7d49e0952f9.png)

bwateratmsft:
[...]

PR #2542, opened 2020-12-04
Fix missing GitLab icon on Linux

bwateratmsft:
Similar to #1106


philliphoff (approved):
Case-sensitivity strikes again! :-)

PR #2543, opened 2020-12-04
Improve messaging for invalid URL error

### LINKED ISSUE #2541: Improve Invalid URL error
When a user gets an Invalid URL error, I believe we could help them fix this issue faster. 

In terms of the error message itself we are given, I think it could be clearer to tell the user the source of the error. 
Either:
1. They have an invalid setting for docker.host attribute in their settings.json
2. They do not have a docker.host attribute (which would override) and instead have an invalid variable for DOCKER_HOST env var

It would help if we changed the popup notification when an invalid URL error occurs from "Report Issue" to "Open Settings".
Then we could open the VS Code Docker settings directly to the docker.host attribute that needs to be modifed. 

In the VS Code Docker Settings, we should provide some examples of the protocols that must be prepended to a url to properly communicate with the docker daemon or your private server. The message could say something like "Depending on the protocol your server is using, you should use a url such as "ssh://192.168.1.1" or "tcp://192.168.1.1". 

This would provide clearer messaging.

bwateratmsft:
Fixes #2541. @ucheNkadiCode what do you think of the below?

Depending on whether the problem is with `DOCKER_HOST` env var or `docker.host` setting, a different message will be shown. In both cases, an "Open Settings" button is shown that takes the user directly to the `docker.host` setting.

image
image

Settings page opened to `docker.host`:
image

### DIFF
--- a/src/docker/ContextManager.ts
+++ b/src/docker/ContextManager.ts
@@ -11,7 +11,7 @@ import * as path from 'path';
 import { URL } from 'url';
 import { commands, Event, EventEmitter, window, workspace } from 'vscode';
 import { Disposable } from 'vscode';
-import { callWithTelemetryAndErrorHandling, IActionContext, UserCancelledError } from 'vscode-azureextensionui';
+import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { AsyncLazy } from '../utils/lazy';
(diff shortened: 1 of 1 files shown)


philliphoff (approved):

ucheNkadiCode:
Only small change is to remove the commas before the "or tcp://" so it would read "...protocol,for example, ssh://myuser@12.3.4  or tcp://1.2.3.4"

Great Job!

ucheNkadiCode:
I checked back and had a few more thoughts, @karolz-ms, from the error code above, are we certain that the user entered a URL with invalid syntax and not that they entered a URL that could not successfully connect to their docker daemon? Is there a separate error for this case?

Sorry for adding more work, but after looking a little harder at other examples of VS Code settings, I'm thinking the messaging on for the docker.host attribute could be even clearer if we added:

Very similar to setting the DOCKER_HOST environment variable in the terminal, however, protocol is required (for example, "ssh://myuser@12.3.4" or "tcp://1.2.3.4") 

Thoughts?

karolz-ms:
@ucheNkadiCode yes, in this case we know that the error occurred because the URL is malformed. Connection errors are different. 

Also, DOCKER_HOST environment variable, and configuration setting, both require the protocol in the URL; protocol is not optional, because without the protocol the URL would be ambiguous.

ucheNkadiCode:
Suggested Revision to docker.host setting:
Equivalent to setting the DOCKER_HOST environment variable, for example, "ssh://myuser@mymachine" or "tcp://1.2.3.4")

PR #2551, opened 2020-12-08
Allow users to download container files

### LINKED ISSUE #2466: Enable downloading files from a running container
An earlier [feature](#2333) enabled viewing (read-only) files in a running container.  This issue builds on that functionality to allow the user to download files from the running container to the local machine (rather than opening them in an editor). This can be useful for non-text files, or just eliminating the need to cut & paste from the editor.

philliphoff:
Currently users can open (text) files of running containers in the VS Code editor.  This change adds a new download command that, like in VS, allows the user to instead extract files from running containers and save them in a user-selected local folder.  Multiple files can be downloaded at a given time.  If the file already exists in the selected folder, the user will be prompted to skip that file, overwrite that file, or to cancel any remaining download.

Resolves #2466.

### DIFF
--- a/src/commands/containers/files/downloadContainerFile.ts
+++ b/src/commands/containers/files/downloadContainerFile.ts
@@ -5,11 +5,24 @@
 
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { IActionContext } from 'vscode-azureextensionui';
+import { IActionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { ext } from '../../../extensionVariables';
 import { localize } from '../../../localize';
 import { FileTreeItem } from '../../../tree/containers/files/FileTreeItem';
 import { multiSelectNodes } from '../../../utils/multiSelectNodes';
+
+async function fileExists(file: vscode.Uri): Promise<boolean> {
+    try {
+        // NOTE: The expectation is that stat() throws when the file does not exist.
+        //       Filed https://github.com/microsoft/vscode/issues/112107 to provide
+        //       a better mechanism than trapping exceptions.
+        await vscode.workspace.fs.stat(file);
+
+        return true;
+    } catch {
+        return false;
+    }
+}
 
 export async function downloadContainerFile(context: IActionContext, node?: FileTreeItem, nodes?: FileTreeItem[]): Promise<void> {
     nodes = await multiSelectNodes(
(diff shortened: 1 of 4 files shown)


src/commands/containers/files/downloadContainerFile.ts
```diff
@@ -0,0 +1,102 @@
...
+    const localFolderUris = await vscode.window.showOpenDialog(
+        {
+            canSelectFiles: false,
+            canSelectFolders: true,
+            canSelectMany: false,
+            openLabel: localize('vscode-docker.commands.containers.files.downloadContainerFile.openLabel', 'Select'),
+            title: localize('vscode-docker.commands.containers.files.downloadContainerFile.openTitle', 'Select folder for download')
+        });
+
+    const localFolderUri = localFolderUris[0];
```
karolz-ms, line 45: According to docs `showOpenDialog` can return `undefined`? Should we handle that?

Is it a possibility that it will return zero-length array?

src/commands/containers/files/downloadContainerFile.ts
```diff
@@ -0,0 +1,102 @@
...
+        async (_, token) => {
+            for (const file of files) {
+                if (token.isCancellationRequested) {
+                    throw new UserCancelledError();
+                }
+
+                const localFileExists = await fileExists(file.localUri);
+
+                if (localFileExists) {
+                    const overwriteFile: vscode.MessageItem = {
```
karolz-ms, line 72: Maybe these messages belong to the top of the file? Or at least outside of the file loop?

src/commands/containers/files/downloadContainerFile.ts
```diff
@@ -0,0 +1,102 @@
...
+                    const result = await vscode.window.showWarningMessage(
+                        localize('vscode-docker.commands.containers.files.downloadContainerFile.existingFileWarning', 'The file \'{0}\' already exists in folder \'{1}\'.', file.fileName, localFolderUri.fsPath),
+                        overwriteFile,
+                        skipFile,
+                        cancelDownload);
+
+                    if (result === skipFile) {
+                        continue;
+                    } else if (result !== overwriteFile) {
+                        throw new UserCancelledError();
```
karolz-ms, line 93: Not `throw`, `continue` instead?

src/commands/containers/files/downloadContainerFile.ts
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                }
+
+                const content = await vscode.workspace.fs.readFile(file.containerUri);
```
karolz-ms, line 97: This might not work so great if the file is big. Can we use `FileSystem.copy()` instead? It has provision for overwriting existing file...

karolz-ms (changes requested):
Overall looks great, just minor suggestions/questions

src/commands/containers/files/downloadContainerFile.ts
philliphoff: Good catch.  I'll accommodate users canceling the dialog.  A zero-length array *shouldn't* happen, but I'll lump that into the cancellation case as well.

src/commands/containers/files/downloadContainerFile.ts
philliphoff: Reasonable; updated;

src/commands/containers/files/downloadContainerFile.ts
philliphoff: Yes, I had missed that in the API.  Presumably `copy()` is much more efficient.  Updated.

src/commands/containers/files/downloadContainerFile.ts
philliphoff: Not sure I follow.  If result is not `skipFile` or `overwriteFile`, then the user must have either explicitly canceled or implicitly canceled by dismissing the notification, in which case we should stop (via the `UserCancelledError()` exception.

karolz-ms: Sorry, never mind--somehow I read the code wrong and thought we throw `UserCancelError` if the user asks to skip a file. This is not the case, obviously.

karolz-ms (approved):

PR #2585, opened 2020-12-26
Upgrade to language server 0.2.0

rcjsuen:
This new release depends on `vscode-languageclient` 7.0.0 which has introduced some API changes. It seems like it no longer re-exports some VS Code API types so I've changed imports from `vscode-languageclient` to `vscode` where applicable.

This update to 0.2.0 brings two small fixes to the editor. As always, just copy/paste the examples below in a Dockerfile editor and do not save (as auto-formatting may change the content that is being parsed).

1. Having an escaped newline as the only argument of an instruction should not be valid. This will now be flagged as an error correctly.
```
FROM alpine
USER \
```

2. This used to cause an infinite loop and break semantic highlighting. This bug has now been fixed.
```
FROM alpine
ENV a \
b
```
```
[Error - 10:15:15 AM] Request textDocument/semanticTokens failed.
  Message: Request textDocument/semanticTokens failed with message: Maximum call stack size exceeded
  Code: -32603 
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3809,51 +3809,41 @@
             }
         },
         "dockerfile-ast": {
-            "version": "0.0.27",
-            "resolved": "https://registry.npmjs.org/dockerfile-ast/-/dockerfile-ast-0.0.27.tgz",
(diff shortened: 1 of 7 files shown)


karolz-ms (approved):
Thank you, Remy

PR #2619, opened 2021-01-12
Fix PATH on Mac systems

### LINKED ISSUE #2578: Need workaround(s) for docker exe missing from path issue (MacOS)
Many customers are suffering from the bug in MacOS that makes the docker executable folder disappear from path, for example: https://github.com/microsoft/vscode-docker/issues/2254 https://github.com/microsoft/vscode-docker/issues/2574 https://github.com/microsoft/vscode-docker/issues/2577

Docker seems to have hard time finding the root cause https://github.com/docker/compose-cli/issues/754

Consider
- following @ke6jjj suggestion and add a setting that contains full path to docker executable (the extension user would set it as a workaround)
- when "docker not found" error happens, probe the default Docker Desktop for Mac installation path and use the executable there if it exists
- (other ideas welcome)

bwateratmsft:
Fixes #2578


src/utils/spawnAsync.ts
```diff
@@ -240,3 +244,26 @@ export function bufferToString(buffer: Buffer): string {
...
+    // Looks for `/usr/local/bin` in the PATH.
+    // Must be whole, i.e. the left side must be the beginning of the string or :, and the right side must be the end of the string or :
+    // Case-insensitive, because Mac is
+    if (/(?<=^|:)\/usr\/local\/bin(?=$|:)/i.test(options?.env?.PATH)) {
+        // Do nothing: PATH already contains `/usr/local/bin`
+        return;
+    }
+
+    options = options ?? {};
+    options.env = options.env ?? process.env;
```
bwateratmsft, line 263:
```suggestion
    options.env = options.env ?? { ...process.env };
```

Might need to be this, to make a clone of `process.env` so that it does not get modified all-up.

karolz-ms (approved):

src/utils/spawnAsync.ts
karolz-ms: Good idea

PR #2620, opened 2021-01-12
Remove SSL support from .NET, use non-root user

bwateratmsft:
This PR removes SSL support for ASP.NET apps. It also switches to a non-root user in the Dockerfile, if all the ports are >1024 (same behavior as Python).


resources/templates/netCore/Dockerfile.template
```diff
@@ -4,6 +4,18 @@ WORKDIR /app
...
 {{/each}}
 
+{{#if (eq platform '.NET: ASP.NET Core')}}
+{{#if ports}}
+ENV ASPNETCORE_URLS=http://+:{{ ports.[0] }}
+{{/if}}
+{{/if}}
+
+{{#unless (isRootPort ports)}}
+# Switching to a non-root user, please refer to https://aka.ms/vscode-docker-python-user-rights
```
bwateratmsft, line 14: TODO: It says "Python" (so do the docs)

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -275,43 +218,10 @@ export class NetCoreTaskHelper implements TaskHelper {
...
             addVolumeWithoutConflicts(volumes, appVolume);
             addVolumeWithoutConflicts(volumes, srcVolume);
             addVolumeWithoutConflicts(volumes, debuggerVolume);
             addVolumeWithoutConflicts(volumes, nugetVolume);
-            if (await fse.pathExists(nugetFallbackVolume.localPath)) {
-                addVolumeWithoutConflicts(volumes, nugetFallbackVolume);
-            }
-        }
-
-        if (userSecrets || ssl) {
```
bwateratmsft, line 294: Do we want to stop mapping the user secrets folder? (Users can manually map it of course)

karolz-ms: If we stop mapping, will that break users that scaffolded the app with SSL (using previous version of the extension)?

karolz-ms (approved):

karolz-ms:
Soo glad to see this finally happening! ☺️

bwateratmsft:
> Soo glad to see this finally happening! ☺️

Same. I agree with @BigMorty though that we owe a meeting with the ASP.NET folks to see what they think, before we merge this.

src/tasks/netcore/NetCoreTaskHelper.ts
bwateratmsft: 🤦‍♂️ Yeah, I should probably try that out.

bwateratmsft:
Closing this since we're changing directions here.

PR #2622, opened 2021-01-12
Capture compose file edit event

### LINKED ISSUE #2613: Add a signal for compose file editing
It would be great to know whether or not users are engaging in editing their compose files. We could send some sort of signal when they hit Ctrl + S on  a `docker-compose.yml` file

This would help us know for sure how popular Compose is becoming.

bwateratmsft:
Fixes #2613


src/telemetry/registerListeners.ts
```diff
@@ -5,25 +5,32 @@
...
+            }
+
+            context.telemetry.properties.lineCount = doc.lineCount.toString();
+
+            void ext.activityMeasurementService.recordActivity('overall');
+        });
+
+        registerEvent('composefilesave', workspace.onDidSaveTextDocument, async (context: IActionContext, doc: TextDocument) => {
+            // If it's not a compose file, skip
+            if (!/docker-compose.*\.(yml|yaml)/i.test(doc.fileName)) {
```
karolz-ms, line 26: This will soon need to change to include `compose.yaml` too. Not sure if you want to change it now, put a `TODO`, or something else.

src/telemetry/registerListeners.ts
```diff
@@ -5,25 +5,32 @@
...
-import { ext } from './extensionVariables';
-
-let lastUploadTime: number = 0;
-const hourInMilliseconds = 1000 * 60 * 60;
+import { ext } from '../extensionVariables';
 
 export function registerListeners(): void {
     if (ext.telemetryOptIn) {
         registerEvent('dockerfilesave', workspace.onDidSaveTextDocument, async (context: IActionContext, doc: TextDocument) => {
-            // If it's not a Dockerfile, or last upload time is within an hour, skip
```
karolz-ms, line 16: So why remove this logic now? Or why did we have it here in the first place?

karolz-ms (approved):

src/telemetry/registerListeners.ts
bwateratmsft: In the past it was done in the name of performance. I had a couple reasons for removing it now:
- VSCode itself is sending (multiple) telemetry events every time a file is saved, with no rate limiting
- For the purposes of experimentation, it's better to send the event every time

src/telemetry/registerListeners.ts
bwateratmsft: Yeah. I wrote this code before picking up that other item. :smile:

Issue #2601, opened 2021-01-07
An error "Unknown parameter structure byName..." occurs in the "Dockerfile Language Server" output panel


v-ruizh:
**OS:** Windows 10
**Build Version:** 20210106.8

**Repro Steps:**
1. Create a .NET Core Console project.
2. Add Docker files to it.
3. Open Dockerfile and check the "Dockerfile Language Server" output panel.

**Expect:**
There is no error in the "Dockerfile Language Server" output panel.

**Actual:**
An error "Unknown parameter structure byName..." occurs in the "Dockerfile Language Server" output panel.

_(node:13968) UnhandledPromiseRejectionWarning: Error: Unknown parameter structure byName
    at computeSingleParam (c:\Users\AzureTest\.vscode-insiders\extensions\ms-azuretools.vscode-docker-1.9.5-alpha\dist\dockerfile-language-server-nodejs\lib\server.js:1:211247)
    at computeMessageParams (c:\Users\AzureTest\.vscode-insiders\extensions\ms-azuretools.vscode-docker-1.9.5-alpha\dist\dockerfile-language-server-nodejs\lib\server.js:1:211410)
    at Object.sendNotification (c:\Users\AzureTest\.vscode-insiders\extensions\ms-azuretools.vscode-docker-1.9.5-alpha\dist\dockerfile-language-server-nodejs\lib\server.js:1:212089)
    at Object.sendDiagnostics (c:\Users\AzureTest\.vscode-insiders\extensions\ms-azuretools.vscode-docker-1.9.5-alpha\dist\dockerfile-language-server-nodejs\lib\server.js:1:292320)
    at c:\Users\AzureTest\.vscode-insiders\extensions\ms-azuretools.vscode-docker-1.9.5-alpha\dist\dockerfile-language-server-nodejs\lib\server.js:1:16962
(node:13968) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:13968) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code._

rcjsuen:
[...]

PR #2625, opened 2021-01-13
Bump version to 1.9.6-alpha

bwateratmsft:
Bumps version to 1.9.6-alpha, along with dependencies. Please don't merge yet.


package.json
```diff
@@ -2778,17 +2778,17 @@
...
         "@azure/arm-appservice": "^6.1.0",
         "@azure/arm-containerregistry": "^8.0.0",
         "@azure/storage-blob": "^12.3.0",
         "@docker/sdk": "^1.0.3",
-        "@grpc/grpc-js": "^1.2.2",
-        "dayjs": "^1.10.2",
-        "dockerfile-language-server-nodejs": "^0.2.1",
+        "@grpc/grpc-js": "^1.2.3",
+        "dayjs": "^1.10.3",
+        "dockerfile-language-server-nodejs": "^0.2.2",
```
rcjsuen, line 2791: Hopefully this will help with #2601.

bwateratmsft: @rcjsuen Should we just grab 0.2.2 for our 1.9.1 release (~Monday)?

rcjsuen: @bwateratmsft It shouldn't hurt. The actual code changes are [minimal](https://github.com/rcjsuen/dockerfile-language-server-nodejs/compare/v0.2.1...v0.2.2), just implementing what was suggested in https://github.com/microsoft/vscode-docker/issues/2601#issuecomment-756134327.

Has anyone else hit #2601? The stack trace indicates it was reproduced on 1.9.5-alpha. 🤔

bwateratmsft: I've got it showing up on our 1.9.1 candidate build. I'll make a PR and we can discuss in our team triage today.

package-lock.json
```diff
@@ -4808,6 +4818,15 @@
                         "read-pkg": "^5.2.0",
                         "type-fest": "^0.8.1"
                     }
+                },
+                "semver": {
```
karolz-ms, line 4822: 😨

karolz-ms (approved):
Just double-check please that the webpacked extension loads after this change

package-lock.json
bwateratmsft: Absolutely! I think that this problem is actually why semver was tripping in the first place--there was one version webpacked in, and another included as a dependency of the language client.

bwateratmsft: Aha, that was _exactly_ it! And now fixed in #2626.

PR #2649, opened 2021-01-20
Update language server

### LINKED ISSUE #2624: --chmod support for COPY and ADD
In docker engine 20.10 support for the `--chmod` flag has been added. Trying to use it in a Dockerfile gives:
```
Unknown flag: chmoddockerfile-utils(40)
```

It would be nice to add support for it.

bwateratmsft:
Fixes #2624. @rcjsuen FYI


glensc:
Why not use a more meaningful title/commit message, so one does not have to clickbait to understand what the cryptic title means:

- `Update dockerfile-language-server-nodejs dependency to ^0.3.0`

PR #2657, opened 2021-01-21
Subset of services experiment

### LINKED ISSUE #2646: Right Click Option to "Select a subset of Compose services"
In order to select a subset of services by right-clicking a Compose file and selecting "Compose Up", a user has to go into their settings.json file and write:

docker.commands.composeUp: "docker-compose ${configurationFile} up ${detached} ${build} ${serviceList}"

Now that we can select services dynamically before Composing up, I'd like an option that makes this dynamic selection more accessible to use

image

It could follow default functionality of this feature such as "none selected" is the same as all selected etc . 

**Question**: Does this composeUp setting in settings.json persist for all other projects the user may open up in VS Code? I feel like this may be problematic if they are a user that has a simpler Project that uses compose and also a more complex project that actually needs the subset

bwateratmsft:
Fixes #2646. This adds a new option to launch a subset of services directly; basically it is equivalent to appending `${serviceList}` to the end of your compose command. This is gated by an experiment.

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
         "onCommand:vscode-docker.compose.down",
         "onCommand:vscode-docker.compose.restart",
         "onCommand:vscode-docker.compose.up",
+        "onCommand:vscode-docker.compose.up.subset",
         "onCommand:vscode-docker.configure",
         "onCommand:vscode-docker.configureCompose",
         "onCommand:vscode-docker.containers.attachShell",
@@ -172,6 +173,10 @@
                 {
                     "command": "vscode-docker.contexts.use",
                     "when": "!vscode-docker:contextLocked"
+                },
+                {
+                    "command": "vscode-docker.compose.up.subset",
+                    "when": "vscode-docker:composeSubsetExp"
                 }
             ],
             "editor/context": [
@@ -201,6 +206,11 @@
                     "group": "docker"
                 },
                 {
+                    "when": "resourceFilename == docker-compose.yml && vscode-docker:composeSubsetExp",
+                    "command": "vscode-docker.compose.up.subset",
+                    "group": "docker"
+                },
+                {
                     "when": "resourceFilename == docker-compose.debug.yml",
                     "command": "vscode-docker.compose.down",
                     "group": "docker"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -200,6 +205,11 @@
                     "command": "vscode-docker.compose.up",
                     "group": "docker"
                 },
+                {
+                    "when": "resourceFilename == docker-compose.yml && vscode-docker:composeSubsetExp",
```
bwateratmsft, line 209: These things will need merging with #2656

src/extension.ts
```diff
@@ -151,6 +154,10 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
         registerListeners();
 
         void openStartPageAfterExtensionUpdate();
+        void activateComposeUpSubsetExperiment();
+
+        // TODO: remove this
```
bwateratmsft, line 159: TODO: Do

src/extension.ts
```diff
@@ -67,6 +68,8 @@ function initializeExtensionVariables(ctx: vscode.ExtensionContext): void {
     registerUIExtensionVariables(ext);
 }
 
+// TODO: Remove this
```
bwateratmsft, line 71: TODO: Do

karolz-ms (approved):
Glad to see that it is relatively easy to set up such experiment! ☺️

bwateratmsft:
@karolz-ms @BigMorty @ucheNkadiCode @philliphoff -- as an alternative, we could keep just one command and show the service checklist every time; we can do this as an experiment so we're not bothering everybody. We can measure success if people are actually using it; we capture the numerator (number of services selected) and denominator (total number of services). If selected < total, it stands to reason they _did_ make use of the feature; otherwise not. If we saw, for example, that 95% were not using it, then combining commands is probably not the way to go. But if, say, 50% were using it--maybe one command is right.

Thoughts?

BigMorty:
Interesting idea @bwateratmsft. I could go either way, but I lean towards having the separate menu/command.

bwateratmsft:
> Interesting idea @bwateratmsft. I could go either way, but I lean towards having the separate menu/command.

That would be my inclination as well.

PR #2661, opened 2021-01-21
Add inline open/download commands and tweak order

philliphoff:
Adds inline commands for container file open/download (which is sometimes quicker than the context menu).  Also makes the open command first (as that's most likely the one being used).

Screenshot 2021-01-21 at 14 25 00

### DIFF
--- a/package.json
+++ b/package.json
@@ -2233,7 +2233,7 @@
                 "command": "vscode-docker.containers.openFile",
                 "title": "%vscode-docker.commands.containers.openFile%",
                 "category": "%vscode-docker.commands.category.dockerContainers%",
-                "icon": "$(file-symlink-file)"
+                "icon": "$(go-to-file)"
             },
             {
                 "command": "vscode-docker.containers.prune",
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

package.json
```diff
@@ -2219,7 +2232,8 @@
             {
                 "command": "vscode-docker.containers.openFile",
                 "title": "%vscode-docker.commands.containers.openFile%",
-                "category": "%vscode-docker.commands.category.dockerContainers%"
+                "category": "%vscode-docker.commands.category.dockerContainers%",
+                "icon": "$(file-symlink-file)"
```
bwateratmsft, line 2236:
```suggestion
                "icon": "$(go-to-file)"
```

`go-to-file` is used in the source control pane for opening a file directly, I'd suggest that over `file-symlink-file`.
image

philliphoff: Yes, I like that one better.  I didn't come across it when searching.

Screenshot 2021-01-25 at 10 23 25

PR #2660, opened 2021-01-21
Add OCI `X-Meta-Source-Client` header to registry requests

bwateratmsft:
Fixes #2659

PR #2663, opened 2021-01-22
Change OCI client ID based on feedback

bwateratmsft:
Changing the OCI client ID based on [this comment](https://github.com/microsoft/vscode-docker/pull/2660#issuecomment-765505114) from @SteveLasker

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -50,4 +50,4 @@ export const dockerComposeHeader = `version: '3.4'
 
 services:`;
 
-export const ociClientId = 'on-prem/vscode/docker';
+export const ociClientId = 'tools/microsoft/vscode/docker';


karolz-ms (approved):

SteveLasker:
LGTM, thank you

BigMorty:
Thanks Brandon. Do we need to make a change in our VS tools also?

bwateratmsft:
I don't _think_ so; I'm not aware of any direct queries to registries.

SteveLasker:
@BigMorty, for the VS tools, are you referring to Visual Studio tools?
As you pull from registries for the FROM statements, it would be great to know the requests came from VS compared to VS Code, Kubernetes, etc. So, yeah. The same way VS Code identifies itself, knowing VS would also be helpful.

bwateratmsft:
@SteveLasker VSCode is only identifying itself here for some select requests but anything through the CLI would not have this header. Is there a way to control the header sent by the CLI?

SteveLasker:
The challenge with the header is it's stored in a central location ./docker directory. I believe Karol or Nathan had come up with a way to manipulate in memory or swap out on context. I don't know if they landed it with a high-degree of confidence in stability. 
How are you setting it with VS Code?

bwateratmsft:
There are a couple of places we make direct HTTP requests to registries; it's for these only we use the header. Some examples--
- Image out-of-date check (once a day, disabled by default)
- Registry explorer
- Docker Hub image search for suggestions when editing Dockerfiles

SteveLasker:
Ahh. Well, if that can be duplicated in VS Code, great. But, I'm not sure how the images in the FROM statements are pre-cached.

bwateratmsft:
Sorry, I should have clarified--all those examples are in the VSCode Docker extension. I don't think any direct HTTP requests exist in the VS tools--all CLI.

BigMorty:
Thanks for checking Brandon.

Issue #2465, opened 2020-11-04
Enable editing of files in a running container


philliphoff:
An earlier [feature](#2333) added viewing (read-only) of files in a running container. This issue builds on top of that functionality to enable editing (and saving) of those files back to the container.

dbreshears:
make sure we can track the usage

aviadcye:
Any news on this?

philliphoff:
@aviadcye It's being worked on.  :-) See #2671.

aviadcye:
cool, thanks :)

bwateratmsft:
This is now released in Docker extension version 1.10.0.

PR #2671, opened 2021-01-25
Adds support for editing files within running containers

philliphoff:
Implements the `writeFile()` method of our container `FileSystemProvider` which allows VS to write changes made to in the editor back to the container.

> As Docker does not support the get/put archive API when using Hyper-V -based container (i.e. Windows container), this functionality is enabled only for Linux-based containers for the time being. While we have workarounds to get files off of Windows containers, I've yet to make the inverse work which means more investigation.  As Windows container usage is small, I think it's better to get the feature out rather than first insist on strict parity.

Related to #2465.


bwateratmsft:
> Related to #2465.

It looks like this just fixes #2465 entirely, right?

bwateratmsft (approved):

philliphoff:
> It looks like this just fixes #2465 entirely, right?

Since it doesn't allow editing of Windows containers, I wouldn't say the issue is completely resolved, though we could close this issue and open a investigatory issue for extending support to Windows containers (if and when I can actually get that to work).

bwateratmsft:
> > It looks like this just fixes #2465 entirely, right?
> 
> Since it doesn't allow editing of Windows containers, I wouldn't say the issue is completely resolved, though we could close this issue and open a investigatory issue for extending support to Windows containers (if and when I can actually get that to work).

Personally I would not shed any tears for ignoring Windows container support for file editing (or files at all for that matter, but you at least got read working).

I guess I'd vote for opening a separate issue and dumping it in the Future milestone until users express interest.

philliphoff:
Sounds good to me; opened #2675.

PR #2691, opened 2021-02-01
Use `HEAD` request to get digest

bwateratmsft:
This change does several things:

1. Uses the `Accept: application/vnd.docker.distribution.manifest.list.v2+json` header to enable doing `HEAD` requests--not subject to rate limiting--to check if an image is out-of-date
1. Re-enables out-of-date checking by default
1. Returns out-of-date checking to once per session instead of once per day. I don't feel strongly that it ought to be once-per-session, so if desired I can keep it at once-per-day.

I'd be OK with putting this in either 1.10.0 or 1.11.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1900,7 +1900,7 @@
                 },
                 "docker.images.checkForOutdatedImages": {
                     "type": "boolean",
-                    "default": false,
+                    "default": true,
                     "description": "%vscode-docker.config.docker.images.checkForOutdatedImages%"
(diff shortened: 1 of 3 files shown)


karolz-ms (approved):

BigMorty:
This is great news!!! I am OK with once a day also.

bwateratmsft:
This is now released in Docker extension version 1.10.0.

PR #2709, opened 2021-02-08
Remove keytar in favor of VSCode API

### LINKED ISSUE #2699: Remove direct use of Keytar and use new `SecretStorage` API
In 1.53.0, VSCode added a new [SecretStorage](https://code.visualstudio.com/api/references/vscode-api#SecretStorage) API. We currently use keytar directly. We should remove our usage of keytar in favor of the `SecretStorage` API.

bwateratmsft:
Fixes #2699

@philliphoff and I talked about it, and since this is a one-time change, that we would not migrate the secrets--instead, just require folks to sign in again. That won't affect ACR since we don't keep any secrets for that, but will affect Docker Hub, GitLab, and generic V2.

If others feel strongly that we _should_ migrate secrets we can investigate that.


karolz-ms (approved):

karolz-ms:
Personally I think it is fine for users to re-authN as part of (this specific) extension upgrade

PR #2720, opened 2021-02-11
Remove use of request package and update other dependencies

### LINKED ISSUE #9: Switch off request package
In order to work with proxies we should be using 'request' library in dockerHubApi

bwateratmsft:
Fixes #9 (wow, that's the smallest bug number I've fixed!)


src/utils/httpRequest.ts
```diff
@@ -4,82 +4,145 @@
...
+export function basicAuthHeader(username: string, password: string): string {
+    const buffer = Buffer.from(`${username}:${password}`);
+    return `Basic ${buffer.toString('base64')}`;
+}
+
+export function bearerAuthHeader(token: string): string {
+    return `Bearer ${token}`;
+}
+
+export async function httpRequest2<T>(url: string, options?: RequestOptionsLike, signRequest?: (request: RequestLike) => Promise<RequestLike>): Promise<HttpResponse2<T>> {
```
philliphoff, line 47: Is the `2` in `httpRequest2` still necessary?

philliphoff (approved):
:shipit:

src/utils/httpRequest.ts
bwateratmsft: 🤦‍♂️ I knew I forgot something!

I had it that way so I could easily refactor and change the name later (because it was distinct from any other symbol in the code base)

bwateratmsft:
> :shipit:

I'll never understand what a squirrel has to do with shipping it

PR #2732, opened 2021-02-24
Remove Python code to override entrypoint

### LINKED ISSUE #2725: Entrypoint for docker-run task
As I understand it, the entrypoint for python debugging is overridden to just "python" automatically to keep the container running.  Then if my "python" module is flask, it will execute flask with the arguments.  This doesn't allow for any room to set a dynamic "LD_LIBRARY_PATH" before my flask app starts.  Typically we execute a shell script as our entrypoint, which then sets the LD_LIBRARY_PATH (based on runtime library mapped into the container) in a .env file which the flask app picks up.

My main question is How can I achieve a dynamic LD_LIBRARY_PATH like I get when I can control the entrypoint?

Version: 1.53.2
Commit: 622cb03f7e070a9670c94bae1a45d78d7181fbd4
Date: 2021-02-11T11:45:54.515Z
Electron: 11.2.1
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Darwin x64 20.3.0

Name: Docker
Id: ms-azuretools.vscode-docker
Description: Makes it easy to create, manage, and debug containerized applications.
Version: 1.10.0
Publisher: Microsoft
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker

bwateratmsft:
Fixes #2725

### DIFF
--- a/src/tasks/python/PythonTaskHelper.ts
+++ b/src/tasks/python/PythonTaskHelper.ts
@@ -104,7 +104,6 @@ export class PythonTaskHelper implements TaskHelper {
 
         // User input is honored in all of the below.
         runOptions.volumes = this.inferVolumes(runOptions, launcherFolder);
-        runOptions.entrypoint = runOptions.entrypoint || 'python';
 
         return runOptions;
     }


karolz-ms (approved):
I agree we should not set entrypoint like this when it is not specified... esp. since `python` is by no means an obvious default for Python workloads.

bwateratmsft:
This may need to get reverted.

Issue #2615, opened 2021-01-09
FastAPI Docker Debugger 


Kludex:
## Description 

As a VSCode user I want to be able to debug my [FastAPI](https://fastapi.tiangolo.com/) application living inside a Docker container. :tada: 

## More information

The web framework is growing, and I think the community will benefit from this feature. :sunglasses:

Besides, a PR was accepted on the [Python VSCode extension](https://github.com/microsoft/vscode-python) with the same purpose, but without Docker: https://github.com/microsoft/vscode-python/issues/14247 

The idea is to have the option `Python: FastAPI` on the `Docker: Initialize for Docker debugging`:
![image](https://user-images.githubusercontent.com/7353520/104109644-610f4d00-52d0-11eb-9135-0a13b88d885b.png)

## Notes
- I'll be working on this. I'll link the PR when I've finished. :factory_worker:
- Please let me know if this feature will not be accepted, so I stop working on it :sunflower:

bwateratmsft:
Thanks for the suggestion! As long as the Python extension supports FastAPI remote debugging (which it probably does), then this should be possible. We'll take a look.

Kludex:
Fwiw, it should be similar to what's on the Flask one.

Basically this: https://github.com/Kludex/fastapi-docker-debug/

karolz-ms:
@Kludex yes, please submit a PR, and thank you for helping us make the extension better!!

Kludex:
@karolz-ms Thank you for accepting the idea! I'll be pushing a PR in the next days.

bwateratmsft:
Hey @Kludex, any updates on this?

Kludex:
@bwateratmsft I've started. But I lost track of it.

I'll be working (and hopefully) finishing it this weekend. :)

bwateratmsft:
This is now released in Docker extension version 1.11.0.

Issue #2498, opened 2020-11-17
The terminal process failed to launch: Could not find the terminal with id 2 on the extension host.


marviobezerra:
Issue Type: <b>Bug</b>

### Environment:
MAC OS Big Sur 11.0.1
.Net 5.0.100
VS Code 1.51.1 & Commit: e5a624b788d92b8d34d1392e4c4d9789406efe8f
VS Code Docker extension v1.8.0
Docker 2.5.0.1

### Steps to reproduce

1. Create a new API using .Net 5 (v 5.0.100)
`dotnet new webapi -o sample`

2. Open the new project on VS Code and add the recommended `launch.json` and `tasks.json` (it prompts up in a popup window a question add these files, answer yes)

> Note that the build and run are working fine at this point.

3. Add a new Dockerfile by using `Command + Shift + P` (select the default options)

4. Select the launch configuration `Docker .NET Core Launch` and press `F5`.

> Console output shows the error **`The terminal process failed to launch: Could not find the terminal with id 2 on the extension host.`**

### Full console output

```
> Executing task: dotnet build {ommited}/Sample.csproj <

Microsoft (R) Build Engine version 16.8.0+126527ff1 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  All projects are up-to-date for restore.
  Sample -> {ommited}/Sample/bin/Debug/net5.0/Sample.dll

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:01.03

Terminal will be reused by tasks, press any key to close it.

> Executing task: docker-build: debug <

The terminal process failed to launch: Could not find the terminal with id 2 on the extension host.

Terminal will be reused by tasks, press any key to close it.
```

Extension version: 1.8.0
VS Code version: Code 1.51.1 (e5a624b788d92b8d34d1392e4c4d9789406efe8f, 2020-11-11T01:11:34.018Z)
OS version: Darwin x64 20.1.0

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz (16 x 2300)|
[...]

PR #2740, opened 2021-02-28
feat: add fastapi docker

Kludex:
Closes https://github.com/microsoft/vscode-docker/issues/2615


src/utils/pythonUtils.ts
```diff
@@ -35,6 +36,11 @@ export function inferPythonArgs(projectType: PythonProjectType, ports: number[])
                 '--nothreading',
                 '--noreload'
```
Kludex, line 37: Do you guys know why when debugging the `noreload` option on both `Flask` and `Django` is activated? :thinking:

src/debugging/python/PythonDebugHelper.ts
```diff
@@ -125,6 +127,8 @@ export class PythonDebugHelper implements DebugHelper {
         switch (projectType) {
             case 'django':
                 return 'Starting development server at (https?://\\S+|[0-9]+)';
+            case 'fastapi':
+                return '(Listening at: (https?://\\S+|[0-9]+) \(113392\))|(Uvicorn running on (https?://\\S+|[0-9]+) \(Press CTRL+C to quit\))' // HEY!
```
Kludex, line 131: This can be cleaned up :sweat_smile:

On the debug, we use `uvicorn`, which is the second (afte the `|`) but on production the first one (before the `|`).

The number `113392` is actually the PID of the master process. I was wondering if I need to match the whole line that identifies that the server is running. :thinking: Like: `Listening at: (https?://\\S+|[0-9]+) \(\d+\)` would be fine?

Kludex:
I'm facing issues to debug it. :cry: 
Can someone help me? I don't have much experience on this setup :disappointed: 

image

If helps:
```bash
➜  vscode-docker git:(feat/fastapi-docker) ✗ node --version
v15.10.0
➜  vscode-docker git:(feat/fastapi-docker) ✗ npm --version 
7.5.3
```

src/utils/pythonUtils.ts
bwateratmsft: This disables the hot-reloading feature. By default, our Docker container build is set up to copy in the files; since the files are copied in instead of being bind mounted, hot-reloading wouldn't work--changes in the source code on the host don't propagate until the image is rebuilt. Our docs do cover how to [enable hot-reloading](https://code.visualstudio.com/docs/containers/debug-python#_how-to-enable-hot-reloading-in-django-or-flask-apps) for Django and Flask, but the user has to remember to uncomment the line in the Dockerfile before they can build an image that's useful for anything other than debugging (i.e., deploying to production).

bwateratmsft:
> I'm facing issues to debug it. 😢
> Can someone help me? I don't have much experience on this setup 😞
> 
> image
> 
> If helps:
> 
> ```shell
> ➜  vscode-docker git:(feat/fastapi-docker) ✗ node --version
> v15.10.0
> ➜  vscode-docker git:(feat/fastapi-docker) ✗ npm --version 
> 7.5.3
> ```

Could be the same issue as what's causing #2498? There's a bug in VSCode somewhere that causes terminals to not work. Seems like a _very_ active issue (100+ comments): https://github.com/microsoft/vscode/issues/105446

src/debugging/python/PythonDebugHelper.ts
bwateratmsft: I'm pretty sure it doesn't need to be the full line, but the code does expect only one group, with that group containing the URL or the port only. Maybe something like this (which uses non-capturing groups for the extra text):

`(?:Listening at: (https?://\\S+|[0-9]+))|(?:Uvicorn running on (https?://\\S+|[0-9]+)`

Kludex:
I'm going to check that in some hours. Thank you!

Do I need to add specific tests? For Django and Flask it looks like they are handled by "python containers" instead of something specific. Am I wrong?

bwateratmsft:
Unfortunately we have pretty little in the way of automated tests, and most of what we do have makes more noise than signal 😞

So nothing in particular to add for now.

src/debugging/python/PythonDebugHelper.ts
Kludex: Perfect. Thank you!

bwateratmsft:
No problem. I'll try to find some time to test these changes out.

bwateratmsft:
Ok, I was able to test it out, bumped into a couple of issues.

1. I need to revert #2732 and fix it a different way. My bad!
2. The grouping logic for the "Now listening" regex wasn't right; non-capturing groups did not work as I'd expected. Also I had a typo (missing parenthesis, fixed it above). This is the correct thing to have:
```typescript
    case 'fastapi':
        return 'Uvicorn running on (https?://\\S+|[0-9]+)';
```
The reason we don't need a listener for the Gunicorn approach is that when debugging--which is the only time this code will be active--we'll always be using Uvicorn to launch (more on that next...)

3. Similar to Flask, rather than running the _file_, we have to run the Uvicorn module in the `docker-run` task, with arguments pointing to the entrypoint. So, instead of the task being this:
```json
        {
            "type": "docker-run",
            "label": "docker-run: debug",
            "dependsOn": [
                "docker-build"
            ],
            "python": {
                "args": [
                    "--host",
                    "0.0.0.0",
                    "--port",
                    "8000"
                ],
                "file": "main.py"
            }
        }
```

It needs to be this:
```json
        {
            "type": "docker-run",
            "label": "docker-run: debug",
            "dependsOn": [
                "docker-build"
            ],
            "python": {
                "args": [
                    "main:app",
                    "--host",
                    "0.0.0.0",
                    "--port",
                    "8000"
                ],
                "module": "uvicorn"
            }
        }
```

The code for generating this task is [here](https://github.com/microsoft/vscode-docker/blob/main/src/tasks/python/PythonTaskHelper.ts#L46-L78), and it contains an example of how the Flask task gets some custom handling--you'd have something very similar, like:
```typescript
// ... (the Flask code)
} else if (options.projectType === 'fastapi') {
    runOptions.args.unshift(`${path.basename(runOptions.file, '.py')}:app`); // Also, an `import * as path from 'path';` at the top

    runOptions.module = 'uvicorn';
    runOptions.file = undefined;
}
```

Kludex:
> I need to revert #2732 and fix it a different way. My bad!

I didn't understand this.

For the other points: got it! It's more clear now.
I'm going to apply those comments at some point between today and Sunday, jfyk.

I'll probably have to bother you to test it again after (if I can't fix the issue I mentioned above).

Thank you for the comments, @bwateratmsft !

bwateratmsft:
> > I need to revert #2732 and fix it a different way. My bad!
> 
> I didn't understand this.

I introduced a bug with #2732 so I gotta fix it or debugging won't work. I have a PR now: #2741

Kludex:
@bwateratmsft I've applied the changes. Thank you for the update @bwateratmsft :)

src/debugging/python/PythonDebugHelper.ts
Kludex: Applied.

bwateratmsft (approved):
> @bwateratmsft I've applied the changes. Thank you for the update @bwateratmsft :)
> 
> Unfortunately I still need to bother you with the testing 😢

Worked like a charm! Didn't encounter any issues on either Window or Mac. Usually Mac means Linux works too, I'm not worried about it.

This is exciting! Thanks @Kludex!

Kludex:
COOL! :tada: 

@bwateratmsft Thank for all the help! :)

PR #2746, opened 2021-03-03
Fix up right-click behavior by not changing tree IDs so much

### LINKED ISSUE #2744: Containers frequently changing tree ID causes right-click actions to not work well
Regression due to #2733. We may want to revert it.

When a container is first started up, the tree ID is relatively rapidly changing (since it depends on the timestamp now). Unfortunately, if the tree ID changes after you right click (which only takes ~2 seconds until the timestamp starts reading things like "X minutes ago" instead of "X seconds ago"), then instead of the action "just working", it drops the quick pick down to find a tree item. This is pretty annoying.

bwateratmsft:
Fixes #2744. When the tree ID changes, if you right clicked the object with the "old" ID, then choose an action--that item is lost now. It will drop the quick pick to make you go find the container again. This is really annoying.

During the first minute of a container's life, the tree ID is changing every ~2 seconds, which makes right click actions very difficult to do.

This change makes it so that we consider the container label for the purposes of deciding whether or not to refresh, but without changing its tree ID, so right click contexts isn't lost.

### DIFF
--- a/src/tree/containers/ContainersTreeItem.ts
+++ b/src/tree/containers/ContainersTreeItem.ts
@@ -128,7 +128,7 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<DockerContainerInf
             return false;
         }
 
-        // Containers' labels/descriptions (status in particular) can change. If they do, we want to cause a refresh. But, we also don't want to change the tree ID based on status.
+        // Containers' labels/descriptions (status in particular) can change. If they do, we want to cause a refresh. But, we also don't want to change the tree ID based on status (in `getTreeId()` in LocalRootTreeItemBase.ts).
         return !array1.some((item, index) => {
             return this.getTreeItemLabel(item) !== this.getTreeItemLabel(array2[index]) ||
                 this.getTreeItemDescription(item) !== this.getTreeItemDescription(array2[index]);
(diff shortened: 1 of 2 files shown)


src/tree/containers/ContainersTreeItem.ts
```diff
@@ -87,7 +88,13 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<DockerContainerInf
             case 'State':
                 return item.State;
             case 'Status':
-                return item.Status;
+                // We make status a sticky value. The result is that in the life of a DockerContainerInfo object, the Status property value will not change after it is first accessed.
```
bwateratmsft, line 91: Don't need this anymore

src/tree/containers/ContainersTreeItem.ts
```diff
@@ -122,6 +122,19 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<DockerContainerInf
         return super.getTreeItemForEmptyList()
     }
 
+    protected areArraysEqual(array1: DockerContainerInfo[] | undefined, array2: DockerContainerInfo[] | undefined): boolean {
+        if (!super.areArraysEqual(array1, array2)) {
+            // If they aren't the same according to the base class implementation, they are definitely not the same according to this either
+            return false;
+        }
+
+        // Containers' labels/descriptions (status in particular) can change. If they do, we want to cause a refresh. But, we also don't want to change the tree ID based on status.
```
karolz-ms, line 131: The part about tree ID: IMO it is a bit unclear how it relates to the code in this file. Suggest mentioning, or adding a reference to `getTreeId()` function.

karolz-ms (approved):

PR #2767, opened 2021-03-10
Add a What's New blurb to the start page

### LINKED ISSUE #2647: Increase awareness for new features
To increase awareness, I suggest we:

- Add Wordings that mention our new Compose Subset of services feature. We can put it in a higher section that is separated from the current messaging. Within that blurb, we should link them to documentation on how to enable it/use it with tasks and whatever else 
- Bump up the amount of users that see our getting started page upon installation/update

Editing to encompass files, subset, and tooltips

bwateratmsft:
Fixes #2647

image

### DIFF
--- a/resources/startPage.html.template
+++ b/resources/startPage.html.template
@@ -94,7 +94,7 @@
                 </div>
                 <div>
                     <ul>
-                        <li>The <a href="command:workbench.view.extension.dockerView">Docker Explorer</a> view now has advanced tooltips! Hover your mouse over an item to see useful info.</li>
+                        <li>The <a href="command:workbench.view.extension.dockerView">Docker Explorer</a> now has advanced tooltips! Hover your mouse over an item to see useful info.</li>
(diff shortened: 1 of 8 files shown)


resources/startPage.html.template
```diff
@@ -83,9 +83,28 @@
             </div>
         </div>
         <hr />
+        {{#if showWhatsNew}}
         <div class="row">
             <div class="icon">
-                <i class="codicon codicon-new-file" aria-label="Add Docker files"></i>
+                <i class="codicon codicon-star-empty" alt=""></i>
```
bwateratmsft, line 89: Since these images are purely decorative, I changed from `aria-label="..."` to `alt=""`, which will cause the images to be skipped by screen readers, which is desirable.

ucheNkadiCode (commented):
Just need to take the "view" out of Docker Explorer view and we're gucci

BigMorty:
I love the new Whats New section!!!

karolz-ms (approved):

PR #2786, opened 2021-03-16
Use global listener for debug start/stop

### LINKED ISSUE #2676: Containers not being removed after debugging when using hotkeys to restart debugging
**Summary:**

When using `removeContainerAfterDebug` in my `launch.json`, containers aren't being removed after debugging when using the Ctrl+Shift+F5 hotkey to restart debugging. They're removed just fine when clicking the restart button, just not when using the hotkey.

**How to reproduce:**

1. Set up docker container debugging with vscode (I'm using node.js)
2. Set `removeContainerAfterDebug` to `true` in `launch.json` in the launch configuration object
3. Start a container and begin debugging
4. Restart debugging using the hotkey (Ctrl+Shift+F5)
5. The previous container is not removed as it should be

**Expected behavior:**

The container should be removed after debugging, the same as when you click the restart debugging button with the mouse.

bwateratmsft:
Fixes #2676

On start: emit ports
On stop: remove debug container (if `removeContainerAfterDebugging` is `true` or `undefined`)

### DIFF
--- a/src/debugging/DockerDebugConfigurationProvider.ts
+++ b/src/debugging/DockerDebugConfigurationProvider.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { CancellationToken, commands, debug, DebugConfiguration, DebugConfigurationProvider, MessageItem, ProviderResult, window, workspace, WorkspaceFolder } from 'vscode';
-import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
+import { CancellationToken, commands, debug, DebugConfiguration, DebugConfigurationProvider, DebugSession, MessageItem, ProviderResult, window, workspace, WorkspaceFolder } from 'vscode';
+import { callWithTelemetryAndErrorHandling, IActionContext, registerEvent } from 'vscode-azureextensionui';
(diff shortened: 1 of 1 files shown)


src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -26,7 +26,21 @@ export interface DockerAttachConfiguration extends NetCoreDockerDebugConfigurati
 export class DockerDebugConfigurationProvider implements DebugConfigurationProvider {
     public constructor(
         private readonly helpers: { [key in DockerPlatform]: DebugHelper }
-    ) { }
+    ) {
+        // Listen for debug termination events to shut down debug containers as needed
+        registerEvent('debugTermination', debug.onDidTerminateDebugSession, async (context: IActionContext, session: DebugSession) => {
```
karolz-ms, line 31: Just one question: should we un-register these events when the extension is deactivated?

karolz-ms (approved):

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: That actually does happen since `registerEvent` pushes it onto `extensionContext.subscriptions`: https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/registerEvent.ts

karolz-ms: Perfect!

PR #2791, opened 2021-03-17
Turn off SSL by default in .NET apps

### LINKED ISSUE #1835: .NET: Investigate creating and using a non-root user by default
Investigate creating and using a non-root user in the container by default.

Related to #1766.

bwateratmsft:
Fixes #1835.

A couple main changes--

- Default ports are changed from 80, 443 to 5000
- `ASPNETCORE_URLS` environment variable goes directly into the Dockerfile (to override the image's default of `http://+:80`)
- A user is created and switched to in the base layer of the Dockerfile
- If SSL or user secrets are enabled, the image is inspected to find a username, and that username used to determine the user's home directory--which is where the stuff needs to be mapped to--so that things can work for both root and non-root seamlessly

If a user wants to re-enable SSL in this new code, there are three steps:
1. Add an `EXPOSE` statement to the Dockerfile, to expose an additional port, e.g.: `EXPOSE 5001`
    - If using ports < 1024, they will need to disable the lines in the Dockerfile that drop down to a low-rights user
1. In tasks.json for the `docker-run: debug` task config, set `"configureSsl": true` under `"netCore"`, e.g.:
```json
"netCore": {
    "appProject": "${workspaceFolder}/NetNoSsl.csproj",
    "enableDebugging": true,
    "configureSsl": true
}
```
3. Also in tasks.json, set an environment variable to override the value in the Dockerfile:
```json
"dockerRun": {
    "env": {
        "ASPNETCORE_URLS": "https://+:5001;http://+:5000"
    }
},
```

# Verifications
### Console apps:
- [x] Debugging works
- [x] Debugging from docker-compose.debug.yml works

### Without SSL:
- [x] Debugging works
- [x] Debugging from docker-compose.debug.yml works
- [x] Run full image locally works
- [x] Deploy to App Service works

### SSL re-enabled:
- [x] Debugging works
- [x] Debugging from docker-compose.debug.yml works
- [x] Deploy to App Service works

### Console apps:
- [x] Debugging works
- [x] Debugging from docker-compose.debug.yml works

### Without SSL:
- [x] Debugging works
- [x] Debugging from docker-compose.debug.yml works
- [x] Run full image locally works
- [x] Deploy to App Service works

### SSL re-enabled:
- [x] Debugging works
- [x] Debugging from docker-compose.debug.yml works
- [x] Deploy to App Service works

## Overall
- [ ] Come up with docs to explain how to turn on SSL (#2805)


resources/templates/netCore/Dockerfile.template
```diff
@@ -4,6 +4,18 @@ WORKDIR /app
...
 {{/each}}
 
+{{#if (eq platform '.NET: ASP.NET Core')}}
+{{#if ports}}
+ENV ASPNETCORE_URLS=http://+:{{ ports.[0] }}
+{{/if}}
+{{/if}}
+
+{{#unless (isRootPort ports)}}
+# Switching to a non-root user, please refer to https://aka.ms/vscode-docker-python-user-rights
```
bwateratmsft, line 14: TODO: fix the link

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -154,19 +146,15 @@ export class NetCoreTaskHelper implements TaskHelper {
         runOptions.os = runOptions.os || 'Linux';
         runOptions.image = inferImageName(runDefinition as DockerRunTaskDefinition, context, context.folder.name, 'dev');
 
-        const ssl = helperOptions.configureSsl !== undefined ? helperOptions.configureSsl : await NetCoreTaskHelper.inferSsl(context.folder, helperOptions);
+        const ssl = !!helperOptions.configureSsl; // SSL will be enabled only if helperOptions.configureSsl is explicitly true
+        context.actionContext.telemetry.properties.netCoreSslSetting = helperOptions.configureSsl === undefined ? 'undefined' : helperOptions.configureSsl.toString();
```
karolz-ms, line 150: Just curious: we prefer to see "undefined" in telemetry, rather than empty string, or "(not set)" or something?

karolz-ms (approved):

resources/templates/netCore/Dockerfile.template
bwateratmsft: Made one, https://aka.ms/vscode-docker-dotnet-configure-containers. Currently points to the Python doc.

src/tasks/netcore/NetCoreTaskHelper.ts
bwateratmsft: I do prefer an explicit value vs. blank, as blank could imply that the telemetry pipeline dropped it, or it's an old version that doesn't set it, etc. As for "undefined" vs "unset" it doesn't matter much, but JS uses "undefined" all over and we're using "true" and "false" for explicitly-set values, so it seems good enough.

PR #2796, opened 2021-03-18
Use managed service identity for pulling from ACR

### LINKED ISSUE #1685: Deploy from ACR to App Service should use MSI with ACRPull role
Many users attempt to enable the *admin mode* on ACR. This is bad for security. At the same time, newly created ACR instances don't provide App Service with the necessary access for deployment, even via the VS Docker extension. Many users wrongly believe enabled admin mode is the only way out.

This is a feature request specific to the deployment from ACR to App Service / Web App which will allow making the deployment seamless.
1. Retrieve Web App Managed Identity or Create (Assign) a new one of it does not exist
1. Assign ACR Pull role to the system managed identity for the ACR in question.

References:
- [App Service Managed Service Identity](https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/services-support-managed-identities#azure-app-service)
- [How to use Managed Service Identities with ACR](https://docs.microsoft.com/azure/container-registry/container-registry-authentication-managed-identity#example-2-access-with-a-system-assigned-identity)

bwateratmsft:
Fixes #1685 so we can finally stop punting it. :smile:

Still need to verify:
- [x] The newly-added `@azure/arm-authorization` package does _not_ get loaded before `Deploy to Azure App Service` is called
    - [x] Unpacked
    - [x] Webpacked
- [x] Unconditionally setting the `identity` field doesn't cause problems with deployments from other container registries (e.g. Docker Hub)

### DIFF
--- a/src/commands/registries/azure/DockerAssignAcrPullRoleStep.ts
+++ b/src/commands/registries/azure/DockerAssignAcrPullRoleStep.ts
@@ -1,0 +1,56 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Progress } from "vscode";
+import { IAppServiceWizardContext } from "vscode-azureappservice"; // These are only dev-time imports so don't need to be lazy
+import { AzureWizardExecuteStep, createAzureClient } from "vscode-azureextensionui";
+import { ext } from "../../../extensionVariables";
+import { localize } from "../../../localize";
+import { AzureRegistryTreeItem } from '../../../tree/registries/azure/AzureRegistryTreeItem';
+import { RemoteTagTreeItem } from '../../../tree/registries/RemoteTagTreeItem';
+
+export class DockerAssignAcrPullRoleStep extends AzureWizardExecuteStep<IAppServiceWizardContext> {
+    public priority: number = 141; // execute after DockerSiteCreateStep
+
+    public constructor(private readonly tagTreeItem: RemoteTagTreeItem) {
+        super();
+    }
+
+    public async execute(context: IAppServiceWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+        const message: string = localize('vscode-docker.commands.registries.azure.deployImage.assigningPullRole', 'Granting permission for App Service to pull image from ACR...');
+        ext.outputChannel.appendLine(message);
+        progress.report({ message: message });
+
+        const armAuth = await import('@azure/arm-authorization');
+        const armContainerRegistry = await import('@azure/arm-containerregistry');
+        const armAppService = await import('@azure/arm-appservice');
+        const authClient = createAzureClient(context, armAuth.AuthorizationManagementClient);
+        const crmClient = createAzureClient(context, armContainerRegistry.ContainerRegistryManagementClient);
+        const appSvcClient = createAzureClient(context, armAppService.WebSiteManagementClient);
+
+        // If we're in execute, then `shouldExecute` passed and `this.treeItem.parent.parent` is guaranteed to be an AzureRegistryTreeItem
+        const registryTreeItem: AzureRegistryTreeItem = this.tagTreeItem.parent.parent as AzureRegistryTreeItem;
+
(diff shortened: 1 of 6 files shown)


src/commands/registries/azure/DockerSiteCreateStep.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { WebSiteManagementClient, WebSiteManagementModels } from '@azure/arm-appservice'; // These are only dev-time imports so don't need to be lazy
+import { Progress } from "vscode";
+import { IAppServiceWizardContext } from "vscode-azureappservice"; // These are only dev-time imports so don't need to be lazy
+import { AzureWizardExecuteStep, createAzureClient } from "vscode-azureextensionui";
+import { ext } from "../../../extensionVariables";
+import { localize } from "../../../localize";
+import { nonNullProp, nonNullValueAndProp } from "../../../utils/nonNull";
+
+export class DockerSiteCreateStep extends AzureWizardExecuteStep<IAppServiceWizardContext> {
```
bwateratmsft, line 14: This code is not new, just refactored out from `deployImageToAzure`.

src/commands/registries/azure/DockerAssignAcrPullRoleStep.ts
```diff
@@ -0,0 +1,56 @@
...
+        progress.report({ message: message });
+
+        const armAuth = await import('@azure/arm-authorization');
+        const armContainerRegistry = await import('@azure/arm-containerregistry');
+        const armAppService = await import('@azure/arm-appservice');
+        const authClient = createAzureClient(context, armAuth.AuthorizationManagementClient);
+        const crmClient = createAzureClient(context, armContainerRegistry.ContainerRegistryManagementClient);
+        const appSvcClient = createAzureClient(context, armAppService.WebSiteManagementClient);
+
+        // If we're in execute, then `shouldExecute` passed and `this.treeItem.parent.parent` is guaranteed to be an AzureRegistryTreeItem
```
bwateratmsft, line 33:
```suggestion
        // If we're in `execute`, then `shouldExecute` passed and `this.tagTreeItem.parent.parent` is guaranteed to be an AzureRegistryTreeItem
```

src/commands/registries/azure/DockerAssignAcrPullRoleStep.ts
```diff
@@ -0,0 +1,56 @@
...
+        const armAppService = await import('@azure/arm-appservice');
+        const authClient = createAzureClient(context, armAuth.AuthorizationManagementClient);
+        const crmClient = createAzureClient(context, armContainerRegistry.ContainerRegistryManagementClient);
+        const appSvcClient = createAzureClient(context, armAppService.WebSiteManagementClient);
+
+        // If we're in `execute`, then `shouldExecute` passed and `this.tagTreeItem.parent.parent` is guaranteed to be an AzureRegistryTreeItem
+        const registryTreeItem: AzureRegistryTreeItem = this.tagTreeItem.parent.parent as AzureRegistryTreeItem;
+
+        // 1. Get the registry resource. We will need the ID.
+        const registry = await crmClient.registries.get(registryTreeItem.resourceGroup, registryTreeItem.registryName);
```
karolz-ms, line 37: As discussed via chat, consider adding some error handling if one of these Azure REST calls fails

karolz-ms (commented):
LGTM, just assume you will attempt to grant ACR access before deploying the image, so waiting with approval to see if that is successful

src/commands/registries/azure/DockerSiteCreateStep.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        const armAppService = await import('@azure/arm-appservice');
+        const client: WebSiteManagementClient = createAzureClient(context, armAppService.WebSiteManagementClient);
+        context.site = await client.webApps.createOrUpdate(nonNullValueAndProp(context.resourceGroup, 'name'), nonNullProp(context, 'newSiteName'), {
+            name: context.newSiteName,
+            kind: 'app,linux',
+            location: nonNullValueAndProp(context.location, 'name'),
+            serverFarmId: nonNullValueAndProp(context.plan, 'id'),
+            siteConfig: this.siteConfig,
+            identity: {
```
bwateratmsft, line 34: Will unconditionally setting this cause problems deploying from other registries? I don't think so but verify.

karolz-ms (approved):

PR #2808, opened 2021-03-23
Support sovereign clouds with ACI features

### LINKED ISSUE #2775: Fail to create an ACI context in Azure China/Azure Germany account
**OS:** Win 10
**Build Version:** 20210312.1
**Docker Desktop:** 3.2.1 (61626)
image

**Repro Steps:**
1. Sign in with Azure China account.
2. Try to create an ACI context.

**Expect:**
Succeed to create an ACI context.

**Actual:**
Fail to create an ACI context.
```
Creating ACI context "serere-ch1"...
Error: Process 'docker context create aci serere-ch1 --subscriptio...' exited with code 1
Error: could not find the requested subscription from your Azure login. You might need to specify a tenant ID with docker login azure --tenant-id xxx 
```
**More Info:**
This issue also reproduces in Azure Germany account.

bwateratmsft:
Fixes #2775.

In the "Create ACI Context" case, the user must already be signed in to the Azure Account extension--we use it to have them pick a subscription and resource group--so knowing the cloud ID is trivial.

In the "Deploy to ACI" case, the Azure Account extension is not required, simply having an existing ACI context is sufficient. There's not a good way to know what region an ACI context belongs to with any `docker context inspect` / etc. calls. So, if deployment fails due to expired token / not signed in / etc., we will have to prompt the user for which cloud to sign in to.

### DIFF
--- a/src/tree/contexts/aci/AciContextCreateStep.ts
+++ b/src/tree/contexts/aci/AciContextCreateStep.ts
@@ -7,6 +7,7 @@ import { Progress } from 'vscode';
 import { AzureWizardExecuteStep, parseError } from 'vscode-azureextensionui';
 import { ext } from '../../../extensionVariables';
 import { localize } from '../../../localize';
+import { promptForAciCloud } from '../../../utils/azureUtils';
 import { executeAsTask } from '../../../utils/executeAsTask';
 import { execAsync } from '../../../utils/spawnAsync';
 import { IAciWizardContext } from './IAciWizardContext';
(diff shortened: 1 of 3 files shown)


src/tree/contexts/aci/AciContextCreateStep.ts
```diff
@@ -30,7 +30,7 @@ export class AciContextCreateStep extends AzureWizardExecuteStep<IAciWizardConte
             if (error.errorType === '5' || /not logged in/i.test(error.message)) {
                 // If error is due to being not logged in, we'll go through login and try again
                 // Because login could involve device auth we do this step in the terminal
-                await executeAsTask(wizardContext, 'docker login azure', localize('vscode-docker.commands.contexts.create.aci.azureLogin', 'Azure Login'), { rejectOnError: true });
+                await executeAsTask(wizardContext, `docker login azure --cloud-name ${wizardContext.environment.name}`, localize('vscode-docker.commands.contexts.create.aci.azureLogin', 'Azure Login'), { rejectOnError: true });
```
bwateratmsft, line 33: TODO: Verify this cloud environment name is the same as what we expect

karolz-ms (approved):

src/tree/contexts/aci/AciContextCreateStep.ts
bwateratmsft: Yep, it works :thumbsup:

PR #2817, opened 2021-03-27
Updating README with guidance to use rootless Docker for linux users

aro5000:
I noticed on the installation page in vscode that linux users are recommended to enable the Docker CLI for non-root users, but this can have some significant security considerations. I am proposing to add a section before this that recommends a rootless Docker configuration. I am running Ubuntu Desktop 20.04 and rootless Docker appears to work well with the extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The Docker extension makes it easy to build, manage, and deploy containerized ap
 
 [Install Docker](https://docs.docker.com/install/) on your machine and add it to the system path.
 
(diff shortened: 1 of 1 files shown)


karolz-ms (approved):
Thank you @aro5000 , makes sense

bwateratmsft (approved):
I'm lacking a Linux VM to try this out with but I'll trust @aro5000 that this works :)

ucheNkadiCode:
I approve this after changing (most secure) to (more secure)

aro5000:
@ucheNkadiCode it looks like only those with write access to the repo can merge. Is there anything else I can provide for this? :)

bwateratmsft:
Merging it. Thanks a ton @aro5000!

PR #2823, opened 2021-03-30
Use `adduser` instead of `useradd`

### LINKED ISSUE #2821: Fail to build the alpine image for a .NET Core project
**OS:** Win 10
**Build Version:** 20210329.3
**.NET Core SDK:** 5.0.201

**Repro Steps:**
1. Open a .NET Core MVC project in VS Code Insiders.
2. Add Docker support for this project.
3. Change the image to "5.0-alpine" in Dockerfile.
4. Right click the Dockerfile -> Select "Build Image...".
5. Check whether the alpine image is built successfully or not.

**Expect:**
The alpine image is built successfully.

**Actual:**
Fail to build the alpine image.
image

**More Info:**
1. This issue does not reproduce for Ubuntu and Debian images.
2. This issue also reproduces for .NET Core Console app.

bwateratmsft:
Fixes #2821 using a command that works on both Alpine and Buster images, for both .NET and Python. However I think it's kind of ugly. Alternatively, we could won't-fix or use documentation. @ucheNkadiCode @karolz-ms what do you think? Personally I'd say just live with the ugliness but I'm interested to hear what you think.

### DIFF
--- a/resources/templates/netCore/Dockerfile.template
+++ b/resources/templates/netCore/Dockerfile.template
@@ -14,7 +14,7 @@ ENV ASPNETCORE_URLS=http://+:{{ ports.[0] }}
 {{#if (eq netCorePlatformOS 'Linux')}}
 # Creates a non-root user with an explicit UID and adds permission to access the /app folder
 # For more info, please refer to https://aka.ms/vscode-docker-dotnet-configure-containers
-RUN useradd -u 5678 appuser && chown -R appuser /app
+RUN adduser -u 5678 --disabled-password --gecos "" appuser && chown -R appuser /app
 USER appuser
 
 {{/if}}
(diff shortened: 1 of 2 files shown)


karolz-ms (approved):

ucheNkadiCode (approved):

ucheNkadiCode:
I would say that we can maybe add documentation or change the scaffolding to at least mention what `disabled-password` and `gecos` is doing. It's not as clean but if it fixes a potential user issue and is best practice, that is the goal

bwateratmsft:
Yeah, I think the docs would be a suitable place for that. `--disabled-password` is saying to create this user without a password--don't need one for a Docker container--and `--gecos ""` is saying to not prompt for a full name or anything like that.

PR #2830, opened 2021-03-31
Update version, changelog, notices for 1.12.0

bwateratmsft:


karolz-ms (approved):
Ship it!

BigMorty:
1.55 of VS Code just came out, aren't we being a little aggressive in requiring it so soon?

bwateratmsft:
> 1.55 of VS Code just came out, aren't we being a little aggressive in requiring it so soon?

@BigMorty It'll still be two weeks before we ship, but additionally--if the user is still on 1.54- of VSCode, then it will just wait to update the Docker extension. Nothing bad happens.

BigMorty:
Thanks, that works for me!

Issue #2764, opened 2021-03-10
An error occurs when building the image for the simple-go-web-app-master project


v-xinda:
**OS:** All
**Build Version:** 20210309.2

**Repro Steps:**

1. Open the simple-go-web-app-master project in VS Code.
2. F1 --> Click "Docker: Add Docker Files to Workspace..." --> Select "Go" as application platform --> Keep default 3000 as the port.
3. Right click on Dockerfile file --> Select "Build Image...".
4. Check whether build the image successfully or not.

**Expect:**
Build the image successfully without any error.

**Actual:**
An error occurs when building the image.
![image](https://user-images.githubusercontent.com/57662131/110604406-048eb800-81c3-11eb-93c7-3c2406af6dd4.png)

karolz-ms:
We should make the app use modules if it does not. Pretty much the whole Go world switched to modules by now. LMK if you need help with this.

bwateratmsft:
I know absolutely nothing about Go so I'd be happy to accept help. :smile:

karolz-ms:
We need to update the testcase to use a more modern app. The sample app the testcase currently uses has not been updated for 3 years.

karolz-ms:
Testcase updated

PR #2835, opened 2021-04-02
Fix Go Dockerfile template

karolz-ms:
Related to https://github.com/microsoft/vscode-docker/issues/2764


bwateratmsft (approved):
Does this fix #2764 entirely or just part of it?

karolz-ms:
This is the only necessary product change; the rest is just updating the testcase.

PR #2836, opened 2021-04-02
New linter settings + mountain of linter error fixes

bwateratmsft:
This simplifies our linter settings to be more or less the recommended with a few exceptions noted in `.eslintrc.json`.

Almost all of the linter errors fixed fall into a few categories:
1. Adding missing semicolons
1. Changing `let` to `const` for variables that never change
1. Removing escaping backslashes that aren't needed
1. Renaming some things to better follow recommended naming scheme
1. Empty blocks (usually a no-op catch), solved by leaving a comment inside

There are a couple of others that were either fixed or suppressed but nothing particularly interesting.

Files with meaningful changes that ought to be looked at:
`.eslintrc.js`, totally rewritten and renamed to...
`.eslintrc.json`
`package.json`, removed some older/deprecated/TSLinty packages


src/docker/Images.ts
```diff
@@ -10,6 +10,12 @@ export interface DockerImage extends DockerObject {
     readonly Size?: number;
 }
 
+export interface ImageInspectionContainers {
```
bwateratmsft, line 13: Slightly more interesting change...defined and exported this here so it can be referenced in `DockerodeUtils.ts`

src/docker/Volumes.ts
```diff
@@ -14,15 +14,17 @@ export interface DockerVolume extends DockerObject {
     readonly Description?: string;
 }
 
+export interface VolumeInspectionContainers {
```
bwateratmsft, line 17: Slightly more interesting change...defined and exported this here so it can be referenced in `DockerodeUtils.ts`

src/dockerCompose/dockerComposeHoverProvider.ts
```diff
@@ -11,32 +11,25 @@ import parser = require('../parser');
...
 
 export class DockerComposeHoverProvider implements HoverProvider {
-    public _parser: parser.Parser;
-    public _keyInfo: KeyInfo;
-
     // Provide the parser you want to use as well as keyinfo dictionary.
-    public constructor(wordParser: parser.Parser, keyInfo: KeyInfo) {
-        this._parser = wordParser;
-        this._keyInfo = keyInfo;
+    public constructor(public readonly parser: parser.Parser, public readonly keyInfo: KeyInfo) {
```
bwateratmsft, line 15: Slightly more interesting change; I don't want `public` variables to start with an underscore. That's weird.

src/scaffolding/wizard/ScaffoldDebuggingStep.ts
```diff
@@ -30,32 +30,38 @@ export class ScaffoldDebuggingStep extends AzureWizardExecuteStep<ScaffoldingWiz
         switch (wizardContext.platform) {
             case 'Node.js':
                 scaffoldContext.platform = 'node';
-                const nodeOptions: NodeScaffoldingOptions = {
-                    package: wizardContext.artifact,
-                };
-                await dockerDebugScaffoldingProvider.initializeNodeForDebugging(scaffoldContext, nodeOptions);
+                await dockerDebugScaffoldingProvider.initializeNodeForDebugging(
```
bwateratmsft, line 33: Slightly more interesting change (here and below). ESLint doesn't like doing declarations inside a `case` (without using `{ }` blocking), because the variable is in scope throughout the `switch`, but is not actually assigned if and until the `case` it is in hits.

philliphoff (approved):

karolz-ms (approved):
`eslintrc.json` looks very reasonable

PR #2844, opened 2021-04-05
Upgrade vscode-azuretools dependencies

bwateratmsft:
Should help with #2833 / #2843

Issue #2843, opened 2021-04-05
Push Docker Image to DockerHub get  Error Message: (n.stack || "").split is not a function


hunglee:
Does this occur consistently?  No 
Repro steps:

1. Push Image to DockerHub
2.

Action: vscode-docker.images.push
Error type: TypeError
Error Message: (n.stack || "").split is not a function

Version: 1.11.0
OS: win32
OS Release: 10.0.19041
Product: Visual Studio Code
Product Version: 1.55.0
Language: en

<details>
<summary>Call Stack</summary>

```
getCallstack extension.bundle.js:16:365761
Object.parseError extension.bundle.js:16:366112
new RegistryConnectErrorTreeItem extension.bundle.js:305:64657
DockerHubAccountTreeItem.loadMoreChildrenImpl extension.bundle.js:859:271956
processTicksAndRejections task_queues.js:97:5
```

</details>

bwateratmsft:
Duplicate of #2833

bwateratmsft:
Changing this to be a duplicate of #2839.

Issue #2833, opened 2021-04-01
(n.stack || "").split is not a function


Dynatrace-Adam-Gardner:
Does this occur consistently? 
Repro steps:

1. Use VSCode extension to push an image
2.

Action: vscode-docker.images.push
Error type: TypeError
Error Message: (n.stack || "").split is not a function

Version: 1.11.0
OS: win32
OS Release: 10.0.19042
Product: Visual Studio Code
Product Version: 1.55.0
Language: en

<details>
<summary>Call Stack</summary>

```
getCallstack extension.bundle.js:16:365761
Object.parseError extension.bundle.js:16:366112
new RegistryConnectErrorTreeItem extension.bundle.js:305:64657
DockerHubAccountTreeItem.loadMoreChildrenImpl extension.bundle.js:859:271956
processTicksAndRejections task_queues.js:97:5
```

</details>

bwateratmsft:
It looks like this is fixed by https://github.com/microsoft/vscode-azuretools/pull/860, which will show up in our next version (1.13.0).

bwateratmsft:
Changing this to be a duplicate of #2839.

PR #2848, opened 2021-04-06
Upgrade `vscode-azure*` dependencies

bwateratmsft:
This upgrades our `vscode-azure*` dependencies. I want to discuss with @karolz-ms / @philliphoff whether we should put this into 1.12.0 or wait until 1.13.0.

This is more or less the same as #2844 but with less bugs. Helps to mitigate #2833 / #2843.


karolz-ms (approved):

philliphoff (approved):
Always good to remove `webpack` configuration complexity.

PR #2853, opened 2021-04-07
Add ellipsis to download

### LINKED ISSUE #2851: [Suggestion] It would be better to add the ellipsis end of "Download"
**OS:** All
Build Version: 20210406.3

**Repro Steps:**

1. Open the Docker Containers view in VS Code -> Expand the "Files" node for a running container -> Hover over a file -> Hover over the download icon.
2. Check the tooltip.

**Actual:**
There is no ellipsis end of "Download".
image

**Suggestion:**
It would be better to add an ellipsis in the end of "Download".

bwateratmsft:
Fixes #2851. Also unrelatedly, disables the `SurveyManager` code since it is not in use and may not be in the future.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -85,7 +85,7 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats: 
         }
         ext.experimentationService = await createExperimentationService(ctx, targetPopulation);
 
-        // Temporarily disabled
+        // Temporarily disabled--reenable if we need to do any surveys
         // (new SurveyManager()).activate();
 
         validateOldPublisher(activateContext);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -86,7 +85,8 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
         }
         ext.experimentationService = await createExperimentationService(ctx, targetPopulation);
 
-        (new SurveyManager()).activate();
+        // Temporarily disabled
```
karolz-ms, line 88: I assume this is "enable if we need to do any surveys" kind of deal? If so, can you update the comment?

"Temporarily" can mean a lot of things 😆

karolz-ms (approved):

PR #2856, opened 2021-04-08
Use `workbench.action.files.openFileFolder` on Mac

### LINKED ISSUE #2846: No dialog pops up after clicking "Open Folder" button
**OS:** Mac
**Build Version:** 20210405.1

**Repro Steps:**

1. Install the docker extension -> F1 -> Type and select "Add Docker Files to Workspace... -> Click "Open Folder" when asking to open a folder or workspace.
2. Check whether open a dialog to select a folder or workspace.

**Expect:**
Pop up a dialog to select a folder or workspace.
Screen Shot 2021-04-06 at 4 08 51 PM

**Actual:**
No dialog pops up.

**More Info:**
This issue doesn't reproduce on Windows/Linux.

bwateratmsft:
Fixes #2846

### DIFF
--- a/src/utils/quickPickWorkspaceFolder.ts
+++ b/src/utils/quickPickWorkspaceFolder.ts
@@ -6,6 +6,7 @@
 import * as vscode from 'vscode';
 import { IActionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { localize } from '../localize';
+import { isMac } from './osUtils';
 
 export async function quickPickWorkspaceFolder(context: IActionContext, noWorkspacesMessage: string): Promise<vscode.WorkspaceFolder> {
     if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length === 1) {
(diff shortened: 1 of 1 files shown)


src/utils/quickPickWorkspaceFolder.ts
```diff
@@ -21,7 +22,12 @@ export async function quickPickWorkspaceFolder(context: IActionContext, noWorksp
         context.errorHandling.buttons = [
             {
                 callback: async () => {
-                    void vscode.commands.executeCommand('workbench.action.files.openFolder');
+                    if (isMac()) {
```
karolz-ms, line 25: So VS Code kept open https://github.com/microsoft/vscode/issues/120752 and indicated that `openFolder` might be registered on the Mac in future. Do you think we should revert to doing things uniformly on all platforms once that happens?

karolz-ms (approved):

src/utils/quickPickWorkspaceFolder.ts
bwateratmsft: That's a good question; I think the answer depends on whether `openFolder` on Mac forces opening _only_ folders, or if you can still open files. If it does enforce it, I definitely want to switch, in order to consistently enforce that folders must be opened. If it is not enforced--like "Open..."--maybe not.

karolz-ms: OK. I will let you decide whether and how you want to track this

bwateratmsft: Sounds good. I'm subscribed to the upstream issue ('cause I opened it), so I'll see it if they close it fixed/won't-fix.

philliphoff (approved):

Issue #2606, opened 2021-01-07
Modernize our webpacking, linting, testing, etc.


bwateratmsft:
Our webpacking strategy is getting dated and needs some TLC. Creating a new VSCode extension now (with `yo code`) results in a very different extension, with webpacking already set up, along with much more. This would allow us to modernize a lot of things and would be preferable to what we're doing now.

Checklist:
- [x] Use recommended linter settings instead of having a long list of rules and exceptions (#2836)
- [x] Use newer webpack loaders and plugins (#2883)
- [x] Use the newer `webpack watch` capabilities, alter tests to support this (#3078)
- [x] GitHub Actions for build (#3097)
- [x] ~~Split up webpack chunks to~~ optimize code loading time to the max (#3103)

bwateratmsft:
We should consider dumping Gulp as part of this. It hasn't been updated in years and results in a lot of Dependabot alerts.

bwateratmsft:
@karolz-ms @philliphoff the current state we're at is webpacking is linting is fixed and webpacking is modernified. The primary remaining work is to switch to `webpack --watch` for the dev loop. This is absolutely not urgent so it can wait.

I have some provisional changes to accomplish that in the [bmw/modernize](https://github.com/microsoft/vscode-docker/tree/bmw/modernize) branch; I wouldn't necessarily consider this work complete--it needs a bit more work--but it's getting there.

What I know is still needed in that branch:

- Make sure tests are running and successful--and also, make sure a failing test fails the build
- Update the AZDO yaml pipelines to use the new NPM scripts
- Probably branch off the GitHub Action this branch adds into a different PR

PR #2883, opened 2021-04-15
Modernize webpacking

bwateratmsft:
Part of #2606. This uses the more modern approach to webpacking that the Yeoman scaffolder creates in extensions by default these days. Also, no more gulp. So much less gulp.


package.json
```diff
@@ -2699,30 +2699,28 @@
...
-        "test-only": "gulp test-only",
-        "unit-test": "gulp unit-test",
-        "webpack-dev": "gulp webpack-dev",
-        "webpack-prod": "gulp webpack-prod"
+        "build": "tsc -p ./",
+        "ci-build": "npm run test",
+        "package": "vsce package",
+        "ci-package": "npm run test && vsce package",
+        "lint": "eslint src --ext ts",
+        "test": "node ./out/test/runTest.js",
```
bwateratmsft, line 2707: Tests are not currently working; the extension isn't loading when running them.

test/commands/selectCommandTemplate.test.ts
```diff
@@ -3,13 +3,14 @@
...
-import { TestInput } from 'vscode-azureextensiondev';
-import { IActionContext } from 'vscode-azureextensionui';
-import { testUserInput } from '../global.test';
-import assert = require('assert');
+// import { runWithExtensionSettings } from '../runWithExtensionSettings';
+// import { CommandTemplate, selectCommandTemplate, defaultCommandTemplates, ext, DockerContext, isNewContextType } from '../../extension.bundle';
+// import { IActionContext } from 'vscode-azureextensionui';
+// import assert = require('assert');
 
+// TODO: re-enable these tests
```
bwateratmsft, line 11: Do

package.json
```diff
@@ -2699,30 +2699,28 @@
...
-        "lint": "gulp lint",
-        "test": "gulp test",
-        "test-only": "gulp test-only",
-        "unit-test": "gulp unit-test",
-        "webpack-dev": "gulp webpack-dev",
-        "webpack-prod": "gulp webpack-prod"
+        "build": "tsc -p ./",
+        "ci-build": "npm run test",
+        "package": "vsce package",
+        "ci-package": "npm run test && vsce package",
```
bwateratmsft, line 2705: This probably ends up running webpack twice

.vscode/launch.json
```diff
@@ -61,27 +61,6 @@
                 "MOCHA_grep": "", // RegExp of tests to run (empty means all)
             }
         },
-        {
-            "name": "Launch Tests (unit)",
```
bwateratmsft, line 65: This launch config isn't really needed anymore. All our tests are very fast, and I have no intention of adding back slow ones. :smile:

package.json
```diff
@@ -2699,30 +2699,27 @@
...
-        "ci-build": "gulp ci-build",
-        "package": "gulp package",
-        "ci-package": "gulp ci-package",
-        "lint": "gulp lint",
-        "test": "gulp test",
-        "test-only": "gulp test-only",
-        "unit-test": "gulp unit-test",
-        "webpack-dev": "gulp webpack-dev",
-        "webpack-prod": "gulp webpack-prod"
+        "build": "tsc -p ./",
```
bwateratmsft, line 2702: These scripts are sorta messy but this isn't even their final form. However, in order to avoid a PR that was super unwieldy, I made them compatible with our current build scripts. I intend to make these neater when we make the switch to `webpack --watch`.

package.json
```diff
@@ -2732,20 +2729,16 @@
...
-        "gulp-eslint": "^6.0.0",
-        "gulp-sourcemaps": "^3.0.0",
-        "gulp-typescript": "^6.0.0-alpha.1",
         "mocha": "^8.2.1",
-        "ts-node": "^9.1.1",
+        "ts-loader": "^8.0.14",
         "typescript": "^4.2.4",
         "umd-compat-loader": "^2.1.2",
         "vsce": "^1.87.1",
-        "vscode-azureextensiondev": "^0.9.2",
```
bwateratmsft, line 2744: This takes out like...600 package dependencies

package.json
philliphoff: If only there were a build orchestration tool that could avoid that... :-) Eh, the pendulum now at "pure scripts" will inevitably swing back the other way again.

bwateratmsft: If only gulp weren't dead :smile:

philliphoff (approved):

webpack.config.js
```diff
@@ -67,8 +68,16 @@ const config = {
             patterns: [
                 './node_modules/vscode-azureextensionui/resources/**/*.svg',
                 './node_modules/vscode-codicons/dist/codicon.{css,ttf}',
+                './node_modules/open/xdg-open*',
```
bwateratmsft, line 71: This is a loose parity change with what's done in vscode-azureextensiondev, but stripped down to only what we need.

philliphoff: Maybe add a little more verbiage about what it's doing (or why it's required)?

philliphoff (approved):

PR #2924, opened 2021-05-12
Deploy container image to Azure Arc

ravipal:
Added support to deploy a container image from a registry to Azure Arc App Service.


eviemoore:
cpmment

src/commands/registries/azure/DockerSiteCreateStep.ts
```diff
@@ -4,37 +4,78 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementClient, WebSiteManagementModels } from '@azure/arm-appservice'; // These are only dev-time imports so don't need to be lazy
+import { Site } from '@azure/arm-appservice/esm/models';
```
karolz-ms, line 7: Should this be a lazy import?

src/commands/registries/azure/DockerSiteCreateStep.ts
```diff
@@ -4,37 +4,78 @@
...
+        const registryTreeItem: RegistryTreeItemBase = this.node.parent.parent;
+        if (registryTreeItem instanceof AzureRegistryTreeItem && context.customLocation) {
+            const appSettings: WebSiteManagementModels.NameValuePair[] = [];
+            const cred = await registryTreeItem.tryGetAdminCredentials();
+            if (!cred) {
+                throw new Error(localize('vscode-docker.commands.registries.azure.dockersitecreatestep.notAdminEnabled', 'Azure App service deployment on Azure Arc only supports running images from Azure Container Registries with admin enabled'));
+            } else {
+                appSettings.push({ name: "DOCKER_REGISTRY_SERVER_URL", value: registryTreeItem.baseUrl });
+                appSettings.push({ name: "DOCKER_REGISTRY_SERVER_USERNAME", value: cred.username });
+                appSettings.push({ name: "DOCKER_REGISTRY_SERVER_PASSWORD", value: nonNullProp(cred, 'passwords')[0].value });
```
philliphoff, line 67: Does `nonNullProp()` guarantee that there's at least one item in the property array, or just that the property itself is non-null?

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -46,17 +46,15 @@ export async function deployImageToAzure(context: IActionContext, node?: RemoteT
         promptSteps.push(subscriptionStep);
     }
 
-    promptSteps.push(...[
-        new vscAzureAppService.SiteNameStep(),
-        new ResourceGroupListStep(),
-        new vscAzureAppService.AppServicePlanListStep()
-    ]);
-    LocationListStep.addStep(wizardContext, promptSteps);
+    promptSteps.push(new vscAzureAppService.SiteNameStep());
```
karolz-ms, line 49: Not sure I understand the purpose of this change... is CustomLocationListStep **not supposed** to see the `AppServicePlanListStep`?

karolz-ms (commented):
Just a couple of questions...

src/commands/registries/azure/DockerSiteCreateStep.ts
ravipal: `notnullProp()` guarantees only the value it not null. The assumption is that if a password is of type array. This was the existing implementation (https://github.com/microsoft/vscode-docker/commit/83f016a589eb247fd2f5d3e3e49096ee73adc8b2#diff-4816881f8a5c146974926a452f283ecd1c0e52099fa13899c7b192ac290e366c)

src/commands/registries/azure/deployImageToAzure.ts
ravipal: CustomLocationListStep automatically creates a plan when a custom location is selected, so the AppServicePlanListStep will be skipped.

src/commands/registries/azure/DockerSiteCreateStep.ts
ravipal: It used for type definition in line 35. So it doesn't have to be lazy import. Added comment.

src/commands/registries/azure/deployImageToAzure.ts
karolz-ms: got it

karolz-ms (approved):

src/commands/registries/azure/DockerSiteCreateStep.ts
karolz-ms: cool

src/commands/registries/azure/deployImageToAzure.ts
karolz-ms: What I was missing is that the `CustomLocationListStep.addStep()` is a static method that creates and adds an instance of `CustomLocationListStep` to `promptSteps`. For some reason this line looked to me like a method on a `CustomLocationListStep` instance, which was taking a list of "prompt steps" 🤪 Now I understand.

ejizba (approved):

PR #2931, opened 2021-05-13
Add extendedLocation property in Plan and Site creation

ravipal:
When an App Service Plan or App Service is created in custom location we need to pass the following property.
    "extendedLocation": {
        "name":  <custom location id>,
        "type": "customLocation"
    },
vscode-azureappservice@0.82.0 sets this property in plan creation and added this property to site creation.


src/commands/registries/azure/DockerSiteCreateStep.ts
```diff
@@ -78,6 +79,38 @@ export class DockerSiteCreateStep extends AzureWizardExecuteStep<IAppServiceWiza
         }
     }
 
+    /**
+     * Has a few temporary workarounds so that the sdk allows some newer properties on the plan
```
karolz-ms, line 83: Please rephrase this comment so that it is clear what needs to be done and when.

1. If this method is to be entirely removed eventually, please say so. Under what circumstances can that happen?
2. If this method will stay, but portions ("temporary workarounds") are to be removed, please indicate which portion is to be removed, and under what circumstances.
3. What is "the sdk" mentioned here? azureextensionui? Some Azure SDK?

karolz-ms (changes requested):

src/commands/registries/azure/DockerSiteCreateStep.ts
ravipal: Updated the comment. Hope this make sense now.

karolz-ms: Yup, much better! thank you!

karolz-ms (approved):

Issue #2820, opened 2021-03-29
Webview is disposed


afaddoul:
Does this occur consistently? 
Repro steps:

1.
2.

Action: vscode-docker.help.openStartPage
Error type: Error
Error Message: Webview is disposed

Version: 1.11.0
OS: darwin
OS Release: 20.3.0
Product: Visual Studio Code
Product Version: 1.54.3
Language: en

<details>
<summary>Call Stack</summary>

```
_.assertNotDisposed extensionHostProcess.js:88:53320
_.set html [as html] extensionHostProcess.js:88:52689
StartPage.createOrShow extension.bundle.js:1053:146275
openStartPage extension.bundle.js:65:393543
```

</details>

bwateratmsft:
@afaddoul does this happen consistently?

afaddoul:
no, actually it was the first time

PavelSosin-320:
WebView? Is it one of 2 pieces of code (WebCiew, Chrome) that were urgently updated by Google and caused a total Internet blackout for 7 hours until people upgraded them in my country?

bwateratmsft:
@PavelSosin-320 this would be the webview for our start page. I don't think this issue is at all related to the Chrome update.

karolz-ms:
Pulling from Future to 1.13 since several people have reported it recently, see related issues, @dbreshears FYI

karolz-ms:
This was actually fixed in 0.13 release

PR #2959, opened 2021-05-21
Fix "webview is disposed" error

karolz-ms:
Fixes issue https://github.com/microsoft/vscode-docker/issues/2820

The poblem is likely due to a handful of async calls done beteween when the start page view is created and when its content is set. That gives the user an opportunity to close the start page before we attempt to set its content. It is especially tempting since during the time the content is computed the already-visible start page panel is empty.

This change front-loads that async processing so that the webview creation becomes fully synchronous.

### DIFF
--- a/src/commands/startPage/StartPage.ts
+++ b/src/commands/startPage/StartPage.ts
@@ -36,6 +36,15 @@ class StartPage {
         const codiconsRoot = vscode.Uri.joinPath(ext.context.extensionUri, ...ext.ignoreBundle ? ['node_modules'] : ['dist', 'node_modules'], 'vscode-codicons', 'dist');
 
         if (!this.activePanel) {
+            const template = await this.getTemplate(resourcesRoot);
+
+            let showWhatsNew = false;
+            try {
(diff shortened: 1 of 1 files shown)


src/commands/startPage/StartPage.ts
```diff
@@ -36,6 +36,15 @@ class StartPage {
         const codiconsRoot = vscode.Uri.joinPath(ext.context.extensionUri, ...ext.ignoreBundle ? ['node_modules'] : ['dist', 'node_modules'], 'vscode-codicons', 'dist');
 
         if (!this.activePanel) {
+            const template = await this.getTemplate(resourcesRoot);
```
philliphoff, line 39: With this `await`, there's a potential for reentrancy (because we haven't yet created the panel instance that would prevent this `if` branch to be taken again, though the calling function may effectively prevent this).  Maybe after the `await` test that `this.activePanel` is still `undefined`?

philliphoff (approved):

src/commands/startPage/StartPage.ts
karolz-ms: Good point. I have refactored the code a bit to prevent this

Issue #2547, opened 2020-12-08
Explorer: dangling images


BercziSandor:
I want an option to be able to see the dangling images as well.
These are currently not present.

karolz-ms:
This has been shipped in release 1.14.

PR #2967, opened 2021-05-25
Option to see dangling images in the explorer

jnsjunior:
#### Reference Issues/PRs

Explorer: dangling images #2547

#### Changes

Implement an option to see dangling images in the explorer

image

image


jnsjunior:
There is one more adjust, new commit.
When i remove dangling image, the fullTag is used to remove the image, or the RepoDigests when it's end with ':none'.
Some cases the digest end with '@none' and this cause an error (bellow).

image

image

image

So, i propose check digest end and use the image ID when digest end with '@none'

image

karolz-ms:
@jnsjunior thank you for your contribution, much appreciated! 

I think adding the ability to show dangling images makes a lot of sense; I am just wondering if a configuration setting is the best way to enable it. An alternative could be to add an enable/disable button to the Images pane in the Docker explorer (next to the existing "Prune" command, see screenshot below). This would be more accessible/discoverable... 🤔 

image

Also @BigMorty @philliphoff please share your thoughts.

jnsjunior:
@karolz-ms yes i agree with your opinion.
i will try to implement this function using a enable/disable button on the panel.

jnsjunior:
Here is,

image

src/tree/images/ImageTreeItem.ts
```diff
@@ -71,6 +71,11 @@ export class ImageTreeItem extends AzExtTreeItemIntermediate {
         if (ref.endsWith(':<none>') && this._item.RepoDigests?.length) {
```
karolz-ms, line 71: Technically the comment in the line above is no longer true (dangling images are shown, depending on the setting), please update it. 

I am also wondering if it makes sense to simplify this code and not use the image digest at all, but switch straight to image ID if the image tag ends with `<none>`

karolz-ms (changes requested):
Looks great, just a couple minor suggestions.

We are super grateful for this change and for following up on the suggestion to make it a command! 🤗

karolz-ms:
@philliphoff please share your thoughts as well if you can

package.json
```diff
@@ -2354,6 +2360,12 @@
                 "category": "%vscode-docker.commands.category.dockerImages%",
                 "icon": "$(clear-all)"
             },
+            {
+                "command": "vscode-docker.images.showDangling",
+                "title": "%vscode-docker.commands.images.showDangling%",
+                "category": "%vscode-docker.commands.category.dockerImages%",
+                "icon": "$(symbol-null)"
```
philliphoff, line 2367: A minor nit, but I wonder if either of the following icons might say "dangling images" more then `symbol-null`:

 - `outline-view-icon` 
Screenshot 2021-05-27 at 09 36 26

 - `testing-show-as-list-icon`

Screenshot 2021-05-27 at 09 37 03

src/commands/images/showDanglingImages.ts
```diff
@@ -0,0 +1,8 @@
+
+import { ext } from "../../extensionVariables";
+import { IActionContext } from 'vscode-azureextensionui';
+
+export async function showDanglingImages(context: IActionContext): Promise<void> {
+    const conf: boolean = await ext.context.globalState.get('vscode-docker.images.showDanglingImages') || false;
```
philliphoff, line 6: Nit: [`get<T>()`](https://code.visualstudio.com/api/references/vscode-api#Memento) has a form which accepts a default value (rather than testing the result).

src/docker/DockerodeApiClient/DockerodeApiClient.ts
```diff
@@ -255,7 +256,12 @@ export class DockerodeApiClient extends ContextChangeCancelClient implements Doc
     }
 
     public async getImages(context: IActionContext, token?: CancellationToken): Promise<DockerImage[]> {
-        const images = await this.callWithErrorHandling(context, async () => this.dockerodeClient.listImages({ filters: { "dangling": ["false"] } }), token);
+        const filters = {};
```
philliphoff, line 259: A semantic question, do we intend the toggle to control whether dangling images are included in all requests, or only when the UI actually *displays* a list of images?  If the latter, then perhaps `getImages()` should accept a `getDanglingImages` argument and the tree view manage getting the setting from global state.

karolz-ms: Great point. My vote would be for the setting to only apply to the UI (the second alternative)

package.json
jnsjunior: Yes, 
outline-view-icon is a good choice. I will commit this change.

src/tree/images/ImageTreeItem.ts
jnsjunior: Yes, the comment need to be changed. I will fix it and remove the ref by digest.

src/commands/images/showDanglingImages.ts
jnsjunior: Good tip. I will fix. Thanks.

src/docker/DockerodeApiClient/DockerodeApiClient.ts
jnsjunior: I vote to only apply to the UI.

karolz-ms: We are in agreement then. 

@jnsjunior please have the `getImages` function take an optional `includeDangling` parameter (defaulting to false), then have the [images tree item (plural)](https://github.com/microsoft/vscode-docker/blob/main/src/tree/images/ImagesTreeItem.ts#L60) check the `showDanglingImages` state parameter and call `getImages` accordingly. 

Thank you for working with us through the review!!

jnsjunior: @karolz-ms and @philliphoff thank you for support me for my first contribution for the vscode-docker :smiley:

karolz-ms (approved):

karolz-ms:
@jnsjunior thanks again for your contribution, much appreciated, and sorry for not merging sooner, we had a holiday here in the US yesterday ☺️

PR #2970, opened 2021-05-25
Merge rel/1.13.0 to main

ravipal:
Merge rel/1.13.0 to main

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-docker",
-    "version": "1.12.5-alpha",
+    "version": "1.13.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-docker",
-    "version": "1.12.5-alpha",
+    "version": "1.13.0",
```
karolz-ms, line 3: Just for the record, this need to be updated to 1.14.0-alpha, but that can happen as a separate PR

karolz-ms (approved):

PR #2971, opened 2021-05-25
Update version to 1.13.1-alpha

ravipal:
Update version to 1.13.1-alpha

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-docker",
-    "version": "1.13.0",
+    "version": "1.13.1-alpha",
     "publisher": "ms-azuretools",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-docker",
-    "version": "1.13.0",
+    "version": "1.13.1-alpha",
```
ravipal, line 3: Should this be 1.13.5-alpha?

philliphoff: `1.13.1-alpha` is fine.  It just needs to be at least `0.0.1` more then the previous version.

philliphoff (approved):

Issue #2004, opened 2020-05-20
What formatter does this extention use?


06kellyjac:
I'm looking to have my contiguous commands be indented extra but I can't find what formatter this extension uses to potentially change it's config file:

```Dockerfile
FROM debian:buster-slim
RUN \
  export DEBIAN_FRONTEND=noninteractive \
  && apt-get update \
  && apt-get install --assume-yes --no-install-recommends --no-install-suggests \
    bash \
    ca-certificates \
    curl \
    gnupg \
    unzip \
  && rm -rf /var/lib/apt/lists/*
```

---

Currently it formats like:

```Dockerfile
FROM debian:buster-slim
RUN \
  export DEBIAN_FRONTEND=noninteractive \
  && apt-get update \
  && apt-get install --assume-yes --no-install-recommends --no-install-suggests \
  bash \
  ca-certificates \
  curl \
  gnupg \
  unzip \
  && rm -rf /var/lib/apt/lists/*
```

bwateratmsft:
@rcjsuen?

06kellyjac:
Thanks for the swift reply, it helped me find it:

It looks like it's this project that holds the formatter https://github.com/rcjsuen/dockerfile-utils
I'll have a look through there for options/config

PR #2976, opened 2021-05-27
Enable formatter to skip instructions that span multiple lines

### LINKED ISSUE #992: Keep indentation for multiline statements
Here an exemple of Dockerfile part the way I like to write:
```dockerfile
RUN     true \
    &&  apt-get --quiet update --yes \
    &&  apt-get --quiet install --yes --no-install-recommends \
            curl \
            dos2unix \
            git \
            htop \
            iftop \
            iotop \
            jq \
            make \
            netcat-openbsd \
            python3-pip \
            vim \
    &&  true
```
I find it really efficient, merge-friendly, extendable, etc. So I want to keep it this way.

If I enable autoformatting and this extension, on save these lines are automatically rewriten to:
```dockerfile
 echo "Acquire::https::Proxy \"http://nexus.tls.renault.fr:911\"\;" >> /etc/apt/apt.conf \

RUN     true \
    &&  apt-get --quiet update --yes \
    &&  apt-get --quiet install --yes --no-install-recommends \
    curl \
    dos2unix \
    git \
    htop \
    iftop \
    iotop \
    jq \
    make \
    netcat-openbsd \
    python3-pip \
    vim \
    &&  true
```

rcjsuen:
Some users want to format instructions that span multiple lines in Dockerfiles themselves. The new setting added here will allow users to decide whether instructions that span multiple lines should be ignored by the formatter or not.

```Dockerfile
FROM scratch
RUN echo && \
            echo && \
                echo && \
        echo
```

1. Create the Dockerfile above.
2. Format the document. The lines in the `RUN` instruction will be formatted.
3. Undo the change so that it's back to its original form.
4. Open the `settings.json` file and add the following setting.
```JSON
{
    "docker.languageserver.formatter.ignoreMultilineInstructions": true
}
```
5. Format the document again. The `RUN` instruction should not be formatted.

This fixes #992 and #2004.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2166,6 +2166,12 @@
                     ],
                     "description": "%vscode-docker.config.docker.languageserver.diagnostics.instructionWorkdirRelative%"
                 },
+                "docker.languageserver.formatter.ignoreMultilineInstructions": {
+                    "scope": "resource",
+                    "type": "boolean",
+                    "default": false,
+                    "description": "%vscode-docker.config.docker.languageserver.formatter.ignoreMultilineInstructions%"
+                },
                 "docker.dockerComposeBuild": {
                     "type": "boolean",
                     "default": true,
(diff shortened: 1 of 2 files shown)


karolz-ms (approved):

karolz-ms:
Thank you, Remy! 🤗

rcjsuen:
@karolz-ms Thanks for the quick review. Have a good day!

Issue #2894, opened 2021-04-23
docker exe missing from path issue (MacOS) (again)


HannesOberreiter:
Description: 
Docker executable cannot be found. It throws following error:

```bash
exec: "com.docker.cli": executable file not found in $PATH
```

This problem was previously reported and fixed. It worked for me without any problems for a few weeks. Now it is broken again (at least for me).

Previous Discussion:
https://github.com/microsoft/vscode-docker/issues/2578

Discussion on Docker:
https://github.com/docker/compose-cli/issues/754

Discussion of possible fixed on Stackoverflow:
https://stackoverflow.com/questions/64120020/exec-com-docker-cli-executable-file-not-found-in-path

------
What I tried so far to fix it again:

- reinstall docker and cli
- reinstall vscode-docker
- (rebooting)

-----

Versions:

vscode-Docker: 1.12.1
Docker Engie: 20.10.5
Docker Desktop: 3.3.1
macOS Big Sur, 11.2.3 (Intel)

My results from PATH Checking:

![PATHS](https://user-images.githubusercontent.com/16878981/115824676-961b5780-a408-11eb-8b87-b5c74a13adc2.png)

Cheers
Hannes

karolz-ms:
@HannesOberreiter sorry to hear that. Did you try disabling "cloud experience" in Docker Desktop, as described by https://github.com/docker/compose-cli/issues/754#issuecomment-724809235. ?

HannesOberreiter:
@karolz-ms the option was removed I think with version 3.2.x.

Cheers
Hannes

karolz-ms:
I see. Hmm. Bummer.

Could you do us a favor and do the following troubleshooting:
1. Try to reproduce the issue. Do not close the VS Code terminal that reports Docker exe not found error.
2. Find the PID of the terminal process that reported the error (it will be one of the VS Code processes that has `--type=ptyHost` argument applied)
3. Do `ps eww <PID>` on that process and report if the PATH environment variable contains `/usr/local/bin`.

Thank you!

HannesOberreiter:
@karolz-ms sure, it is included in the PATH env.

![image](https://user-images.githubusercontent.com/16878981/115916077-905b5b80-a474-11eb-8e9a-fb114c269865.png)

LeaveMyYard:
I can confirm to have the same issue, both on macOS Big Sur on Intel chip and on macOS Big Sur on Apple chip.

[...]

PR #2984, opened 2021-06-03
Add setting for Docker CLI executable path

karolz-ms:
Provides a workaround for [issue 2894](https://github.com/microsoft/vscode-docker/issues/2894)


src/commands/containers/attachShellContainer.ts
```diff
@@ -46,7 +47,7 @@ export async function attachShellContainer(context: IActionContext, node?: Conta
             // If so use it, otherwise use sh
             try {
                 // If this succeeds, bash is present (exit code 0)
-                await execAsync(`docker exec -i ${node.containerId} sh -c "which bash"`);
+                await execAsync(`${dockerExePath()} exec -i ${node.containerId} sh -c "which bash"`);
```
philliphoff, line 50: Should the Docker path be quoted to allow for paths with whitespace?

karolz-ms: Good question. The user can include quotes in the path as necessary, for example the following works:

```json
"docker.dockerPath": "'/Users/karolz/docker local'"
```

(the above being hard link to `/usr/local/bin/docker`)

philliphoff: Definitely true, but do we expect the user to *know* that they should do that?  (We don't generally require users to self-quote their paths, for example, in the task/launch JSON properties.)

karolz-ms: Given that we do not expect it to be a common occurrence (that the path to Docker executable contains whitespace), I just updated the description of the setting to provide some guidance for the user.

package.nls.json
```diff
@@ -188,6 +188,7 @@
     "vscode-docker.config.docker.showRemoteWorkspaceWarning": "Set to true to prompt to switch from \"UI\" extension mode to \"Workspace\" extension mode if an operation is not supported in UI mode.",
     "vscode-docker.config.docker.scaffolding.templatePath": "The path to use for scaffolding templates.",
     "vscode-docker.config.docker.showStartPage": "Show the Docker extension Start Page when a new update is released.",
+    "vscode-docker.config.docker.dockerPath": "Absolute path to Docker client executable ('docker' command). If the path contains whitespace, it need to be quoted appropriately.",
```
philliphoff, line 191: Nit: "...it needs to be quoted..."

philliphoff (approved):

Issue #2884, opened 2021-04-15
Support for .NET debugging on Apple M1 (and ARM64)


philliphoff:
Updated debuggers for .NET are coming soon that support `arm64` (in particular, the Apple M1).  We need some changes to our debugging infrastructure to ensure the proper debugger is used within a `arm64` container.  See the [PoC](https://github.com/microsoft/vscode-docker/tree/philliphoff-arm64).  The logic could be a bit more nuanced to eliminate downloading debuggers unnecessarily, and the "attach" scenario needs updating as well.

Test matrix for launch and attach scenarios:

| Tested | OS | OS Arch | Container Arch | Notes |
| --- | --- | --- | --- | --- |
| ✅ | Mac | amd64 (Intel) | amd64 | Tested successfully. |
| ✅ | Mac | amd64 (Intel) | amd64 (alpine) | Tested successfully. |
| ✅ | Mac | arm64 (M1) | arm64 | Tested successfully. |
| ✅ | Mac | arm64 (M1) | arm64 (alpine) | Tested successfully. |
| ❓ | Mac | arm64 (M1) | amd64 (emulated) | Application runs in the debugger, but no breakpoints set are hit. Note: Error on building image: `qemu: uncaught target signal 11 (Segmentation fault) - core dumped` |
| ❓ | Mac | arm64 (M1) | amd64 (emulated) (alpine) | Application runs in the debugger, but no breakpoints set are hit. Note: Error on building image: `qemu: uncaught target signal 11 (Segmentation fault) - core dumped` |
| ✅ | Linux | amd64 (Intel) | amd64 | Tested successfully. |
| ✅ | Linux | amd64 (Intel) | amd64 (alpine) | Tested successfully. |
| ❌ | Linux | aarch64 | aarch64 | Not yet supported by OmniSharp. |
| ❌  | Linux | aarch64 | aarch64 (alpine) | Not yet supported by OmniSharp. |
| ✅ | Windows | amd64 (Intel) | amd64 | Tested successfully. |
| ✅ | Windows | amd64 (Intel) | amd64 (alpine) | Tested successfully. |
| ❓ | Windows | arm64 | arm64 | Need special hardware as Parallels does not support nested virtualization with the M1 Mac |

bwateratmsft:
Is the 'blocked' tag for awaiting the release of vsdbg?

philliphoff:
Yes.

Jonatthu:
[...]

PR #2985, opened 2021-06-03
Enable debugging .NET applications in `arm64` containers

philliphoff:
Enable debugging of .NET applications in `arm64` containers by installing new `arm64` flavors of the .NET debugger as well as new in-container detection script to choose the correct debugger from a now larger matrix of platforms/architectures.

> NOTE: This is intentionally limited to unblock users on `arm64` (e.g. M1) Macs; there are external limitations to enabling `arm64` support in general.

Related to #2884.


karolz-ms (approved):
The number of user OS - Docker daemon OS - container arch combinations makes my head spin a little, but your change looks good ☺️

philliphoff:
@karolz-ms Yes and, admittedly, this doesn't address all of them (e.g. still need to investigate why `arm64` `alpine` containers won't stop at breakpoints but otherwise run successfully).  Still, this at least unblocks some M1 scenarios without (fingers-crossed) breaking all of the original `amd64` combinations.

Issue #2953, opened 2021-05-18
Docker Start Page pops up for non-docker related activations


ucheNkadiCode:
I think that although this error happens once per update... it happens EVERY UPDATE of VS Code. So I feel like this bug directly influences customer satisfaction and is important to tackle.

Repro video: https://channel9.msdn.com/Shows/VS-Code-Livestreams/1-56-Release-Party#time=03m47s

karolz-ms:
The current implementation is that the page will pop up when we detect an updated version of the extension was activated. It should not happen just because of VS Code update, but it will happen for every VS Code copy that the user is using (the result of Docker extension upgrade check is not synced between machines).

As the extension can be activated for various reasons, including running *any* tasks (because it contributes a few custom tasks) it might seem that the start page pops up semi-randomly. IMO this happens too eagerly, and I propose that we show the start page for specific gestures instead of all extension activations. Here is what this gesture list might look like:

1. Show Docker view
2. Open a Dockerfile or Docker Compose file

The intention behind making the list so short and specific is to limit the (somewhat disruptive) showing of the start page to situations where the user already has decides to show a new piece of UI, and automatically showing another piece of UI will not be very jarring. 

@ucheNkadiCode @BigMorty @philliphoff @dbreshears  please share your thoughts

philliphoff:
There are a couple of [asks](https://github.com/microsoft/vscode/issues/44711#issuecomment-369183434) in VS Code to allow extensions to have more insight into why they're being activated, which would allow us to better tailor our experience, but they've not gotten much traction.

I agree with @karolz-ms that we may need to switch to a model where we show the start page only for a select set of commands (or language service initializations, etc.).  If we think that set is large, we might find a way to make it fairly generic in our command registration, so that we can just turn *off* the few commands/providers that we don't want to induce the start page.

karolz-ms:
We had a chat over email and this is the RC for the gestures that might trigger start page showing:

1. Show Docker Explorer (Docker view)
1. Open a Dockerfile or Docker Compose file
1. Add Docker / Docker Compose files to workspace
1. Install Docker
1. Container inspect
1. Image inspect

karolz-ms:
Fixed in release 1.14

PR #2990, opened 2021-06-07
Narrow the set of cases when we pop up start page after extension upgrade

karolz-ms:
Fixes [issue 2953](https://github.com/microsoft/vscode-docker/issues/2953)

### DIFF
--- a/src/scaffolding/wizard/OpenStartPageStep.ts
+++ b/src/scaffolding/wizard/OpenStartPageStep.ts
@@ -12,15 +12,13 @@ export class OpenStartPageStep extends AzureWizardExecuteStep<ScaffoldingWizardC
         super();
     }
 
-    // eslint-disable-next-line @typescript-eslint/naming-convention
-    public async execute(_wizardContext: ScaffoldingWizardContext, _progress: never): Promise<void> {
+    public async execute(): Promise<void> {
         // Wait for the page to be shown. This helps ensure that, if scaffolded files are to be opened,
         // they are opened last and are visible after the wizard is finished.
         await openStartPageAfterExtensionUpdate();
     }
 
-    // eslint-disable-next-line @typescript-eslint/naming-convention
-    public shouldExecute(_wizardContext: ScaffoldingWizardContext): boolean {
+    public shouldExecute(): boolean {
         return true;
     }
 }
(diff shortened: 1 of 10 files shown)


src/scaffolding/wizard/OpenStartPageStep.ts
```diff
@@ -0,0 +1,26 @@
...
+import { AzureWizardExecuteStep } from "vscode-azureextensionui";
+import { openStartPageAfterExtensionUpdate } from "../../commands/startPage/openStartPage";
+import { ScaffoldingWizardContext } from "./ScaffoldingWizardContext";
+
+export class OpenStartPageStep extends AzureWizardExecuteStep<ScaffoldingWizardContext> {
+    public constructor(public readonly priority: number) {
+        super();
+    }
+
+    // eslint-disable-next-line @typescript-eslint/naming-convention
```
karolz-ms, line 15: I am not entirely sure what is the best practice for unused function arguments...

philliphoff: If you don't use *any* of the arguments, you can simply omit them.  JS passes all arguments given it by the caller to the called function regardless of whether the function actually declares/uses them.

src/scaffolding/wizard/OpenStartPageStep.ts
```diff
@@ -0,0 +1,26 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardExecuteStep } from "vscode-azureextensionui";
+import { openStartPageAfterExtensionUpdate } from "../../commands/startPage/openStartPage";
+import { ScaffoldingWizardContext } from "./ScaffoldingWizardContext";
+
+export class OpenStartPageStep extends AzureWizardExecuteStep<ScaffoldingWizardContext> {
+    public constructor(public readonly priority: number) {
```
philliphoff, line 11: Is `priority` used for anything?

philliphoff (approved):

src/scaffolding/wizard/OpenStartPageStep.ts
karolz-ms: It is used by the wizard framework to order the steps, and is an abstract member, so TS is unhappy if it is not there.

Issue #2968, opened 2021-05-25
There is an error under the repository of Azure registry after deleting the image


v-xinda:
**OS:** All
**Build Version:** 1.13.0

**Repro Steps:**
1. Make sure connected Azure registry.
2. Right click a image -> Push -> Azure -> Select a registry (e.g. nicoledocker) -> Enter default value.
3. Navigate to Registry view -> Expand the Azure -> Expand "nicoledocker" registry ->Expand the repository -> image.
4. Right click the image -> Delete Image.
5. Check whether there is an error under the repository of Azure registry or not.

**Expect:**
There is no error under the repository of Azure registry after deleting the image.

**Actual:**
There is an error under the repository of Azure registry after deleting the image.
![image](https://user-images.githubusercontent.com/57662131/119456597-6942c300-bd6d-11eb-8374-cf2cb538d02a.png)

**More Info:**
1. The image is deleted successfully.
![image](https://user-images.githubusercontent.com/57662131/119457019-ce96b400-bd6d-11eb-9989-2ab3184d749f.png)
2. This issue can be fixed by refreshing the Registry view.

karolz-ms:
Bumping up the priority. I see this every time I delete an image from a registry and it makes us look sloppy.

PR #3043, opened 2021-07-07
Handle repositories that return 404 when all tags are removed

karolz-ms:
Fixes https://github.com/microsoft/vscode-docker/issues/2968

### DIFF
--- a/src/utils/httpRequest.ts
+++ b/src/utils/httpRequest.ts
@@ -8,7 +8,17 @@ import { default as fetch, Request, RequestInit, Response } from 'node-fetch';
 import { URL, URLSearchParams } from 'url';
 import { localize } from '../localize';
 
-export async function httpRequest<T>(url: string, options?: RequestOptionsLike, signRequest?: (request: RequestLike) => Promise<RequestLike>): Promise<HttpResponse<T>> {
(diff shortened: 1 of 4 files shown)


src/utils/httpRequest.ts
```diff
@@ -23,18 +33,37 @@ export async function httpRequest<T>(url: string, options?: RequestOptionsLike,
...
+export class HttpResponse<T> implements ResponseLike {
     private bodyPromise: Promise<T> | undefined;
-    private normalizedHeaders: { [key: string]: string } | undefined;
+    public readonly headers: HeadersLike;
+    public readonly status: number;
+    public readonly statusText: string;
+    public readonly ok: boolean;
+
+    public constructor(private readonly innerResponse: Response, public readonly url: string) {
+        // Unfortunately Typescript will not consider a getter accessor when checking whether a class implements an interface.
```
philliphoff, line 51: That's surprising!

philliphoff (approved):

src/utils/httpRequest.ts
karolz-ms: Yes, that's unfortunate. Also adding a setter accessor, or prepending interface members with `readonly` does not help either. 😞

PR #3074, opened 2021-07-18
chore: replace deprecated vscode-test with @vscode/test-electron

iChenLei:
vscode-test be marked as [deprecated](https://www.npmjs.com/package/vscode-test) on npmjs.org(yarnpkg.com), so just replace vscode-test with latest @vscode/test-electron for `vscode-docker`'s CI. cc @karolz-ms @bwateratmsft


bwateratmsft (approved):
Looks good to me. I verified from the build logs that tests are still running so everything should be good. Thanks @iChenLei! We'll merge this later today after our release.

PR #3077, opened 2021-07-19
Update dependencies and fix resulting issues

bwateratmsft:
This PR has a huge number of file changes but most of them are trivial. Some highlights on why:

- `vscode-azureextensionui` no longer has a shared `IAzureUserInput` (at least not externally exposed), but instead passes it along in `IActionContext`. So, our `ext.ui` had to go and be replaced. In some places this resulted in `IActionContext` bubbling a little further.
- `vscode-azureextensionui` is updated to TS 4.0, so the intermediate `AzExtTreeItemIntermediate` and `AzExtParentTreeItemIntermediate` weren't needed anymore. However, we also used it for wrapping the `resolveTooltip` call, so I created new classes for taking care of that.
- `telemetry.properties.cancelStep` is deprecated, and there's a few different ways of effecting that now--as an argument to `UserCancelledError`, a `stepName` option given to quick picks and dialogs, etc.
- New codicons package, `@vscode/codicons`.


src/extension.ts
```diff
@@ -191,27 +183,6 @@ async function getDockerInstallationIDHash(): Promise<string> {
     return 'unknown';
 }
 
-/**
- * Workaround for https://github.com/microsoft/vscode/issues/76211 (only necessary if people are on old versions of VS Code that don't have the fix)
- */
-function validateOldPublisher(activateContext: IActionContext): void {
```
bwateratmsft, line 197: This didn't need to be removed.

However should we anyway? What do you think @karolz-ms?

karolz-ms: I'd say let's remove this code; it is ancient history by now.

karolz-ms (approved):
Looks good. Praying to webpack gods for safe passage 😬

PR #3081, opened 2021-07-21
Add support for Workspace Trust

### LINKED ISSUE #2829: Use VSCode's Workspace Trust features
https://github.com/microsoft/vscode/issues/106488
https://github.com/microsoft/vscode/issues/120251

VSCode is adding a way to denote whether a workspace is trusted or not, and to gate features based on that. We should investigate using these APIs to increase users' security.

Features that should require a trusted workspace:
- [x] Tasks (always workspace)
- [x] Debug (always workspace)
- [x] Compose build / up (always workspace)
- [x] Image build (always workspace)
- [x] Command customization (_sometimes_ workspace, need to inspect)
- [x] Dockerode / `DOCKER_HOST` config (_sometimes_ workspace, need to inspect)
- [x] Scaffolding
- [x] Azure image build, run file as task

**Note:** VSCode is probably going to implement trust in such a way that any settings from untrusted workspaces are _not_ applied, which would take care of all of these except image build / compose up.

---

Features that require trust (of an image, not a file) but can't be implemented since it's not files:
- Image run (and compose up)

bwateratmsft:
Fixes #2829. Please let me know if you think of any other actions or settings that should require trust!

**Actions that will not be enabled in an untrusted workspace**
- Building images and compose up/down/etc.
- Scaffolding
- Building images in Azure or running files as tasks in Azure
- Running tasks (disabled by VSCode entirely, no additional work for us)
- Debugging (disabled by VSCode entirely, no additional work for us)

Note, actions are disabled both by visibility (`when` clauses in package.json) and programmatically (checks to `vscode.workspace.isTrusted`).

**Settings that will not be enabled in an untrusted workspace**
- Anything pertaining to Docker target host--`docker.host`, `docker.context`, `docker.dockerodeOptions`, `docker.certPath`, `docker.tlsVerify`, `docker.machineName`
- Anything pertaining to command customization--`docker.commands.build`, `docker.commands.run`, `docker.commands.runInteractive`, `docker.commands.attach`, `docker.commands.logs`, `docker.commands.composeUp`, `docker.commands.composeDown`
- Scaffolding template path--`docker.scaffolding.templatePath`
- Docker EXE location--`docker.dockerPath`

**Things _allowed_ in an untrusted workspace**
- Working with local/remote Docker server, e.g. with explorer (this is not something the workspace itself provides, after all)
- Working with registries
- Language server. @rcjsuen do you have any thoughts about this? There's obviously denial-of-service potential with massive or somehow-maliciously-busted Dockerfiles, but that's probably something I'm least concerned about...

### DIFF
--- a/src/commands/compose/compose.ts
+++ b/src/commands/compose/compose.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
-import { IActionContext } from 'vscode-azureextensionui';
+import { IActionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { rewriteComposeCommandIfNeeded } from '../../docker/Contexts';
 import { localize } from "../../localize";
 import { executeAsTask } from '../../utils/executeAsTask';
@@ -14,6 +14,10 @@ import { selectComposeCommand } from '../selectCommandTemplate';
 import { getComposeServiceList } from './getComposeServiceList';
 
 async function compose(context: IActionContext, commands: ('up' | 'down' | 'upSubset')[], message: string, dockerComposeFileUri?: vscode.Uri, selectedComposeFileUris?: vscode.Uri[]): Promise<void> {
+    if (!vscode.workspace.isTrusted) {
+        throw new UserCancelledError('enforceTrust');
+    }
+
     // If a file is chosen, get its workspace folder, otherwise, require the user to choose
     // If a file is chosen that is not in a workspace, it will automatically fall back to quickPickWorkspaceFolder
     const folder: vscode.WorkspaceFolder = (dockerComposeFileUri ? vscode.workspace.getWorkspaceFolder(dockerComposeFileUri) : undefined) ||
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
                 "@types/semver": "^7.3.4",
                 "@types/tar-stream": "^2.2.0",
                 "@types/uuid": "^8.3.0",
-                "@types/vscode": "1.55.0",
+                "@types/vscode": "1.56.0",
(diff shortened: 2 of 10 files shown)


src/commands/compose/compose.ts
```diff
@@ -14,6 +14,10 @@ import { selectComposeCommand } from '../selectCommandTemplate';
 import { getComposeServiceList } from './getComposeServiceList';
 
 async function compose(context: IActionContext, commands: ('up' | 'down' | 'upSubset')[], message: string, dockerComposeFileUri?: vscode.Uri, selectedComposeFileUris?: vscode.Uri[]): Promise<void> {
+    if (!vscode.workspace.isTrusted) {
+        throw new UserCancelledError('enforceTrust');
```
bwateratmsft, line 18: Here and elsewhere, we could throw a descriptive error (with error reporting button disabled) instead. Thoughts?

Note, the only way to even get into this code path in an untrusted workspace--because the commands are not shown, via `when` clauses--is keyboard shortcuts directly to the command, or programmatically.

karolz-ms: I would do that. After spending some weeks cleaning up issues coming from users hitting the "Report issue" button without providing any context, I am all for using issue reporting in a "judicious" way ☺️

karolz-ms (approved):
Nice!

src/commands/compose/compose.ts
bwateratmsft: As I have it written now, it will just...do nothing. `UserCancelledError` doesn't show anything at all. Since it shouldn't even be possible to execute these commands in the first place, I felt it was OK.

I don't feel strongly one way or the other. Do you have a preference?

karolz-ms: If it does not show the UI for submitting an issue, then I think it is good as-is

PR #3083, opened 2021-07-21
Support interactive `docker-run` tasks

### LINKED ISSUE #2978: Run docker task in interactive mode
`docker-run` tasks are always run in detached mode. I would like to specify which tasks should run in interactive mode, so I don't have to manually attach to the docker containers to see the output.

I have tried making `docker.commands.run` interactive, but it seems that tasks doesn't use those templates.

bwateratmsft:
Fixes #2978.

- [ ] TODO: a doc update PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -1049,6 +1049,10 @@
                                 "type": "string",
                                 "description": "%vscode-docker.tasks.docker-run.dockerRun.image%"
                             },
+                            "interactive": {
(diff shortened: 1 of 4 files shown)


karolz-ms (approved):

bwateratmsft:
Ugh. Turns out this probably isn't possible unless we dump `CustomExecution` which is probably not happening.

PR #3086, opened 2021-07-22
Use the inspect API to get user settings vs. defaults

### LINKED ISSUE #2858: Refactor `selectCommandTemplate.ts` to use the configuration `inspect` API instead of hardcoded default templates
There is an `inspect` API for configuration that allows the extension to inspect the default, global, workspace, and workspace folder values of a configuration. Using this API here would be ideal vs. duplicating the values in hardcoded templates.

~~This is probably required to do #2829 effectively.~~

bwateratmsft:
Fixes #2858. Also fixes an unrelated compile error that came up sometimes.

### DIFF
--- a/src/commands/selectCommandTemplate.ts
+++ b/src/commands/selectCommandTemplate.ts
@@ -13,35 +13,20 @@ import { dockerExePath, DefaultDockerPath } from '../utils/dockerExePathProvider
 
 type TemplateCommand = 'build' | 'run' | 'runInteractive' | 'attach' | 'logs' | 'composeUp' | 'composeDown';
 
-export type TemplatePicker = (items: IAzureQuickPickItem<CommandTemplate>[], options: IAzureQuickPickOptions) => Promise<IAzureQuickPickItem<CommandTemplate>>;
+interface CommandSettings {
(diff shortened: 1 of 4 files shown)


src/commands/selectCommandTemplate.ts
```diff
@@ -108,45 +94,39 @@ export async function selectCommandTemplate(
...
-    } else {
-        settingsTemplates = templateSetting;
-    }
+    const commandSettings = getCommandSettings();
+    const userTemplates: CommandTemplate[] = toCommandTemplateArray(commandSettings.workspaceFolderValue ?? commandSettings.workspaceValue ?? commandSettings.globalValue);
+    const defaultTemplates: CommandTemplate[] = toCommandTemplateArray(commandSettings.defaultValue);
 
-    // Get a template array from hardcoded defaults
-    const hardcodedTemplates = defaultCommandTemplates[command];
+    // Defense-in-depth: Reject if the workspace is untrusted but user templates from a workspace or workspace folder showed up somehow
```
karolz-ms, line 109: 👍

karolz-ms (approved):

Issue #3055, opened 2021-07-12
Change our start page to the new Getting Started experience


bwateratmsft:
Announced here: https://code.visualstudio.com/updates/v1_57#_new-getting-started-experience, https://code.visualstudio.com/updates/v1_56#_welcome-page-walkthroughs

Extensions can contribute walkthroughs.

bwateratmsft:
Tagging this experimentation. The C++ extension team has a way to properly set up these walkthroughs as experiments, and this is obviously a screamingly good candidate for experimentation.

bwateratmsft:
This may have to wait for VSCode 1.59, as it contains fixes that are necessary to make start pages experiment-able.

BigMorty:
Sounds good.

bwateratmsft:
Steps we might like:

- Add Dockerfiles
- Build image
- Run image
- Docker explorer
- Publish to cloud (e.g. Azure if Azure Account extension installed?)

ucheNkadiCode:
Add dockerfiles to workspace 
A Dockerfile is a text document that contains all the commands a user would sequentially call on the command line to assemble an image. Open the command pallete (CTRL + Shift + P) and choose Docker:Add Dockerfiles to Workspace - We can show a gif of editing a dockerfile

Build an image
An image is a read-only template with instructions for creating a Docker container. It can be based off a Dockerfile or another image. Navigate to your workspace and right-click your Dockerfile. Click Build Image. - Show a gif of right click build on a Dockerfile. **Most right-click actions are available from the command pallette**.

Run a container
A container is a runnable instance of an image. Navigate to the Docker Explorer (Whale Icon), right-click on your image, and select **Run Container**. gif navigating to the docker explorer, right clicking on an image, select run container

Use the Docker Explorer
The Docker extension makes it easy to build, manage, and deploy containerized applications. You can examine and manage Docker assets such as containers, images, volumes, networks, and container registries. Also, if the Azure Account extension[link to azure account extension in vs code] is installed, you can browse your Azure Container Registries as well.

The right-click menu provides access to commonly used commands for each type of asset.

Gif to show off Tool tips and files.  

Push an image to the cloud
The Docker Extension allows you to push your Docker image to Azure Container Registries, Docker Hub, AWS, GCP, and other third party providers. From the Images pane, right click on an image and select **Push**. 

Deploy to App Service
[...]

PR #3090, opened 2021-07-27
[remote] sync version across different machines

akosyakov:
to avoid showing the same update page to the same user on different machines over and over again

It happens in context of remote development, see for instance gitpod-io/gitpod#4566 (comment)
Also see https://github.com/microsoft/vscode-docs/issues/4699

### DIFF
--- a/src/commands/startPage/openStartPage.ts
+++ b/src/commands/startPage/openStartPage.ts
@@ -10,7 +10,7 @@ import { extensionVersion } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { startPage } from './StartPage';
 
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):
Looks fine to me! Though we'll probably be eliminating this start page and replacing it with the new VSCode standardized welcome experience, in #3055. @karolz-ms I think maybe what would make sense is to keep this PR but not complete it, unless #3055 gets pushed to the next release?

karolz-ms:
#3055 is P2, so it is likely to be pushed out. My suggestion is to take this PR now. And thanks @akosyakov for the contribution! 🤗

bwateratmsft:
Sounds good! Thanks @akosyakov!

PR #3096, opened 2021-07-29
Add support for compose profiles

### LINKED ISSUE #2777: Make use of profiles for compose up subset / compose tasks
https://docs.docker.com/compose/profiles/

bwateratmsft:
Fixes #2777.


package.json
```diff
@@ -1300,24 +1300,20 @@
...
-                                    "enum": [
-                                        "file"
-                                    ]
-                                }
+                                ]
                             }
-                        ]
+                        ],
+                        "not": {
+                            "required": [
```
bwateratmsft, line 1312: Cleaner way to denote one XOR the other.

package.json
```diff
@@ -1737,6 +1742,61 @@
                     ],
                     "description": "%vscode-docker.config.template.composeUp.description%"
                 },
+                "docker.commands.composeUpSubset": {
```
bwateratmsft, line 1745: Only good way I could work out to deal with putting `${profileList}` toward the beginning and `${serviceList}` at the end was to make a new customizable command.

src/commands/compose/getComposeSubsetList.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    return chosenSubsets.map(c => c.data);
+}
+
+async function getServiceSubsets(workspaceFolder: vscode.WorkspaceFolder, composeCommand: string, type: SubsetType): Promise<string[] | undefined> {
+    // TODO: if there are any profiles, then only services with no profiles show up when you query `config --services`. This makes for a lousy UX.
+
+    try {
+        // Start by getting a new command with the exact same files list (replaces the "up ..." or "down ..." with "config --services" or "config --profiles")
+        const configCommand = composeCommand.replace(composeCommandReplaceRegex, `config --${type}`);
```
bwateratmsft, line 113: We should use JSON output.

package.json
bwateratmsft: Double check this though.

package.json
```diff
@@ -1300,22 +1300,12 @@
                             {
                                 "required": [
```
bwateratmsft, line 1301: Tasks in this screenshot: incorrect (both), incorrect (neither), correct (file), correct (module)
image

package.json
```diff
@@ -1401,22 +1404,12 @@
...
```
bwateratmsft, line 1405: Tasks in this screenshot: incorrect (both), incorrect (neither), correct (up), correct (down)
image

package.json
```diff
@@ -1354,10 +1344,23 @@
...
+                                        "items": {
+                                            "type": "string"
+                                        }
+                                    },
                                     "customOptions": {
                                         "type": "string",
                                         "description": "%vscode-docker.tasks.docker-compose.dockerCompose.up.customOptions%"
                                     }
+                                },
+                                "not": {
```
bwateratmsft, line 1359: Tasks in this screenshot: incorrect (both), correct (services), correct (profiles), correct (neither)
image

karolz-ms (approved):

src/commands/compose/getComposeSubsetList.ts
bwateratmsft: The only way to get all the information would be to use `docker-compose config --profiles` to get a list of profiles, then use `docker-compose config --format json` once for each profile, then dedupe. IMO that's just silly, so I'll leave it as it is now.

PR #3102, opened 2021-07-30
Add experiment for API -> CLI fallback

### LINKED ISSUE #3099: Make CLI vs. API fallback to CLI an experiment
We changed the context lookup logic to go from CLI (slow) to API first, falling back to CLI if needed (maybe faster, maybe not--have to load the gRPC stuff which we didn't before). We should make it an experiment so we can pretty clearly determine if it is faster or not on average.

bwateratmsft:
Fixes #3099.


kzadora (approved):

kzadora:
Ehm, sorry, approved from wrong account.

karolz-ms (approved):

PR #3103, opened 2021-07-30
Code "splitting"

bwateratmsft:
Part of #2606. After some testing with webpack I've determined there's really nothing more to do. As far as I can tell, the kind of codesplitting their docs describe applies more to older versions of webpack and/or ES and/or web-based things where network calls happen. The current approach we're using, with `await import(...)`, works as well as we need it to.

This PR will update some comments that were missed but doesn't do anything else.

We've reached a point of substantially diminishing returns on lazy loading, so at this point I'm willing to call it done. I'll keep an eye on Kensho to ensure nothing regresses.


karolz-ms (approved):
👍

PR #3115, opened 2021-08-04
Use low-rights `node` user in Node.js apps

### LINKED ISSUE #1834: Node.js: Investigate using "node-user" by default
Investigate using "node-user" in the container by default. Of course, Node ecosystem is huge so this won't work all the time, but probably will most of the time.

Related to #1766.

bwateratmsft:
Fixes #1834.

### DIFF
--- a/resources/templates/node/Dockerfile.template
+++ b/resources/templates/node/Dockerfile.template
@@ -1,8 +1,6 @@
 FROM node:14-alpine
 ENV NODE_ENV=production
-{{#unless (isRootPort ports)}}
 USER node
-{{/unless}}
 WORKDIR /usr/src/app
 COPY ["package.json", "package-lock.json*", "npm-shrinkwrap.json*", "./"]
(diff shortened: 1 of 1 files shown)


resources/templates/node/Dockerfile.template
```diff
@@ -1,5 +1,8 @@
 FROM node:14-alpine
 ENV NODE_ENV=production
+{{#unless (isRootPort ports)}}
```
bwateratmsft, line 3: Incidentally I found that the Node image let me use ports < 1024 while not being root, so if we want we can probably remove this guard.

philliphoff (approved):

resources/templates/node/Dockerfile.template
philliphoff: I think I'd remove that just from a consistency point of view (i.e. we scaffold using `node` user but then user changes the port being used or vice-versa).

PR #3144, opened 2021-08-16
Bump version to 1.16.5-alpha

bwateratmsft:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-docker",
-    "version": "1.16.0",
+    "version": "1.16.5-alpha",
```
philliphoff, line 3: You don't like `1.16.1-alpha`? :-)

philliphoff (approved):

package.json
bwateratmsft: :smile:

My thinking is that if I do `1.16.1-alpha`, and we end up needing to ship a `1.16.1` out of a `rel/` branch, now we need to update `main` _again_ (i.e. to `1.16.2-alpha`).

If we get all the way to 5 servicing releases, we have bigger problems...

PR #3188, opened 2021-09-05
Upgrade to language server 0.6.0

rcjsuen:
The new 0.6.0 release of the language server adds support for heredocs in `ADD` and `COPY` as well as allowing multiple heredocs to be defined in a single instruction. There is also a small change to flag relevant linting issues as being "unnecessary". I'm not sure how VS Code renders this and if it takes the theme into consideration at all.
```Dockerfile
#escape=`
#escape=`
FROM node:alpine
# the linting issues above and below will now be slightly faded
# (depending on your theme?) to note that they are "unnecessary" given
# that there are duplicates and one of them should be removed
CMD [ "executable" ]
CMD [ "executable" ]
ENTRYPOINT [ "executable" ]
ENTRYPOINT [ "executable" ]
HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 CMD [ "executable" ]
HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 CMD [ "executable" ]

# multiple heredocs can now be defined in a single instruction without
# being reported as errors
RUN <<FILE1 cat > file1 && <<FILE2 cat > file2
first heredoc
FILE1
second heredoc
FILE2

# heredoc syntax support is available for ADD and COPY so the following
# lines will no longer be reported as errors
ADD <<EOF /test.txt
echo "Hello world"
EOF

COPY <<EOF /test.txt
echo "Hello world"
EOF

# invoke Intellisense on lines 36 and 40, nothing should be prompted,
# previously all instructions would be prompted as the system was not
# aware that it was in a heredoc
ADD <<eof test.txt

eof

COPY <<eof test.txt

eof
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
                 "@docker/sdk": "^1.0.3",
                 "@grpc/grpc-js": "^1.2.12",
                 "dayjs": "^1.10.4",
-                "dockerfile-language-server-nodejs": "^0.5.0",
+                "dockerfile-language-server-nodejs": "^0.6.0",
                 "dockerode": "^3.2.1",
                 "fs-extra": "^10.0.0",
                 "glob": "^7.1.6",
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):
> I'm not sure how VS Code renders this and if it takes the theme into consideration at all.

Looks like it fades the text a bit:
Old:
image

New:
image

Everything looks great, thanks @rcjsuen!

Issue #2539, opened 2020-12-03
Add intellisense for docker compose files


bwateratmsft:
Creating this umbrella issue since there are a bunch of different things that basically boil down to creating a language server + intellisense + etc. for compose files.

- [ ] Linting support (Compose schema validation) (https://github.com/microsoft/compose-language-service/issues/84)
- [ ] Volume pathing intellisense (https://github.com/microsoft/compose-language-service/issues/85)
- [ ] Snippets for some well-known services (postgres, redis, etc.), including common ports (https://github.com/microsoft/compose-language-service/issues/86) 
- [ ] Image autocomplete suggestions (https://github.com/microsoft/compose-language-service/issues/87)
- [x] Port snippets (#2535)
- [x] Autocompletions for subsequent parts--e.g., when typing 're', complete to 'restart:', then suggest 'always' and other valid suggestions (#2713)
- [x] Linting support (YAML validation) (#37)
- [x] Volume snippets (#2536)
- [x] URL click behavior for base images (#2200)
- [x] Auto format indentation on save/while typing (#2714)
- [x] Updated icon theme: update https://github.com/microsoft/vscode/blob/main/extensions/theme-seti/icons/vs-seti-icon-theme.json (so all compose files can get pink whale that currently only appears for `docker-compose.yml` and `docker-compose.override.yml`) (https://github.com/microsoft/vscode/commit/554182620f43390075d8c7e7fa36634288ef4e2d) (should be available in VSCode 1.61.0)
- [x] Syntax highlighting
- [x] Language ID for `file.associations` (#206) (https://github.com/microsoft/vscode/pull/118042)
- [x] Defining a language with more specific file names would help with spurious activations. `onLanguage:yaml` accounts for about 60% of all our activations; I'm sure many (or most) of those are not actually docker-compose files

Maybe/currently not possible:
- ~~Semantic highlighting~~ (I can't think of a use case)
- ~~Change the [file picking logic](https://github.com/microsoft/vscode-docker/blob/main/src/utils/quickPickFile.ts) to be better and language-based~~ (I don't think currently possible)
- ~~Port-in-use intellisense~~ (Gross, port scanning)

Other:
- ~~docker-stack.yml (#520)~~

[...]

PR #3206, opened 2021-09-13
Add the Compose language service

bwateratmsft:
Part of #2539.


package.json
```diff
@@ -2308,15 +2308,11 @@
                     "default": "docker",
                     "description": "%vscode-docker.config.docker.dockerPath%",
                     "scope": "machine-overridable"
-                }
-            }
-        },
-        "configurationDefaults": {
-            "[dockercompose]": {
```
bwateratmsft, line 2315: This may need to come back

rcjsuen:
Congratulations and thank you for contributing to the LSP community!

bwateratmsft:
> Congratulations and thank you for contributing to the LSP community!

Thanks @rcjsuen! We're excited. Still a long ways to go on it, but gotta start somewhere!

package.json
```diff
@@ -2308,15 +2308,11 @@
...
-                "editor.quickSuggestions": {
-                    "other": true,
-                    "comments": false,
-                    "strings": true
+                },
+                "docker.enableDockerComposeLanguageService": {
```
karolz-ms, line 2312: Would it make sense to roll Compose langserv as an experiment?

I am not sure if I would keep the old completion provider etc... but maybe just default `enableDockerComposeLanguageService` to false and gradually increase the population that defaults to true.

bwateratmsft: Yeah. I definitely considered it. I'm not counting it out, but it's tricky to do an experiment where the population can _choose_ to be in treatment or control; and I think it _is_ important for them to be able to choose on this.

karolz-ms: Agreed, but maybe we can treat this experiment in a special way, foremost as means to turn the default to off quickly if we detect that treatment population is running into unforeseen issues. At least at the beginning.

package.json
bwateratmsft: I don't think so. It is controlling always-on completions which are not viable (yet).

package.json
bwateratmsft: I created #3208 to look into this. For now I think we should check in as-is, so we can get a build for CTI tonight.

karolz-ms (approved):

PR #3231, opened 2021-09-23
Add a step to open a folder at the beginning

### LINKED ISSUE #3223: Show an "Open Folder" step before running image, if no workspace is open
Additionally, might need to change verbiage on "Now that you've built an image..."

bwateratmsft:
Fixes #3223.

image


bwateratmsft:
TODO: Should we change the verbiage on the "Run container" step to not refer back to previous steps that may or may not exist?

karolz-ms (approved):
Nice

ucheNkadiCode:
> the verbiage on the "Run container" step t

I'm not understanding which steps WOULD NOT exist before "Run containers". Wouldn't that be to create a dockerfile?

bwateratmsft:
@ucheNkadiCode if no workspace is open, "Add Dockerfiles" and "Build Image" will not show up because it isn't possible for them to succeed. So, that would make "Run containers" the first step, and the prose would begin with "Now that you've built an image..."

bwateratmsft:
Screenshots for the new verbiage @ucheNkadiCode came up with:

image
image

Issue #3189, opened 2021-09-05
Strange Makefile issue


JulioBarros:
I'm having a strange issue with the docker extension installed when I try to edit a Makefile. It used to work fine but starting a day or so ago, when I open a Makefile the syntax highlighting turns everything red and the formatting is messed up (it the tabs below the make targets are removed) when I save.

If I remove the extension the makefiles behave/edit correctly.

Could the extension be confusing Makefiles with docker files somehow?

Thanks in advance.

rcjsuen:
@JulioBarros Please provide screenshots of the behaviour with and without the extension installed.

rcjsuen:
@JulioBarros Also, please provide the following information:
- operating system
- version of Visual Studio Code
- version of extension installed

JulioBarros:
@rcjsuen Thanks for the quick reply ... uninstalling and reinstalling (a couple of times) may have fixed the issue because I don't see it now.

I'll close this for now and take screenshots if I see it again.

Btw, I'm on a Mac 11.52, VSCode 1.60.0 and it reinstalled Docker plug in v1.16.1 (I don't know what it was before).

Sorry for the trouble. Thanks again.

mcauto:
same here.

![image](https://user-images.githubusercontent.com/15080806/132179083-cafce4e2-7381-46bd-b793-deffd93b6f5a.png)

![image](https://user-images.githubusercontent.com/15080806/132179112-010f075b-d8e8-46e4-8a0c-046bebbc6103.png)

![image](https://user-images.githubusercontent.com/15080806/132179140-afefb5c6-1d18-4212-908f-caf891113c80.png)

![image](https://user-images.githubusercontent.com/15080806/132179209-8a0db99d-bfaa-4b06-ac66-34fa354ee44e.png)

rcjsuen:
@mcauto It is strange that your screenshot shows the extension as being disabled but then there is an error.

1. Please share the contents of your Makefile.
2. Please share the output. I want to see what the `32603` error is.

rcjsuen:
@JulioBarros @mcauto This looks like a duplicate of https://github.com/microsoft/vscode/issues/132005 which is a duplicate of https://github.com/microsoft/vscode/issues/132173.

@mcauto I am still interested in what I outlined in https://github.com/microsoft/vscode-docker/issues/3189#issuecomment-913521844 so I can reproduce the problem and see what the error is. Thank you.

mcauto:
[...]

PR #3239, opened 2021-09-25
Upgrade Dockerfile language server to 0.7.1

rcjsuen:
The 0.7.1 release of the Dockerfile language server has the following improvements:
- updates linter to consider syntax valid in BuildKit
- improve robustness of the parser to not get in invalid loops (inadvertently discovered by https://github.com/microsoft/vscode-docker/issues/3189#issuecomment-913979668)
- ignore valid instructions but to instead only flag the incorrect duplicates

Please use the following Dockerfiles for testing purposes. As always, please just copy/paste the content in to a `Dockerfile` without saving to preserve the formatting (to reproduce the incorrect parsing behaviour).
```Dockerfile
#escape=`
#escape=`
# only the second escape directive above is flagged as the first is valid
FROM node:alpine

# only the first set of instructions will be flagged as they are duplicated and unnecessary
CMD [ "executable" ]
ENTRYPOINT [ "executable" ]
HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 CMD [ "executable" ]
# the second set here is what will be used so they are no longer flagged
CMD [ "executable" ]
ENTRYPOINT [ "executable" ]
HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 CMD [ "executable" ]

# RUN can have no arguments or only flags in BuildKit, these two instructions should not be errors
RUN
RUN --mount=type=cache,target=/var/cache/foo,uid=1,gid=1
```
```Dockerfile
# the Output panel will complain about an infinite loop in
# textDocument/semanticTokens/full when the file is opened, if it does 
# not happen, just edit the comments, the new release fixes this
# https://github.com/microsoft/vscode-docker/issues/3189#issuecomment-913979668
FROM alpine
# hover over the error, it should no longer say RU\NRUN merging two into one,
# it should just say RU\N
RU\N
RUN echo
FROM node:alpine
# this error is merged into RUNRUNRUN, there should be no errors here
RU\
N
RUN
RUN echo

# this should not be an error, the embedded comment should be ignored
FR\
#
OM alpine
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
                 "@grpc/grpc-js": "^1.2.12",
                 "@microsoft/compose-language-service": "^0.0.1-alpha",
                 "dayjs": "^1.10.4",
-                "dockerfile-language-server-nodejs": "^0.6.0",
+                "dockerfile-language-server-nodejs": "^0.7.1",
                 "dockerode": "^3.2.1",
                 "fs-extra": "^10.0.0",
                 "glob": "^7.1.6",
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):
Looks great, thanks @rcjsuen!

PR #3248, opened 2021-09-30
Remove start page in favor of walkthrough

### LINKED ISSUE #3235: Got both the walkthrough and Get Started page
Tested on Insiders:

- Installed Docker → Walkthrough opened
- Opened Docker sidebar
 
🐛 _Docker - Getting Started_ opens

bwateratmsft:
Fixes #3235. This removes the old start page since the walkthrough is clearly more successful and less intrusive.

~~I looked into replacing the tree view node for "Open Start Page" with one that would launch straight to the walkthrough, but it doesn't seem possible unfortunately.~~
I figured it out!


package.json
```diff
@@ -2838,7 +2827,6 @@
             {
                 "id": "dockerStart",
                 "title": "%vscode-docker.walkthrough.dockerStart.title%",
-                "when": "never",
```
bwateratmsft, line 2841: This line makes the walkthrough no-longer-an-experiment.

package.json
```diff
@@ -2997,7 +2985,6 @@
         "@types/xml2js": "^0.4.8",
         "@typescript-eslint/eslint-plugin": "^4.28.4",
         "@typescript-eslint/parser": "^4.28.4",
-        "@vscode/codicons": "^0.0.22",
```
bwateratmsft, line 3000: Codicons package is no longer used.

src/telemetry/registerListeners.ts
```diff
@@ -26,11 +25,11 @@ export function registerListeners(): void {
     }
 
     registerEvent('dockerfileopen', vscode.workspace.onDidOpenTextDocument, handleDocEvent('dockerfile', () => {
```
bwateratmsft, line 27: Should we maybe refactor these to be a bit more like the old code?

package.json
```diff
@@ -225,6 +225,10 @@
                 {
                     "command": "vscode-docker.registries.azure.runFileAsTask",
                     "when": "isWorkspaceTrusted"
+                },
+                {
+                    "command": "vscode-docker.help.openWalkthrough",
+                    "when": "never"
```
bwateratmsft, line 231: There's already a VSCode command to open the walkthrough so I don't see a need for this to show in the palette.

bwateratmsft: The only reason we need this command defined at all is because the tree view command can't provide an argument to `workbench.action.openWalkthrough`.

philliphoff (approved):

Issue #3257, opened 2021-10-07
Support Docker rootless mode on linux


VeraSimon:
While the vscode-docker extension supports running as a non-root user on Linux as [documented here](https://code.visualstudio.com/docs/containers/overview#_installation), Docker's documentation on [setting that up](https://docs.docker.com/engine/install/linux-postinstall/) is very explicit about the security implications of doing so. In the same piece of documentation, they link to how to [set up Docker to run in rootless mode](https://docs.docker.com/engine/security/rootless/), but the vscode-docker extension doesn't see the running daemon due to the socket not being located at /var/run/docker.sock, but under a directory in /run/user associated with the users ID. Adding support for rootless mode would allow a setup path that doesn't needlessly reduce the security of the user's system in order to run the vscode-docker extension.

VeraSimon:
Found workaround by setting docker.host to the unix socket path in settings.json

bwateratmsft:
I agree, I think we should update the README [section](https://github.com/microsoft/vscode-docker#installation) and maybe the docs as well where rootless is mentioned to include setting up either a Docker context, `DOCKER_HOST` environment variable, or `docker.host` extension setting to the correct socket path.

PR #3259, opened 2021-10-07
Clarify install instructions for Linux rootless mode

VeraSimon:
Brief clarification to note that Docker rootless mode needs to have the generated context enabled before the vscode-docker extension will see it. Partial fix for issue #3257


ucheNkadiCode:
Looks good to me!

PR #3269, opened 2021-10-20
Upgrade Dockerfile language server to 0.7.2

rcjsuen:
This small update prevents two cases of infinite loops from being thrown in a `textDocument/semanticTokens/full` request. One of the two cases is the one reported in #3268.

Open these two files as Dockerfiles and you should see the exception error pop up. Do not save or format the file as the bug is contingent on the second line not having any whitespace in the front.
```Dockerfile
FROM $image\
abc
```
```Dockerfile
R\ 
# abc
```
You can also test with the Dockerfile described in https://github.com/microsoft/vscode-docker/issues/3268#issuecomment-946637035. Opening that file should not cause any problems either.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
                 "@grpc/grpc-js": "^1.2.12",
                 "@microsoft/compose-language-service": "^0.0.1-alpha",
                 "dayjs": "^1.10.4",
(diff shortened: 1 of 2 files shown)


karolz-ms:
Thank you, Remy 🤗

ucheNkadiCode:
Thanks so much @rcjsuen! We always appreciate your help 😁

Issue #3281, opened 2021-10-28
Docker tasks should obey docker.dockerPath setting


karolz-ms:
See https://github.com/docker/compose-cli/issues/754#issuecomment-953965746 for context

bwateratmsft:
Related to #2977

bwateratmsft:
This is now released in Docker extension version 1.19.0.

PR #3289, opened 2021-11-02
Make volumes accessible on SELinux systems by adding relabel option.

tmds:
The volumes mounted by the plugin are not accessible in the container on SELinux systems like Fedora and RHEL.
They need to be relabeled.

For more info see the _Labeling Volume Mounts_ section in https://docs.podman.io/en/latest/markdown/podman-run.1.html.

@bwateratmsft ptal.

### DIFF
--- a/src/tasks/DockerRunTaskDefinitionBase.ts
+++ b/src/tasks/DockerRunTaskDefinitionBase.ts
@@ -21,7 +21,7 @@ export interface DockerContainerPort {
 export interface DockerContainerVolume {
     localPath: string;
     containerPath: string;
-    permissions?: 'ro' | 'rw';
+    permissions?: 'ro' | 'rw' | 'ro,z' | 'rw,z';
 }
 
 export interface DockerRunOptions {
--- a/src/tasks/DockerRunTaskProvider.ts
+++ b/src/tasks/DockerRunTaskProvider.ts
@@ -102,9 +102,22 @@ export class DockerRunTaskProvider extends DockerTaskProvider {
     }
 
     private getVolumeOptions(volume: DockerContainerVolume, isWindows: boolean): string {
-        return !volume.permissions ? '' :
-            !isWindows ? `:${volume.permissions}` :
-                // for Windows, filter out the unsupported 'z'/'Z' options.
-                `:${volume.permissions.split(',').filter(s => s !== 'z' && s !== 'Z').join(',')}`;
+        if (!volume.permissions) {
+            return '';
+        } else if (!isWindows) {
+            return ':' + volume.permissions;
+        } else {
+            // The 'z' and 'Z' options aren't supported on Windows containers, normalize to simply ro / rw
+            switch (volume.permissions as string) {
+                case 'ro,Z':
+                case 'ro,z':
+                    return ':ro';
+                case 'rw,Z':
+                case 'rw,z':
+                    return ':rw';
+                default:
+                    return ':' + volume.permissions;
+            }
+        }
     }
 }
--- a/src/tasks/DockerRunTaskProvider.ts
+++ b/src/tasks/DockerRunTaskProvider.ts
@@ -9,7 +9,7 @@ import { localize } from '../localize';
 import { cloneObject } from '../utils/cloneObject';
 import { CommandLineBuilder } from '../utils/commandLineBuilder';
 import { dockerExePath } from '../utils/dockerExePathProvider';
-import { DockerRunOptions } from './DockerRunTaskDefinitionBase';
+import { DockerRunOptions, DockerContainerVolume } from './DockerRunTaskDefinitionBase';
 import { DockerTaskProvider } from './DockerTaskProvider';
 import { NetCoreRunTaskDefinition } from './netcore/NetCoreTaskHelper';
 import { NodeRunTaskDefinition } from './node/NodeTaskHelper';
@@ -91,7 +91,7 @@ export class DockerRunTaskProvider extends DockerTaskProvider {
             .withKeyValueArgs('-e', runOptions.env)
             .withArrayArgs('--env-file', runOptions.envFiles)
             .withKeyValueArgs('--label', getAggregateLabels(runOptions.labels, defaultVsCodeLabels))
-            .withArrayArgs('-v', runOptions.volumes, volume => `${volume.localPath}:${volume.containerPath}${volume.permissions ? ':' + volume.permissions : ''}`)
+            .withArrayArgs('-v', runOptions.volumes, volume => `${volume.localPath}:${volume.containerPath}${this.getVolumeOptions(volume, runOptions.os === 'Windows')}`)
             .withArrayArgs('-p', runOptions.ports, port => `${port.hostPort ? port.hostPort + ':' : ''}${port.containerPort}${port.protocol ? '/' + port.protocol : ''}`)
             .withArrayArgs('--add-host', runOptions.extraHosts, extraHost => `${extraHost.hostname}:${extraHost.ip}`)
             .withNamedArg('--entrypoint', runOptions.entrypoint)
@@ -99,5 +99,12 @@ export class DockerRunTaskProvider extends DockerTaskProvider {
             .withArg(runOptions.customOptions)
             .withQuotedArg(runOptions.image)
             .withArgs(runOptions.command);
+    }
+
+    private getVolumeOptions(volume: DockerContainerVolume, isWindows: boolean): string {
+        return !volume.permissions ? '' :
+            !isWindows ? `:${volume.permissions}` :
+                // for Windows, filter out the unsupported 'z'/'Z' options.
+                `:${volume.permissions.split(',').filter(s => s !== 'z' && s !== 'Z').join(',')}`;
     }
 }
--- a/src/tasks/netcore/NetCoreTaskHelper.ts
+++ b/src/tasks/netcore/NetCoreTaskHelper.ts
@@ -204,31 +204,31 @@ export class NetCoreTaskHelper implements TaskHelper {
             const appVolume: DockerContainerVolume = {
                 localPath: path.dirname(helperOptions.appProject),
                 containerPath: runOptions.os === 'Windows' ? 'C:\\app' : '/app',
-                permissions: 'rw'
+                permissions: 'rw,z'
             };
 
             const srcVolume: DockerContainerVolume = {
                 localPath: folder.uri.fsPath,
                 containerPath: runOptions.os === 'Windows' ? 'C:\\src' : '/src',
-                permissions: 'rw'
+                permissions: 'rw,z'
             };
 
             const debuggerVolume: DockerContainerVolume = {
                 localPath: vsDbgInstallBasePath,
                 containerPath: runOptions.os === 'Windows' ? 'C:\\remote_debugger' : '/remote_debugger',
-                permissions: 'ro'
+                permissions: 'ro,z'
             };
 
             const nugetRootVolume: DockerContainerVolume = {
                 localPath: path.join(os.homedir(), '.nuget', 'packages'),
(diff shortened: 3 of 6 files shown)


bwateratmsft:
I'm reluctant to have this flag added by default to all volume mounts in all `docker-run` tasks, especially given how ominously the [docs warn](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label) about the `z` and `Z` options ("Use **extreme** caution...", emphasis theirs). An alternative option would be to add `rw,z` and `ro,z` as options [here](https://github.com/microsoft/vscode-docker/blob/main/package.json#L1212-L1215), which would allow end users to specify these flags if they wish. It should be noted, you *can* specify those values today, it will work but it will give a yellow squiggle:

image
image

We have logic to implicitly add several volume mappings as needed for debugging .NET and Python. If there are conflicts; i.e. if the user specifies a volume mapping to the same destination in the container, the user's input will be respected. In the example above I'm taking advantage of that logic to get `/src` mapped with `rw,z`.

Do you think updating package.json with `rw,z` and `ro,z` as enum options is sufficient? Or is this likely a narrow enough scenario that the yellow squiggles can just be ignored, with no changes to the extension?

tmds:
> We have logic to implicitly add several volume mappings as needed for debugging .NET and Python.

I'm unable to debug .NET applications because the volumes used by the plugin don't set the `z` flag.
More changes are needed besides updating `package.json` to allow the 'z' option.

> especially given how ominously the docs warn about the z and Z options

I'll look into this a bit deeper and get back to you.

bwateratmsft:
In addition to the changes to `package.json`, you'll need to update your `tasks.json` to override the default volume mappings. Something like this:

```json
        {
            "type": "docker-run",
            "label": "docker-run: debug",
            "dependsOn": [
                "docker-build: debug"
            ],
            "dockerRun": {
                "volumes": [
                    {
                        "containerPath": "/src",
                        "localPath": "${workspaceFolder}",
                        "permissions": "rw,z"
                    },
                    {
                        "containerPath": "/app",
                        "localPath": "${workspaceFolder}",
                        "permissions": "rw,z"
                    },
                    {
                        "containerPath": "/remote_debugger",
                        "localPath": "~/.vsdbg",
                        "permissions": "ro,z"
                    },
                    {
                        "containerPath": "/root/.nuget/packages",
                        "localPath": "~/.nuget/packages",
                        "permissions": "ro,z"
                    },
                ]
            },
            "netCore": {
                "appProject": "${workspaceFolder}/Net6.csproj",
                "enableDebugging": true
            }
        },
```

This should work today albeit with the warning squiggles. For a Windows host, just replace the `~` in the `localPath`'s with `%USERPROFILE%`.

tmds:
> I'll look into this a bit deeper and get back to you.

The comment in the docker docs is out-of-date with the implementation. It will no accept any of the paths mentioned for relabeling.

Trying to relabel them will cause docker to fail.

That means it is not a good idea to unconditionally add the 'z' as the implementation is currently doing.

> We have logic to implicitly add several volume mappings as needed for debugging .NET and Python.

We should make a change that the volumes used for debugging are relabeled. Then they work out of the box on systems with SELinux.

@bwateratmsft do you have a suggestion on how to change `DockerContainerVolume` to support the relabling?

```ts
export interface DockerContainerVolume {
    localPath: string;
    containerPath: string;
    permissions?: 'ro' | 'rw';
}
```

bwateratmsft:
Should we set it so that that flag is only added when `enableDebugging` is set to true? I suppose it's sorta moot since all these implicit volumes are only added if it's true.

For `DockerContainerVolume` you could just add values `ro,z` and `rw,z`, same as for `package.json`.

tmds:
@bwateratmsft I've added the `,z` flag to the debug volumes.

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -156,6 +156,11 @@ export class NetCoreTaskHelper implements TaskHelper {
 
         runOptions.volumes = await this.inferVolumes(context.folder, runOptions, helperOptions, ssl, userSecrets); // Volumes specifically are unioned with the user input (their input does not override except where the container path is the same)
 
+        // Use a command that keeps the container running when started with the '-t' option.
+        if (runOptions.os !== 'Windows') {
+            runOptions.command = runOptions.command || '/bin/sh';
+        }
```
tmds, line 162: This is an additional change.

To test this, I had used an image that by default prints usage instructions. This caused the debug session to fail because it assumed the default command would keep the container running.

This explicitly specifies a default command so it has the desired behavior.

@bwateratmsft what do you think about this change?

bwateratmsft: My gut feeling is that it should be fine. We do rely on that behavior (of the container staying up and idle); I can't see it being harmful to make it explicit instead of assuming the base images (or in this case, the start command) remain unchanged. Similarly, in Visual Studio, the containers get started with `tail -f /dev/null` as the do-nothing command.

@karolz-ms or @philliphoff can you think of any issues that might arise?

src/tasks/DockerRunTaskDefinitionBase.ts
```diff
@@ -21,7 +21,7 @@ export interface DockerContainerPort {
 export interface DockerContainerVolume {
     localPath: string;
     containerPath: string;
-    permissions?: 'ro' | 'rw';
+    permissions?: 'ro' | 'rw' | 'ro,z' | 'rw,z';
```
bwateratmsft, line 24: Should we add the `ro,z` and `rw,z` options to the [enum](https://github.com/microsoft/vscode-docker/blob/main/package.json#L1212-L1215) in `package.json` and maybe also the [description string](https://github.com/microsoft/vscode-docker/blob/main/package.nls.json#L85) in `package.nls.json`? This would enable end users to use these options without the warning squiggles.

src/tasks/netcore/NetCoreTaskHelper.ts
karolz-ms: Hmm. I am not sure about this change. It means that, without additional information, VS Code will not use default command specified by the Docker image. This violates principle of least surprise IMO. And if the image requires a nonstandard command to be "useful", well, we already have a `command` property for that on the `docker-run` task.

bwateratmsft: The alternative way to make it explicit would be to scaffold it in, but I don't want to stuff the tasks with so much stuff that it gets overwhelming. I'd worry that scaffolding something like that would cause more problems/confusion than it would solve.

src/tasks/DockerRunTaskDefinitionBase.ts
tmds: I have added this. ptal.

bwateratmsft: It's in `package.json` now but got removed from `DockerRunTaskDefinitionBase.ts`, was that intentional? It would need to be in both.

bwateratmsft: Oh, also the changes to `NetCoreTaskHelper.ts` with the `,z` label on each volume mapping got removed.

tmds:
@bwateratmsft I had reverted the wrong commit. I've fixed it now.

bwateratmsft (approved):
Looks good to me!

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -208,25 +208,25 @@ export class NetCoreTaskHelper implements TaskHelper {
             const appVolume: DockerContainerVolume = {
                 localPath: path.dirname(helperOptions.appProject),
                 containerPath: runOptions.os === 'Windows' ? 'C:\\app' : '/app',
-                permissions: 'rw'
+                permissions: 'rw,z'
```
bwateratmsft, line 211: I found one last issue; if Windows containers are used, the `,z` option is not valid. Can you add checks so that `,z` is only added for not-Windows containers? e.g.:
```suggestion
                permissions: runOptions.os === 'Windows' ? 'rw' : 'rw,z'
```

bwateratmsft (changes requested):

src/tasks/netcore/NetCoreTaskHelper.ts
bwateratmsft: That's a great idea, I like it!

tmds: I've added the filter.

You may have some suggestions about the style/implementation. (I don't write TypeScript often.)

bwateratmsft: Sorry for the slow response, it's been a very busy week. I'm taking today and next week off so I won't be able to get to this until after the US Thanksgiving holiday.

tmds: @bwateratmsft it would be nice if this, and podman support (https://github.com/microsoft/vscode-docker/pull/3306) are part of the next release (`v1.19`).

I'd like to write a blog post about how use this plugin to debug a .NET app running in a container on Fedora.
Fedora uses podman and SELinux.

tmds: > this, and podman support (#3306)

and https://github.com/microsoft/vscode-docker/issues/3281 also.

bwateratmsft: Sorry again for the slow response. Last week was really hectic. I will get to this PR this week; we are planning to release 1.19 tentatively 17 Jan so this will definitely be a part of it.

tmds: No problem. As the changes go in, I'll verify they work well on Fedora.

src/tasks/DockerRunTaskProvider.ts
```diff
@@ -99,4 +99,11 @@ export class DockerRunTaskProvider extends DockerTaskProvider {
             .withQuotedArg(runOptions.image)
             .withArgs(runOptions.command);
     }
+
+    private getVolumeOptions(volume: DockerContainerVolume, isWindows: boolean): string {
```
bwateratmsft, line 103: I'd maybe do this; it's a bit hard to follow the nested ternaries:

```typescript
    private getVolumeOptions(volume: DockerContainerVolume, isWindows: boolean): string {
        if (!volume.permissions) {
            return '';
        } else if (!isWindows) {
            return ':' + volume.permissions;
        } else {
            // The 'z' and 'Z' options aren't supported on Windows containers, normalize to simply ro / rw
            switch (volume.permissions as string) {
                case 'ro,Z':
                case 'ro,z':
                    return ':ro';
                case 'rw,Z':
                case 'rw,z':
                    return ':rw';
                default:
                    return ':' + volume.permissions;
            }
        }
    }
```

src/tasks/DockerRunTaskDefinitionBase.ts
bwateratmsft, line 24: Waiting to hear from @philliphoff if the `,z` option is allowed on Mac. I suspect it is since it works on Linux containers on Windows, but I wanna check for sure...

bwateratmsft: Yep, it works. 👍

src/tasks/DockerRunTaskProvider.ts
bwateratmsft: I guess there are also some merge conflicts due to the fix for #3281.

tmds: Thanks. I've applied this.

tmds:
@bwateratmsft I've addressed your feedback. ptal.

bwateratmsft (approved):
Looks good to me, thanks @tmds!

bwateratmsft:
This is now released in Docker extension version 1.19.0.

Issue #3263, opened 2021-10-13
Make a CLI-only `DockerApiClient`


bwateratmsft:
We need an implementation of [`DockerApiClient`](https://github.com/microsoft/vscode-docker/blob/main/src/docker/DockerApiClient.ts) that uses CLI only. This will work better with a wider range of scenarios than the CLI + HTTP daemon through Dockerode + gRPC daemon. Plus it mostly removes the need to look up the Docker context during activation...

Potentially resolves:
- #2058
- #3295, #3590, #3605
- #3241, #1590, #3306, #3367, #3411
- #2941, #2933, #3578
- #2391
- #3475
- #3596
- _Maybe_ #2442

Related to #2977, #3281

bwateratmsft:
I've noticed that `com.docker.cli` tends to be drastically faster than `docker`.

bwateratmsft:
Work is ongoing for this, it will need to be moved to the 1.21.0 milestone.

ucheNkadiCode:
We are continuing to work on a path forward for supporting other Container runtimes. This will be moved to the 1.22 milestone.

bwateratmsft:
I know a lot of folks are interested so I wanted to give an update. Work on this is proceeding in earnest right now (as in, we're past most of the planning and really writing code!). We have rather ambitious plans so it will take some time.

benatshippabo:
> I know a lot of folks are interested so I wanted to give an update. Work on this is proceeding in earnest right now (as in, we're past most of the planning and really writing code!). We have rather ambitious plans so it will take some time.

Thanks for the update @bwateratmsft! I'm so glad this is underway

bwateratmsft:
Thanks for your thoughts, we appreciate the input! I'll try to address some of your questions here:

> Do want to start a branch for this? would you accept contributions/ideas?

The key work on the container runtimes library is still in a private repository but we should be opening it up soon; certainly some time before release. I'll update on this issue once it is opened. We love contributions!

> so can multiple extensions be installed and the first one to run, or do they all run

[...]

PR #3306, opened 2021-11-10
Add support for detecting podman.

tmds:
Further improves support for `podman` based systems by not requiring the user to manually change the settings.

Relates to https://github.com/microsoft/vscode-docker/issues/1590, https://github.com/microsoft/vscode-docker/pull/2984.

@bwateratmsft @karolz-ms ptal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2317,7 +2317,7 @@
                 },
                 "docker.dockerPath": {
                     "type": "string",
-                    "default": "docker",
+                    "default": "",
                     "description": "%vscode-docker.config.docker.dockerPath%",
                     "scope": "machine-overridable"
                 },
--- a/src/docker/ContextManager.ts
+++ b/src/docker/ContextManager.ts
@@ -33,7 +33,6 @@ const dockerConfigFile = path.join(os.homedir(), '.docker', 'config.json');
 const dockerContextsFolder = path.join(os.homedir(), '.docker', 'contexts', 'meta');
 
 const WindowsLocalPipe = 'npipe:////./pipe/docker_engine';
-const UnixLocalPipe = 'unix:///var/run/docker.sock';
 
 const defaultContext: Partial<DockerContext> = {
     Id: 'default',
@@ -249,7 +248,7 @@ export class DockerContextManager implements ContextManager, Disposable {
                 contextList = [{
                     ...defaultContext,
                     Current: true,
-                    DockerEndpoint: isWindows() ? WindowsLocalPipe : UnixLocalPipe,
+                    DockerEndpoint: isWindows() ? WindowsLocalPipe : this.getUnixLocalPipe(),
                 } as DockerContext];
             }
 
@@ -311,7 +310,7 @@ export class DockerContextManager implements ContextManager, Disposable {
             return {
                 ...defaultContext,
                 Current: true,
-                DockerEndpoint: isWindows() ? WindowsLocalPipe : UnixLocalPipe,
+                DockerEndpoint: isWindows() ? WindowsLocalPipe : this.getUnixLocalPipe(),
             } as DockerContext;
         }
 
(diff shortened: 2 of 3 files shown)


bwateratmsft:
We may want to hold off on this, since I think #3263 will dramatically change how we approach it. I could foresee having a selector for Docker path, that could allow for `docker`, `podman`, `com.docker.cli`, or a custom path, etc.

tmds:
> We may want to hold off on this, since I think #3263 will dramatically change how we approach it. I could foresee having a selector for Docker path, that could allow for docker, podman, com.docker.cli, or a custom path, etc.

Probably you'll want some logic as what is added here in `dockerExePath`, and get rid of `getUnixLocalPipe`?

What is `com.docker.cli`?

bwateratmsft:
> Probably you'll want some logic as what is added here in `dockerExePath`, and get rid of `getUnixLocalPipe`?

Yeah, if a setting isn't specified we can try to detect what is present.

> What is `com.docker.cli`?

More recent versions of Docker have `docker` as a wrapper script around `com.docker.cli`. I don't know why but I've found that `com.docker.cli` is way, _way_ faster--like, 50ms instead of 500+ms. For things like the explorer view where we have to refresh every few seconds, it might be imperative to use `com.docker.cli` in order to keep things snappy.

tmds:
> We may want to hold off on this, since I think #3263 will dramatically change how we approach it.

We could determine how to detect docker/podman in this PR so it can be used as part of #3263.
I'm also fine waiting for #3263.

karolz-ms:
> > What is `com.docker.cli`?
> 
> More recent versions of Docker have `docker` as a wrapper script around `com.docker.cli`. I don't know why but I've found that `com.docker.cli` is way, _way_ faster--like, 50ms instead of 500+ms. For things like the explorer view where we have to refresh every few seconds, it might be imperative to use `com.docker.cli` in order to keep things snappy.
> 
> In places where we intentionally show the user the command line--like tasks--we would not want to use `com.docker.cli`, but for background stuff it's probably the best bet.

We should bring it up with Docker, i.e. have them weigh in on how safe it is to use `com.docker.cli` directly, what are the ramifications etc.

BigMorty:
I would like to wait for #3263.

tmds:
@bwateratmsft you've updated the milestone for https://github.com/microsoft/vscode-docker/issues/3263 to 1.20.

Can you consider including this change in 1.19 so podman gets detected automagically?

dbreshears:
@tmds.   1.19 RC build is in final testing.  This will need to wait for a future milestone.   Sorry about that.

tmds:
@dbreshears @bwateratmsft do you have a rough idea when 1.20 is expected?

BigMorty:
@tmds, we like to release approximately once a month. Sometimes this must slide a week or two, but only occasionally.

tmds:
@bwateratmsft if https://github.com/microsoft/vscode-docker/issues/3263 doesn't make it in 1.20, I hope we can get this PR in so `podman` gets detected automatically.

As I said before, I think we can merge this even when https://github.com/microsoft/vscode-docker/issues/3263 will change/remove parts of it.

BigMorty:
@tmds - We've been discussing this internally at length. We don't want to take this PR, as we are apprehensive about automatically detecting Podman from our Docker extension. That said, Podman should work if it is configured in the extension's settings--we do not intentionally have any roadblocks to its functionality. If you run into blockers getting it to work via settings, create a new issue please.

We do appreciate the support and contributions you have made, and hope you continue to do so.
Thanks!

tmds:
> That said, Podman should work if it is configured in the extension's settings

As `podman` is getting more adoption, I regret it doesn't _just work_ for the developer using this plugin.

tigerinus:
There is also `nerdctl` which is claimed to be `docker` cli compatible.

Hopefully whatever coming up works for `nerdctl` as well.

tmds:
I published a blog post that includes instructions on using this extension with `podman`: https://developers.redhat.com/articles/2022/02/22/debug-net-applications-running-local-containers-vs-code#.

bwateratmsft:
Great blog, thanks @tmds!

ucheNkadiCode:
Great job @tmds! It might be helpful for the title to mention something like "Debug with Podman" just for easier discoverability. Thanks again for writing this up!

PR #3318, opened 2021-11-15
update read me

ucheNkadiCode:
made small nit changes to align with the VS Code overview page


bwateratmsft (approved):

bwateratmsft:
Just FYI, I shipped 1.18 this morning, so this won't show up in the extension page until 1.19 unless we want to do a dot fix, but that seems like overkill for this.

ucheNkadiCode:
I do not need this to be a dot fix at all. Totally fine to see this in 1.19 in the near future. 

Thanks!

PR #3322, opened 2021-11-17
Convert dev imports to true dev imports only

### LINKED ISSUE #3319: Change our dev-time imports to `import type`
TS 3.8 has an [`import type`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) feature, which allows us to specify that an import is _only_ a dev-time import. We have a lot of such imports in the code, this would help us make sure they don't accidentally become runtime imports.

Thanks @alexweininger for telling me about this!

bwateratmsft:
Fixes #3319. Using `import type` will prevent these dev-time imports from accidentally becoming runtime imports. Also turns some `require`s into `import`s.

### DIFF
--- a/src/commands/registries/azure/DockerAssignAcrPullRoleStep.ts
+++ b/src/commands/registries/azure/DockerAssignAcrPullRoleStep.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Progress } from "vscode";
-import { IAppServiceWizardContext } from "vscode-azureappservice"; // These are only dev-time imports so don't need to be lazy
+import type { IAppServiceWizardContext } from "vscode-azureappservice"; // These are only dev-time imports so don't need to be lazy
 import { AzureWizardExecuteStep, createAzureClient } from "vscode-azureextensionui";
 import { ext } from "../../../extensionVariables";
 import { localize } from "../../../localize";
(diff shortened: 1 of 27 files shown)


src/commands/registries/azure/DockerAssignAcrPullRoleStep.ts
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Progress } from "vscode";
-import { IAppServiceWizardContext } from "vscode-azureappservice"; // These are only dev-time imports so don't need to be lazy
+import type { IAppServiceWizardContext } from "vscode-azureappservice"; // These are only dev-time imports so don't need to be lazy
```
karolz-ms, line 7: Now that we use `import type` maybe we do not need these "it is only a dev-time import" comments anymore--you decide.

karolz-ms (approved):
Good change. 

Also shows that it would really help if TS had a language spec. Oh well 😒

src/commands/registries/azure/DockerAssignAcrPullRoleStep.ts
bwateratmsft: I guess I want to leave them, just for clarity on _why_ it's `import type`.

Issue #3290, opened 2021-11-03
Change Functionality of Show Dangling Images to Show Only Dangling/Show All Images


ucheNkadiCode:
When I click Show dangling images - right now it toggles between either showing both used containers and unused (dangling) containers OR used containers. 

I propose that the show dangling containers button should show ONLY the images that are unsused even if it's an empty list, then clicking the button once more should show ONLY the used containers. This would help a user know exactly what would get pruned. 

This would include conditional text based on what the user had selected

bwateratmsft:
I looked into this further and it would require changing the interface' meaning: https://github.com/microsoft/vscode-docker/blob/main/src/docker/DockerApiClient.ts#L37

It's  _probably_ safe to do so but I'm not real excited about that. Though perhaps it would be more practical once #3263 is done.

ucheNkadiCode:
> I looked into this further and it would require changing the interface' meaning: https://github.com/microsoft/vscode-docker/blob/main/src/docker/DockerApiClient.ts#L37
> 
> It's _probably_ safe to do so but I'm not real excited about that. Though perhaps it would be more practical once #3263 is done. A slightly-updated interface could have a trinary where it's `'NonDanglingOnly' | 'DanglingOnly' | 'All'`.

That's definitely an option, we could also do "isDangling: boolean" which would make it more of a toggle. How would that be represented in the extension? Would it be a dropdown menu pretty much? 

I like the trinary idea through since we want there to be a way for the user to see all of their images, all the dangling images (which totally could be empty if all images are being used), or just the NON-dangling aka currently used images (which could be empty as well )

bwateratmsft:
I don't think a trinary in the extension UX is necessary, just in the interface.

Honestly I'm a little skeptical of showing dangling images in general...they are all tagged "<none>:<none>" which is not helpful. All the clearly-identifying information (namely the tag) has been stripped; this is Docker's design. But, we have the feature now, so a takeback isn't a good idea.

bwateratmsft:
Per discussion we'll won't-fix this, the changes in #3336 are good enough for now.

PR #3336, opened 2021-11-29
Change show/hide dangling to a proper toggle

### LINKED ISSUE #3270: Hint still shows "Show dangling images" instead of "Hide dangling images"
It's not a serious bug, however I can't figure out whether dangling images are shown in the list or not. It would be great to change the icon as well, based on the selected option.
I attached a demonstration

https://skr.sh/vAdlTJ1CL13

bwateratmsft:
Fixes #3270. I intentionally did _not_ include #3290 due to [this](https://github.com/microsoft/vscode-docker/issues/3290#issuecomment-981920116).

The icons (and their tooltips) look like this:
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -350,12 +350,12 @@
                 {
                     "command": "vscode-docker.images.showDangling",
                     "when": "view == dockerImages && !vscode-docker:newSdkContext && !vscode-docker:danglingShown",
-                    "group": "navigation@2"
+                    "group": "navigation@3"
                 },
                 {
                     "command": "vscode-docker.images.hideDangling",
                     "when": "view == dockerImages && !vscode-docker:newSdkContext && vscode-docker:danglingShown",
-                    "group": "navigation@2"
+                    "group": "navigation@3"
                 },
                 {
                     "command": "vscode-docker.images.configureExplorer",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -229,6 +230,14 @@
                 {
                     "command": "vscode-docker.help.openWalkthrough",
                     "when": "never"
+                },
+                {
+                    "command": "vscode-docker.images.showDangling",
```
bwateratmsft, line 235: These two shouldn't show up in the command palette.

karolz-ms (approved):

ucheNkadiCode:
@bwateratmsft, 

So this changes the behavior to "show only dangling images" and "show all images?" That's basically the extent of what I was hoping for. The trinary would just be a bit mor work on top

bwateratmsft:
> So this changes the behavior to "show only dangling images" and "show all images?" That's basically the extent of what I was hoping for. The trinary would just be a bit mor work on top

@ucheNkadiCode This PR is to make it a toggle, so the button flips from "Show dangling images" to "Hide dangling images". In terms of what is shown, by default it is only _non-dangling_ images, clicking "Show dangling" shows all. There is not currently a way to show only _dangling_ images. That would be what #3290 is asking for but would require interface changes I'm somewhat reluctant to make.

PR #3353, opened 2021-12-08
Use PowerShell instead of WMI in Windows container file stat

### LINKED ISSUE #3282: Switch to PowerShell WMI provider
On Windows containers , the extension uses WMI to acquire file metadata, via the `wmic` utility.  This utility has been deprecated in Windows 11, in favor of using PowerShell WMI command-lets instead (e.g. `Get-WmiObject`).  At some point, Windows 11 container images will no longer ship with `wmic`; we need to move to the PowerShell equivalent.  (Windows 10 images should ship with both utilities.)

bwateratmsft:
Fixes #3282. The diff that Git produces is pretty ugly, so it may be easier to just look at the rewritten `statWindowsContainerItem()` function as a whole: https://github.com/microsoft/vscode-docker/blob/29aa5d45f25834ac180bb7a5d9eafe897e1155f0/src/docker/files/ContainerFilesUtils.ts#L224-L266

---

The PowerShell stat script is pretty complex and worth breaking down.

```powershell
$finfo = Get-Item -Path '${itemPath}'; Write-Output ('{0};{1};{2};{3}' -f ([System.DateTimeOffset]$finfo.CreationTimeUtc).ToUnixTimeMilliseconds(), ([System.DateTimeOffset]$finfo.LastWriteTimeUtc).ToUnixTimeMilliseconds(), $finfo.Length, @('file','directory')[$finfo.PSIsContainer])
```

Gets and stores the FS info in a variable, `$finfo`:

```powershell
$finfo = Get-Item -Path '${itemPath}'; 
```

Outputs the data in the same format we used for Linux containers' stat:

```powershell
Write-Output ('{0};{1};{2};{3}' -f ...
```

Casts `CreationTimeUtc` to `DateTimeOffset`, which has the very helpful `.ToUnixTimeMillseconds()` method. Since it's already in milliseconds we do not need to multiple by 1000 later, unlike Linux. Same thing is used for `LastWriteTimeUtc`.

```powershell
([System.DateTimeOffset]$finfo.CreationTimeUtc).ToUnixTimeMilliseconds()
```

PowerShell 5 is present in containers I checked (20H2, 1809, Nanoserver LTSC2022) and lacks a ternary operator, so I used the insane-but-functional workaround suggested by many users online of creating a two-element array and using the true/false result to index into it:

```powershell
@('file','directory')[$finfo.PSIsContainer]
```

The result of the above is the same as the Linux stat output we are using (except the x1000 to milliseconds thing):
A file: `1638890621016;1638901466884;276;file`
A directory: `1636031191290;1638890621025;1;directory`

### DIFF
--- a/src/docker/files/ContainerFilesUtils.ts
+++ b/src/docker/files/ContainerFilesUtils.ts
@@ -232,7 +232,7 @@ export async function statWindowsContainerItem(executor: DockerContainerExecutor
             }
         );
 
-        const statRegex = /^(?<ctime>\d+|[W]);(?<mtime>\d+);(?<size>\d+);(?<type>.+)$/g;
+        const statRegex = /^(?<ctime>\d+);(?<mtime>\d+);(?<size>\d+);(?<type>.+)$/g;
 
         const statMatch = statRegex.exec(result);
 
--- a/src/docker/files/ContainerFilesUtils.ts
+++ b/src/docker/files/ContainerFilesUtils.ts
@@ -221,161 +221,39 @@ export async function statLinuxContainerItem(executor: DockerContainerExecutor, 
     return undefined;
 }
 
-function parseWmiList(wmiList: string): { [key: string]: string } | undefined {
-    const lines = wmiList.replace(/[\r]+/g, '').split('\n');
+export async function statWindowsContainerItem(executor: DockerContainerExecutor, itemPath: string, itemType: DirectoryItemType): Promise<DirectoryItemStat | undefined> {
+    const command: string[] = ['powershell', '-Command', `$finfo = Get-Item -Path '${itemPath}'; Write-Output ('{0};{1};{2};{3}' -f ([System.DateTimeOffset]$finfo.CreationTimeUtc).ToUnixTimeMilliseconds(), ([System.DateTimeOffset]$finfo.LastWriteTimeUtc).ToUnixTimeMilliseconds(), $finfo.Length, @('file','directory')[$finfo.PSIsContainer])`];
 
-    let parsedObject: { [key: string]: string };
-
-    for (const line of lines) {
-        const index = line.indexOf('=');
-
-        if (index > 0) {
-            const name = line.substr(0, index);
-            const value = line.substr(index + 1);
-
-            if (parsedObject === undefined) {
-                parsedObject = {};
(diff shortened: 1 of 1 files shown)


src/docker/files/ContainerFilesUtils.ts
```diff
@@ -221,161 +221,39 @@ export async function statLinuxContainerItem(executor: DockerContainerExecutor,
...
-    }
-
-    const parsedPath = path.win32.parse(itemPath);
-
-    const drive = parsedPath.root.replace(/\\/, '');
-    const wmipath = parsedPath.dir.concat('\\');
-    const filename = parsedPath.base;
-    const command = ['cmd', '/C', `wmic fsdir where "drive='${drive}' and path='${wmipath}' and filename='${filename}'" get ${CreationDate}, ${LastModified} /format:list`];
+export async function statWindowsContainerItem(executor: DockerContainerExecutor, itemPath: string, itemType: DirectoryItemType): Promise<DirectoryItemStat | undefined> {
+    const command: string[] = ['powershell', '-Command', `$finfo = Get-Item -Path '${itemPath}'; Write-Output ('{0};{1};{2};{3}' -f ([System.DateTimeOffset]$finfo.CreationTimeUtc).ToUnixTimeMilliseconds(), ([System.DateTimeOffset]$finfo.LastWriteTimeUtc).ToUnixTimeMilliseconds(), $finfo.Length, @('file','directory')[$finfo.PSIsContainer])`];
```
bwateratmsft, line 225: PowerShell 5 is present in containers I checked (20H2, 1809, Nanoserver LTSC2022) and lacks a ternary operator, so I used the insane-but-functional workaround suggested by many users online of creating a two-element array and using the true/false result to index into it:
```powershell
@('file','directory')[$finfo.PSIsContainer]
```

bwateratmsft:
The build failure is during test, seems to be something failing during downloading the VSCode release. Probably they are releasing right now.

karolz-ms (approved):
Looks good. 

My only suggestion would be to add the explanation of the script workings from this PR into the code, as a comment (maybe abbreviated a bit)

philliphoff (approved):
Thanks for getting rid of all that WMI ugliness! :-)

Issue #3349, opened 2021-12-06
Language server errors whilst writing `COPY --from` line


henrahmagix:
This means VS Code alerts me with a popup in the bottom right, saying:
>A request has failed. See the output for more information.

Viewing the output for "Dockerfile Language Server", I see multiple repetitions of the following:
```
[Error - 13:44:19] Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: Cannot read property 'getRange' of undefined
  Code: -32603 
```

To reproduce, in a Dockerfile, start writing `COPY --from=something`: the line will be highlighted red and the tooltip info will say "COPY requires at least two arguments - dockerfile-utils(7)" – this is helpful =) Then start typing a space to start writing your two arguments, and the error above occurs immediately after typing the space. If you continue to write out the two arguments, no further error occurs; but if you go back to the `--from` argument and retype the space, then the error occurs every time you do that.

- No error except tooltip highlight: `COPY --from=something`
- Error occurs: `COPY --from=something ` (I typed a non-breaking space here because the highlight wasn't showing the extra space: don't copy this)
- No error except tooltip highlight: `COPY --from=something /path/to/something`
- All good: `COPY --from=something /path/to/something ./something`

rcjsuen:
@henrahmagix Thank you for the bug report. I was able to successfully reproduce this and will take a look at it.

rcjsuen:
I've opened https://github.com/rcjsuen/dockerfile-language-server-nodejs/issues/258 to track this.

bwateratmsft:
Thanks @rcjsuen! Like #3327, I'll close this one since language server updates are regularly picked up.

henrahmagix:
Thanks so much!

bwateratmsft:
This is now released in Docker extension version 1.19.0.

Issue #3347, opened 2021-12-03
Dockerfile language server hangs


willfrey:
I'm having the Dockerfile language server hang repeatedly. The only fix is for me to manually kill the process in a terminal.

Does this occur consistently?  Yes
Repro steps:

1. Start writing an instruction in a Dockerfile that requires two or more arguments, e.g.
```Dockerfile
# ...
COPY . .
# ...
```
2. Watch the Dockerfile language server eat up 99% of the CPU

Version: 1.18.0
OS: darwin
OS Release: 21.1.0
Product: Visual Studio Code
Product Version: 1.62.3
Language: en

rcjsuen:
@willfrey Thanks for the bug report. Something similar was reported a few weeks back in #3312 but the other person could not reproduce it anymore. Given that we're not seeing widespread reports of this and a Dockerfile with an instruction that has two arguments are a dime a dozen (like your example, a lot of people use `COPY` to move source code and build artifacts around), I'm not sure it is purely contingent on just an instruction that has two arguments.

Would you be able to share the following information?

1. Please attach a screenshot that shows the 99% CPU usage. Since you are on macOS I presume you were observing this via Activity Monitor?
2. Please share a complete copy of your Dockerfile so I can try to open it and reproduce the problem.

Thank you for your assistance.

willfrey:
@rcjsuen Thanks for the reply!

It seems to be better behaved today but I will keep an eye out for if it acts up again. If it does, I'll follow up on here.

bwateratmsft:
I guess I'll close this for now, please let us know if it reoccurs @willfrey! Thanks!

willfrey:
It's happening 🙃 

```Dockerfile
# syntax=docker/dockerfile:1.3-labs

ARG PYTHON_VERSION=3.9.9
ARG POETRY_VERSION=1.1.12

FROM python:${PYTHON_VERSION} AS python

FROM python AS poetry
SHELL [ "/bin/bash", "-o", "pipefail", "-c" ]
ARG POETRY_VERSION
RUN <<EOF
curl --silent --show-error --location https://install.python-poetry.org | PYTHONUSERBASE=/usr/local XDG_DATA_HOME=/usr/local/libexec python3 -
rm --recursive /root/.cache
EOF
CMD [ "/bin/sh" ]

WORKDIR /
FROM poetry AS requirements
WORKDIR /var/tmp
COPY pyproject.toml poetry.lock ./
[...]

Issue #3327, opened 2021-11-20
Autoformatting breaks Dockerfile with heredoc


wedi:
Using the new heredoc syntax in Dockerfiles used to trigger error messages in VS Code but autoformatting did not interfere with it. Unfortunately this has changed (unfortunately I cannot tell when): since some time autoformatting indents the end marker and following lines which makes it impossible to use autoformatting on save.

This dockerfile

```dockerfile
# syntax=docker/dockerfile:1.3-labs
FROM alpine
RUN <<EOF
    echo "Heredocs! Finally!"
EOF
CMD ["sh", "-c", "sleep 5"]
```

becomes

```dockerfile
# syntax=docker/dockerfile:1.3-labs
FROM alpine
RUN <<EOF
    echo "Heredocs! Finally!"
    EOF
    CMD ["sh", "-c", "sleep 5"]
```

 after formatting.

rcjsuen:
@wedi Thanks for the bug report.

For a workaround, you can set the `"docker.languageserver.formatter.ignoreMultilineInstructions"` preference to `true`. See #2976.

wedi:
Thanks! I definitely prefer fixing multiline commands myself over fixing the dockerfile in an external editor. :)

bwateratmsft:
@rcjsuen do you want to create an issue in https://github.com/rcjsuen/dockerfile-language-server-nodejs for this? Or keep it here?

rcjsuen:
> @rcjsuen do you want to create an issue in https://github.com/rcjsuen/dockerfile-language-server-nodejs for this?

I've opened https://github.com/rcjsuen/dockerfile-utils/issues/105 to track this.

bwateratmsft:
Sounds good. Since there's a workaround and we pick up the latest language server regularly I'll close this one.

bwateratmsft:
This is now released in Docker extension version 1.19.0.

PR #3355, opened 2021-12-12
Upgrade Dockerfile language server to 0.7.3

rcjsuen:
This update to the Dockerfile language server fixes three issues that have been raised by the community:
- the formatter will now ignore heredocs (https://github.com/microsoft/vscode-docker/issues/3327)
- the server will no longer enter an infinite loop when using Intellisense in `ADD` or `COPY` instructions (https://github.com/microsoft/vscode-docker/issues/3347)
- errors will no longer be thrown when using Intellisense after a `--chown` or `--from` flag in `ADD` or `COPY` instructions (https://github.com/microsoft/vscode-docker/issues/3349)

```Dockerfile
FROM alpine as alpine
# invoke Intellisense at the end of COPY, the CPU usage of the
# Dockerfile language server should not jump to 99% (issue #3347)
COPY . . 
```
```Dockerfile
FROM node:alpine
# invoke Intellisense at the end of this ADD instruction,
# there should be no more errors (issue #3349)
ADD --chown=user:group
# invoke Intellisense at the end of this COPY instruction,
# there should be no more errors (issue #3349)
COPY --chown=user:group

# issue #3327
# 1. format the document, these three lines should not change
# 2. highlight and select the three lines and use format selection,
# the three lines should not change
# 3. add a backslash (\) after "cat", the EOT line should not move
RUN <<EOT
    cat
EOT
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
                 "@grpc/grpc-js": "^1.2.12",
                 "@microsoft/compose-language-service": "^0.0.4-alpha",
                 "dayjs": "^1.10.4",
-                "dockerfile-language-server-nodejs": "^0.7.2",
+                "dockerfile-language-server-nodejs": "^0.7.3",
                 "dockerode": "^3.2.1",
                 "fs-extra": "^10.0.0",
                 "glob": "^7.1.6",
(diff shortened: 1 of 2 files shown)


rcjsuen:
image

I've contributed before but it looks like the build doesn't think so. Let me know if you all need me to do something different or whatever.

bwateratmsft:
> I've contributed before but it looks like the build doesn't think so. Let me know if you all need me to do something different or whatever.

Weird...oh well, I approved it to run. Maybe its memory only extends back to the point where we added this GitHub Action.

bwateratmsft (approved):
LGTM!

rcjsuen:
@bwateratmsft Thanks for the review!

Is the team planning to cut a new release this month or next month?

bwateratmsft:
Given the holidays and lots of people taking time off, we're planning to skip a month, so releasing tentatively 17 Jan.

PR #3399, opened 2022-01-23
Upgrade Dockerfile language server to 0.8.0

rcjsuen:
This new release improves the semantic highlighting support for variables in addition to correcting a false positive linting error about empty continuation lines on here-documents. Please test these two new changes with the Dockerfile below.

```Dockerfile
FROM node:alpine
# the variables should now
# have symbols in a
# different color
RUN echo ${variable:+word}
RUN echo ${variable:-word}
# there should not be an
# empty continuation line
# error here
RUN <<EOT
    abc

    def
EOT
```

Screenshot taken with the "Tomorrow Night Blue" theme:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
                 "@grpc/grpc-js": "^1.5.0",
                 "@microsoft/compose-language-service": "^0.0.5-alpha",
                 "dayjs": "^1.10.4",
-                "dockerfile-language-server-nodejs": "^0.7.3",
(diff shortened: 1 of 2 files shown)


rcjsuen:
Same as last month (https://github.com/microsoft/vscode-docker/pull/3355#issuecomment-992433882), GitHub Actions doesn't want to run the build... 🤔

bwateratmsft (approved):
Awesome work as always @rcjsuen, thank you!

PR #3402, opened 2022-01-25
Infer more default args, show as required correctly

### LINKED ISSUE #3400: Cannot configure build args using `tasks.json`
I'm trying to add `--build-arg` to the `docker build` command triggered by the `Docker Images: Build Image...` task.

I created the following `.vscode/tasks.json` file:

```
{
    "version": "2.0.0",
    "tasks": [
        {
            "type": "docker-build",
            "dockerBuild": {
                "buildArgs": {
                    "SOME_ARG": "foo",
                },
            },
        }
    ],
}
```

I can reproduce the problem with a dummy `Dockerfile`:

```
FROM centos:7
```

When I execute the `Build Image` task, the following command gets called:

    docker build --pull --rm -f "Dockerfile" -t vscodedockerissue:latest "."

I expected the following to be called:

    docker build --build-arg SOME_ARG=foo --pull --rm -f "Dockerfile" -t vscodedockerissue:latest "."

VSCode: 1.63.2
Docker Extension: v1.19.0

bwateratmsft:
Fixes #3400. Additionally, this includes an `npm audit fix` and `npm update` to clear some more alerts.

There's a few changes encompassed in this.

- I tried to make the task's required properties equivalent to the CLI's required properties--namely, context for `docker build` needed to be the only required parameter. Tag and Dockerfile should be optional.
- Required properties should show up in `tasks.json` now.
- I changed the logic for official build task resolution to consider default arguments better.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,7 +57,7 @@
                 "ts-loader": "^9.2.3",
                 "typescript": "^4.5.4",
                 "umd-compat-loader": "^2.1.2",
-                "vsce": "^2.5.1",
+                "vsce": "^2.6.4",
                 "webpack": "^5.65.0",
                 "webpack-bundle-analyzer": "^4.4.1",
                 "webpack-cli": "^4.6.0"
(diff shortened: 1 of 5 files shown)


bwateratmsft:
I had to undo making `dockerRun.image` required as that made all of our default-scaffolded tasks throw warnings. We could update the scaffolding but that would not help existing projects, and I think warning on something that we scaffolded is a bad look. It's worth noting, that if after full task resolution that property is still not defined, it will throw a clear error message.

PR #3403, opened 2022-01-25
Try to get port from launch settings

### LINKED ISSUE #3381: Avoid defaulting to port 5000
macOS Monterey (12.1) includes support for AirPlay, enabling the machine to act as a display for other Apple devices, which is enabled by default.  AirPlay happens to use port 5000 (among others). This means that applications using 5000 may run into "port already in use" errors that are difficult to diagnose, as use of this port by the system is not widely known.  This will impact Mac users running either .NET and Python (Flask) applications which traditionally used port 5000, and will become more prevalent as Monterey proliferates.

Prior to .NET 6, ASP.NET applications were scaffolded to use port 5000 by default.  In accordance with that convention, the Docker tooling prompts users to use port 5000 when scaffolding Docker support.  In addition, the tooling will attempt to map the same ports exposed by an image when running an image (e.g. so an image exposing port 5000 will be run attempting to map the port to the local 5000 port).  With .NET 6, however, ASP.NET applications are scaffolded to use a [randomly chosen port](https://docs.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-6.0?view=aspnetcore-6.0#template-generated-ports-for-kestrel) (i.e. the port is chosen at scaffold-time and then used thereafter). (Note that, absent an explicitly-defined port, ASP.NET applications will ultimately fall back to port 5000.)

The Docker tooling should consider avoiding defaulting to 5000.  In the case of .NET applications, the tooling could attempt to sniff the Kestrel launch profile in the `Properties/launchSettings.json` file, extracting the `applicationUrl` property, which will contain the port chosen at scaffold time.  (This should also generally work for pre-.NET6 applications.) Alternatively, the tooling could default to its own random port (since it's fairly arbitrary) excluding 5000.  Alternatively, the tooling could just avoid mapping ports to the local 5000 port (e.g. when running a container), or do so only when it detects 5000 already in use.  Or, perhaps simply generate a warning/error if on Mac and 5000 is already in use (presumably by the AirPlay service).

bwateratmsft:
Fixes #3381.

TODO:

- [x] .NET, uses 5000
- [x] Flask, uses 5000
- ~~Django, FastAPI? Use 8000~~
- ~~Node? Uses 3000~~

### DIFF
--- a/src/commands/containers/browseContainer.ts
+++ b/src/commands/containers/browseContainer.ts
@@ -20,6 +20,7 @@ const commonWebPorts = [
     3001,   // (Node.js) Sails.js
     5001,   // (.NET Core) ASP.NET SSL
     5000,   // (.NET Core) ASP.NET HTTP and (Python) Flask
+    5002,   // (Python) Flask (newer Flask apps)
     8000,   // (Python) Django and FastAPI
     8080,   // (Node.js)
     8081    // (Node.js)
--- a/src/scaffolding/wizard/ChoosePortsStep.ts
+++ b/src/scaffolding/wizard/ChoosePortsStep.ts
@@ -14,10 +14,13 @@ export class ChoosePortsStep extends TelemetryPromptStep<ScaffoldingWizardContex
     }
 
     public async prompt(wizardContext: ScaffoldingWizardContext): Promise<void> {
+        // If there are random suggested ports, show those, otherwise show the default
+        const suggestedPorts = wizardContext.suggestedRandomPorts?.length ? wizardContext.suggestedRandomPorts : this.defaultPorts;
+
         const opt: vscode.InputBoxOptions = {
-            placeHolder: this.defaultPorts.join(', '),
+            placeHolder: suggestedPorts.join(', '),
             prompt: localize('vscode-docker.scaffold.choosePortsStep.whatPorts', 'What port(s) does your app listen on? Enter a comma-separated list, or empty for no exposed port.'),
-            value: this.defaultPorts.join(', '),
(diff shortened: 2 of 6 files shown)


bwateratmsft:
> * [ ]  Flask, uses 5000

Regarding Flask, there's nothing like the .NET scaffolded port. I'm thinking about just picking a random port, or otherwise leaving it as-is and awaiting user feedback. Any thoughts @philliphoff?

karolz-ms (approved):

ucheNkadiCode:
@bwateratmsft , so is the point of this PR to avoid overlapping with port 5000 in flask? or are we planning to take the port from launch settings, otherwise provide a different default port than 5000

bwateratmsft:
> @bwateratmsft , so is the point of this PR to avoid overlapping with port 5000 in flask? or are we planning to take the port from launch settings, otherwise provide a different default port than 5000

No, the issue is that on newer macOS versions, port 5000 is always taken by an AirPlay component. .NET apps have a way around this in that at the time of scaffolding, they choose a random port and stamp it into launchSettings.json. However, Flask has nothing like that--there is no "scaffold" experience for Flask apps AFAIK. That basically leaves two choices:

1. Do nothing
1. Choose a port randomly, perhaps within some range, to use for Flask apps

bwateratmsft:
Discussion leans to fixed 5001. (option 2 above)

bwateratmsft:
I ended up going with 5002 instead. Our "Open in Browser" logic expects 5001 to be SSL (it's .NET's default SSL port), so it would open to `https://localhost:5001`, which would result in an error.

PR #3412, opened 2022-02-04
Improve Django scaffolding experience

### LINKED ISSUE #3410: Improve Add Dockerfiles to Django projects
Scenario:

User selects add dockerfiles to workspace
Select Application Platform -> no change
Python: Django -> no change
Choose the app's entry point (e.g. manage.py, app.py) 

My suggestion is to add a dialog so the user can (optionally) select where their wsgi.py file is in their application. We should try to look for wsgi file in the root folder as well as the [workspace name]\[workspace_folder of same name]\wsgi.py > ...(e.g. wsgi.py, asgi.py) and pop those to the top, but then also allow them to type it in 

Port: 8000 -> no change

If they choose manage.py, need to attempt to locate the wsgi.py, but if they choose wsgi.py or asgi.py then use the path exactly

bwateratmsft:
Fixes #3410. Two things in this:

- For Django apps only, when asking for input on the primary artifact we will no longer suggest WSGI/ASGI files.
- When searching for the WSGI file for Django apps, now look in any immediate subfolder from the primary artifact, rather than one of the exact name of the workspace folder.

### DIFF
--- a/src/scaffolding/wizard/python/ChoosePythonArtifactStep.ts
+++ b/src/scaffolding/wizard/python/ChoosePythonArtifactStep.ts
@@ -12,19 +12,26 @@ import { PythonScaffoldingWizardContext } from './PythonScaffoldingWizardContext
 const moduleRegex = /([a-z_]+[.])*([a-z_])/i;
 
 // Exclude Python files in the .venv folder from showing in the pick list
-const excludePattern = '.venv/**';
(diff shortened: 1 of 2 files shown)


bwateratmsft:
Build failures will be fixed by #3414.

karolz-ms (approved):

PR #3415, opened 2022-02-06
Include different delimiter for Date

left-code:
In case that Windows container is created with different localization function `parseWindowsDirectoryItems` won't return any folder.
Tested on container created using German localization.

### DIFF
--- a/src/docker/files/ContainerFilesUtils.ts
+++ b/src/docker/files/ContainerFilesUtils.ts
@@ -109,7 +109,7 @@ function parseWindowsType(dirOrSize: string): DirectoryItemType {
 }
 
 function parseWindowsDirectoryItems(input: string, parentPath: string): DirectoryItem[] {
-    const expression = /^(?<date>\d{1,2}(\/|.)\d{1,2}(\/|.)\d{4})\s+(?<time>\d{1,2}:\d{1,2}( (AM|PM))?)\s+(?<dirOrSize><DIR>|<SYMLINKD>|\d+)\s+(?<name>.*)$/gm;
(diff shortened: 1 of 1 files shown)


philliphoff (approved):
That's a great catch!  Thanks for the contribution!

left-code:
> @left-code what does the date field look like on German localization?

@bwateratmsft  It's in format `dd.mm.yyyy` delimited by dot.

Maybe my solution won't work for all localization, but I would say that dot delimiter is pretty common in Europe.

src/docker/files/ContainerFilesUtils.ts
```diff
@@ -109,7 +109,7 @@ function parseWindowsType(dirOrSize: string): DirectoryItemType {
 }
 
 function parseWindowsDirectoryItems(input: string, parentPath: string): DirectoryItem[] {
-    const expression = /^(?<date>\d{1,2}\/\d{1,2}\/\d{4})\s+(?<time>\d{1,2}:\d{1,2}( (AM|PM))?)\s+(?<dirOrSize><DIR>|<SYMLINKD>|\d+)\s+(?<name>.*)$/gm;
+    const expression = /^(?<date>\d{1,2}(\/|.)\d{1,2}(\/|.)\d{4})\s+(?<time>\d{1,2}:\d{1,2}( (AM|PM))?)\s+(?<dirOrSize><DIR>|<SYMLINKD>|\d+)\s+(?<name>.*)$/gm;
```
bwateratmsft, line 112: I guess I'd make this change, since the `.` as written accepts any character, not just a ".".

```suggestion
    const expression = /^(?<date>\d{1,2}(\/|\.)\d{1,2}(\/|\.)\d{4})\s+(?<time>\d{1,2}:\d{1,2}( (AM|PM))?)\s+(?<dirOrSize><DIR>|<SYMLINKD>|\d+)\s+(?<name>.*)$/gm;
```

bwateratmsft:
Everything looks good, thanks for the contribution @left-code!

bwateratmsft:
This is now released in Docker extension version 1.20.0. Thanks for your contribution @left-code!

Issue #3407, opened 2022-01-31
Update to track 2 Azure SDK


bwateratmsft:
Azure SDK for JS is being significantly upgraded, and the upgraded version called "track 2". We need to update to use it as the old one will be out-of-support before long.

Depends on:
- https://github.com/microsoft/vscode-azuretools/pull/1050
- https://github.com/microsoft/vscode-azuretools/pull/1054
- https://github.com/microsoft/vscode-azuretools/pull/1061

bwateratmsft:
This has ended up being a very far-reaching change. Almost all of the changes are just to imports, but there are so many files touched that I don't feel comfortable trying to push this into 1.20.

PR #3422, opened 2022-02-14
Add a survey to get some inspiration for features

bwateratmsft:
@ucheNkadiCode FYI

Requirement for eligibility is to have 10 active sessions ever. We'll also target the experiment to English-only.

Prompt text: How can we make the Docker extension better?
Options: <kbd>Take survey</kbd> <kbd>Don't ask again</kbd>

Popup will come 60 seconds after extension activation.


karolz-ms (approved):

src/extension.ts
```diff
@@ -69,13 +70,14 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
             targetPopulation = tas.TargetPopulation.Team;
         } else if (/alpha/ig.test(extensionVersion.value)) {
             targetPopulation = tas.TargetPopulation.Insiders;
+        } else if (/Insiders/ig.test(vscode.env.appName)) {
```
bwateratmsft, line 73: This is changed in the same way as https://github.com/microsoft/vscode-azuretools/pull/1067, but that is in the new `@microsoft/vscode-azext-utils` package, which won't make it in to here until we do the T2 upgrade in #3407. Rather than publish a new deprecated `vscode-azureextensionui` I just copied the code here.

PR #3434, opened 2022-02-17
Engineering improvements enabled by VSCode 1.64

### LINKED ISSUE #3408: Use `onTaskType` activation event
VSCode *finally* added an `onTaskType` activation event, so we should definitely make use of it to reduce spurious activations. https://code.visualstudio.com/updates/v1_64#_ontasktype-activation-event

### LINKED ISSUE #3409: Use VSCode's built-in file watcher
VSCode recently [opened up the filesystem watcher](https://code.visualstudio.com/updates/v1_64#_vscodeworkspacecreatefilesystemwatcher-now-supports-any-path) to be able to watch anything, including outside the workspace. This would be a great thing to replace the context watcher with.

bwateratmsft:
Fixes #3408, fixes #3409

TODO:
- [x] Verify the file watcher behavior
- [x] Verify the task activation behavior


src/docker/ContextManager.ts
```diff
@@ -80,16 +80,27 @@ export class DockerContextManager implements ContextManager, Disposable {
...
         // that are done in CLI. Worst case, a user would have to restart VSCode.
         try {
-            if (fse.existsSync(dockerConfigFile)) {
-                this.configFileWatcher = fs.watch(dockerConfigFile, async () => this.refresh());
+            if (fse.existsSync(path.join(dockerConfigFolder, dockerConfigFile))) {
+                this.configFileWatcher = workspace.createFileSystemWatcher(
+                    new RelativePattern(dockerConfigFolder, dockerConfigFile),
+                    true, // Don't expect file to be created (since we already verified its existence)
+                    false, // Do expect file to change
+                    true // Don't expect file to be deleted (are they uninstalling?!)
```
karolz-ms, line 88: Are you sure we want to ignore deletions? If Docker is uninstalled, we probably want to "refresh"...

karolz-ms (approved):
Looks good; just one question

src/docker/ContextManager.ts
karolz-ms: Yes, the latter is much better. My suggestion is then to keep the code as-is, but just massage the comment a bit, explaining why we do not need to watch for file deletion/Docker uninstall.

karolz-ms (approved):

PR #3438, opened 2022-02-18
Dedupe ports and put IP in pick description

### LINKED ISSUE #3431: Pop up a dialog contained two same ports when opening a .NET running container in browser
**OS:** Ubuntu 18.04
**Build Version:** 20220214.1
**Regression:** Previous Release

**Repro Steps:**

1. Open the .NET Core project and add Docker files for this project in VS Code.
2. F5 with "Docker .NET Core Launch" configuration.
3. Right click the running container in the CONTAINERS view -> Open in Browser.
4. Check whether opens a web page directly.

**Expect:**
Open a web page directly.

**Actual:**
Pop up a dialog contained two same ports.
image

**More Info:**
1. No responding after selecting the second port on Ubuntu. 
2. This issue also reproduce on Codespaces. 
3. This issue doesn't reproduce on Windows and Mac(M1).

bwateratmsft:
Fixes #3431.

- [x] Need to do some final testing.


src/commands/containers/browseContainer.ts
```diff
@@ -43,10 +43,22 @@ export async function browseContainer(context: IActionContext, node?: ContainerT
...
         .filter(port => port.IP); // IP must be defined (not falsy)
 
+    // There can be multiple ports that are bound to localhost, so let's remove duplicates
+    const possiblePorts: DockerPort[] = Array.from(new Set(
+        // Type cleanup in order to dedupe properly
+        dupedPossiblePorts.map(port => {
+            return {
+                ...port,
+                Type: 'tcp', // Already established that it's either undefined or tcp, let's make it always tcp
+                IP: isLocalhostBinding(port.IP) ? 'localhost' : port.IP, // Turn any localhost IP bindings into "localhost"
```
karolz-ms, line 57: 0.0.0.0 is "every IP address on the machine"
127.0.0.1 and localhost are loopback addresses, meaning the server is only accessible from the same host. 
In general they should not be treated the same. In particular, 0.0.0.0 (INADDR_ANY) should not be changed to "localhost", even though in the context of the bug that prompted the change (related to port binding), they have the same impact.

Also, technically, the IPv4 loopback address range is 127.0.0.0/8, so anything from 127.0.0.1 to 127.0.254.254; and `loopback` DNS name might be bound to any of these addresses (although I have yet to experience a case when it is not 127.0.0.1).

So, when looking for port conflicts, 
1. 0.0.0.0 will affect all addresses on the machine, but different loopback addresses can have same ports
2. 0.0.0.0 should not be replaced with `localhost`

karolz-ms (changes requested):

src/commands/containers/browseContainer.ts
bwateratmsft: Also it's probably worth defining an interface similar to `DockerPort`, but used only here, to make it clear that we are playing with IPs/etc. _on purpose_ in order to make a usable URL out of a port binding.

bwateratmsft: Did some updates. What do you think?

karolz-ms (approved):

src/commands/containers/browseContainer.ts
karolz-ms: Looks good!! Thanks

PR #3440, opened 2022-02-19
Support alternative project names in Docker Compose tasks

### LINKED ISSUE #3436: Unable to set --project-name using docker-compose task
First of all, thank you for creating such a brilliant plugin.

We are in the process of setting up Visual Studio Code tasks to simplify complex `docker compose` commands within my team. Because we use a mono-repository, we are layering multiple compose files on top of one another, the base file which is located in a `docker` repository like this:

```none
root/
├─ docker/
│  ├─ docker-compose.yaml
│  ├─ .env
├─ project1/
│  ├─ docker-compose.yaml
```

`docker-compose.yaml` files inside the project directories include modifications to the base compose file. Since `docker/docker-compose.yaml` is always the first compose file in our docker compose commands, the project name is inferred as `docker` regardless of which product we spin up.

**We would instead like the ability to specify the `--project-name` argument on a task-by-task basis, such as:**

```json
{
    "label": "Compose up: Project 1",
    "type": "docker-compose",
    "dockerCompose": {
        "up": {
            "detached": true,
            "build": true,
            "projectName": "Project1"
        },
        "files": [
            "${workspaceFolder}/docker/docker-compose.yaml",
            "${workspaceFolder}/project1/docker-compose.yaml"
        ],
        "envFile": "${workspaceFolder}/docker/.env"
    }
}
```

Using `"customOptions": "--project-name product1"` is not possible, since that will place the argument incorrectly in the resulting command:

```bash
docker compose up -f docker/docker-compose.yaml -f project1/docker-compose.yaml --env-file docker/.env up -d --project-name product1
```

As opposed to:

```bash
docker compose up -f docker/docker-compose.yaml -f project1/docker-compose.yaml --env-file docker/.env --project-name product1 up -d
```

If this is within scope, I would be happy to open a PR.

Xitric:
This PR fixes #3436 by adding the `projectName` property on compose up and down tasks:

```json
{
  "label": "Compose up: Project1",
  "type": "docker-compose",
  "dockerCompose": {
    "projectName": "myproject",
    "up": {
      "detached": true,
      "build": true,
    },
    "files": [
      "${workspaceFolder}/docker-compose.yaml"
    ]
  }
}
```

When using an alternative project name to compose up, the same project name must be specified in order to compose down again. Thus, if composing up a `docker-compose.yaml` file with an alternative project name, right-clicking that file in the file tree and selecting the `Compose Down` command will not work.

I decided not to perform any validation logic on the project name character set, since that is best performed by Docker itself.

I will open a PR to update the documentation as well.


package.json
```diff
@@ -1436,6 +1436,10 @@
                                 "items": {
                                     "type": "string"
                                 }
+                            },
+                            "projectName": {
+                                "type": "string",
+                                "description": "%vscode-docker.tasks.docker-compose.dockerCompose.projectName%"
```
bwateratmsft, line 1442: Just a nit-pick since as-written this will work perfectly, but adding `.description` will make it consistent with the other root-level properties (`up`, `down`, `files`, `envFile`)

```suggestion
                                "description": "%vscode-docker.tasks.docker-compose.dockerCompose.projectName.description%"
```

package.nls.json
```diff
@@ -115,6 +115,7 @@
     "vscode-docker.tasks.docker-compose.dockerCompose.envFiles.description": "(DEPRECATED) Use `envFile` instead.",
     "vscode-docker.tasks.docker-compose.dockerCompose.envFile.description": "File of environment variables read in and applied to the Docker containers.",
     "vscode-docker.tasks.docker-compose.dockerCompose.files.description": "The docker-compose files to include, in order.",
+    "vscode-docker.tasks.docker-compose.dockerCompose.projectName": "Alternate project name to use when naming and labeling Docker objects. If using an alternate project name when composing up, the same project name must be specified when composing down.",
```
bwateratmsft, line 118: Corresponds to above change

```suggestion
    "vscode-docker.tasks.docker-compose.dockerCompose.projectName.description": "Alternate project name to use when naming and labeling Docker objects. If using an alternate project name when composing up, the same project name must be specified when composing down.",
```

bwateratmsft (approved):
Looks good, I'll try to give it a test run this week. Thanks @Xitric!

package.nls.json
Xitric: Ah yes, of course! I did not notice this when I moved it from within the `up` property and out to the root level

bwateratmsft (approved):

bwateratmsft:
Tested it out, everything works great. Thanks again for the contribution @Xitric!

PR #3478, opened 2022-03-30
Add a setting to control the compose command used

### LINKED ISSUE #2977: Add a flag to use new Docker Compose command
The latest releases of Docker Desktop include the new implementation of Docker Compose command (`docker compose` with no dash). Currently our extension only uses the old `docker-compose` command when running in the local context. 

The new command is destined to replace the old one eventually, but at the moment it still may have some bugs, or incompatibilities, and some incompatibilities might never be reconciled. The proposal is then to introduce a setting allowing the user to opt in for the new command, with the idea that sometime in the future we will switch to the new command as a default, and the setting could be then used to opt out of the new command. This should provide the smoothest migration path to the new Docker Compose implementation for our users.

bwateratmsft:
Fixes #2977. A string setting, `docker.composeCommand` is added. If unset, the extension will try to use `docker compose` by running `docker compose version`; if it succeeds we assume we can use `docker compose`, otherwise we use `docker-compose`. If the context or settings are altered, we will re-run that check, since different contexts have different requirements (for example, ACI contexts require `docker compose`, not supporting `docker-compose` at all).

I expect few people will need to force `docker-compose` when `docker compose` does work, but for them, they can use the setting in either workspace or user settings.

I marked #2977 as having documentation impact, because it will require documenting the new `${composeCommand}` placeholder in command customization.

Scenarios to test:
- [x] Login/logout
- [x] Push/pull image
- [x] Azure CLI
- [x] Deploy to ACI
- [x] View ACI containers
- [x] .NET debug
- [x] Python debug
- [x] Container files
    - [x] Windows
    - [x] Linux
- [x] `docker-build` task
- [x] `docker-run` task
- [x] `docker-compose` task
- [x] Compose group commands
- [x] Custom commands
- [x] Fallback on systems not supporting `docker compose`


package.json
```diff
@@ -1604,7 +1604,7 @@
                             "type": "string"
                         }
                     ],
-                    "default": "docker build --pull --rm -f \"${dockerfile}\" -t ${tag} \"${context}\"",
+                    "default": "${config:docker.dockerPath} build --pull --rm -f \"${dockerfile}\" -t ${tag} \"${context}\"",
```
bwateratmsft, line 1607: Rather than always running `docker ...`, we can pick up the setting they apply with `docker.dockerPath`. In contrast to `docker.composeCommand`, that setting is _not_ unset by default, because we aren't attempting to do any autodetection.

package.json
```diff
@@ -1932,14 +1932,7 @@
                     "default": [
                         {
                             "label": "Compose Down",
-                            "template": "docker-compose ${configurationFile} down",
```
bwateratmsft, line 1935: Since we now intelligently choose the compose command, we no longer need separate command templates based on context for `compose down`. It is still needed for the two `compose up` commands, because ACI contexts lack a `--build` flag and will error if it is present.

src/tasks/DockerTaskProvider.ts
```diff
@@ -51,7 +50,6 @@ export abstract class DockerTaskProvider implements TaskProvider {
                 context.platform = getPlatform(task.definition);
 
                 context.actionContext.telemetry.properties.dockerPlatform = context.platform;
-                context.actionContext.telemetry.properties.orchestration = 'single' as DockerOrchestration; // TODO: docker-compose, when support is added
```
bwateratmsft, line 54: No longer needed, as the task name is an exact hint as to whether we're dealing with single or compose orchestration.

src/utils/dockerExePathProvider.ts
```diff
@@ -5,13 +5,61 @@
...
+    }
+});
+
+// Register a listener to clear the value when the context changes
+// Separately, because changes to the relevant settings (like `docker.composeCommand`) will automatically result in a context refresh, that will indirectly cause this to fire too
+ext.context.subscriptions.push(
+    ext.dockerContextManager.onContextChanged(() => composeCommandLazy.clear())
+);
+
+export async function getComposeCommand(context?: IActionContext): Promise<string> {
```
bwateratmsft, line 53: I'm still contemplating, but maybe this (and `dockerExePath()`) should move to `DockerContextManager`.

src/utils/dockerExePathProvider.ts
```diff
@@ -5,13 +5,61 @@
...
+    } catch {
+        // Otherwise fall back to the old command
+        return OldComposeCommand;
+    }
+});
+
+// Register a listener to clear the value when the context changes
+// Separately, because changes to the relevant settings (like `docker.composeCommand`) will automatically result in a context refresh, that will indirectly cause this to fire too
+ext.context.subscriptions.push(
+    ext.dockerContextManager.onContextChanged(() => composeCommandLazy.clear())
```
bwateratmsft, line 50: Does this go boom if this value isn't set by the time this file is imported?

src/commands/selectCommandTemplate.ts
```diff
@@ -166,16 +165,7 @@ export async function selectCommandTemplate(
     actionContext.telemetry.properties.commandContextType = `[${selectedTemplate.contextTypes?.join(', ') ?? ''}]`;
     actionContext.telemetry.properties.currentContextType = currentContextType;
 
-    let resolvedCommand = resolveVariables(selectedTemplate.template, folder, additionalVariables);
```
bwateratmsft, line 169: Since the command template is now based on the setting, this is no longer needed to rewrite commands to match what they used in settings.

src/docker/ContextManager.ts
```diff
@@ -106,13 +116,18 @@ export class DockerContextManager implements ContextManager, Disposable {
...
+        this.composeCommandLazyDisposable = this.onContextChanged(() => this.composeCommandLazy.clear());
+
         // Set the initial DockerApiClient to be the context loading client
         ext.dockerClient = new ContextLoadingClient(this.loadingFinishedEmitter.event);
     }
 
     public dispose(): void {
         void this.configFileWatcher?.dispose();
         void this.contextFolderWatcher?.dispose();
+        void this.composeCommandLazyDisposable?.dispose();
```
karolz-ms, line 130: Maybe instead of having separate members for the file watchers and the composeCmdLazyDisposable, the ContextManager could just have one "disposables array"? These file watchers etc do not seep to be used outside of class constructor

karolz-ms (dismissed):
Very nice fix/refactoring. Just one minor suggestion to consider.

src/docker/ContextManager.ts
bwateratmsft: Good idea!

karolz-ms (approved):

PR #3481, opened 2022-04-04
Add a command to copy full tag for remote images

stuartthomson:
There are already commands for copying image digest, or copying the full tag for local images. This just adds the ability to the registries view.

### DIFF
--- a/package.json
+++ b/package.json
@@ -588,24 +588,24 @@
                     "group": "regs_tag_1_general@1"
                 },
                 {
-                    "command": "vscode-docker.registries.copyImageDigest",
-                    "when": "view == dockerRegistries && viewItem =~ /DockerV2;Tag;/",
-                    "group": "regs_tag_1_general@2"
-                },
-                {
                     "command": "vscode-docker.registries.copyRemoteFullTag",
                     "when": "view == dockerRegistries && viewItem =~ /DockerV2;Tag;/",
                     "group": "regs_tag_1_general@2"
                 },
                 {
+                    "command": "vscode-docker.registries.copyImageDigest",
+                    "when": "view == dockerRegistries && viewItem =~ /DockerV2;Tag;/",
+                    "group": "regs_tag_1_general@3"
+                },
+                {
                     "command": "vscode-docker.registries.deployImageToAzure",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -591,6 +592,11 @@
                     "when": "view == dockerRegistries && viewItem =~ /DockerV2;Tag;/",
                     "group": "regs_tag_1_general@2"
                 },
+                {
+                    "command": "vscode-docker.registries.copyRemoteFullTag",
+                    "when": "view == dockerRegistries && viewItem =~ /DockerV2;Tag;/",
+                    "group": "regs_tag_1_general@2"
```
bwateratmsft, line 598: This has the same value as for `copyImageDigest` so I'm not sure what order the UI will choose. We should adjust the other values. Personally I think "Copy Full Tag" should come before "Copy Image Digest", so that would mean keeping `regs_tag_1_general@2` for "Copy Full Tag", and reindexing "Copy Image Digest" and the others in the `regs_tag_1_general` section. Also, reordering the entries so that the order here in the manifest is the same as the order in the UI.

src/commands/registries/copyRemoteFullTag.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export async function copyRemoteFullTag(context: IActionContext, node?: DockerV2TagTreeItem): Promise<string> {
+    if (!node) {
+        node = await ext.registriesTree.showTreeItemPicker<DockerV2TagTreeItem>(registryExpectedContextValues.dockerV2.tag, {
+            ...context,
+            noItemFoundErrorMessage: localize('vscode-docker.commands.registries.copyRemoteFullTag.noImages', 'No remote images are available to copy the full tag')
+        });
+    }
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    vscode.env.clipboard.writeText(node.fullTag);
```
bwateratmsft, line 21: At the time we did these linter supressions we didn't know that you can do this instead:
```suggestion
    // Don't wait
    void vscode.env.clipboard.writeText(node.fullTag);
```
(I included the comment in the suggestion too, we usually like to indicate that we're intentionally not waiting on the call)

bwateratmsft (approved):
Looks great, thanks @stuartthomson!

bwateratmsft:
This is now released in version 1.22 of the Docker extension. Thanks for the contribution @stuartthomson!

Issue #3450, opened 2022-03-01
Opening DockerFiles in the editor crashes VS Code


MDPolaris:
I recently upgraded to Docker extension 1.20.0 and every time I try to try to view a DockerFile VS Code hangs and the following window appears. 
<img width="264" alt="image" src="https://user-images.githubusercontent.com/17074189/156168798-2c82d795-3d8a-4101-b379-71ba203151ff.png">

I have tried waiting but the window never loads. The only thing that works is to Reopen the window. I reverted to 1.19.0 and the issue is not present. I also disabled the 1.20.0 extension version and the issue is also not present.

MDPolaris:
I apologize, I missed the request for more information.  Here is the sanitized Docker File

```
FROM python:3.8-slim-buster

WORKDIR /app

RUN export JAVA_HOME=/usr/bin/java

# install OS packages
RUN apt-get update 
RUN apt-get install -y sudo git unzip wget ca-certificates curl apt-transport-https lsb-release gnupg less locales libicu63 libssl1.1 libc6 libgcc1 libgssapi-krb5-2 liblttng-ust0 libstdc++6 zlib1g jq default-jre

# Install powershell
RUN echo Installing Powershell and modules
RUN wget https://github.com/PowerShell/PowerShell/releases/download/v7.1.3/powershell_7.1.3-1.debian.10_amd64.deb && dpkg -i powershell_7.1.3-1.debian.10_amd64.deb && apt-get install -y -f 
RUN pwsh -c Install-Module -Name Az -AllowClobber -SkipPublisherCheck -Force
RUN pwsh -c Install-Module -Name Az.Dns -AllowClobber -SkipPublisherCheck -Force
RUN pwsh -c Install-Module -Name Az.IoTHub -AllowClobber -SkipPublisherCheck -Force
RUN pwsh -c Install-Module -Name Az.Sql -AllowClobber -SkipPublisherCheck -Force

RUN echo installing az-cli
RUN curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor | tee /etc/apt/trusted.gpg.d/microsoft.gpg > /dev/null 
RUN export AZ_REPO=$(lsb_release -cs) && echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $AZ_REPO main" | tee /etc/apt/sources.list.d/azure-cli.list &&  apt-get update && apt-get install azure-cli -y 

# Install kubectl
[...]

PR #3508, opened 2022-05-04
Upgrade Dockerfile language server to 0.9.0

rcjsuen:
This update includes support for the parsing the new `--link` flag in `ADD` and `COPY` instructions as well as a fix for negative semantic token values that seemed to cause Visual Studio Code to hang and crash (#3450).

To test the `--link` flag validating support, please use the Dockerfile below.
```Dockerfile
FROM scratch
# these are okay
ADD --link . .
ADD --link=true . .
ADD --link=false . .
ADD --link=tRUe . .
ADD --link=FALse . .
COPY --link . .
COPY --link=true . .
COPY --link=false . .
COPY --link=tRUe . .
COPY --link=FALse . .
#these are not okay
ADD --link= . .
ADD --link=abc . .
COPY --link= . .
COPY --link=abc . .
```

To test the semantic tokens crash, please use the Dockerfile provided in https://github.com/microsoft/vscode-docker/issues/3450#issuecomment-1069259251.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
                 "@microsoft/vscode-azext-azureutils": "^0.1.3",
                 "@microsoft/vscode-azext-utils": "^0.1.1",
                 "dayjs": "^1.10.4",
-                "dockerfile-language-server-nodejs": "^0.8.0",
+                "dockerfile-language-server-nodejs": "^0.9.0",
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):
Awesome as always, thanks @rcjsuen!

PR #3516, opened 2022-05-09
Switch back to default .NET images (undo #2996, #3029)

### LINKED ISSUE #3512: Fail to build image for Dockerfile with the default 7.0 image
**OS:** Win 10
**Docker Extension Build Version:** 20220506.1
**.NET Version:** 7.0.100-preview.3.22179.4

**Repro Steps:**
1. Create and open a .NET 7.0 MVC project in VS Code.
2. Add Docker files to this project (Linux Container) -> Right click the "Dockerfile" file -> Select "Build Image...".
3. Check whether succeeds to build image of the "Dockerfile" file.

**Expect:**
Succeed to build image of the "Dockerfile" file.

**Actual:**
Fail to build image of the "Dockerfile" file.
image

**More Info:**
1. This issue only reproduces with the default Docker image "7.0-focal".
2. This issue does not reproduce with other images "7.0 and 7.0-alpine".
3. This issue does not reproduce for a .NET 6.0 project.

bwateratmsft:
~~With the release of Ubuntu 22.04, the new Ubuntu tag for images is `-jammy` instead of `-focal`. There is no `7.0-focal` at all, only `7.0-jammy`. There is both for 6.0, so we do not need to service a fix.~~

Undoes #2996, #3029. The .NET attach scenario no longer relies on `ps` so we don't need to use Ubuntu-based images by default.

Fixes #3512.

### DIFF
--- a/src/scaffolding/wizard/netCore/NetCoreGatherInformationStep.ts
+++ b/src/scaffolding/wizard/netCore/NetCoreGatherInformationStep.ts
@@ -20,7 +20,7 @@ const minCSharpVersionString = '1.23.9';
 const aspNetBaseImage = 'mcr.microsoft.com/dotnet/aspnet';
 const consoleNetBaseImage = 'mcr.microsoft.com/dotnet/runtime';
 const netSdkImage = 'mcr.microsoft.com/dotnet/sdk';
-const linuxTagSuffix = '-focal';
+const linuxTagSuffix = '-jammy';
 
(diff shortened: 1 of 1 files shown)


bwateratmsft:
@karolz-ms I'm not sure if we even need this suffix anymore. I just tried with the default image and the attach scenario worked successfully, which IIRC was the original motivation for adding this suffix in #2996.

karolz-ms:
@bwateratmsft agreed, let's try switching back to default image (undo #2996)

PR #3521, opened 2022-05-11
Update dependencies

bwateratmsft:


src/tree/containers/files/DirectoryTreeItem.ts
```diff
@@ -33,9 +33,7 @@ export class DirectoryTreeItem extends AzExtParentTreeItem {
     }
 
     public get iconPath(): vscode.ThemeIcon {
-        return (this as vscode.TreeItem).collapsibleState === vscode.TreeItemCollapsibleState.Expanded
-            ? new vscode.ThemeIcon('folder-opened')
-            : new vscode.ThemeIcon('folder');
+        return new vscode.ThemeIcon('folder');
```
bwateratmsft, line 36: This never worked because `collapsibleState` is not updated by VSCode, so it never changed to expanded. We added a utility method to track the expansion state and keep it up to date, but that involves listening to some fast-firing events, so in theory is at least somewhat expensive--and in this case offers very little value. I figured it was fine to just fix to only the folder icon.

karolz-ms (approved):

PR #3533, opened 2022-05-24
Fix 'Add Docker Files to Workspace' command

IIFelix:
When using the 'Add Docker Files to Workspace' command within a fastapi app where the main.py is in a subfolder the resulting 'tasks.json' will not include the subfolder which breaks the debug mode.

Example where main.py is in a directory called 'app':
Before:
```json
{
	"version": "2.0.0",
	"tasks": [
		{
			"type": "docker-build",
			"label": "docker-build",
			"platform": "python",
			"dockerBuild": {
				"tag": "guitartabsdocker:latest",
				"dockerfile": "${workspaceFolder}/Dockerfile",
				"context": "${workspaceFolder}",
				"pull": true
			}
		},
		{
			"type": "docker-run",
			"label": "docker-run: debug",
			"dependsOn": [
				"docker-build"
			],
			"python": {
				"args": [
					"main:app",
					"--host",
					"0.0.0.0",
					"--port",
					"8000"
				],
				"module": "uvicorn"
			}
		}
	]
}
```

After:

```json
{
	"version": "2.0.0",
	"tasks": [
		{
			"type": "docker-build",
			"label": "docker-build",
			"platform": "python",
			"dockerBuild": {
				"tag": "guitartabsdocker:latest",
				"dockerfile": "${workspaceFolder}/Dockerfile",
				"context": "${workspaceFolder}",
				"pull": true
			}
		},
		{
			"type": "docker-run",
			"label": "docker-run: debug",
			"dependsOn": [
				"docker-build"
			],
			"python": {
				"args": [
					"app.main:app",
					"--host",
					"0.0.0.0",
					"--port",
					"8000"
				],
				"module": "uvicorn"
			}
		}
	]
}
```

PR #3534, opened 2022-05-26
Fix Python paths in subfolders

bwateratmsft:
Related to #3533 I found this issue while testing. On Linux/Mac, the Python file path would have a forward slash, and this would work. On Windows however, it had a backslash, and the backslash was not replaced with a `.` as it should have been.


karolz-ms (approved):

bwateratmsft:
This has now been released in Docker extension version 1.23.0.

PR #3555, opened 2022-06-27
Deprecate and migrate old environment variable settings

### LINKED ISSUE #3539: Consider deprecating `docker.host`, `docker.context`, etc. settings
We're thinking about deprecating these settings, in favor of using VSCode's standard `terminal.env.<os>` option. I think that we would want to have a migration experience to help users get over to the new settings, as well as some documentation to describe how to set those.

VSCode will apply those environment variables automatically to all terminal windows, but we would need to pick up and apply the appropriate setting for any shell tasks that we execute behind-the-scenes.

bwateratmsft:
Closes #3539.

Settings migration could be a sticky issue. I want to verify the following, to be confident this feature will not break anything:

- [x] No notification if settings aren't used (covers vast majority of users)
- [x] Setting changes get picked up by context manager
- [x] Verify user -> user settings
- [x] Verify workspace -> workspace settings
- [x] Verify workspace folder -> workspace folder settings
- [x] Verify multiple setting targets combined
- [x] Verify multiple settings combined
- [x] ~~Verify an empty workspace folder setting with non-empty user setting works?~~ Nah


package.json
```diff
@@ -3081,14 +3059,11 @@
...
                 "docker.dockerodeOptions",
-                "docker.host",
-                "docker.context",
-                "docker.certPath",
-                "docker.tlsVerify",
-                "docker.machineName",
                 "docker.scaffolding.templatePath",
-                "docker.dockerPath"
+                "docker.dockerPath",
+                "docker.composeCommand"
```
bwateratmsft, line 3066: `docker.composeCommand` should have been added here originally.

package.json
```diff
@@ -2231,39 +2231,17 @@
...
+                    },
+                    "description": "%vscode-docker.config.docker.environment%",
                     "scope": "machine-overridable"
                 },
-                "docker.machineName": {
-                    "type": "string",
-                    "default": "",
-                    "description": "%vscode-docker.config.docker.machineName%",
-                    "scope": "machine-overridable"
+                "docker.dockerodeOptions": {
```
bwateratmsft, line 2242: `docker.dockerodeOptions` is unchanged (for now, will go away with #3263), but the diff is being weird.

src/extension.ts
```diff
@@ -208,6 +216,17 @@ namespace Configuration {
...
 /* eslint-enable @typescript-eslint/no-namespace, no-inner-declarations */
 
+function setEnvironmentVariableContributions(ctx: vscode.ExtensionContext): void {
+    const settingValue: NodeJS.ProcessEnv = vscode.workspace.getConfiguration('docker').get<NodeJS.ProcessEnv>('environment', {});
+
+    ctx.environmentVariableCollection.clear();
+    ctx.environmentVariableCollection.persistent = true;
+
+    for (const key of Object.keys(settingValue)) {
+        ctx.environmentVariableCollection.replace(key, settingValue[key]);
```
bwateratmsft, line 226: @Tyriar is this the right way to use this API? We call this `setEnvironmentVariableContributions` at extension activation and anytime the `docker.environment` setting is changed.

src/utils/addDockerSettingsToEnv.ts
```diff
@@ -4,30 +4,18 @@
...
-
-function addDockerSettingToEnv(settingKey: string, envVar: string, env: NodeJS.ProcessEnv, oldEnv: NodeJS.ProcessEnv): void {
-    const value = workspace.getConfiguration(configPrefix).get<string>(settingKey, '');
+export function addDockerSettingsToEnv(newEnv: NodeJS.ProcessEnv, oldEnv: NodeJS.ProcessEnv): void {
+    const environmentSettings: NodeJS.ProcessEnv = workspace.getConfiguration('docker').get<NodeJS.ProcessEnv>('environment', {});
 
-    const expectedType = "string";
-    const actualType = typeof value;
-    if (expectedType !== actualType) {
-        ext.outputChannel.appendLine(localize('vscode-docker.utils.env.ignoring', 'WARNING: Ignoring setting "{0}.{1}" because type "{2}" does not match expected type "{3}".', configPrefix, settingKey, actualType, expectedType));
```
bwateratmsft, line 25: The setting definition will assert strings only as values; the user could _still_ ignore the resulting warning, but that's on them.

karolz-ms (approved):

karolz-ms (approved):

Issue #3475, opened 2022-03-25
Don't communicate with docker until necessary


guilhem:
Currently, docker plugin communicate with docker at each run, even if not necessary.
When using socket activation, it spawns docker daemon + all networks, and it has impact on battery consumption.

Postponing docker communication will prevent it.

Another way can be to separate `Dockerfile` linting plugin from Docker management, itself.
User being able to disable Docker communication without losing linting.

bwateratmsft:
This would probably be solved by #3263. Right now our initial communications with Docker are to get the context so we can set up Dockerode; if the CLI alone was used then none of that would be needed.

guilhem:
thanks @bwateratmsft, I subscribed.

bwateratmsft:
Fixed by #3263

bwateratmsft:
This has now been released in Docker extension version 1.23.0.

PR #3586, opened 2022-07-21
Use Docker CLI only

### LINKED ISSUE #3263: Make a CLI-only `DockerApiClient`

bwateratmsft:
## Overview
It's finally time ~~(sorta, starting as just a draft)~~! This PR is _very_ big but it does a couple of key things:

- Uses the Docker (and compose) CLI only for everything--no more Dockerode, no more HTTP endpoint, none of that
- Dumps the old `ContextManager` since it's not really necessary anymore
- Depends on #3637
- Closes #3263 and presumably a ton of the linked issues within

## Verification Tasks
- [x] Windows
    - [x] Debug Python
        - [x] Launch
        - [x] Attach
    - [x] Debug NodeJS
        - [x] Launch
        - [x] Attach
    - [x] Debug .NET
        - [x] Launch
        - [x] Attach
    - [x] Explorer
        - [x] Files
            - [x] Linux containers
            - [x] Windows containers
        - [x] Outdated images
        - [x] Tooltips
        - [x] Start/stop/etc. containers
        - [x] Inspect
        - [x] Logs
        - [x] Compose groups
        - [x] Performance
    - [x] Scaffolding
    - [x] Compose subsets
    - [x] Compose tasks
    - [x] Command customization
    - [x] Change runtime info
    - [x] No communication w/ Docker before necessary (#3475)
    - [x] Activation performance
- [x] The same thing but Mac
- [x] The same thing but Linux

## Bugs as I find them, checked off as I fix them
- [x] Clean up remaining `// TODO: runtimes` items
- [x] "Run Azure CLI Image" command not working
- [x] "Pull repository" command not working
- [x] Erroneous warnings about Docker not being installed https://github.com/microsoft/vscode-docker-extensibility/issues/122
- [x] Error in .NET attach config
- [x] Excessive warnings in output log about use of `DOCKER_HOST` environment variable
- [x] Seems like changes to env vars are not making it into VSCode terminals
- [x] Tasks are overescaped and not executing
- [x] Spurious prompt when doing compose up
- [x] Image "full tag" is not including tag https://github.com/microsoft/vscode-docker-extensibility/issues/119
- [x] Image names with `-` are not handled properly https://github.com/microsoft/vscode-docker-extensibility/issues/120
- [x] The "Show Dangling Images" option is showing _only_ dangling images
- [x] Images without a tag cannot be removed
- [x] Shell isn't being set up properly
- [x] ~~ACI doesn't really work at all https://github.com/microsoft/vscode-docker-extensibility/issues/90~~ Removing ACI: #3642
- [x] ~~FS performance is diminished due to ~~https://github.com/microsoft/vscode/issues/156691~~ need for LRU cache to maintain tree item stat info~~ I don't find any apparent difference in behavior if `stat` returns no useful info (except file type)
- [x] Only `sh` getting opened in attach shell, never `bash`
- [x] Can't run processes on Mac
- [x] File ops fail on Windows containers https://github.com/microsoft/vscode-docker-extensibility/pull/95
- [x] Writing files on Linux fails due to incorrect destination and unterminated tar stream
- [x] Outdated image checking is not working
- [x] Command visibility VSCode contexts not being set when Docker context changes
- [x] `nodeUtils.test.ts` test failures
- [x] Last-chosen service subset is not being remembered https://github.com/microsoft/vscode-azuretools/issues/1216
- [x] Related to https://github.com/microsoft/vscode-docker-extensibility/issues/90, too frequent context checks
- [x] Bad startup performance on explorer view
- [x] Issues parsing some port strings https://github.com/microsoft/vscode-docker-extensibility/issues/41
- [x] Compose groups aren't getting grouped (label not making it?) https://github.com/microsoft/vscode-docker-extensibility/pull/92
- [x] Volume listing error: cannot read properties of undefined (reading 'valueOf')
- [x] NodeJS debugging not working
- [x] Debugging does not launch browser automatically https://github.com/microsoft/vscode-docker-extensibility/pull/93
- [x] Tooltip failing on images w/ containers
- [x] Image sizes are wrong https://github.com/microsoft/vscode-docker-extensibility/pull/91
- [x] Compose group commands not working (cannot read properties of undefined, reading 'commandName')

Note: I found the explorer view performance to be degraded due to three things:
- The first `docker` CLI call is a little bit slower, likely due to some esoteric OS caching things
- Too many calls to check Docker context (item added above for that)
- The biggest thing: VSCode is activating some extensions (according to the spinner in the status bar at the bottom), but it's really not clear to me what. The extension host log did not show anything, nor did the running extensions. I've noticed that if I wait a while after startup, and then go to the explorer view, performance is much better.


package.json
```diff
@@ -1250,9 +1250,7 @@
                                             "description": "%vscode-docker.tasks.docker-run.dockerRun.volumes.permissions%",
                                             "enum": [
                                                 "rw",
-                                                "ro",
-                                                "rw,z",
```
karolz-ms, line 1254: the ",z" variants were for SE-Linux or something. Any particular reason for removing them?

src/runtimes/runners/TaskCommandRunnerFactory.ts
```diff
@@ -0,0 +1,79 @@
...
+    rejectOnError?: boolean;
+    focus?: boolean;
+    env?: never; // Environment is not needed and should not be used, because VSCode adds it already (due to using `ExtensionContext.environmentVariableCollection`)
+}
+
+export class TaskCommandRunnerFactory implements ICommandRunnerFactory {
+    public constructor(private readonly options: TaskCommandRunnerOptions) {
+    }
+
+    public getCommandRunner(): CommandRunner {
```
karolz-ms, line 24: I am wondering if `getCommandRunner` should be just a factory method, taking options as a parameter. The code consuming it could be simpler then (no need to create a separate "factory" object).

src/runtimes/runners/TaskCommandRunnerFactory.ts
```diff
@@ -0,0 +1,79 @@
...
+            const commandResponse: CommandResponse<T> = await normalizeCommandResponseLike(commandResponseLike);
+            await executeAsTask(this.options, commandResponse.command, commandResponse.args);
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            return undefined!;
+        };
+    }
+}
+
+async function executeAsTask(options: TaskCommandRunnerOptions, command: string, args?: vscode.ShellQuotedString[]): Promise<void> {
+    const shellExecutionOptions = { cwd: options.cwd || options.workspaceFolder?.uri?.fsPath || os.homedir() };
```
karolz-ms, line 35: Do you also want to allow for passing additional environment variables?

karolz-ms (commented):
I like where this is going--the more abstract API for interacting with the container orchestrator feels right

(this is part 1 of the review, will continue)

package.json
bwateratmsft: Great question. @danegsta opted to use the newer `--mount` syntax for volume mapping, which does not support the SELinux relabel feature. Docker Inc's guidance on the topic was basically "We don't really want to be in the business of relabeling things, so _you_ do that on your own, before we get involved".

src/runtimes/runners/TaskCommandRunnerFactory.ts
bwateratmsft: It was originally implemented that way, but I cannot remember why we changed it. @danegsta do you remember?

src/runtimes/runners/TaskCommandRunnerFactory.ts
bwateratmsft: We can probably put off adding that until we need it. The environment variables in the `containers.environment` setting will be automatically applied by VSCode itself.

karolz-ms: sounds good

package.json
karolz-ms: Roger

src/runtimes/runners/TaskCommandRunnerFactory.ts
danegsta: If I remember correctly, it was a combination of using a factory object making the typings much more straightforward and making it easier to pass around/re-use a given factory multiple times in a given UI operation as the config should only need to be set once.

karolz-ms: Not sure about the typing part, but for the re-use, given that the factory's only function is to hold onto options, one could just create the "options" object and re-use it as needed. But in majority of use cases I have seen so far during the review, there is only a single call to `getCommandRunner`, and so the options could be created inline, making the code simpler.

Just sayin ☺️ No biggie, I will let you and Brandon decide what to do.

karolz-ms (dismissed):
FWIIW, looks good. I tried to focus on changes to foundational classes, but even with that, it is a lot of changes to review, so I reserve the right to review some "how could I miss this?!" changes in future ☺️

src/runtimes/ContainerRuntimeManager.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export class ContainerRuntimeManager extends RuntimeManager<IContainersClient> {
+    public readonly contextManager: IContextManager = new ContextManager();
+    public readonly onContainerRuntimeClientRegistered = this.runtimeClientRegisteredEmitter.event;
+
+    public constructor() {
+        super('containerClient');
+    }
+
+    // TODO: runtimes: temporarily just return the Docker client, always
```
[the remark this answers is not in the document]
bwateratmsft: That's right. Currently it's Docker only, but future plans have a setting to control the runtime client that gets used.

src/runtimes/OrchestratorRuntimeManager.ts
```diff
@@ -0,0 +1,24 @@
...
+import { RuntimeManager } from './RuntimeManager';
+
+export class OrchestratorRuntimeManager extends RuntimeManager<IContainerOrchestratorClient> {
+    public readonly onOrchestratorRuntimeClientRegistered = this.runtimeClientRegisteredEmitter.event;
+
+    public constructor() {
+        super('orchestratorClient');
+    }
+
+    // TODO: runtimes: temporarily just return the Docker Compose client, always
```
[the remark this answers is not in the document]
bwateratmsft: That's right. Currently it's Docker only, but future plans have a setting to control the runtime client that gets used.

src/debugging/DockerServerReadyAction.ts
```diff
@@ -213,11 +215,30 @@ class DockerLogsTracker extends vscode.Disposable {
...
+                    stdOutPipe: this.logStream,
+                }
+            ).then(
+                () => {
+                    // Do nothing on fulfilled
+                },
+                () => {
+                    // Do nothing on reject
+                    // Needed because the usual termination path is for a cancellation error to be thrown,
+                    // because disposal of this object cancels the token and thus the process
```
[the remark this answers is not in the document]
bwateratmsft: I'll add some details. In short, cancelling the token kills the process which makes `runWithDefaultShell` throw a `CommandCancelledError`. If you `void` something and it throws, it will bubble up to the extension host process. The host process eats it, but in the "Show Running Extensions" view it leaves warnings, and also spits them to the host process' console.

Nelkins3 (approved):

bwateratmsft:
Enough things have changed since the last test pass that I'm going to uncheck all the items to test, so I can take another pass.

src/tree/settings/CommonProperties.ts
```diff
@@ -30,14 +30,12 @@ export const sortByProperties: ITreePropertyInfo<CommonSortBy>[] = [
     { property: 'Label', description: localize('vscode-docker.tree.settings.label', 'Sort alphabetically by label') }
 ];
 
-export function getCommonPropertyValue(item: DockerObject, property: CommonProperty): string {
+export function getCommonPropertyValue(item: { createdAt?: Date, size?: number }, property: CommonProperty): string {
```
danegsta, line 33: It might be cleaner and less error prone to split this up into "getNormalizedCreatedTime" and "getNormalizedSize" or something along those lines, since this is effectively two different functions glued together with a switch statement.

src/tree/images/ImageTreeItem.ts
```diff
@@ -71,32 +74,48 @@ export class ImageTreeItem extends ToolTipTreeItem {
...
     public async resolveTooltipInternal(actionContext: IActionContext): Promise<MarkdownString> {
         actionContext.telemetry.properties.tooltipType = 'image';
-        return resolveTooltipMarkdown(imageTooltipTemplate, { NormalizedName: this.fullTag, NormalizedSize: getCommonPropertyValue(this._item, 'Size'), ...await ext.dockerClient.inspectImage(actionContext, this.imageId) });
+
+        const imageInspection = (await ext.runWithDefaultShell(client =>
+            client.inspectImages({ imageRefs: [this.imageId] })
+        ))?.[0];
+        const associatedContainers = await ext.runWithDefaultShell(client =>
+            client.listContainers({ imageAncestors: [this.imageId] })
+        );
```
danegsta, line 93: Any reason not to wrap this in a Promise.all to allow some parallelization of the CLI calls?

danegsta (commented):
A few small comments, but so far everything looks great; it's exciting to see all the work coming together.

src/tree/settings/CommonProperties.ts
bwateratmsft: This is used by the grouping/labeling system in the tree views. Since the property name to look at (e.g. `CreatedTime` or `Size`) comes from VSCode settings, there would inevitably be a switch statement _somewhere_--either here, where the two "function bodies" are, or in a method reaching out to those two separated functions.

All of the rest of them use something like `getContainerPropertyValue` / `getImagePropertyValue` / etc. and contain all the "bodies" in big switch statements. We could of course refactor that but is there much value?

danegsta: Makes sense; I'd initially found this method via cases where it was being called with a hardcoded value for the property, but if it gets used dynamically as well, the potential benefit of a refactor is pretty minimal.

danegsta (approved):

danegsta (dismissed):

danegsta (approved):

PR #3603, opened 2022-08-18
Added option for running docker container in interactive

kai-nguyen-aligent:
I would like to propose this change to support interactive mode while running docker container from a task. Currently, there is no way to remove the `-d` flag as it is hard-coded. I tried to use the customRun option and passing in `-i` but it has no effect.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1128,6 +1128,11 @@
                                 "type": "string",
                                 "description": "%vscode-docker.tasks.docker-run.dockerRun.containerName%"
                             },
+                            "detached": {
+                                "type": "boolean",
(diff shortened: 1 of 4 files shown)


bwateratmsft:
Unfortunately, due to limitations with VSCode's `CustomExecution`, this isn't possible, at least not in a way that we feel is "right" (we tried in #3083). "Interactive" shells can be opened, but they are one-way--you can read, but you cannot write or interact in any way. For example, if the `docker run -it ...` call created an interactive shell, it would not really be interactive as you could not enter anything.

If all that is needed is to _read_, e.g. view logs, you can accomplish this by running the "Docker Containers: View Logs" command after starting the container.

PR #3669, opened 2022-10-25
Autodetect compose configuration

### LINKED ISSUE #3668: Auto-detect Compose V2 vs. V1?
Pertaining to https://github.com/microsoft/vscode-docker/blob/24fd906a9c22d596cc427743feaf91ac02b21205/src/extension.ts#L208, we need to decide whether we want to continue auto-detecting Compose V2 vs. V1. Our docs [imply](https://code.visualstudio.com/docs/containers/reference#_docker-compose-up) that we do this, so if not we should update the docs too.

bwateratmsft:
Fixes #3668

### DIFF
--- a/src/runtimes/clients/AutoConfigurableDockerComposeClient.ts
+++ b/src/runtimes/clients/AutoConfigurableDockerComposeClient.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ext } from '@microsoft/vscode-azext-azureappservice/out/src/extensionVariables';
 import * as vscode from 'vscode';
+import { ext } from '../../extensionVariables';
 import { execAsync } from '../../utils/execAsync';
 import { AsyncLazy } from '../../utils/lazy';
 import { DockerComposeClient, IContainerOrchestratorClient } from '../docker';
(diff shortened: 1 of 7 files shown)


karolz-ms (dismissed):

src/runtimes/clients/AutoConfigurableDockerComposeClient.ts
```diff
@@ -0,0 +1,83 @@
...
+            return {
+                commandName: composeCommand,
+                composeV2: isComposeV2,
+            };
+        } else {
+            // User has not set a compose command, so we will attempt to autodetect it
+
+            try {
+                ext.outputChannel.appendLine('Attempting to autodetect Docker Compose command...');
+                await execAsync('docker compose version');
```
bwateratmsft, line 61: Interested in others' thoughts on an idea I have. If we successfully detect Compose V2, we could stamp a memento, to avoid ever running this again, saving on perf. This makes the assumption that once a user installs a CLI version with Compose V2, they will never go back to one without it--if they do, they must use the setting to override the conclusion from that memento.

karolz-ms: How bad is the perf hit?

bwateratmsft: Seemed like something in the vicinity of half a second, once per session, and only if you use compose features. Not bad.

karolz-ms: In that case my vote would be to err on the side of more robust/frequent detection

src/utils/execAsync.ts
```diff
@@ -20,7 +20,7 @@ export async function execAsync(command: string, options?: cp.ExecOptions & { st
     const stderrFinal = new AccumulatorStream();
 
     let stdinPipe: stream.Readable | undefined;
-    if (options.stdin) {
+    if (options?.stdin) {
```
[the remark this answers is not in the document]
bwateratmsft: This fixes #3677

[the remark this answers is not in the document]
tuxtvixx: Thanks for helping

PR #3672, opened 2022-10-26
Bump minimum VSCode version

### LINKED ISSUE #3671: Change the minimum VS Code version that the Docker extension supports
**OS:** Win10
**Build Version:** 20221021.9

**Repro Steps:**

1. Figure out the oldest vs code version that the extension supports, the version is 1.66.0 as below.
image
2. Install VS Code 1.66.2 from https://code.visualstudio.com/updates.
3. Install the Docker extension from VSIX.
4. Check whether no error occurs.

**Expect:**
No error occurs.

**Actual:**
An error occurs.
image

bwateratmsft:
Fixes #3671

### DIFF
--- a/package.json
+++ b/package.json
@@ -2899,7 +2899,7 @@
         ]
     },
     "engines": {
-        "vscode": "^1.66.0"
+        "vscode": "^1.67.0"
     },
     "capabilities": {
         "virtualWorkspaces": false,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2944,7 +2944,7 @@
         "@types/node": "16.x",
         "@types/node-fetch": "^2.5.12",
         "@types/semver": "^7.3.9",
-        "@types/vscode": "1.66.0",
+        "@types/vscode": "1.67.0",
```
bwateratmsft, line 2947: I keep this in sync with the minimum VSCode version because `vsce` can get mad if there's a mismatch.

karolz-ms (approved):

PR #3680, opened 2022-10-28
Base explorer refreshes on streamed events instead of polling so frequently

### LINKED ISSUE #3674: Use `docker events` to avoid making so many CLI calls
When idling on the Docker explorer view, there's a small but noticeable increase in CPU usage, presumably due to all the CLI calls from polling. The `docker events` feature enables us to avoid making unnecessary polling calls. Probably the simplest approach (and what the container tools window in VS does) is to use it just as an event notification, but not actually use the data from the events. This would eliminate the need to poll on a schedule while still keeping the implementation of the current views essentially unchanged.

bwateratmsft:
Fixes #3674

### DIFF
--- a/src/runtimes/docker/contracts/CommandRunner.ts
+++ b/src/runtimes/docker/contracts/CommandRunner.ts
@@ -5,33 +5,40 @@
 
 import { CommandLineArgs } from '../utils/commandLineBuilder';
 
-type CommandResponseBase = {
+/**
+ * A command response includes the command (i.e., the executable) to execute, and arguments to pass
+ */
+export type CommandResponseBase = {
     command: string;
     args: CommandLineArgs;
 };
 
+/**
+ * A {@link CommandResponseBase} that also includes a method to parse the output of the command
+ */
 export type PromiseCommandResponse<T> = CommandResponseBase & {
     parse: (output: string, strict: boolean) => Promise<T>;
 };
 
+/**
+ * A {@link CommandResponseBase} that also includes a method to parse streaming output of the command
+ * as an {@link AsyncGenerator}
+ */
 export type GeneratorCommandResponse<T> = CommandResponseBase & {
     parseStream: (output: NodeJS.ReadableStream, strict: boolean) => AsyncGenerator<T>;
 };
 
+/**
+ * A {@link CommandResponseBase} that cannot include parsing methods--i.e. the output is `void`
+ */
 export type VoidCommandResponse = CommandResponseBase & {
     parse?: never;
     parseStream?: never;
 };
 
 /**
- * A CommandResponse record provides instructions on how to invoke a command
- * and a parse callback that can be used to parse and normalize the standard
- * output from invoking the command. This is the standard type returned by all
- * commands defined by the IContainersClient interface.
+ * A helper type that allows for several simple ways to resolve an item
  */
-export type CommandResponse<T> = PromiseCommandResponse<T> | GeneratorCommandResponse<T> | VoidCommandResponse;
-
-
 export type Like<T> = T | Promise<T> | (() => T | Promise<T>);
 
 /**
@@ -44,7 +51,7 @@ export type CommandRunner =
 /**
  * A {@link StreamingCommandRunner} provides instructions on how to invoke a streaming command
  */
-export type StreamingCommandRunner = <T>(commandResponseLike: Like<GeneratorCommandResponse<T>>) => Promise<AsyncGenerator<T>>;
+export type StreamingCommandRunner = <T>(commandResponseLike: Like<GeneratorCommandResponse<T>>) => AsyncGenerator<T>;
 
 /**
  * A {@link ICommandRunnerFactory} is used to build a CommandRunner instance
@@ -55,6 +62,11 @@ export interface ICommandRunnerFactory {
     getStreamingCommandRunner(): StreamingCommandRunner;
 }
 
+/**
+ * Converts a `Like<CommandResponse>` into a `CommandResponse`
+ * @param commandResponseLike The command response-like to normalize
+ * @returns The command response
+ */
 export function normalizeCommandResponseLike<TCommandResponse extends CommandResponseBase>(commandResponseLike: Like<TCommandResponse>): Promise<TCommandResponse> {
     if (typeof commandResponseLike === 'function') {
         return Promise.resolve(commandResponseLike());
--- a/src/runtimes/docker/contracts/CommandRunner.ts
+++ b/src/runtimes/docker/contracts/CommandRunner.ts
@@ -5,27 +5,46 @@
 
 import { CommandLineArgs } from '../utils/commandLineBuilder';
 
+type CommandResponseBase = {
+    command: string;
+    args: CommandLineArgs;
+};
+
+export type PromiseCommandResponse<T> = CommandResponseBase & {
+    parse: (output: string, strict: boolean) => Promise<T>;
+};
+
+export type GeneratorCommandResponse<T> = CommandResponseBase & {
+    parseStream: (output: NodeJS.ReadableStream, strict: boolean) => AsyncGenerator<T>;
+};
+
+export type VoidCommandResponse = CommandResponseBase & {
+    parse?: never;
+    parseStream?: never;
+};
+
 /**
  * A CommandResponse record provides instructions on how to invoke a command
  * and a parse callback that can be used to parse and normalize the standard
  * output from invoking the command. This is the standard type returned by all
  * commands defined by the IContainersClient interface.
- *
- * Parse will not be implemented for streaming operations, like container logs
- * or files.
  */
-export type CommandResponse<T> = {
-    command: string;
-    args: CommandLineArgs;
-    parse?: (output: string, strict: boolean) => Promise<T>;
-};
+export type CommandResponse<T> = PromiseCommandResponse<T> | GeneratorCommandResponse<T> | VoidCommandResponse;
 
-export type CommandResponseLike<T> = CommandResponse<T> | Promise<CommandResponse<T>> | (() => CommandResponse<T> | Promise<CommandResponse<T>>);
+
+export type Like<T> = T | Promise<T> | (() => T | Promise<T>);
 
 /**
  * A {@link CommandRunner} provides instructions on how to invoke a command
  */
-export type CommandRunner = <T>(commandResponse: CommandResponseLike<T>) => Promise<T>;
+export type CommandRunner =
+    (<T>(commandResponseLike: Like<PromiseCommandResponse<T>>) => Promise<T>) &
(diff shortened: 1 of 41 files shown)


src/tree/RefreshManager.ts
```diff
@@ -0,0 +1,169 @@
...
+export class RefreshManager extends vscode.Disposable {
+    private readonly refreshEmitter = new vscode.EventEmitter<[RefreshTarget, RefreshReason]>();
+    private readonly disposables: vscode.Disposable[] = [];
+    private readonly cts: vscode.CancellationTokenSource = new vscode.CancellationTokenSource();
+
+    public constructor() {
+        super(() => vscode.Disposable.from(...this.disposables).dispose());
+
+        this.disposables.push(
+            this.refreshEmitter.event(async ([target, reason]) => {
```
bwateratmsft, line 28: This is needlessly complex, an event isn't necessary to forward to an always-singleton listener.

src/tree/RefreshManager.ts
```diff
@@ -0,0 +1,246 @@
...
+        void callWithTelemetryAndErrorHandling('vscode-docker.tree.eventRefresh', async (context: IActionContext) => {
+            context.errorHandling.suppressDisplay = true;
+            context.telemetry.suppressIfSuccessful = true;
+
+            // Try at most `eventListenerTries` times to (re)connect to the event stream
+            for (let i = 0; i < eventListenerTries; i++) {
+                try {
+                    const eventGenerator = await ext.streamWithDefaultShell(client =>
+                        client.getEventStream({
+                            types: ['container', 'image', 'network', 'volume'],
```
bwateratmsft, line 64: An issue I've found is that events are getting fired for `docker exec` calls in containers, which we do a lot of for files...which forces it into a refresh loop. These exec calls need to be ignored.

src/runtimes/files/ContainerFilesProvider.ts
```diff
@@ -80,19 +80,18 @@ export class ContainerFilesProvider extends vscode.Disposable implements vscode.
...
                         operatingSystem: containerOS,
-                    }),
-                    ext.runtimeManager,
-                    {
-                        stdOutPipe: containerOS === 'windows' ? accumulator : tarUnpackStream(accumulator),
-                    }
+                    })
                 );
 
+                stream.Readable.from(generator).pipe(targetStream);
```
bwateratmsft, line 93: This is never returning

src/runtimes/docker/contracts/CommandRunner.ts
```diff
@@ -5,37 +5,57 @@
 
 import { CommandLineArgs } from '../utils/commandLineBuilder';
 
+type CommandResponseBase = {
+    command: string;
+    args: CommandLineArgs;
+};
+
+export type PromiseCommandResponse<T> = CommandResponseBase & {
```
karolz-ms, line 13: The old `CommandResponse` was already returning a promise from `parse()`, so I suggest keeping the old name. That will reduce the number of changes in this PR and also avoid having "promise of a promise" in the signatures, which does not read very comfortably.

src/runtimes/docker/contracts/CommandRunner.ts
```diff
@@ -5,37 +5,57 @@
...
- *
- * Parse will not be implemented for streaming operations, like container logs
- * or files.
  */
-export type CommandResponse<T> = {
-    command: string;
-    args: CommandLineArgs;
-    parse?: (output: string, strict: boolean) => Promise<T>;
-};
+export type CommandResponse<T> = PromiseCommandResponse<T> | GeneratorCommandResponse<T> | VoidCommandResponse;
```
karolz-ms, line 32: Where is this used? Do we need this? Would `CommandResponseBase` be sufficient?

src/runtimes/docker/contracts/CommandRunner.ts
```diff
@@ -5,37 +5,57 @@
...
 
-export type CommandResponseLike<T> = CommandResponse<T> | Promise<CommandResponse<T>> | (() => CommandResponse<T> | Promise<CommandResponse<T>>);
+
+export type Like<T> = T | Promise<T> | (() => T | Promise<T>);
```
karolz-ms, line 35: This is convenient

karolz-ms (commented):
Sharing my thoughts so far, will continue tomorrow

src/tree/RefreshManager.ts
```diff
@@ -0,0 +1,255 @@
...
+            context.errorHandling.suppressDisplay = true;
+            context.telemetry.suppressIfSuccessful = true;
+
+            const eventTypesToWatch: EventType[] = ['container', 'image', 'network', 'volume'];
+            const eventActionsToWatch: EventAction[] = Array.from(new Set<EventAction>([...ContainerEventActions, ...ImageEventActions, ...NetworkEventActions, ...VolumeEventActions]));
+
+            // Try at most `eventListenerTries` times to (re)connect to the event stream
+            for (let i = 0; i < eventListenerTries; i++) {
+                try {
+                    const eventGenerator = await ext.streamWithDefaultShell(client =>
```
karolz-ms, line 70: Can the events come faster than we can process them? Should we "debounce" the refresh invocations?

karolz-ms (approved):
Overall looks very good, just a few suggestions/questions

src/runtimes/docker/contracts/ContainerClient.ts
```diff
@@ -152,7 +152,82 @@ type InfoCommand = {
...
-    info(options: InfoCommandOptions): Promise<CommandResponse<InfoItem>>;
+    info(options: InfoCommandOptions): Promise<PromiseCommandResponse<InfoItem>>;
+};
+
+// Event Stream Command Types
+
+/**
+ * Types of objects that can be listened for events to
+ */
+export type EventType = 'container' | 'image' | 'network' | 'volume' | 'daemon' | 'plugin' | 'config' | 'secret' | 'service' | 'node' | 'task' | 'engine';
```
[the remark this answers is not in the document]
bwateratmsft: That's a good idea. It wouldn't affect parsing--at runtime this is just a string--but it accounts for the possibility of types beyond these. The change needs to first be made upstream, so I'll do that, and then update in this PR.

src/runtimes/runners/runWithDefaultShell.ts
```diff
@@ -3,70 +3,134 @@
...
+    const factory = new DefaultEnvShellStreamCommandRunnerFactory(additionalOptions);
 
+    // Get the active client
+    const client: TClient = await runtimeManager.getClient();
+
+    try {
+        // Flatten the callback
+        const response = await normalizeCommandResponseLike(callback(client));
+
+        if (response.parse) {
```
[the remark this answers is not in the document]
bwateratmsft: It won't*

*Except for e.g. https://github.com/microsoft/vscode-docker/pull/3680/files#diff-a4dea273c47cea16ce8ce1a56393835548720d231dc73cbf1dbbab8e2e02ed02R38, but we intentionally do _not_ want to treat it like a `GeneratorCommandResponse<T>` in that case.

src/tree/RefreshManager.ts
bwateratmsft: They do come pretty fast in certain circumstances and we probably should debounce. The net effect if we don't is that we spawn a couple of CLI calls at once, which can return in an arbitrary order, but since we just update the data to whatever is returned it isn't harmful (aside from the wasted CPU time).

I'm still deciding what the best debouncing approach is. There's basically two ways I can think of:

1. Ignore events of the same type that come in within X milliseconds of the last one
2. Collapse rapidly-fired events of the same type into one refresh that occurs X milliseconds after the last one (more or less how we do diagnostics in the compose language service--diagnostics get sent ~1 second after the user stops typing)

The disadvantage of the first strategy is that we could miss relevant changes to the data that occur after the first call; the disadvantage of the second is the implementation is a little more complex and it adds latency before real changes are reflected in the UI. I think the second approach is still better though.

src/runtimes/docker/contracts/CommandRunner.ts
bwateratmsft: IMO we should keep it as is; calling it "Promise" is to contrast it from `GeneratorCommandResponse<T>` which has an `AsyncGenerator<T>` return from its parsing method.

src/runtimes/docker/contracts/ContainerClient.ts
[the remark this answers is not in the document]
bwateratmsft: (And, with https://github.com/microsoft/vscode-docker-extensibility/pull/133, it's no longer a `Promise<AsyncGenerator<T>>`)

src/tree/RefreshManager.ts
karolz-ms: Agreed. Thanks for good explanation

src/runtimes/docker/contracts/CommandRunner.ts
karolz-ms: I understand, but "linguistics" PoV it makes sense to call the most common case just "Foo", and the special case "SpecialFoo". E.g glider vs motor glider. 

I do not feel strongly about this, but still think it would be a simplification.

src/tree/RefreshManager.ts
```diff
@@ -0,0 +1,275 @@
...
+            context.errorHandling.suppressDisplay = true;
+            context.telemetry.suppressIfSuccessful = true;
+
+            const eventTypesToWatch: EventType[] = ['container', 'image', 'network', 'volume'];
+            const eventActionsToWatch: EventAction[] = Array.from(new Set<EventAction>([...ContainerEventActions, ...ImageEventActions, ...NetworkEventActions, ...VolumeEventActions]));
+
+            // Try at most `eventListenerTries` times to (re)connect to the event stream
+            for (let i = 0; i < eventListenerTries; i++) {
+                try {
+                    const eventGenerator = ext.streamWithDefaultShell(client =>
```
bwateratmsft, line 72: The process isn't being cancelled

src/tree/RefreshManager.ts
```diff
@@ -0,0 +1,275 @@
...
+    private readonly disposables: vscode.Disposable[] = [];
+    private readonly cts: vscode.CancellationTokenSource = new vscode.CancellationTokenSource();
+
+    public constructor() {
+        super(() => vscode.Disposable.from(...this.disposables).dispose());
+
+        // VSCode does *not* cancel by default on disposal of a CancellationTokenSource, so we need to manually cancel
+        this.disposables.unshift(new vscode.Disposable(() => this.cts.cancel()));
+
+        this.setupRefreshOnInterval();
```
bwateratmsft, line 37: An ill effect of this is that we're now initiating communication with Docker before needed, which would regress #3475. Need to fix this, so that none of these events are set up until the tree is refreshed for the first time.

karolz-ms (approved):
nice

PR #3693, opened 2022-11-09
Restart the event listener every 5 minutes

### LINKED ISSUE #3692: In certain circumstances the `docker events` process can be leaked and left running
I've been doing some experimentation and found that the `docker events` process started for #3674 can be "leaked" and left running. In normal VSCode shutdowns, the extension deactivates and the process is killed as expected. However, if I do "Reload Window", it seems that the deactivation pathway never occurs. The PowerShell process (on Windows) is killed properly, but the `docker events` process underneath it persists.

I'm going to see if I can run this process _not_ in a shell, to avoid this issue.

bwateratmsft:
Fixes #3692

### DIFF
--- a/src/tree/RefreshManager.ts
+++ b/src/tree/RefreshManager.ts
@@ -170,6 +170,8 @@ export class RefreshManager extends vscode.Disposable {
 
                 // If the event generator terminates on its expected lifecycle it will exit without an error
                 // Continue the loop as normal, starting the next event stream at the previous stop time
+                // We intentionally do not reset the `errorCount`, so that at most 3 failures occur before
+                // giving up, regardless of whether or not they are consecutive.
             }
         });
 
--- a/src/tree/RefreshManager.ts
+++ b/src/tree/RefreshManager.ts
@@ -174,7 +174,6 @@ export class RefreshManager extends vscode.Disposable {
                 // giving up, regardless of whether or not they are consecutive.
             }
         });
-
     }
 
     private setupRefreshOnConfigurationChange(): void {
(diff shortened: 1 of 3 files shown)


src/tree/RefreshManager.ts
```diff
@@ -156,7 +163,13 @@ export class RefreshManager extends vscode.Disposable {
 
                         throw error;
                     }
+                } finally {
+                    // Move the next start timestamp to the current end timestamp
+                    eventsSinceTimestamp = eventsUntilTimestamp;
                 }
+
+                // If the event generator terminates on its expected lifecycle it will exit without an error
+                // Continue the loop as normal, starting the next event stream at the previous stop time
```
karolz-ms, line 172: Very much like how you fixed this with such a small change.

Should we reset `errorCount` here?

karolz-ms (approved):

src/tree/RefreshManager.ts
bwateratmsft: That's a good thought. I'm still inclined toward "no", we should not reset it. Here's my thinking--

1. This is a "best-effort" sort of feature, it isn't essential to the extension or explorer working
2. The `errorCount` thing is basically to limit the maximum number of tries we make. Though I'm not sure how it could actually happen, say we're in some scenario where every other call fails. If we reset the `errorCount`, we'll keep doing this indefinitely; if we don't reset it, we'll encounter at most 3 failures before giving up.

karolz-ms: Sounds reasonable. Maybe add 1-line comment that giving up after max tries is the intention, so no resetting of error count.

karolz-ms (approved):

PR #3700, opened 2022-11-15
Add root folder flag to build ACR image

OliverMKing:
Hello! Hope you're doing well.

This PR adds an optional flag to the Build ACR Image command. If used, it will swap the root folder from the `workspace folder` to the directory that the Dockerfile is stored in. Currently, this flag is only accessed by another extension passing it in with the `executeCommand` function.

We want to use this in the `ms-kubernetes-tools.aks-devx-tools` extension like the following:
```TypeScript
vscode.commands.executeCommand(
      'vscode-docker.registries.azure.buildImage',
      undefined,
      RootStrategy.DOCKERFILE_FOLDER
);
```

This change was tested by generating a .vsix of the Docker extension and using it with the above snippet. It will successfully change the root folder for the image.

While this new flag is only accessible through executing the command with another extension it would be trivial to implement it in this extension. The assumption that the built image should always use the `workspace folder` as the root was incorrect for our usecase and is likely something that other users would like to specify as well.

This is the last change that the `ms-kubernetes-tools.aks-devx-tools` extension needs to effectively consume this extension. If needed, I'm more than happy to discuss more or discuss alternative solutions.

Thank you for your time and review!

### DIFF
--- a/src/commands/registries/azure/tasks/scheduleRunRequest.ts
+++ b/src/commands/registries/azure/tasks/scheduleRunRequest.ts
@@ -62,8 +62,6 @@ export async function scheduleRunRequest(context: IActionContext, requestType: '
             // changes the root to the folder where the Dockerfile is
             // it is used by the ms-kubernetes-tools.aks-devx-tools extension (https://github.com/Azure/aks-devx-tools)
             rootUri = vscode.Uri.file(path.dirname(fileItem.absoluteFilePath));
-            const fileUri = vscode.Uri.file(fileItem.absoluteFilePath);
-            fileItem = createFileItem(rootUri, fileUri);
         }
 
         const uploadedSourceLocation: string = await uploadSourceCode(await node.getClient(context), node.registryName, node.resourceGroup, rootUri, tarFilePath);
@@ -77,7 +75,7 @@ export async function scheduleRunRequest(context: IActionContext, requestType: '
                 isPushEnabled: true,
                 sourceLocation: uploadedSourceLocation,
                 platform: { os: osType },
-                dockerFilePath: fileItem.relativeFilePath
+                dockerFilePath: path.relative(rootUri.fsPath, fileItem.absoluteFilePath)
             };
         } else {
             runRequest = {
--- a/src/commands/registries/azure/tasks/scheduleRunRequest.ts
+++ b/src/commands/registries/azure/tasks/scheduleRunRequest.ts
@@ -15,7 +15,7 @@ import { localize } from "../../../../localize";
 import { AzureRegistryTreeItem } from '../../../../tree/registries/azure/AzureRegistryTreeItem';
 import { registryExpectedContextValues } from "../../../../tree/registries/registryContextValues";
 import { delay } from '../../../../utils/promiseUtils';
-import { Item, quickPickDockerFileItem, quickPickYamlFileItem } from '../../../../utils/quickPickFile';
+import { Item, quickPickDockerFileItem, quickPickYamlFileItem, createFileItem } from '../../../../utils/quickPickFile';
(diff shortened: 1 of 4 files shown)


src/commands/registries/azure/tasks/scheduleRunRequest.ts
```diff
@@ -24,7 +24,12 @@ import { getStorageBlob } from '../../../../utils/lazyPackages';
 const idPrecision = 6;
 const vcsIgnoreList = ['.git', '.gitignore', '.bzr', 'bzrignore', '.hg', '.hgignore', '.svn'];
 
-export async function scheduleRunRequest(context: IActionContext, requestType: 'DockerBuildRequest' | 'FileTaskRunRequest', uri: vscode.Uri | undefined): Promise<() => Promise<AcrRun>> {
+export enum RootStrategy {
```
bwateratmsft, line 27: I'd make this enum explicit in its typing instead of the implicit default integer enum. Personally I prefer strings because when debugging, having variables with values `0` or `1` just mean I have to go look at the source code to find out what they really mean. You can make it explicitly string like so:

```typescript
export enum RootStrategy {
    Default = 'Default',
    DockerfileFolder = 'DockerfileFolder',
}
```

An added benefit is that there is zero ambiguity about whether the AKS DevX extension should pass an integer or string.

bwateratmsft, line 27: It's repeated in the code below but can we also add a comment here to note that this enum is shared with the AKS DevX extension?

src/commands/registries/azure/tasks/scheduleRunRequest.ts
```diff
@@ -51,7 +56,16 @@ export async function scheduleRunRequest(context: IActionContext, requestType: '
         // Prepare to run.
         ext.outputChannel.show();
 
-        const uploadedSourceLocation: string = await uploadSourceCode(await node.getClient(context), node.registryName, node.resourceGroup, rootFolder, tarFilePath);
+        let rootUri = rootFolder.uri;
+        if (rootStrategy === RootStrategy.DockerfileFolder) {
```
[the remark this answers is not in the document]
OliverMKing: I like this change! I used `path.relative` because ideally it will work on all OS.

OliverMKing:
Thanks for the initial review! I followed all suggestions and have pushed the modifications.

src/commands/registries/azure/tasks/scheduleRunRequest.ts
```diff
@@ -62,7 +75,7 @@ export async function scheduleRunRequest(context: IActionContext, requestType: '
                 isPushEnabled: true,
                 sourceLocation: uploadedSourceLocation,
                 platform: { os: osType },
-                dockerFilePath: fileItem.relativeFilePath
+                dockerFilePath: path.relative(rootUri.fsPath, fileItem.absoluteFilePath)
             };
         } else {
             runRequest = {
```
bwateratmsft, line 81: Can you change `taskFilePath` in this one to match the `path.relative(...)` above used for the Dockerfile path?

OliverMKing: Done

bwateratmsft (approved):

bwateratmsft:
Thanks again @OliverMKing! Ready for me to merge?

OliverMKing:
> Thanks again @OliverMKing! Ready for me to merge?

Yes please! Thank you so much for your help.

This isn't urgent or anything but what is the typical release cycle for this extension? Is there an estimate on when these things will get cut into a release?

bwateratmsft:
Our normal release schedule is monthly on the third Monday, but since the US Thanksgiving holiday is next week we plan to delay it a week. That would put our next release on November 28.

Issue #3709, opened 2022-11-21
Fail to add docker files to the .NET projects


v-chenzha:
**OS:** Win10
**Build Version:** 20221116.1

**Repro steps:**

1. Create a .NET MVC project with `dotnet new mvc` command and open it in VS Code.
2. F1 -> Select "Add Docker Files to Workspaces..." -> Select ".NET: ASP.NET Core" -> Select "Linux" -> Enter.
3. Check whether the next step is "Include optional Docker Compose files?".

**Expect:**
The next step is "Include optional Docker Compose files?".

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/107091828/203023143-1f82f12d-6f1a-4568-b974-32a3ec3aff33.png)

**More Info:**
1. The issue also reproduces when adding docker files to a .NET Core console project.
2. This issue does not reproduce on Linux and Mac.

karolz-ms:
I cannot reproduce the issue. 😑 At least not on my Win 11 machine. I do not have access to a Win 10 machine at the moment, but I would be surprised if the OS version made a difference. 

The process that is launched between the OS/port selection and the "Include optional Compose files?" question is `dotnet build`, with a custom target.  The code is here:  https://github.com/microsoft/vscode-docker/blob/main/src/utils/netCoreUtils.ts#L17

@v-chenzha 
1. Can you verify that you can `dotnet build` the project from command line, after step 1?
2. Can you try to repro the issue on a different machine and/or Win 11 machine?

v-chenzha:
@karolz-ms 
1. Succeed to build the project on windows 10. This issue doesn't exist after executing `dotnet build` command.
2. This issue only reproduces on windows 10, it doesn't reproduce on Windows 11.
3. This issue is unstable; it is easy to reproduce on new windows 10 machines.

karolz-ms:
@v-chenzha I tried on a freshly set up Windows 10 machine with .NET SDK version 7.0.100 installed, extension build 20221116.8, and I cannot reproduce the issue. What version of .NET are you using (provide output of `dotnet --version` please).

v-chenzha:
@karolz-ms 
Today I create two new windows 10 machines, we can reproduce this issue on one machine, but can't reproduce it on the other one. This is an unstable issue.

This is my dotnet version:
![image](https://user-images.githubusercontent.com/107091828/203485562-a3c84bc7-e8c7-4056-b30a-81ce60eda4c9.png)

[...]

PR #3712, opened 2022-11-23
Fix issue 3709

karolz-ms:
Fixes https://github.com/microsoft/vscode-docker/issues/3709 by:
- increasing the timeout to 20 seconds (the previous 10 seconds might be a bit short in a case of a freshly-started machine with low RAM and slow disk)
- improving error handling in case a process is terminated due to a signal

### DIFF
--- a/src/runtimes/docker/utils/spawnStreamAsync.ts
+++ b/src/runtimes/docker/utils/spawnStreamAsync.ts
@@ -195,6 +195,8 @@ export async function spawnStreamAsync(
             disposable.dispose();
             if (code === 0) {
                 resolve();
+            } else if (signal) {
+                reject(new ChildProcessError(`Process exited due to signal ${signal}`, code, signal));
             } else {
                 reject(new ChildProcessError(`Process exited with code ${code}`, code, signal));
             }
--- a/src/utils/netCoreUtils.ts
+++ b/src/utils/netCoreUtils.ts
@@ -9,6 +9,7 @@ import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { getTempFileName } from './osUtils';
 import { execAsync } from './execAsync';
+import { parseError } from '@microsoft/vscode-azext-utils';
 
 export async function getNetCoreProjectInfo(target: 'GetBlazorManifestLocations' | 'GetProjectProperties', project: string): Promise<string[]> {
     const targetsFile = path.join(ext.context.asAbsolutePath('resources'), 'netCore', `${target}.targets`);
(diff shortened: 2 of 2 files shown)


src/utils/netCoreUtils.ts
```diff
@@ -17,7 +18,13 @@ export async function getNetCoreProjectInfo(target: 'GetBlazorManifestLocations'
     const command = `dotnet build /r:false /t:${target} /p:CustomAfterMicrosoftCommonTargets="${targetsFile}" /p:CustomAfterMicrosoftCommonCrossTargetingTargets="${targetsFile}" /p:InfoOutputPath="${outputFile}" "${project}"`;
 
     try {
-        await execAsync(command, { timeout: 10000 });
+        try {
+            await execAsync(command, { timeout: 20000 });
+        } catch (err) {
+            const error = parseError(err);
+            throw new Error(localize('vscode-docker.netCoreUtils.noProjectInfo', 'Unable to determine project information for target \'{0}\' on project \'{1}\' {2}', target, project, error.message));
```
karolz-ms, line 25: I am not entirely sure if `localize()` requires the message key and text to be spelled out verbatim (so tools can fish them out of code and generate localization files automatically), or if they can be defined as constants somewhere... 

Assumed the former, hence the duplication on line 25 and 37.

danegsta (dismissed):

src/runtimes/docker/utils/spawnStreamAsync.ts
```diff
@@ -195,6 +195,8 @@ export async function spawnStreamAsync(
             disposable.dispose();
             if (code === 0) {
                 resolve();
+            } else if (signal) {
+                reject(new ChildProcessError(`Process exited due to signal ${signal}`, code, signal));
```
bwateratmsft, line 199: This change should be made upstream in https://github.com/microsoft/vscode-docker-extensibility as well

src/utils/netCoreUtils.ts
bwateratmsft: Generally I would just do two separate strings (even if identical), two separate keys here. It might change with the `l10n` API though. Ultimately I don't think it really matters though, unless we ever _actually_ do localization.

bwateratmsft (dismissed):

src/runtimes/docker/utils/spawnStreamAsync.ts
karolz-ms: OK I will submit a separate PR for that

bwateratmsft (approved):

PR #3737, opened 2022-11-30
Bumps default python version to Python 3.10

### LINKED ISSUE #3736: Python Version bump to Python 3.10
Python 3.11 is out and the default object passed is Python 3.8. While I'm not sure if 3.11 is the right move. Bumping the python version from 3.8 to 3.10 seems reasonable.

kjaymiller:
resolves #3736 by changing the image from python 3.8-slim to 3.10-slim


bwateratmsft (approved):
Looks fine to me. Thanks @kjaymiller!

PR #3747, opened 2022-12-05
Run most commands without a shell

### LINKED ISSUE #3725: `Failed to set up event listener: Unexpected token W in JSON at position 0` error
Originally from https://github.com/microsoft/vscode-docker/issues/3724#issuecomment-1330889789

/cc @brainz80

bwateratmsft:
Fixes #3725 either mostly or entirely. There are still scenarios where we will run in a shell; some of these can't plausibly be avoided. For example, because tasks support the `customOptions` argument which is a single string, but could contain multiple arguments that can't easily be separated, it is still necessary to run those through a shell. However, the output is never parsed for those, so I think that everything should work fine.


src/extensionVariables.ts
```diff
@@ -59,8 +59,6 @@ export namespace ext {
     // Container runtime related items
     export let runtimeManager: ContainerRuntimeManager;
     export let orchestratorManager: OrchestratorRuntimeManager;
-    export const runWithDefaultShell = runWithDefaultShellInternal;
-    export const streamWithDefaultShell = streamWithDefaultShellInternal;
-    export const runOrchestratorWithDefaultShell = runOrchestratorWithDefaultShellInternal;
```
bwateratmsft, line 64: The orchestrator ones weren't actually being used, so I opted to just remove them. If we need them in the future they can be added back.

src/runtimes/docker/clients/DockerClientBase/DockerListNetworkRecord.ts
```diff
@@ -1,79 +1,79 @@
-/*---------------------------------------------------------------------------------------------
```
bwateratmsft, line 1: Git wants to change the line endings. No idea why, but it's going to keep trying to do it, so may as well just check it in.

src/runtimes/docker/clients/DockerClientBase/parseDockerLikeEnvironmentVariables.ts
```diff
@@ -1,26 +1,26 @@
-/*---------------------------------------------------------------------------------------------
```
bwateratmsft, line 1: Git wants to change the line endings. No idea why, but it's going to keep trying to do it, so may as well just check it in.

src/runtimes/docker/utils/dayjs.ts
```diff
@@ -1,54 +1,54 @@
-/*---------------------------------------------------------------------------------------------
```
bwateratmsft, line 1: Git wants to change the line endings. No idea why, but it's going to keep trying to do it, so may as well just check it in.

karolz-ms (dismissed):
Looks good, just one question

src/runtimes/docker/commandRunners/wslStream.ts
```diff
@@ -22,20 +22,14 @@ export type WslShellCommandRunnerOptions = ShellStreamCommandRunnerOptions & {
...
         const command = this.options.wslPath ?? 'wsl.exe';
-        const args = [
-            ...(this.options.distro ? ['-d', this.options.distro] : []),
-            '--',
-            commandResponse.command,
-            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            ...Shell.getShellOrDefault(this.options.shellProvider).quote(commandResponse.args),
-        ];
+        const args = composeArgs(
+            withNamedArg('-d', this.options.distro),
```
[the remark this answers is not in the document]
bwateratmsft: IIRC, the `withNamedArg()` function will internally noop if distro is absent.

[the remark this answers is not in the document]
bwateratmsft: Yes, double-checked and that is indeed the behavior: https://github.com/microsoft/vscode-docker-extensibility/blob/main/packages/container-runtimes/src/utils/commandLineBuilder.ts#L99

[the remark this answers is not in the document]
karolz-ms: cool

src/runtimes/runners/runWithDefaults.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export function streamWithDefaultsInternal<T>(callback: StreamingClientCallback<IContainersClient, T>, additionalOptions?: DefaultEnvStreamCommandRunnerOptions): AsyncGenerator<T> {
+    return streamWithDefaults(
+        callback,
+        ext.runtimeManager,
+        additionalOptions
+    );
+}
+
+async function runWithDefaults<TClient extends ClientIdentity, T>(
```
danegsta, line 37: It might be worth defining the method overloads so the type system can properly determine the expected return type based on the value of callback:
```
function runWithDefaults<TClient extends ClientIdentity, T>(callback: ClientCallback<TClient, T>, runtimeManager: RuntimeManager<TClient>, additionalOptions?: DefaultEnvStreamCommandRunnerOptions): Promise<T>;
function runWithDefaults<TClient extends ClientIdentity>(callback: VoidClientCallback<TClient>, runtimeManager: RuntimeManager<TClient>, additionalOptions?: DefaultEnvStreamCommandRunnerOptions): Promise<void>;
async function runWithDefaults<TClient extends ClientIdentity, T>(callback: ClientCallback<TClient, T> | VoidClientCallback<TClient>, runtimeManager: RuntimeManager<TClient>, additionalOptions?: DefaultEnvStreamCommandRunnerOptions): Promise<T | void> { ...
```

bwateratmsft: The definitions of the `runWithDefaultsInternal` / `streamWithDefaultsInternal` accomplish that

danegsta: So they do; should the names of runWithDefaults and runWithDefaultsInternal be swapped? I feel like the current naming is a bit confusing.

bwateratmsft: I think that's reasonable.

danegsta (approved):

PR #3759, opened 2022-12-08
Fix some "shell" quoting bugs

bwateratmsft:
Fixes #3758. Creating as draft since the upstream PR isn't merged yet.

Things to test:
- [x] Windows
    - [x] Container list and inspect
    - [x] Image list and inspect
    - [x] Network list and inspect
    - [x] Volume list and inspect
    - [x] Context list and inspect
    - [x] .NET debug attach
    - [x] .NET debug w/ SSL
    - [x] Files view
        - [x] Windows container #3761
        - [x] Linux container
    - [x] Container actions (start, stop, remove, prune, etc.)
    - [x] Compose subset
    - [x] Unrelated container tooltip change
- [x] Unix
    - [x] Container list and inspect
    - [x] Image list and inspect
    - [x] Network list and inspect
    - [x] Volume list and inspect
    - [x] Context list and inspect
    - [x] .NET debug attach
    - [x] .NET debug w/ SSL
    - [x] Files view
        - [x] Linux container
    - [x] Container actions (start, stop, remove, prune, etc.)
    - [x] Compose subset
    - [x] Unrelated container tooltip change

Issue #3739, opened 2022-12-01
Docker extension does not find docker installed: Failed to connect. Is Docker installed? 


lopezvoliver:
Docker extension does not find docker installed. 

Does this occur consistently? Yes
Repro steps:
1. I installed docker desktop (Docker version 20.10.21, build baeda1f) 
2. I opened a new vscode window and opened the Docker extension. Both "Images" and "Containers" show the error message `Failed to connect. Is Docker installed?`

3. I also tried adding the docker Path directly in the user settings. In my case, the path is:

`C:\Program Files\Docker\Docker\resources\bin\docker.exe`

4. Opening a terminal in vscode shows that the docker command is available. 
4.1 Using a git bash terminal:

```
$ which docker
/c/Program Files/Docker/Docker/resources/bin/docker
```

4.2 Using a powershell terminal:

```
Get-Command docker
CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     docker.exe                                         0.0.0.0    C:\Program Files\Docker\Docker\resources\bin\docker.exe
```

Version: 1.23.1
OS: win32
OS Release: 10.0.19045
Product: Visual Studio Code
Product Version: 1.73.1
Language: en

lopezvoliver:
Rolling back to the extension version v1.22.2 solves the issue for me.

bwateratmsft:
@lopezvoliver does the issue persist if you restart Docker Desktop and VSCode? Is there any other error message or is it just the "not installed" message?

lopezvoliver:
Yes. I tried restarting Docker desktop and vscode, I also tried restarting the PC, and even uninstalling then installing docker again, as well as uninstalling and reinstalling the vscode docker extension. Yes, it was the only error message.

Finally, after installing the previous version of vscode docker extension mentioned above now it is working as expected.

bwateratmsft:
@lopezvoliver A couple of clarifying questions.

1. What do you see if you run this in a PowerShell or CMD terminal? This is the exact process we run to check if Docker is installed.
```sh
powershell.exe -c "docker -v"
```
2. Do you have a value for the setting `docker.dockerPath` in either workspace or user settings?

dsgthb:
@sebeyruv the extension rollback to v.1.22.2 solved the issue also for me. Thank you @lopezvoliver 

Docker version 20.10.21, build baeda1f

VS Code 1.73.1

[...]

PR #3769, opened 2022-12-13
Another attempt at doing things without a shell...

bwateratmsft:
Partially reverts #3759 (the change in `NetCoreDebugHelper.ts`), and pulls down upstream changes.


bwateratmsft:
This might fix some of the causes of #3739.

PR #3772, opened 2022-12-13
Change branding to .NET instead of .NET Core

### LINKED ISSUE #3770: .NET Core 3.1 end-of-life
With the end of life of .NET Core 3.1 being today, we should remove the branding where we call it ".NET Core", and if needed remove code that explicitly supports it.

bwateratmsft:
Fixes #3770. We're not going to change the places where we have `netCore` in the source or in scaffolded user files, though.


karolz-ms (approved):

resources/templates/netCore/Dockerfile.template
```diff
@@ -4,7 +4,7 @@ WORKDIR /app
 EXPOSE {{ . }}
 {{/each}}
 
-{{#if (eq platform '.NET: ASP.NET Core')}}
```
bwateratmsft, line 7: Turns out the branding for ASP.NET is still "ASP.NET Core".

karolz-ms (approved):

PR #3794, opened 2023-01-09
Support deployment to Azure Container Apps with ACA extension

bwateratmsft:
Closes #3535, plus a bunch of tracking issues in ADO. Functionality E2E depends on https://github.com/microsoft/vscode-azurecontainerapps/pull/212 / https://github.com/microsoft/vscode-azurecontainerapps/issues/209.

/cc @MicroFish91

PR #3795, opened 2023-01-10
Update ACA deployment contract & use `registryPath` as `loginServer`

MicroFish91:
To merge into #3794 

@bwateratmsft Let me know if you think something like this could work, thanks!

### DIFF
--- a/src/commands/registries/azure/deployImageToAca.ts
+++ b/src/commands/registries/azure/deployImageToAca.ts
@@ -47,7 +47,6 @@ export async function deployImageToAca(context: IActionContext, node?: RemoteTag
     if (registry instanceof AzureRegistryTreeItem) {
         // No additional work to do; ACA can handle this on its own
     } else {
-        // Any other V2 compliant registry
         const { auth, registryPath } = await registry.getDockerCliCredentials() as { auth?: { username?: string, password?: string }, registryPath: string };
 
         if (!auth?.username || !auth?.password || !registryPath) {
--- a/src/commands/registries/azure/deployImageToAca.ts
+++ b/src/commands/registries/azure/deployImageToAca.ts
@@ -21,7 +21,7 @@ const minimumAcaExtensionVersion = '0.4.0'; // TODO: get the exact minimum versi
 
 // The interface of the command options passed to the Azure Container Apps extension's deployImageToAca command
 interface DeployImageToAcaOptionsContract {
-    imageName: string;
+    image: string;
     loginServer?: string;
     username?: string;
     secret?: string;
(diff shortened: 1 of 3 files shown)


src/tree/registries/dockerHub/DockerHubNamespaceTreeItem.ts
```diff
@@ -57,7 +59,7 @@ export class DockerHubNamespaceTreeItem extends RegistryTreeItemBase {
 
     public async getDockerCliCredentials(): Promise<IDockerCliCredentials> {
         return {
-            registryPath: '',
```
bwateratmsft, line 60: Can you verify that the Docker Login / Logout palette commands still work against Docker Hub? In the past we've never put the registry URL into the command line (the CLI assumes Docker Hub if no registry is given), but with this change, we will be.

src/commands/registries/azure/deployImageToAca.ts
```diff
@@ -38,27 +38,32 @@ export async function deployImageToAca(context: IActionContext, node?: RemoteTag
...
-    addImageTaggingTelemetry(context, commandOptions.imageName, '');
+    addImageTaggingTelemetry(context, commandOptions.image, '');
 
     const registry: RegistryTreeItemBase = node.parent.parent;
     if (registry instanceof AzureRegistryTreeItem) {
         // No additional work to do; ACA can handle this on its own
-    } else if (registry instanceof DockerHubNamespaceTreeItem || registry instanceof DockerV2RegistryTreeItemBase) {
-        const { auth } = await registry.getDockerCliCredentials() as { auth?: { username?: string, password?: string } };
+    } else {
+        // Any other V2 compliant registry
```
bwateratmsft, line 50: This comment is a _little_ misleading since GitLab is only sorta kinda is V2 compliant. Enough so that the Docker CLI can pull images, but that's about it.

src/tree/registries/dockerHub/DockerHubNamespaceTreeItem.ts
MicroFish91: There is one thing I noticed though, monitoring the config file, if we don't add `index.docker.io`, we get the first line being added: 

image

Would it be worth adding that whole url as the `registryPath` so they are consistent?

My url filter in ACA should be able to strip that back down to `index.docker.io` for the SDK so we'd still be good there as well.

bwateratmsft: I'm not sure if the CLI treats them differently. The best thing to do would be to put a private image in Docker Hub, sign in using the new URL, and see if it can be pulled normally.

MicroFish91: Okay tested it, it looks like it is very specific about the url it wants.  I've updated the registry url accordingly!

bwateratmsft (approved):

PR #3805, opened 2023-01-20
Add `registryName` to the ACA contract

MicroFish91:
My interpretation of what needed to change to support [this](https://github.com/microsoft/vscode-azurecontainerapps/pull/229#discussion_r1067151186).  Let me know if I was off or on the ball :)


bwateratmsft:
How do login server and registry differ? Are they different just for individual cases (e.g. Docker Hub) or more generally?

MicroFish91:
For **ACR**: From what I remember, `loginServer` parameter currently comes in as undefined for ACR stuff via contract; however, at its core the `registryName` should be no different than the `loginServer` here.

For **Docker Hub** it will come in as:
loginServer: index.docker.io
registryName: docker.io

**GitLabs**
They are the same.

bwateratmsft:
If Docker Hub is the only one that needs special casing of login vs. registry, I think it's best to just send registry, and have ACA break it into the two arguments to supply to the deployment

MicroFish91:
Are you opposed to keeping the contract how it was then?  I could just parse from the entire `imageName` but use stricter RegEx.

For ACR: `/\.azurecr\.io\//i`

For Docker Hub:  `/^docker\.io/i`?

PR #3809, opened 2023-01-24
Add diagnostic logging option to help with advanced troubleshooting

danegsta:


src/extension.ts
```diff
@@ -94,6 +104,29 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
...
+                ext.outputChannel.appendLine(`DOCKER_HOST: ${process.env.DOCKER_HOST}`);
+            } catch {
+                // Do not throw for diagnostic logging
+            }
+        }
+
+        // Additional diagnostic logging for Linux systems
+        if (ext.diagnosticLogging && isLinux()) {
+            try {
+                await execAsync('uname -a');
```
karolz-ms, line 118: Consider adding a small utility function that would let you write this code in one line, e.g.

```ts
noThrow(() => await execAsync('uname -a"));
```

karolz-ms (approved):
Looks great. Just one small suggestion for better code readability.

src/extension.ts
danegsta: Realized trying to implement this pattern as a method would be pretty tricky if not impossible due to typescript/eslint requirements for promise handling, so I'll leave as explicit try/catch blocks.

package.json
```diff
@@ -2244,6 +2244,12 @@
                     "type": "boolean",
                     "default": true,
                     "description": "%vscode-docker.config.docker.enableDockerComposeLanguageService%"
+                },
+                "docker.diagnosticLogging": {
```
bwateratmsft, line 2248: If we're actually intending to merge this, I would rely on VSCode's new log level setting instead, along with the log output channel.

bwateratmsft (changes requested):

src/runtimes/docker/utils/spawnStreamAsync.ts
```diff
@@ -229,7 +231,41 @@ export async function spawnStreamAsync(
...
+            }
+        });
+
+        if (stdErrPipe) {
+            diagnosticStdErrPipe.pipe(stdErrPipe);
+        }
+        stdErrPipe = diagnosticStdErrPipe;
     }
 
     const childProcess = spawn(
```
bwateratmsft, line 271: I think we should dump the entire environment as well, i.e. `{...process.env, ...options.env}`.

bwateratmsft: At a minimum, any `DOCKER_*` environment variables and `PATH`.

dnegstad: Logging the full environment (process and configured) makes sense; I’ll update what we output here. While theoretically a user could change the docker environment setting mid run, I’m leaning towards logging once at activation as a reasonable trade off between fidelity and noise in diagnostic logging.

src/commands/registries/logInToDockerCli.ts
```diff
@@ -30,7 +30,7 @@ export async function logInToDockerCli(context: IActionContext, node?: RegistryT
     }
 
     if (!username || !password) {
-        ext.outputChannel.appendLine(localize('vscode-docker.commands.registries.logIn.skipping', 'WARNING: Skipping login for "{0}" because it does not require authentication.', creds.registryPath));
+        ext.outputChannel.warn(localize('vscode-docker.commands.registries.logIn.skipping', 'WARNING: Skipping login for "{0}" because it does not require authentication.', creds.registryPath));
```
bwateratmsft, line 33: Nit: since VSCode prepends the log level to each line we can just take the "WARNING" out of the string
```suggestion
        ext.outputChannel.warn(localize('vscode-docker.commands.registries.logIn.skipping', 'Skipping login for "{0}" because it does not require authentication.', creds.registryPath));
```

package.json
danegsta: Updated to use the LogOutputChannel implementation

src/extension.ts
```diff
@@ -190,6 +223,14 @@ function registerDockerClients(): void {
         if (e.affectsConfiguration('docker.composeCommand')) {
             composeClient.reconfigure();
         }
+
+        if (ext.diagnosticLogging && isLinux()) {
+            try {
+                await execAsync(`which ${dockerClient.commandName}`);
```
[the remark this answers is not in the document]
danegsta: Where seems to be somewhat shell dependent in Windows, so it's going to be more difficult to get a useful result from it. I think it's worth revisiting, but not something I'd want to delay this PR on.

src/runtimes/docker/utils/spawnStreamAsync.ts
danegsta: Updated to log environment at debug level.

src/commands/registries/logInToDockerCli.ts
danegsta: Removed the redundant label

src/runtimes/clients/AutoConfigurableDockerClient.ts
```diff
@@ -17,5 +20,15 @@ export class AutoConfigurableDockerClient extends DockerClient implements AutoCo
         const config = vscode.workspace.getConfiguration('docker');
         const dockerCommand = config.get<string | undefined>('dockerPath') || 'docker';
         this.commandName = dockerCommand;
+
+        if (ext.outputChannel.debugLoggingEnabled && isLinux()) {
```
bwateratmsft, line 24: Can this go in with the other debug-only code?

src/extension.ts
```diff
@@ -85,6 +89,18 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
             )
         );
 
+        if (ext.outputChannel.debugLoggingEnabled) {
```
bwateratmsft, line 92: This has me wondering. You mentioned VSCode doesn't really expose anything about this channel until it is registered. If the log level is changed _after_ activation, all of this code will be missed. I'm fine with that if the log level change persists to the next session, but if not then none of this code could ever run.

danegsta: There is a hook to detect log level changes, but it doesn't offer any way of knowing what the previous log level was. We could potentially add a check to see if we've logged these diagnostics before and only show them once, but I think for the handful of startup properties it's reasonable to ask users to enable logging then restart VS Code in order to capture the full debug context.

bwateratmsft: Right. I'm totally fine requiring a restart to get the full output. I just don't know if VSCode "forgets" inbetween sessions that an extension has had its logging level bumped up.

src/runtimes/docker/utils/spawnStreamAsync.ts
```diff
@@ -229,7 +231,39 @@ export async function spawnStreamAsync(
     }
 
     if (options.onCommand) {
-        options.onCommand([command, ...args].join(' '));
+        options.onCommand([command, ...normalizedArgs].join(' '));
+    }
+
+    ext.outputChannel.debug([command, ...normalizedArgs].join(' '));
+
+    let stdOutPipe: NodeJS.WritableStream | undefined = options.stdOutPipe;
```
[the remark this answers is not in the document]
danegsta: I don't know that it makes sense to do a diagnostics only stream runner, since things like error messages we potentially want to log (as an error) no matter what. Some of the diagnostics can be split out into methods to keep the main code cleaner, but I think it makes sense to have debug/diagnostics somewhat inline with the code being logged. My main reason for doing this in spawnStreamAsync was that it's the lowest level in how we spawn processes, since we use it in the command runners as well as execAsync. Doing the logging here ensures that it applies to all the commands we run.

src/utils/AzExtLogOutputChannelWrapper.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    public get logLevel() {
+        return this._logOutputChannel.logLevel;
+    }
+
+    public get debugLoggingEnabled() {
+        return this.logLevel > 0 && this.logLevel <= vscode.LogLevel.Debug;
+    }
+
+    private;
```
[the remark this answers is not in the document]
danegsta: ‽

src/extension.ts
```diff
@@ -94,6 +110,21 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
             ext.orchestratorManager = new OrchestratorRuntimeManager()
         );
 
+        // Additional diagnostic logging for Linux systems
+        if (ext.outputChannel.debugLoggingEnabled && isLinux()) {
```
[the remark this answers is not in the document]
danegsta: I'll add a utility entry for some of these general diagnostic calls (as well as making sure we don't block on these background diagnostic calls)

src/runtimes/clients/AutoConfigurableDockerClient.ts
danegsta: Refactored this to use a method from the new diagnostics utility functions

src/runtimes/docker/utils/spawnStreamAsync.ts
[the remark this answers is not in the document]
bwateratmsft: Having it here is going to be problematic for refactoring to use this library as a package, as well.

src/utils/diagnostics.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function logSystemInfo(outputChannel: vscode.LogOutputChannel): void {
+    logEnvironment(outputChannel, process.env, '--- Process Environment ---');
+
+    if (isLogLevelEnabled(outputChannel, vscode.LogLevel.Debug)) {
+        // Linux specific system diagnostic logging
+        if (isLinux()) {
+            outputChannel.debug('--- System Info ---');
+            try {
+                execAsync('uname -a').catch(() => {/* Do not throw */ });
```
bwateratmsft, line 38: Wait, does executing like this actually send output to the output channel?

src/runtimes/docker/utils/spawnStreamAsync.ts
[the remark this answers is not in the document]
danegsta: Very good point.

src/extension.ts
```diff
@@ -150,6 +155,8 @@ function registerEnvironmentVariableContributions(): void {
         actionContext.telemetry.suppressAll = true;
         actionContext.errorHandling.suppressDisplay = true;
 
+        logDockerEnvironment(ext.outputChannel);
```
bwateratmsft, line 158: I'd move this line down into the `if` below, since this event gets called for _all_ settings changes. It could become quite noisy.

src/extension.ts
```diff
@@ -179,7 +186,7 @@ function registerDockerClients(): void {
     );
 
     // Register an event to watch for changes to config, reconfigure if needed
-    registerEvent('docker.command.changed', vscode.workspace.onDidChangeConfiguration, (actionContext: IActionContext, e: vscode.ConfigurationChangeEvent) => {
+    registerEvent('docker.command.changed', vscode.workspace.onDidChangeConfiguration, async (actionContext: IActionContext, e: vscode.ConfigurationChangeEvent) => {
```
bwateratmsft, line 189: Is this line change needed?

src/utils/diagnostics.ts
```diff
@@ -11,14 +11,12 @@ export function isLogLevelEnabled(outputChannel: vscode.LogOutputChannel, logLev
...
+export function logProcessEnvironment(outputChannel: vscode.LogOutputChannel): void {
     if (isLogLevelEnabled(outputChannel, vscode.LogLevel.Debug)) {
         try {
-            if (title) {
-                outputChannel.debug(title);
-            }
+            outputChannel.debug(`--- Process Environment (${Object.getOwnPropertyNames(process.env).length}) ---`);
 
-            for (const key in environment) {
+            for (const key in process.env) {
```
bwateratmsft, line 19: I think we'll want to do this, instead of `process.env`, since this is what gets used in VSCode terminals automatically, and in processes we launch thanks to `DefaultEnvStreamCommandRunnerFactory`:
`withDockerEnvSettings(process.env)`

bwateratmsft: Also I think it needs to be `Object.keys(theThing)`

danegsta: I've broken the logging up by process environment and Docker environment from settings in order to clearly see where environment settings are coming from (see logProcessEnvironment and logDockerEnvironment) with the Docker environment being re-emitted if the change event fires.

I'll update to `for of Object.keys` to be safe and only echo own keys for the environment objects.

src/extension.ts
danegsta: Nope, I'll remove that async.

src/utils/diagnostics.ts
danegsta: I've updated execAsync to use a logging enabled wrapper around spawnStreamAsync; if debug logs aren't enabled it's basically the same as spawnStreamAsync, but I wrap these calls in a check to ensure we'll actually output debug logs before running them to avoid any potential slowdowns if debug logging isn't turned on.

bwateratmsft (approved):
Don't forget to copy the changes upstream!

PR #3826, opened 2023-02-07
Clean up activation events

### LINKED ISSUE #3755: Clean up activation events
Enabled by https://code.visualstudio.com/updates/v1_74#_implicit-activation-events-for-declared-extension-contributions, we can remove most of our activation events. Will require VSCode 1.74.0+.

bwateratmsft:
VSCode has gotten quite pushy about cleaning up activation events--showing warnings in the workspace even when `package.json` isn't open--so I'm just going to fix this now. Fixes #3755.


package.json
```diff
@@ -29,110 +29,13 @@
...
-        "onCommand:vscode-docker.contexts.configureExplorer",
-        "onCommand:vscode-docker.contexts.refresh",
-        "onCommand:vscode-docker.contexts.help",
         "onTaskType:docker-build",
         "onTaskType:docker-run",
         "onTaskType:docker-compose",
         "onDebugInitialConfigurations",
         "onDebugResolve:docker",
         "onFileSystem:docker",
-        "onLanguage:dockerfile",
```
bwateratmsft, line 127: I verified that activation still occurs when a Dockerfile is opened. The activation event for compose is not removed, but the comma at the end of the line has to be removed so it shows up as an added line later.

karolz-ms (approved):

PR #3845, opened 2023-02-24
Changes to Scaffolding for .NET 8 Preview #3834

alexyaang:
This PR addresses issue #3834 . 

This PR supports the change in naming convention for .NET 8 Docker tags & the change in `adduser` command for Debian 12.

### DIFF
--- a/resources/templates/netCore/Dockerfile.template
+++ b/resources/templates/netCore/Dockerfile.template
@@ -12,11 +12,15 @@ ENV ASPNETCORE_URLS=http://+:{{ ports.[0] }}
 {{/if}}
 {{#unless (isRootPort ports)}}
 {{#if (eq netCorePlatformOS 'Linux')}}
+{{#if netCoreBaseImageDefaultUser }}
+USER {{ netCoreBaseImageDefaultUser }}
+{{else}}
 # Creates a non-root user with an explicit UID and adds permission to access the /app folder
 # For more info, please refer to https://aka.ms/vscode-docker-dotnet-configure-containers
 RUN adduser -u 5678 --disabled-password --gecos "" appuser && chown -R appuser /app
 USER appuser
 
+{{/if}}
 {{/if}}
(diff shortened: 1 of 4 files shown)


resources/templates/netCore/Dockerfile.template
```diff
@@ -12,11 +12,15 @@ ENV ASPNETCORE_URLS=http://+:{{ ports.[0] }}
 {{/if}}
 {{#unless (isRootPort ports)}}
 {{#if (eq netCorePlatformOS 'Linux')}}
+{{#if netCoreBaseImageDefaultUser }}
+USER {{ netCoreBaseImageDefaultUser }}
+{{else}}
 # Creates a non-root user with an explicit UID and adds permission to access the /app folder
 # For more info, please refer to https://aka.ms/vscode-docker-dotnet-configure-containers
 RUN adduser -u 5678 --disabled-password --gecos "" appuser && chown -R appuser /app
```
bwateratmsft, line 20: Does `adduser` still work on newer Debian but .NET 7 (and below)?

bwateratmsft (approved):

resources/templates/netCore/Dockerfile.template
alexyaang: No, `adduser` will not work on Debian with .NET 7 or below. This is the error I get from running the `adduser` command from .NET 6 scaffolding with a Debian 12 image: `adduser: not found`

But in this case, since all .NET 7 images are based on Debian 11 and below, the case you described theoretically wouldn't happen.

bwateratmsft: That's what I'm concerned about--will .NET 7 images get updated to Debian 12 at some point?

alexyaang: I agree, it's definitely concerning. We'll have to reach out to the Docker image team to get a sure answer.

PR #3846, opened 2023-02-28
Refactored localize to use new l10n API

alexyaang:
This PR addresses issue #3693 

The changes replace `vscode-nls` with the vscode localization (`l10n`) API as introduced in https://code.visualstudio.com/updates/v1_72#_localization-as-part-of-the-api.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2891,7 +2891,6 @@
         "tar": "^6.1.11",
         "tree-kill": "^1.2.2",
         "vscode-languageclient": "^8.0.2",
-        "vscode-nls": "^5.0.0",
         "vscode-tas-client": "^0.1.22",
         "xml2js": "^0.4.23"
     }
--- a/package.json
+++ b/package.json
@@ -2809,7 +2809,7 @@
         ]
     },
     "engines": {
-        "vscode": "^1.74.0"
+        "vscode": "^1.75.0"
     },
     "capabilities": {
         "virtualWorkspaces": false,
@@ -2854,7 +2854,7 @@
         "@types/node": "16.x",
         "@types/node-fetch": "^2.5.12",
         "@types/semver": "^7.3.9",
-        "@types/vscode": "1.74.0",
+        "@types/vscode": "1.75.0",
         "@types/xml2js": "^0.4.8",
         "@typescript-eslint/eslint-plugin": "^5.19.0",
         "@typescript-eslint/parser": "^5.19.0",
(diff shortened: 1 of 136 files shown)


src/commands/registries/azure/deleteAzureRegistry.ts
```diff
@@ -6,7 +6,7 @@
 import { DialogResponses, IActionContext } from '@microsoft/vscode-azext-utils';
 import { ProgressLocation, window } from 'vscode';
 import { ext } from '../../../extensionVariables';
-import { localize } from "../../../localize";
+import { l10n } from 'vscode';
```
danegsta, line 9: There's an existing `vscode` import on line 7; when adding an import, double check that we're not already importing from the module. If there's an existing import, update it to add any additional exports you need to import.

danegsta (changes requested):
The PR looks good for switching to the new API; only changes would be cleaning up the imports to avoid importing `vscode` multiple times (and to try to keep the import sorting consistent for easier readability).

src/commands/compose/compose.ts
```diff
@@ -6,7 +6,7 @@
 import { IActionContext, UserCancelledError } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
 import { ext } from '../../extensionVariables';
-import { localize } from "../../localize";
+import { l10n } from 'vscode';
```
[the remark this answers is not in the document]
alexyaang: That makes sense, I'll make the change for all similar files.

src/commands/registries/azure/deleteAzureRegistry.ts
alexyaang: Just pushed the changes to reduce redundant imports, organize imports on files I touched. Let me know if there's anything else I should pay attention to!

package.json
```diff
@@ -2891,7 +2891,6 @@
         "tar": "^6.1.11",
         "tree-kill": "^1.2.2",
         "vscode-languageclient": "^8.0.2",
-        "vscode-nls": "^5.0.0",
```
bwateratmsft, line 2894: I can't put the comment in the right location, but we need to update `@types/vscode` to 1.75.0, and also update the VSCode engine dependency to match.

alexyaang: I see. Good catch!

bwateratmsft (approved):

danegsta (approved):

PR #3852, opened 2023-03-07
Yangalex/docker swarm

alexyaang:


package.nls.json
```diff
@@ -153,6 +153,7 @@
     "vscode-docker.config.template.composeDown.match": "The regular expression for choosing the right template. Checked against docker-compose YAML files, folder name, etc.",
     "vscode-docker.config.template.composeDown.description": "Command templates for `docker-compose down` commands.",
     "vscode-docker.config.docker.containers.groupBy": "The property to use to group containers in Docker view: ContainerId, ContainerName, CreatedTime, FullTag, ImageId, Networks, Ports, Registry, Repository, RepositoryName, RepositoryNameAndTag, State, Status, Tag, or None",
+    "vscode-docker.config.docker.containers.groupByLabel": "The property to use for grouping containers by label in Docker Containers view",
```
alexyaang, line 156: Is this description good enough?

bwateratmsft: How about this?

```suggestion
    "vscode-docker.config.docker.containers.groupByLabel": "The items will be grouped by the value of this Docker label",
```

src/tree/containers/ContainerGroupTreeItem.ts
```diff
@@ -46,6 +46,8 @@ export class ContainerGroupTreeItem extends LocalGroupTreeItemBase<DockerContain
                 return getContainerStateIcon(this.group);
             case 'Image':
                 return new ThemeIcon('multiple-windows');
+            case 'Label':
```
bwateratmsft, line 49: This is valid, but an equivalent that's also valid and a little shorter is:

```typescript
            case 'Image':
            case 'Label':
                return new ThemeIcon('multiple-windows');
```

You can keep it as is or change it--it's up to you--just wanted to make alternatives known

src/tree/containers/ContainerProperties.ts
```diff
@@ -20,6 +20,7 @@ export const containerProperties: ITreePropertyInfo<ContainerProperty>[] = [
     { property: 'State', exampleValue: 'exited' },
     { property: 'Status', exampleValue: 'Exited (0) 2 hours ago' },
     { property: 'Compose Project Name', description: l10n.t('Value used to associate containers launched by a \'docker-compose up\' command') },
+    { property: 'Label', exampleValue: 'com.microsoft.created-by' },
```
bwateratmsft, line 23: I'd include the label value so it's clear that we're grouping by label _and_ value
```suggestion
    { property: 'Label', exampleValue: 'com.microsoft.created-by=visual-studio-code' },
```

src/tree/settings/CommonProperties.ts
```diff
@@ -52,3 +53,28 @@ export function getCommonGroupIcon(property: CommonProperty | CommonGroupBy): Th
 
     return new ThemeIcon(icon);
 }
+
+export const NonLabelGroupName = l10n.t('others');
```
bwateratmsft, line 57:
```suggestion
export const NonLabelGroupName = l10n.t('Others');
```
Or maybe "Unlabeled containers"? Though that's not quite right, since they could have other labels.

src/tree/containers/ContainersTreeItem.ts
```diff
@@ -82,6 +82,15 @@ export class ContainersTreeItem extends LocalRootTreeItemBase<DockerContainerInf
                 return ti1.label === NonComposeGroupName ? 1 : -1;
             }
         }
+        if (this.groupBySetting === 'Label'
```
bwateratmsft, line 85: I'd turn this into an `else if (...)` grouped with the `if` above, and maybe put in an extra newline before and after the block that should've always been there.

src/tree/containers/ContainerProperties.ts
```diff
@@ -87,3 +90,5 @@ export function getComposeProjectName(container: ListContainersItem): string {
         return NonComposeGroupName;
     }
 }
+
```
bwateratmsft, line 93: Nit: some extra newlines

src/tree/settings/CommonProperties.ts
```diff
@@ -52,3 +53,28 @@ export function getCommonGroupIcon(property: CommonProperty | CommonGroupBy): Th
...
+    const labels = Object.keys(containerLabels).map(label => ({ label: label, value: containerLabels[label] }));
+
+    const composeProject = labels.find(l => l.label === labelName);
+    if (composeProject) {
+        return composeProject.value;
+    } else {
+        return NonLabelGroupName;
+    }
+}
+
```
bwateratmsft, line 78: Nit: some extra newlines

src/tree/containers/ContainerProperties.ts
```diff
@@ -65,6 +66,8 @@ export function getContainerPropertyValue(item: ListContainersItem, property: Co
             return getComposeProjectName(item);
         case 'Image':
             return item.image.originalName;
+        case 'Label':
+            return getLabel(item?.labels, 'containers');
```
bwateratmsft, line 70: A couple of changes I'd suggest to this method--

1. Change the name from `getLabel` to `getConfiguredLabelGroup`, since it's not taking the label to get as an input, but rather looking it up from settings
2. Pass in just `item` and in the method body, go into `item.labels`

src/tree/settings/CommonProperties.ts
```diff
@@ -52,3 +53,28 @@ export function getCommonGroupIcon(property: CommonProperty | CommonGroupBy): Th
...
+    if (!containerLabels) {
+        return NonLabelGroupName;
+    }
+
+    const labelName = workspace.getConfiguration('docker')?.get<string | undefined>(`${itemType}.groupByLabel`, undefined);
+    if (!labelName) {
+        return NonLabelGroupName;
+    }
+
+    const labels = Object.keys(containerLabels).map(label => ({ label: label, value: containerLabels[label] }));
```
bwateratmsft, line 69: JS lets you directly index map objects, since they're ultimately just regular old objects. So you can shorten this block + the rest of the method body--

```typescript
return containerLabels?.[labelName] || NonLabelGroupName;
```
Essentially what that line means is, from the `containerLabels` object, if it has a property with name matching `labelName`'s value, then it will return that--or, if it doesn't, or if the value is null, undefined, or falsy, then it will return `NonLabelGroupName`'s value instead.

src/tree/containers/ContainersTreeItem.ts
bwateratmsft: Semantically it's the same, but looks a little neater IMO.

package.nls.json
alexyaang: Much better :)

src/tree/containers/ContainerGroupTreeItem.ts
alexyaang: Simpler, I like it.

src/tree/containers/ContainerProperties.ts
alexyaang: Wouldn't this encourage users to put `com.microsoft.created-by=visual-studio-code` instead of `com.microsoft.created-by` in the settings? which will cause the Group by Label feature to not work?

src/tree/settings/CommonProperties.ts
alexyaang: Yeah... either way it's tricky. I'll go with "Others" for now

src/tree/containers/ContainerProperties.ts
alexyaang: oops

src/tree/containers/ContainerProperties.ts
alexyaang: That makes sense. Makes the code more readable for sure.

src/tree/settings/CommonProperties.ts
alexyaang: That simplifies the method by a lot, thanks Brandon :)

src/tree/containers/ContainerProperties.ts
bwateratmsft: That's fair. The others are an example end value though. Maybe we could put the example setting value in the description for the setting?

package.nls.json
```diff
@@ -153,6 +153,7 @@
     "vscode-docker.config.template.composeDown.match": "The regular expression for choosing the right template. Checked against docker-compose YAML files, folder name, etc.",
     "vscode-docker.config.template.composeDown.description": "Command templates for `docker-compose down` commands.",
     "vscode-docker.config.docker.containers.groupBy": "The property to use to group containers in Docker view: ContainerId, ContainerName, CreatedTime, FullTag, ImageId, Networks, Ports, Registry, Repository, RepositoryName, RepositoryNameAndTag, State, Status, Tag, or None",
+    "vscode-docker.config.docker.containers.groupByLabel": "The items will be grouped by the value of this Docker label",
```
bwateratmsft, line 156:
```suggestion
    "vscode-docker.config.docker.containers.groupByLabel": "The items will be grouped by the value of this container label",
```
More generic and equally or more accurate, IMO.

src/tree/settings/CommonProperties.ts
bwateratmsft: I think these lines need to stay, before that return:
```typescript
    if (!labelName) {
        return NonLabelGroupName;
    }
```

bwateratmsft: (_Technically_ they don't, as `item?.[undefined]` evaluates to `undefined` which would be falsy, so falling into that `NonLabelGroupName`, but that's a sort of esoteric behavior, so keeping these lines is better for clarity)

src/tree/containers/ContainerProperties.ts
alexyaang: That's a great idea actually. I'll do that

package.nls.json
alexyaang: Agreed

src/tree/settings/CommonProperties.ts
alexyaang: I agree, that is a bit confusing. I didn't think about that at all.

bwateratmsft (approved):

PR #3857, opened 2023-03-09
Don't show compose start/stop when not in compose group mode

### LINKED ISSUE #3495: The error message is unclear after executing the command "Docker Containers: Compose Stop/Start/Logs" from command palette
**OS:** All
**Build Version:** 20220412.2
**Regression:** Not a Regression

**Repro Steps:**

1. Right click a docker.compose file -> Compose Up -> There is a running container in the Container view.
2. Click the "Configure Explorer…" button on the Containers view -> Group By -> Select a property (don't select "Compose Project Name").
3. F1 -> Docker Containers: Compose Stop/Start/Logs.
4. Check whether the error message is clear.

**Expect:**
The error message is clear.

**Actual:**
The error message is unclear. 

> "No Docker Compose projects are running" looks like the “Compose Up” command is not executed, there is no running container in the Container view.
> image

bwateratmsft:
Fixes #3495

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,14 @@
                     "when": "config.docker.containers.groupBy == 'Compose Project Name'"
                 },
                 {
+                    "command": "vscode-docker.containers.composeGroup.start",
+                    "when": "config.docker.containers.groupBy == 'Compose Project Name'"
+                },
+                {
+                    "command": "vscode-docker.containers.composeGroup.stop",
+                    "when": "config.docker.containers.groupBy == 'Compose Project Name'"
+                },
+                {
                     "command": "vscode-docker.containers.composeGroup.restart",
                     "when": "never"
                 },


package.json
```diff
@@ -65,6 +65,14 @@
...
+                {
+                    "command": "vscode-docker.containers.composeGroup.start",
+                    "when": "config.docker.containers.groupBy == 'Compose Project Name'"
+                },
+                {
+                    "command": "vscode-docker.containers.composeGroup.stop",
+                    "when": "config.docker.containers.groupBy == 'Compose Project Name'"
+                },
                 {
                     "command": "vscode-docker.containers.composeGroup.restart",
```
bwateratmsft, line 77: Restart and down overlap with commands that are executed on compose files--not tree items--so we'll continue to never show those in the command palette.

karolz-ms (approved):

Issue #3492, opened 2022-04-12
Add ability to remove image and container group


bwateratmsft:
For images:

1. Ability to remove an entire image group
2. Removal will still constitute the current behavior of untagging, which may remove the image altogether or just the tag

For containers:
1. If the container group is a compose group, no removal option should be shown

PR #3867, opened 2023-03-15
Yangalex/removeGroup

alexyaang:
This PR addresses issue #3492 by adding the ability to remove an entire container or image group (excluding container compose groups)


~/.vscode/extensions/theme-dracula
bwateratmsft, line 1: What's this?

package.json
```diff
@@ -436,6 +436,11 @@
                     "when": "view == dockerImages && viewItem == image",
                     "group": "images_3_destructive@1"
                 },
+                {
+                    "command": "vscode-docker.images.group.remove",
+                    "when": "view == dockerImages && viewItem =~ /imageGroup$/i",
```
bwateratmsft, line 441: Thinking about it, if the user sees both "Image Group: Remove Image..." and "Image: Remove Image..." in the command palette, confusion is likely to result. I'd suggest a couple things--

1. Don't show this command in the command palette--i.e., in the `menus.commandPalette` part of this file, add one with `"when": "never"` (there are many other examples)
2. Use the same `"category"` property as the other image commands (i.e. `"%vscode-docker.commands.category.dockerImages%"`)
3. Remove the newly-added category string from `package.nls.json`

We did a similar thing when doing commands for compose down--which is a different command if you run it on a file vs. run it on a tree item.

src/commands/images/removeImageGroup.ts
```diff
@@ -0,0 +1,30 @@
...
+        ext.imagesTree,
+        ImageGroupTreeItem.contextValue,
+        node,
+        nodes
+    );
+
+    const confirmRemove = vscode.l10n.t('Are you sure you want to remove the selected image groups? If there are other tags or child images for images within this group, only the tags will be removed.');
+
+    // no need to check result - cancel will throw a UserCancelledError
+    await context.ui.showWarningMessage (confirmRemove, { modal: true }, { title: 'Remove' });
```
bwateratmsft, line 24:
```suggestion
    await context.ui.showWarningMessage (confirmRemove, { modal: true }, { title: vscode.l10n.t('Remove') });
```

I see the same mistake was already made in `removeImage.ts`, can you fix it there as well?

bwateratmsft (commented):
Implementation looks good so far. I think more-or-less the same pattern applies for container groups as well. As discussed, we'll do just containers and images for now, since the need and value is probably far lower for volumes and networks.

~/.vscode/extensions/theme-dracula
alexyaang: Great catch! I downloaded a new theme for vscode called Dracula and it somehow showed up here...

src/commands/images/removeImageGroup.ts
alexyaang: Absolutely!

package.json
```diff
@@ -73,6 +73,10 @@
                     "command": "vscode-docker.containers.composeGroup.stop",
                     "when": "config.docker.containers.groupBy == 'Compose Project Name'"
                 },
+                {
```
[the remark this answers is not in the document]
alexyaang: liek right above `vscode-docker.debugging.initializeForDebugging` ?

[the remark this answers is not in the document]
bwateratmsft: Right above this bit that's added at line 140:
```json
                {
                    "command": "vscode-docker.images.group.remove",
                    "when": "never"
                }
```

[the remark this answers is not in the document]
alexyaang: ohh yup that makes sense

bwateratmsft (approved):

PR #3879, opened 2023-03-23
Yangalex/SBOM

alexyaang:


package.nls.json
```diff
@@ -242,6 +242,7 @@
     "vscode-docker.commands.images.runInteractive": "Run Interactive",
     "vscode-docker.commands.images.tag": "Tag...",
     "vscode-docker.commands.images.copyFullTag": "Copy Full Tag",
+    "vscode-docker.commands.images.sbom": "Generate Sbom",
```
karolz-ms, line 245:
```suggestion
    "vscode-docker.commands.images.sbom": "Generate Software Bill of Materials (SBOM)",
```

src/commands/images/generateSbom.ts
```diff
@@ -0,0 +1,29 @@
...
+import { ext } from '../../extensionVariables';
+import { ImageSbomItem } from '../../runtimes/docker';
+import { ImageTreeItem } from '../../tree/images/ImageTreeItem';
+
+export async function generateSbom(context: IActionContext, node?: ImageTreeItem): Promise<void> {
+    if (!node) {
+        await ext.imagesTree.refresh(context);
+        node = await ext.imagesTree.showTreeItemPicker<ImageTreeItem>(ImageTreeItem.contextValue, {
+            ...context,
+            noItemFoundErrorMessage: vscode.l10n.t('No images are available for generating sbom')
```
karolz-ms, line 17:
```suggestion
            noItemFoundErrorMessage: vscode.l10n.t('No images are available for generating Software Bill of Materials')
```

src/commands/images/generateSbom.ts
```diff
@@ -0,0 +1,29 @@
...
+        });
+    }
+
+    const generating: string = vscode.l10n.t('Generating sbom...');
```
karolz-ms, line 21:
```suggestion
    const generating: string = vscode.l10n.t('Generating Software Bill of Materials...');
```

karolz-ms: In general, you should say "Software Bill of Materials" to the user, unless the space is really tight. In the latter case use "SBOM" (all caps).

src/extension.ts
```diff
@@ -100,7 +101,7 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats:
         );
 
         // Set up Docker clients
-        registerDockerClients();
+        await registerDockerClients();
```
karolz-ms, line 104: Do we really need to slow down extension activation here? Why?

src/runtimes/docker/clients/DockerClientBase/DockerClientBase.ts
```diff
@@ -651,6 +654,42 @@ export abstract class DockerClientBase extends ConfigurableClient implements ICo
...
+        const sboms = new Array<ImageSbomItem>();
+        sboms.push({raw: output});
+        return sboms;
+    }
+
+    async imageGenerateSbom(options: ImageGenerateSbomCommandOptions): Promise<PromiseCommandResponse<Array<ImageSbomItem>>> {
+        return {
+            command: this.commandName,
+            args: this.getGenerateSbomCommandArgs(options),
+            parse: (output, strict) => this.parseImageGenerateSbomCommandOutput(options, output, strict),
```
karolz-ms, line 687: Consider creating the return Array in place and eliminating the `parseImageGenerateSbomCommandOutput`

karolz-ms (commented):
Thanks for taking this up, Alex. Looks generally good, but I am not sure about the new delay introduced during the extension activation part...

src/runtimes/docker/clients/DockerClientBase/DockerInfoRecord.ts
```diff
@@ -5,17 +5,30 @@
...
 
     if (typeof info.OSType === 'string' && !isContainerOS(info.OSType)) {
         return false;
     }
-
-    return true;
+    else if (!info.ClientInfo) {
+        return false;
+    }
+    else {
```
danegsta, line 31: The return true shouldn't need to be wrapped in an else statement here.

src/runtimes/docker/clients/DockerClientBase/DockerInfoRecord.ts
```diff
@@ -5,17 +5,30 @@
...
 export function isDockerInfoRecord(maybeInfo: unknown): maybeInfo is DockerInfoRecord {
+
     const info = maybeInfo as DockerInfoRecord;
 
     if (typeof info.OSType === 'string' && !isContainerOS(info.OSType)) {
         return false;
     }
-
-    return true;
+    else if (!info.ClientInfo) {
```
danegsta, line 28: `if (info.ClientInfo && typeof info.ClientInfo !== 'object')` otherwise ClientInfo won't actually be optional.

danegsta: Look at DockerInspectImageRecord.ts for an example of the pattern we've been using for complex record validation.

src/runtimes/docker/contracts/ContainerClient.ts
```diff
@@ -609,6 +614,30 @@ type InspectImagesCommand = {
     inspectImages(options: InspectImagesCommandOptions): Promise<PromiseCommandResponse<Array<InspectImagesItem>>>;
 };
 
+export type ImageSbomItem = {
+    /**
+     * The RAW inspect output
```
danegsta, line 619: Unless we plan on trying to parse the SBOM output, we should probably just return a string rather than an ImageSbomItem.

src/runtimes/docker/contracts/ContainerClient.ts
```diff
@@ -609,6 +614,30 @@ type InspectImagesCommand = {
...
+     */
+    imageRef: string;
+};
+
+type ImageGenerateSbomCommand = {
+    /**
+     * Generate a CommandResponse for pruning images
+     * @param options Command options
+     */
+    imageGenerateSbom(options: ImageGenerateSbomCommandOptions): Promise<PromiseCommandResponse<Array<ImageSbomItem>>>;
```
danegsta, line 638: It doesn't make sense for this to return an array unless the `docker sbom` command can take multiple images; from what I can see, it only accepts a single image tag. This should probably be Promise<PromiseCommandResponse<string>> for the command return type.

src/runtimes/docker/contracts/ContainerClient.ts
```diff
@@ -609,6 +614,30 @@ type InspectImagesCommand = {
...
+ * Options for inspecting images
+ */
+export type ImageGenerateSbomCommandOptions = CommonCommandOptions & {
+    /**
+     * The image names/IDs/etc. to inspect, passed directly to the CLI
+     */
+    imageRef: string;
+};
+
+type ImageGenerateSbomCommand = {
```
danegsta, line 633: SbomForImageCommand to stay consistent with the naming convention for the rest of the commands.

src/runtimes/docker/contracts/ContainerClient.ts
```diff
@@ -609,6 +614,30 @@ type InspectImagesCommand = {
...
+export type ImageSbomItem = {
+    /**
+     * The RAW inspect output
+     */
+    raw: string;
+};
+/**
+ * Options for inspecting images
+ */
+export type ImageGenerateSbomCommandOptions = CommonCommandOptions & {
```
danegsta, line 626: SbomForImageCommandOptions

src/runtimes/docker/contracts/ContainerClient.ts
danegsta, line 638: sbomForImage

src/runtimes/docker/contracts/ContainerClient.ts
```diff
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { FileType, ShellQuotedString } from 'vscode';
+import { DockerClientInfo } from '../clients/DockerClientBase/DockerInfoRecord';
```
danegsta, line 7: Don't import directly from the client types; any types referenced in ContainerClient should be defined in this file.

src/runtimes/docker/contracts/ContainerClient.ts
```diff
@@ -145,6 +146,10 @@ export type InfoItem = {
      * The OS type for the container runtime
      */
     osType?: ContainerOS;
+    /**
+     * Docker Client info record
+     */
+    clientInfo?: DockerClientInfo;
```
danegsta, line 152: Define a ClientInfo type in this file rather than importing the type from DockerInfoRecord.

@bwateratmsft we should probably add a work item to implement a "feature support" check like we've discussed in the past. It would let us cleanup this and several other cases where we have to leak Docker implementation details into the extension UI logic.

src/runtimes/docker/clients/DockerClientBase/DockerClientBase.ts
```diff
@@ -651,6 +654,42 @@ export abstract class DockerClientBase extends ConfigurableClient implements ICo
 
     //#endregion
 
+    //#region ImagesGenerateSbom Command
+
+    /**
+     * Generate the command line arguments for invoking a pull image command on
+     * a Docker-like client
```
danegsta, line 661: Update the pull image references in the comments

src/runtimes/docker/clients/DockerClientBase/DockerClientBase.ts
```diff
@@ -651,6 +654,42 @@ export abstract class DockerClientBase extends ConfigurableClient implements ICo
...
+    /**
+     * Generate the command line arguments for invoking a pull image command on
+     * a Docker-like client
+     * @param options Pull image command options
+     * @returns Command line arguments for pulling a container image
+     */
+    protected getGenerateSbomCommandArgs(options: ImageGenerateSbomCommandOptions): CommandLineArgs {
+        return composeArgs(
+            withArg('sbom'),
+            withNamedArg('--format', 'spdx-json', {shouldQuote: false}),
```
danegsta, line 668: We should probably let the user pick the appropriate format rather than hard coding spdx-json; at the very least we should add a format property to the command options and allow setting a specific output type as a vscode setting.

src/runtimes/docker/clients/DockerClientBase/DockerClientBase.ts
```diff
@@ -651,6 +654,42 @@ export abstract class DockerClientBase extends ConfigurableClient implements ICo
...
+    //#region ImagesGenerateSbom Command
+
+    /**
+     * Generate the command line arguments for invoking a pull image command on
+     * a Docker-like client
+     * @param options Pull image command options
+     * @returns Command line arguments for pulling a container image
+     */
+    protected getGenerateSbomCommandArgs(options: ImageGenerateSbomCommandOptions): CommandLineArgs {
+        return composeArgs(
```
danegsta, line 666: We should consider whether we want to support the --exclude, --platform, and/or --version arguments.

danegsta: Additionally, there's a --quiet argument that we should look into

package.json
```diff
@@ -444,6 +444,11 @@
                     "when": "view == dockerImages && viewItem == image",
                     "group": "images_2_general@5"
                 },
+                {
+                    "command": "vscode-docker.images.sbom",
+                    "when": "view == dockerImages && viewItem == image && vscode-docker:hasSbom",
```
danegsta, line 449: `vscode-docker:supportsSbom`?

src/commands/images/generateSbom.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    let sbomResult: ImageSbomItem[];
+    await vscode.window.withProgress( { location: vscode.ProgressLocation.Notification, title: generating }, async () => {
+        sbomResult = await ext.runWithDefaults(client => client.imageGenerateSbom({ imageRef: node.imageId }));
+    });
+
+    await openReadOnlyJson(node, JSON.parse(sbomResult[0].raw));
```
danegsta, line 28: We can't necessarily assume the SBOM will always be json

src/extension.ts
```diff
@@ -352,3 +358,22 @@ function activateComposeLanguageClient(ctx: vscode.ExtensionContext): void {
 }
 
 //#endregion Language services
+
+const hasSbomContextKey = 'vscode-docker:hasSbom';
+async function checkDockerVersionForSbom(): Promise<void> {
```
danegsta, line 363: `checkSbomSupport`?

src/extension.ts
```diff
@@ -352,3 +358,22 @@ function activateComposeLanguageClient(ctx: vscode.ExtensionContext): void {
...
+
+    let hasSbom:boolean = false;
```
danegsta, line 365: nit: missing space in type declaration

danegsta: supportsSbom?

src/extension.ts
```diff
@@ -352,3 +358,22 @@ function activateComposeLanguageClient(ctx: vscode.ExtensionContext): void {
...
+    const info = await ext.runWithDefaults(client =>
+        client.info({})
+    );
+
+    const dockerPlugins: DockerPlugins[] = info.clientInfo?.Plugins;
+    if (!dockerPlugins || dockerPlugins.length === 0) {
```
danegsta, line 371: You can just do:
supportsSbom = toArray(dockerPlugins || []).some((plugin: DockerPlugins) => (plugin.Name || '').toLowerCase() === 'sbom')

package.nls.json
alexyaang: I like this!

src/commands/images/generateSbom.ts
alexyaang: That makes sense

src/runtimes/docker/clients/DockerClientBase/DockerInfoRecord.ts
alexyaang: I agree

src/commands/images/generateSbom.ts
bwateratmsft: It may be sufficient to just run it in the terminal

src/extension.ts
```diff
@@ -184,8 +185,11 @@ function registerDockerClients(): void {
         ext.orchestratorManager.registerRuntimeClient(composeClient)
     );
 
+    // Sets context value for whether Docker version supports sbom
+    await checkDockerVersionForSbom();
```
bwateratmsft, line 189: IMO it's not worth slowing the activation to prevent this command from showing up on unsupported Docker versions. I'd be fine with just trying the command and showing an error if it doesn't work for any reason.

src/runtimes/docker/clients/DockerClientBase/DockerClientBase.ts
bwateratmsft: I think of it more as informative than as part of a workflow--e.g., you do it to look at something, but not to copy and paste the output somewhere. So IMO it's fine to always use JSON. If a different format is desired to archive the file or similar, they should do so directly from the CLI.

src/runtimes/docker/contracts/ContainerClient.ts
alexyaang: Yeah that make more sense!

src/runtimes/docker/contracts/ContainerClient.ts
alexyaang: Will do. But is there any particular reason why we do this?

src/runtimes/docker/clients/DockerClientBase/DockerClientBase.ts
alexyaang: Good catch

src/runtimes/docker/contracts/ContainerClient.ts
alexyaang: yeah I agree with you here

Issue #3690, opened 2022-11-09
Current docker context name in status bar


khssnv:
Hello. This is a feature request.

Is it possible to add an optional field with Docker context name to the Status bar?

It is convenient to inspect and update compose stacks and swarm clusters by switching Docker context. But user should keep in mind which context is active currently to not mess it up.

Now the current context is shown in "contexts" section of the Docker Explorer tab Side bar only.
![image](https://user-images.githubusercontent.com/833019/200756153-1d80b916-f12a-4c01-80bc-beda032c5890.png)

But it would be helpful to have an option to add a current context name in the status bar as well. Status bar looks like a suitable place for the context name because it is already possible to add remote machine name, current VCS branch name or DB connection name to the status bar with other plugins.
![image](https://user-images.githubusercontent.com/833019/200757780-64ffc70a-7f7c-4b4a-912e-85bb5b665efb.png)

bwateratmsft:
I like this idea. In fact I might like it better than the existing Contexts tree view. :smile:

alexyaang:
@khssnv  This has now been released in Docker extension version 1.26.0. You can enable the status bar item by: 

1. Going to VSCode Settings -> Docker Settings
2. Navigate to the setting `Docker › Contexts: Show in Status Bar` 
3. Check the checkbox and close the setting page
4. You should see the current Docker context in the left corner of your status bar 

> <img width="199" alt="image" src="https://github.com/microsoft/vscode-docker/assets/59073590/f9a3481d-3b7e-44d5-8ac7-57336ff34582">

You can switch context by clicking on the status bar. 

> <img width="671" alt="image" src="https://github.com/microsoft/vscode-docker/assets/59073590/c2f57656-878a-4269-b555-6a11bf669dbf">

PR #3883, opened 2023-03-28
Add Docker Context to Status Bar

alexyaang:
This PR closes issue #3690 by adding an option to display the current Docker context in VSCode's status bar.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -159,7 +159,7 @@
     "vscode-docker.config.docker.containers.sortBy": "The property to use to sort containers in Docker view: CreatedTime or Label",
     "vscode-docker.config.docker.contexts.description": "Any secondary properties to display for a Docker context (an array). Possible elements include: Name, Description and DockerEndpoint",
     "vscode-docker.config.docker.contexts.label": "The primary property to display for a Docker context: Name, Description or DockerEndpoint",
-    "vscode-docker.config.docker.contexts.showInStatusBar": "Show Docker context in the status bar",
+    "vscode-docker.config.docker.contexts.showInStatusBar": "Show current Docker context in the status bar",
     "vscode-docker.config.docker.images.groupBy": "The property to use to group images in Docker view: CreatedTime, FullTag, ImageId, None, Registry, Repository, RepositoryName, RepositoryNameAndTag, or Tag",
     "vscode-docker.config.docker.images.description": "Any secondary properties to display for a image (an array). Possible elements include: CreatedTime, FullTag, ImageId, Registry, Repository, RepositoryName, RepositoryNameAndTag, Tag, and Size",
     "vscode-docker.config.docker.images.label": "The primary property to display for a image: CreatedTime, FullTag, ImageId, Registry, Repository, RepositoryName, RepositoryNameAndTag, Tag, or Size",
--- a/package.nls.json
+++ b/package.nls.json
@@ -159,6 +159,7 @@
     "vscode-docker.config.docker.containers.sortBy": "The property to use to sort containers in Docker view: CreatedTime or Label",
     "vscode-docker.config.docker.contexts.description": "Any secondary properties to display for a Docker context (an array). Possible elements include: Name, Description and DockerEndpoint",
     "vscode-docker.config.docker.contexts.label": "The primary property to display for a Docker context: Name, Description or DockerEndpoint",
+    "vscode-docker.config.docker.contexts.showInStatusBar": "Show Docker context in the status bar",
     "vscode-docker.config.docker.images.groupBy": "The property to use to group images in Docker view: CreatedTime, FullTag, ImageId, None, Registry, Repository, RepositoryName, RepositoryNameAndTag, or Tag",
     "vscode-docker.config.docker.images.description": "Any secondary properties to display for a image (an array). Possible elements include: CreatedTime, FullTag, ImageId, Registry, Repository, RepositoryName, RepositoryNameAndTag, Tag, and Size",
     "vscode-docker.config.docker.images.label": "The primary property to display for a image: CreatedTime, FullTag, ImageId, Registry, Repository, RepositoryName, RepositoryNameAndTag, Tag, or Size",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -28,7 +28,7 @@ import { AzureAccountExtensionListener } from './utils/AzureAccountExtensionList
 import { logDockerEnvironment, logSystemInfo } from './utils/diagnostics';
 import { DocumentSettingsClientFeature } from './utils/DocumentSettingsClientFeature';
 import { migrateOldEnvironmentSettingsIfNeeded } from './utils/migrateOldEnvironmentSettingsIfNeeded';
-import { registerDockerContextStatusBarEvent, registerDockerContextStatusBarItems } from './utils/registerDockerContextStatusBarItems';
+import { registerDockerContextStatusBarEvent } from './utils/registerDockerContextStatusBarItems';
 
 export type KeyInfo = { [keyName: string]: string };
 
@@ -120,7 +120,6 @@ export async function activateInternal(ctx: vscode.ExtensionContext, perfStats: 
         registerCommands();
 
         // Set up docker context status bar items
-        void registerDockerContextStatusBarItems(ctx);
         registerDockerContextStatusBarEvent(ctx);
 
         registerDebugProvider(ctx);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -200,28 +200,44 @@ function registerDockerClients(): void {
         if (e.affectsConfiguration('docker.composeCommand')) {
             composeClient.reconfigure();
         }
+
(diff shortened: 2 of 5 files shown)


package.nls.json
```diff
@@ -159,6 +159,7 @@
     "vscode-docker.config.docker.containers.sortBy": "The property to use to sort containers in Docker view: CreatedTime or Label",
     "vscode-docker.config.docker.contexts.description": "Any secondary properties to display for a Docker context (an array). Possible elements include: Name, Description and DockerEndpoint",
     "vscode-docker.config.docker.contexts.label": "The primary property to display for a Docker context: Name, Description or DockerEndpoint",
+    "vscode-docker.config.docker.contexts.showInStatusBar": "Show Docker context in the status bar",
```
bwateratmsft, line 162:
```suggestion
    "vscode-docker.config.docker.contexts.showInStatusBar": "Show current Docker context in the status bar",
```

src/extension.ts
```diff
@@ -196,9 +200,49 @@ function registerDockerClients(): void {
...
+
+        if (e.affectsConfiguration('docker.contexts.showInStatusBar')) {
+            dockerContextStatusBarItem.dispose();
+            // Don't wait
+            void registerContextStatusBarItems(ext.context);
+        }
     });
 }
 
+async function registerContextStatusBarItems({ subscriptions }: vscode.ExtensionContext) {
```
bwateratmsft, line 212: In retrospect I think we _should_ put this into its own file, along with the configuration change event listener. `utils` folder is fine I think.

src/extension.ts
```diff
@@ -196,9 +200,49 @@ function registerDockerClients(): void {
         if (e.affectsConfiguration('docker.composeCommand')) {
             composeClient.reconfigure();
         }
+
+        if (e.affectsConfiguration('docker.contexts.showInStatusBar')) {
```
bwateratmsft, line 204: Along with moving to another file, we can do another event registration. Overall it makes for very slightly less efficient code, but much neater.

src/extension.ts
alexyaang: good suggestion, will do!

src/extension.ts
```diff
@@ -196,9 +200,49 @@ function registerDockerClients(): void {
...
     });
 }
 
+async function registerContextStatusBarItems({ subscriptions }: vscode.ExtensionContext) {
+    const currentContext = await ext.runtimeManager.contextManager.getCurrentContext();
+    const showInStatusBar = vscode.workspace.getConfiguration('docker').get('contexts.showInStatusBar', false);
+
+    // if it's undefined, it means there is no context set, so we don't need to show the status bar item
+    // if user don't want context to clutter up status bar, we don't need to show
+    if (!currentContext || !showInStatusBar) {
```
bwateratmsft, line 217: I'd set this up so that we check `showInStatusBar` first, and don't check `currentContext` if `showInStatusBar` is false. The context lookup could potentially be slow, and will result in communicating with Docker before it's necessary. Something like this:

```typescript
const config = vscode.workspace.getConfiguration('docker');
if (!config.get('contexts.showInStatusBar', false)) {
    dockerContextStatusBarItem?.dispose();
    return;
}

const currentContext = await ext.runtimeManager.contextManager.getCurrentContext();
if (!currentContext) {
    dockerContextStatusBarItem?.dispose();
    return;
}
```

or

```typescript
const config = vscode.workspace.getConfiguration('docker');
let currentContext: ListContextItem | undefined;

if (!config.get('contexts.showInStatusBar', false) ||
    !(currentContext = await ext.runtimeManager.contextManager.getCurrentContext())) { // Intentional assignment and boolean check
    dockerContextStatusBarItem?.dispose();
    return;
}
```

alexyaang: that's a great point, context checking does take around 1~ second and this would avoid unnecessary slower runtime

src/utils/registerDockerContextStatusBarItems.ts
```diff
@@ -0,0 +1,62 @@
...
+        currentDockerContext = await ext.runtimeManager.contextManager.getCurrentContext();
+        const currentContextName = `Context: ${currentDockerContext.name}`;
+        ext.dockerContextStatusBarItem.text = currentContextName;
+        ext.dockerContextStatusBarItem.tooltip = vscode.l10n.t('Change Docker Context');
+        ext.dockerContextStatusBarItem.show();
+    }
+
+    subscriptions.push(
+        ext.dockerContextStatusBarItem,
+        vscode.workspace.onDidChangeConfiguration(updateStatusBar),
```
bwateratmsft, line 56:
```suggestion
```

This line can be removed. The `registerEvent` method in `registerDockerContextStatusBarEvent` is a utility that automatically will add the disposable to `context.subscriptions`.

src/utils/registerDockerContextStatusBarItems.ts
```diff
@@ -0,0 +1,62 @@
...
+        ext.runtimeManager.contextManager.onContextChanged(updateStatusBar)
```
bwateratmsft, line 57: I'd move this registration up into `registerDockerContextStatusBarEvent()` as a second registration, having it also call `registerDockerContextStatusBarItems`.

src/utils/registerDockerContextStatusBarItems.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    const dockerContextUseCommand = 'vscode-docker.contexts.use';
+
+    // Register the status bar item for the current context
+    ext.dockerContextStatusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, 20);
+    ext.dockerContextStatusBarItem.command = dockerContextUseCommand;
+    ext.dockerContextStatusBarItem.name = vscode.l10n.t('Docker Contexts');
+
+    async function updateStatusBar() {
```
bwateratmsft, line 46: This can be flattened out into the parent function

bwateratmsft:
I guess my thinking for the structure is different than what you have. To be clear, this it looks like this approach should work fine, although line 56 where _all_ config changes lead to an update is too permissive.

My thinking was to tie up the two trigger events--config change and context change--and have them call a `register` / `setup` / whatever `StatusBarItem`, which would, regardless of the prior state, make things match the correct state--that is, dispose of the old status bar item (if it is present), if the setting is false or there's no context then bail out, otherwise create a new status bar and set up the text and command and such.

However, if disposing and the reinitializing of the status bar item cause flashing or causes things to jump around, then it's definitely _not_ the way to go.

romanbuten (changes requested):
#

src/utils/registerDockerContextStatusBarItems.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // if dockerContextStatusBarItem is created, then we dispose
+    if (!config.get(dockerContextStatusBarSetting, false) ||
+        !(currentDockerContext = await ext.runtimeManager.contextManager.getCurrentContext())) { // Intentional assignment and boolean check
+        return;
+    }
+
+    const dockerContextUseCommand = 'vscode-docker.contexts.use';
+    const currentContextName = `Context: ${currentDockerContext.name}`;
+    currentDockerContext = await ext.runtimeManager.contextManager.getCurrentContext();
```
bwateratmsft, line 46: This line is redundant; the above `!(currentDockerContext = await ext.runtimeManager.contextManager.getCurrentContext()))` will have already set `currentDockerContext`

```suggestion
```

src/utils/registerDockerContextStatusBarItems.ts
```diff
@@ -0,0 +1,55 @@
...
+    ext.dockerContextStatusBarItem?.dispose();
+
+    // if dockerContextStatusBarItem is created, then we dispose
+    if (!config.get(dockerContextStatusBarSetting, false) ||
+        !(currentDockerContext = await ext.runtimeManager.contextManager.getCurrentContext())) { // Intentional assignment and boolean check
+        return;
+    }
+
+    const dockerContextUseCommand = 'vscode-docker.contexts.use';
+    const currentContextName = `Context: ${currentDockerContext.name}`;
```
bwateratmsft, line 45: This is subjective, but personally I'd rather it show just the context name, without the "Context: " before. I don't feel strongly though. Can you ask for PM input?

src/utils/registerDockerContextStatusBarItems.ts
```diff
@@ -0,0 +1,63 @@
...
+    ctx.subscriptions.push(
+        ext.dockerContextStatusBarItem,
+        ext.runtimeManager.contextManager.onContextChanged(scheduleUpdateStatusBar)
+    );
+
+    // Don't wait
+    void scheduleUpdateStatusBar();
+}
+
+export async function showStatusBarItemIfNeeded() {
```
bwateratmsft, line 35:
```suggestion
async function showStatusBarItemIfNeeded() {
```
Since this function is not being used externally we don't need to export it

bwateratmsft (approved):

PR #3908, opened 2023-04-17
Bumped version to 1.25.5-alpha

alexyaang:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-docker",
-    "version": "1.25.0",
+    "version": "1.25.5-alpha",
```
bwateratmsft, line 3: The changes in `package-lock.json` are actually supposed to be there.

bwateratmsft (approved):

PR #3921, opened 2023-04-27
Support native .NET 8 image build

alexyaang:
To test:
- [X] Dotnet:
    - [x] SDK build
    - [X] Dockerfile build 
    - [x] Run container
- [x] Python:
     - [x] Dockerfile build 
- [x] Node.js:
     - [x] Dockerfile build


src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -32,6 +30,12 @@ export class DockerBuildTaskProvider extends DockerTaskProvider {
 
     // TODO: Skip if image is freshly built
     protected async executeTaskInternal(context: DockerBuildTaskContext, task: DockerBuildTask): Promise<void> {
+
+        if (task.definition.netCore?.useSdkBuild) {
```
danegsta, line 34: Any .NET specific behavior should be implemented by overriding the new build function in NetCoreTaskHelper; DockerBuildTaskProvider shouldn't implement any custom behavior for .NET specific build actions.

src/tasks/DockerPseudoterminal.ts
```diff
@@ -120,10 +120,37 @@ export class DockerPseudoterminal implements Pseudoterminal {
             }
         );
     }
+
+    public async executeCommandInTerminal(command: string, options?: ExecuteCommandInTerminalOptions): Promise<ExecAsyncOutput> {
```
danegsta, line 124: Consider naming this execAsyncInTerminal or similar; using Command has to much potential for confusion with CommandProvider/CommandResponse.

src/tasks/DockerPseudoterminal.ts
```diff
@@ -120,10 +120,37 @@ export class DockerPseudoterminal implements Pseudoterminal {
...
+
+    public async executeCommandInTerminal(command: string, options?: ExecuteCommandInTerminalOptions): Promise<ExecAsyncOutput> {
+
+        // Output what we're doing, same style as VSCode does for ShellExecution/ProcessExecution
+        this.write(`> ${command} <\r\n\r\n`, DEFAULTBOLD);
+
+        return await execAsync(
+            command,
+            {
+                cwd: this.resolvedDefinition.options?.cwd || options.folder.uri.fsPath,
```
danegsta, line 132: The env logic from executeCommandResponseInTerminal would still apply here, as the dotnet sdk build still uses Docker internally to add the resulting image to the local image registry.

src/tasks/DockerPseudoterminal.ts
```diff
@@ -120,10 +120,37 @@ export class DockerPseudoterminal implements Pseudoterminal {
...
 }
 
-type ExecuteCommandInTerminalOptions = {
+type ExecuteCommandResponseInTerminalOptions = {
     commandResponse: VoidCommandResponse | PromiseCommandResponse<unknown>;
     folder: WorkspaceFolder;
     token?: CancellationToken;
 };
+
+type ExecuteCommandInTerminalOptions = {
```
danegsta, line 153: ExecuteCommandResponseInTerminalOptions should extend this type (since it's a superset of this type)

src/tasks/TaskHelper.ts
```diff
@@ -62,45 +59,39 @@ export interface DockerRunTaskContext extends DockerTaskExecutionContext {
 // This doesn't need to be extended so redefining for parity is simplest
 export type DockerComposeTaskContext = DockerTaskExecutionContext;
 
-export interface TaskHelper {
+export abstract class TaskHelper {
```
danegsta, line 62: We'll need a "DefaultTaskHelper" or similar that will be used if there isn't a specific project type TaskHelper (dotnet, python, etc.) since build is now part of the TaskHelper. Otherwise we won't be able to build anything but .NET, Node, and Python projects. Basically the helper should be one of the specific helpers if one applies, otherwise it should be a default/generic helper.

src/tasks/DockerBuildTaskProvider.ts
alexyaang: Good catch! I forgot to delete it after I moved the logic to `NetCoreTaskHelper` :(

src/tasks/DockerPseudoterminal.ts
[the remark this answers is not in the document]
alexyaang: yeah this makes sense

alexyaang: That makes sense, done!

src/tasks/DockerPseudoterminal.ts
alexyaang: Got it!

src/tasks/DockerPseudoterminal.ts
alexyaang: :+1:

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -287,6 +296,13 @@ export class NetCoreTaskHelper implements TaskHelper {
 
         return volumes;
     }
+
+    private async buildWithDotnetSdk(context: DockerBuildTaskContext): Promise<void> {
+        //reference: https://learn.microsoft.com/en-us/dotnet/core/docker/publish-as-container
+        const publishFlag = NetCoreTaskHelper.isWebApp ? '-p:PublishProfile=DefaultContainer' : '/t:PublishContainer';
+        const sdkBuildCommand = `dotnet publish --os linux --arch x64 -c Release ${publishFlag}`;
```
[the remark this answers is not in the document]
alexyaang: What should the image tag be set to match the settings in `launch.json`? I don't see an image tag in launch :sweat_smile:

src/tasks/registerTaskProviders.ts
```diff
@@ -0,0 +1,37 @@
+import { ExtensionContext, tasks } from 'vscode';
+import { DockerBuildTaskProvider } from './DockerBuildTaskProvider';
+import { DockerComposeTaskProvider } from './DockerComposeTaskProvider';
+import { DockerRunTaskProvider } from './DockerRunTaskProvider';
+import { netCoreTaskHelper } from './netcore/NetCoreTaskHelper';
+import { nodeTaskHelper } from './node/NodeTaskHelper';
+import { pythonTaskHelper } from './python/PythonTaskHelper';
+
+
+export function registerTaskProviders(ctx: ExtensionContext): void {
```
danegsta, line 10: Consider moving this file into the "utils" folder with other helper functions.

src/tasks/TaskHelper.ts
```diff
@@ -62,45 +59,55 @@ export interface DockerRunTaskContext extends DockerTaskExecutionContext {
...
+            path: options.context,
+        });
+
+        const runner = context.terminal.getCommandRunner({
+            folder: context.folder,
+            token: context.cancellationToken,
+        });
+
+        await runner(command);
+    }
```
danegsta, line 93: Ensure you have new lines between method definitions

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -132,6 +133,14 @@ export class NetCoreTaskHelper implements TaskHelper {
...
+    public async build(context: DockerBuildTaskContext, buildDefinition: DockerBuildTaskDefinition) {
+        if (buildDefinition.netCore?.useSdkBuild) {
+            await this.buildWithDotnetSdk(context, buildDefinition); // build with .NET SDK
+        } else {
+            await super.build(context, buildDefinition); // Dockerfile build
+        }
+    }
+
     public async getDockerRunOptions(context: DockerRunTaskContext, runDefinition: NetCoreRunTaskDefinition): Promise<DockerRunOptions> {
         const runOptions = runDefinition.dockerRun;
```
danegsta, line 145:
```suggestion
        const runOptions = super.getDockerRunOptions(context, runDefinition);
```

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -287,6 +296,49 @@ export class NetCoreTaskHelper implements TaskHelper {
 
         return volumes;
     }
+
+    private async buildWithDotnetSdk(context: DockerBuildTaskContext, buildDefinition: DockerBuildTaskDefinition): Promise<void> {
+        //reference: https://learn.microsoft.com/en-us/dotnet/core/docker/publish-as-container
```
danegsta, line 301: Consider adding this comment as a JSDoc @see entry on the method (https://jsdoc.app/tags-see.html)

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -287,6 +296,49 @@ export class NetCoreTaskHelper implements TaskHelper {
...
+        await context.terminal.execAsyncInTerminal(
+            sdkBuildCommand,
+            {
+                folder: context.folder,
+                token: context.cancellationToken,
+            }
+        );
+    }
+
+    private getImageNameAndTag(buildDefinition: DockerBuildTaskDefinition): [string, string] {
```
danegsta, line 316: Consider using the existing parseDockerLikeImageName method as it handles a number of image name/tag format edge cases.

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -287,6 +296,49 @@ export class NetCoreTaskHelper implements TaskHelper {
...
         return volumes;
     }
+
+    private async buildWithDotnetSdk(context: DockerBuildTaskContext, buildDefinition: DockerBuildTaskDefinition): Promise<void> {
+        //reference: https://learn.microsoft.com/en-us/dotnet/core/docker/publish-as-container
+        const publishFlag = NetCoreTaskHelper.isWebApp ? '-p:PublishProfile=DefaultContainer' : '/t:PublishContainer';
+        const [runtimeOs, runtimeArch] = this.getOsAndArch(buildDefinition);
+        const [imageName, imageTag] = this.getImageNameAndTag(buildDefinition);
+
+        const sdkBuildCommand = `dotnet publish --os ${runtimeOs} --arch ${runtimeArch} ${publishFlag} -c Release -p:ContainerImageName=${imageName} -p:ContainerImageTag=${imageTag}`;
```
danegsta, line 306: The build configuration should probably default to Debug instead of Release. It would make sense to have a discussion about how/if we would want to make the property configurable (without breaking existing behavior).

```suggestion
        const sdkBuildCommand = `dotnet publish --os ${runtimeOs} --arch ${runtimeArch} ${publishFlag} -c Debug -p:ContainerImageName=${imageName} -p:ContainerImageTag=${imageTag}`;
```

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -33,50 +31,21 @@ export class DockerBuildTaskProvider extends DockerTaskProvider {
     // TODO: Skip if image is freshly built
     protected async executeTaskInternal(context: DockerBuildTaskContext, task: DockerBuildTask): Promise<void> {
         const definition = cloneObject(task.definition);
-        definition.dockerBuild = definition.dockerBuild || {};
```
danegsta, line 36: Make sure you're invoking the new base getDockerBuildOptions method in any of the child class method overrides to avoid the dockerBuild property being undefined.

src/tasks/DockerRunTaskProvider.ts
```diff
@@ -31,7 +31,6 @@ export class DockerRunTaskProvider extends DockerTaskProvider {
     // TODO: Skip if container is freshly started, but probably depends on language
     protected async executeTaskInternal(context: DockerRunTaskContext, task: DockerRunTask): Promise<void> {
         const definition = cloneObject(task.definition);
-        definition.dockerRun = definition.dockerRun || {};
```
danegsta, line 34: Make sure you're invoking the new base getDockerRunOptions method in any of the child class method overrides to avoid the dockerRun property being undefined.

src/tasks/DockerBuildTaskProvider.ts
danegsta: Or ensure you initialize to a default value if undefined/null.

src/tasks/netcore/NetCoreTaskHelper.ts
alexyaang: That makes sense, I'll change it to debug for now and I'll bring this up at standup tomorrow.

src/tasks/registerTaskProviders.ts
alexyaang: :satisfied:

src/tasks/DockerBuildTaskProvider.ts
alexyaang: This makes sense, god catch! :sweat:

src/tasks/netcore/NetCoreTaskHelper.ts
alexyaang: Nice! will do

src/tasks/DockerBuildTaskProvider.ts
alexyaang: Will do

src/tasks/netcore/NetCoreTaskHelper.ts
alexyaang: Will do

src/tasks/TaskHelper.ts
```diff
@@ -62,45 +59,56 @@ export interface DockerRunTaskContext extends DockerTaskExecutionContext {
...
+
+    async postBuild(context: DockerBuildTaskContext, buildDefinition: DockerBuildTaskDefinition): Promise<void> {
+        return;
+    }
+
+    async preRun(context: DockerRunTaskContext, runDefinition: DockerRunTaskDefinition): Promise<void> {
+        return;
+    }
```
danegsta, line 101: More missing new lines

alexyaang: :disappointed_relieved: I should've looked closer

src/tasks/DockerBuildTaskProvider.ts
```diff
@@ -7,14 +7,12 @@ import * as fse from 'fs-extra';
...
 import { cloneObject } from '../utils/cloneObject';
 import { resolveVariables } from '../utils/resolveVariables';
 import { DockerBuildOptions } from './DockerBuildTaskDefinitionBase';
 import { DockerTaskProvider } from './DockerTaskProvider';
 import { NetCoreBuildTaskDefinition } from './netcore/NetCoreTaskHelper';
 import { NodeBuildTaskDefinition } from './node/NodeTaskHelper';
-import { defaultVsCodeLabels, getAggregateLabels } from './TaskDefinitionBase';
 import { DockerBuildTaskContext, TaskHelper, throwIfCancellationRequested } from './TaskHelper';
 
 export interface DockerBuildTaskDefinition extends NetCoreBuildTaskDefinition, NodeBuildTaskDefinition {
```
danegsta, line 18: We should revisit the build/run task inheritance, as the current task definition inheritance is the opposite of the TaskHelper inheritance. Theoretically we should be able to structure the types such that we won't need to explicitly cast between types when calling the base methods on TaskHelper.

alexyaang:
We also have plans for the `configuration` flag in .NET Dockerfile builds. More details are documented here: https://github.com/microsoft/vscode-docker/issues/3927

package.nls.json
```diff
@@ -60,6 +60,8 @@
     "vscode-docker.tasks.docker-build.platform": "The target platform for the application.",
     "vscode-docker.tasks.docker-build.netCore.description": "Options for building .NET projects in Docker.",
     "vscode-docker.tasks.docker-build.netCore.appProject": "The .NET project (.csproj, .fsproj, etc.) to build.",
+    "vscode-docker.tasks.docker-build.netCore.useSdkBuild": "Whether to use the .NET Core SDK to build the project.",
+    "vscode-docker.tasks.docker-build.netCore.configuration": "The configuration flag in .NET that modifies runtime behavior.",
```
danegsta, line 64:
```suggestion
    "vscode-docker.tasks.docker-build.netCore.configuration": "The .NET build configuration.",
```

package.nls.json
```diff
@@ -60,6 +60,8 @@
     "vscode-docker.tasks.docker-build.platform": "The target platform for the application.",
     "vscode-docker.tasks.docker-build.netCore.description": "Options for building .NET projects in Docker.",
     "vscode-docker.tasks.docker-build.netCore.appProject": "The .NET project (.csproj, .fsproj, etc.) to build.",
+    "vscode-docker.tasks.docker-build.netCore.useSdkBuild": "Whether to use the .NET Core SDK to build the project.",
```
danegsta, line 63:
```suggestion
    "vscode-docker.tasks.docker-build.netCore.useSdkBuild": "Build the image using the .NET SDK (instead of a Dockerfile).",
```

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -287,6 +298,44 @@ export class NetCoreTaskHelper implements TaskHelper {
...
     }
+    /**
+     * @see {@link https://learn.microsoft.com/en-us/dotnet/core/docker/publish-as-container} for further information
+     */
+    private async buildWithDotnetSdk(context: DockerBuildTaskContext, buildDefinition: DockerBuildTaskDefinition): Promise<void> {
+        const configuration = buildDefinition.netCore?.configuration ? `-c ${buildDefinition.netCore?.configuration} ` : '';
+        const publishFlag = NetCoreTaskHelper.isWebApp ? '-p:PublishProfile=DefaultContainer' : '/t:PublishContainer';
+        const [runtimeOs, runtimeArch] = this.getOsAndArch(buildDefinition);
+        const imageNameInfo = parseDockerLikeImageName(buildDefinition.dockerBuild.tag);
+        const sdkBuildCommand = `dotnet publish --os ${runtimeOs} --arch ${runtimeArch} ${publishFlag} ${configuration}-p:ContainerImageName=${imageNameInfo.image} -p:ContainerImageTag=${imageNameInfo.tag}`;
```
danegsta, line 309: I'd suggest using the commandLineBuilder methods to make this code easier to maintain/debug.

Something like:

const sdkBuildCommand = composeArgs(
  withArg('dotnet', 'publish'),
  withNamedArg('-c', buildDefinition.netCore?.configuration),
  ...
)();

await context.terminal.execAsyncInTerminal(
  Shell.getShellOrDefault().quote(sdkBuildCommand),
  ...
);

src/tasks/netcore/NetCoreTaskHelper.ts
```diff
@@ -287,6 +298,44 @@ export class NetCoreTaskHelper implements TaskHelper {
...
+            }
+        );
+    }
+
+    private getOsAndArch(buildDefinition: DockerBuildTaskDefinition): [string, string] {
+        let runtimeOs: string = "linux";
+        let runtimeArch: string = "x64";
+        if (buildDefinition.dockerBuild?.platform) {
+            const [dockerOs, dockerArch] = buildDefinition.dockerBuild.platform.split('/');
+            runtimeOs = dockerOs;
```
danegsta, line 324: It looks like we will need to map between Docker and .NET for the OS; it's windows-* for Docker, but win-* for .NET.

bwateratmsft (commented):
Having the `docker-build` task run `dotnet build` seems out of its charter IMO. @danegsta @alexyaang is it plausible to scaffold a proper `dotnet publish` shell(ish) task and make _that_ the `dependsOn` for our `docker-run` task? It might remove the need to make almost all of these changes.

alexyaang:
> IIRC, the main thing the `docker-run` task needs from the `docker-build` task before it is the image name. We could just scaffold something into the `dotnet publish` task, and if they break it it's their fault. Or, alternatively, devise something like Docker's build args for that `dotnet publish` task, and allow the .NET task helper's `getDockerRunOptions` look into those args.

After discussing this offline w/ @danegsta  and @bwateratmsft. The new approach proposed by Brandon is creating a new ambient task and a new `TaskProvider`, I'm going to experiment with this by creating a dummy extension and see it this approach is possible. If so, this is definitely a route we can explore.

alexyaang:
I was able to build a proof-of-concept ambient task that can be called by a real run-task. So this is a potential option that we can explore for this feature.

PR #3928, opened 2023-05-04
Add back `onLanguage:dockerfile` activation event

stackia:
Previously we rely on the inferred activation event for `dockerfile`, but https://github.com/microsoft/vscode/pull/179287 breaks this behavior for us.

This PR adds back the `onLanguage:dockerfile` activation event in package.json.

Fixes https://github.com/microsoft/vscode-docker/issues/3899


bwateratmsft (approved):
Thank you for the PR!

Issue #3576, opened 2022-07-16
HEREDOC parsing results in Empty Continuation Line output


rbairwell:
Possibly a reoccurrence of https://github.com/microsoft/vscode-docker/pull/3399 / https://github.com/microsoft/vscode-docker/pull/3355

If a HEREDOC section of a Dockerfile (as introduced in Buildx 1.4 https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#here-documents , if there is an empty line in the HEREDOC/HERE_DOCUMENT section, this extension reports "**Empty continuation line** dockerfile-utils(11)"

### Docker File snippet:
```
# syntax=docker/dockerfile:1.4
....
FROM debian:bullseye
...
# see https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/
COPY <<EOF ${MY_PATH}/file.patch
--- file	2022-07-15 18:10:27.286229900 +0000
+++ file.two	2022-07-15 18:09:57.204518600 +0000
@@ -3,11 +3,11 @@
 # DEBUG_CFLAGS=-g -O0

 # LOCALCFLAGS= -I/usr/local/include/libxml2
```
with the report happening on the empty line between the two commented out sections.

### Screenshot:
![image](https://user-images.githubusercontent.com/775269/179346921-cb1aed52-1b82-4c9e-9814-bde947d8ee89.png)

### Environment
**Extension version:** v1.22.1 ( last updated 11/07/2022, 15:43:38 )
**VS Code Version:**
Version: 1.69.1 (user setup)
Commit: b06ae3b2d2dbfe28bca3134cc6be65935cdfea6a
Date: 2022-07-12T08:21:24.514Z
Electron: 18.3.5
Chromium: 100.0.4896.160
Node.js: 16.13.2
V8: 10.0.139.17-electron.0
OS: Windows_NT x64 10.0.19044

bwateratmsft:
@rcjsuen can you take a look?

rcjsuen:
I can confirm that this is a problem. Heredocs are only supported for `RUN` instructions at the moment. I will update this issue once I have created the relevant issues for tracking.

rcjsuen:
This is now being tracked by https://github.com/rcjsuen/dockerfile-utils/issues/112.

bwateratmsft:
[...]

Issue #3950, opened 2023-05-24
Autoformatting breaks Dockerfile with heredoc (for COPY instruction)


moha-gh:
Hi,

I have a problem that is very similar to #3327. Using `v1.25.1` of the extension, a simple `Dockerfile` that uses [heredocs](https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/) in `COPY` instructions like this:

```
FROM alpine:3.18

COPY <<EOF /foo/bar
x
y
z
EOF

COPY <<EOF /bar/foo
a
b
c
EOF

```

is re-formatted to

```
FROM alpine:3.18

COPY <<EOF /foo/bar
    x
    y
    z
    EOF

COPY <<EOF /bar/foo
    a
    b
    c
    EOF

```

which breaks things completely (due to the indentation of the `EOF` marker).

From the previously mentioned bug report I gather that https://github.com/rcjsuen/dockerfile-utils is used for formatting. If I use the latest upstream release ([0.11.0](https://github.com/rcjsuen/dockerfile-utils/releases/tag/v0.11.0))  via `docker run rcjsuen/dockerfile-utils`, the file is formatted correctly.

Would it be possible to pull in the newer version? Between [0.9.0](https://github.com/microsoft/vscode-docker/blob/c77c14fa57d6cd25f6543046fe2d58440a029c02/package-lock.json#L2381) which is in use right now AFAIU and 0.11.0 there have been a number of heredoc-related fixes (see [here](https://github.com/rcjsuen/dockerfile-utils/compare/v0.9.0...v0.11.0)) - and other as well.

bwateratmsft:
Unfortunately there isn't a version of `dockerfile-language-server-nodejs` / `dockerfile-language-service` that uses v0.11.0 of the `dockerfile-utils` package. So no compatible version exists. It's possible we could override those two packages to force them to use `dockerfile-utils` 0.11.0, assuming it actually does work.

alexyaang:
The fix has now been released in Docker extension version 1.26.0

Issue #3836, opened 2023-02-16
Request textDocument/semanticTokens failed (on edge case)


scchow:
# Problem Description

I am encountering the same issue as #2055, which was reported to be fixed a while ago, but has popped up for me in a particular edge case.

This seems to occur when I first end a `RUN` command with `&\`, then add a comment with a markdown link of form `[txt](url)`, and continue the typing below. See the example below for details.

When opening/editing a Dockerfile with this bug, I get repeated notifications that `Request textDocument/semanticTokens/full failed.` Examining the full output, I see:
``` 
[Error - 6:18:50 PM] Request textDocument/semanticTokens/full failed.
  Message: Request textDocument/semanticTokens/full failed with message: Maximum call stack size exceeded
  Code: -32603 
...
<Omitted for brevity>
...
[Error - 6:18:51 PM] Request textDocument/semanticTokens/full failed.
  Message: Request textDocument/semanticTokens/full failed with message: Maximum call stack size exceeded
  Code: -32603 
```
It seems that with every keypress, I get a new message. 

# Sample Dockerfile
``` Dockerfile
FROM ubuntu 

RUN echo "hi there" &\
# Let's add a comment with a markdown link [here](https://www.google.com) hope nothing breaks...
echo "oops I broke something and I'm getting a lot of popup messages"
```

This is a very particular edge case as it only appears if the following conditions are met:
1. End a line with `&\` exactly. Note if you end it with `& \`, this bug does not occur
2. Add a comment with a standard markdown link of style `[txt](url)`
3. Continue the command below the comment

# Setup Details
I was able to reproduce this bug on two setups.
1. Just my primary machine
2. Using Remote-SSH to ssh into a remote machine and editing a Dockerfile on that machine.

## Primary Machine
- OS: Microsoft Windows 10
- VS Code Version: 1.75.1
- Docker Extension Version: v1.23.3
- Remote-SSH Version: v0.96.0

## Remote Machine
- OS: Ubuntu 22.04
- Docker Extension Version: v1.23.3

dbreshears:
@rcjsuen.  Can you take a look at this one?

rcjsuen:
Thank you for the bug report. I have opened https://github.com/rcjsuen/dockerfile-language-service/issues/110 to track this.

alexyaang:
The fix has now been released in Docker extension version 1.26.0

PR #3955, opened 2023-05-28
Update Dockerfile language server to 0.10.1

rcjsuen:
First of all, apologies to everyone for having just disappeared into thin air and also thank you to everyone for your continued bug reports (even if I did not respond to them until now...) as it helps me identify things to fix to make everyone's Dockerfile editing experience a little better. I can't promise any turnaround times but I hope to be more responsive than I have been in the last year or so after seemingly dropping off the face of the earth. :P

---

This update adds heredoc support for `COPY` so that it does not try to format it or lint it for Dockerfile-specific errors. The update also includes a fix caused by quotes in comments that can cause the semantic highlighter to get into an infinite loop.

This should fix the following issues:
- #3576
- #3836
- #3950

This does *not* fix https://github.com/microsoft/vscode-docker/issues/3576#issuecomment-1437199184. I need to wrap my head around that one and I would suggest creating a separate issue about that.

The following file should trigger the above three problems so it can be verified on the current build. With the fix, you should:
1. not be spammed by the "Request textDocument/semanticTokens/full failed." error whenever you try to type anything in the file
2. formatting should indent `/folder` in the second `COPY` but ignore the first `COPY`
3. as noted in the file, line 11 should not get flagged by a false positive
```Dockerfile
FROM ubuntu 

RUN echo "hi there" &\
# Let's add a comment with a markdown link [here](https://www.google.com) hope nothing breaks...
echo "oops I broke something and I'm getting a lot of popup messages"

# there should not be an empty continuation line error on line 11
COPY <<EOF /bar/foo
a
b

c
EOF

COPY a b \
/folder/
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
                 "@microsoft/vscode-azext-azureutils": "^1.1.5",
                 "@microsoft/vscode-azext-utils": "^1.2.2",
                 "dayjs": "^1.11.7",
-                "dockerfile-language-server-nodejs": "^0.10.0",
+                "dockerfile-language-server-nodejs": "^0.10.1",
                 "fs-extra": "^11.1.1",
                 "gradle-to-js": "^2.0.1",
                 "handlebars": "^4.7.7",
@@ -2366,9 +2366,9 @@
             "integrity": "sha512-c68LpLbO+7kP/b1Hr1qs8/BJ09F5khZGTxqxZuhzxpmwJKOgRFHJWIb9/KmqnqHhLdO55aOxFH/EGBvUQbL/RQ=="
         },
         "node_modules/dockerfile-ast": {
-            "version": "0.4.2",
-            "resolved": "https://registry.npmjs.org/dockerfile-ast/-/dockerfile-ast-0.4.2.tgz",
-            "integrity": "sha512-k770mVWaCm3KbyOSPFizP6WB2ucZjfAv8aun4UsKl+IivowK7ItwBixNbziBjN05yNpvCL1/IxBdZiSz6KQIvA==",
+            "version": "0.5.0",
+            "resolved": "https://registry.npmjs.org/dockerfile-ast/-/dockerfile-ast-0.5.0.tgz",
+            "integrity": "sha512-YsRrWww6mKRS1HK32gbherPlgfvwS593ZeDegb5glNCBJgByICAYZCP5F+njY6TSB0eiPdFgCU9RkuO516mLFQ==",
             "dependencies": {
                 "vscode-languageserver-textdocument": "^1.0.1",
                 "vscode-languageserver-types": "^3.17.0-next.3"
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -2394,24 +2394,24 @@
...
         },
         "node_modules/dockerfile-language-service": {
-            "version": "0.9.0",
-            "resolved": "https://registry.npmjs.org/dockerfile-language-service/-/dockerfile-language-service-0.9.0.tgz",
-            "integrity": "sha512-sDUkTR4LUimEAWXDIbUTEx2CytCBlx+XlJkg4B2Ptvak9HkwPD4JpXesaWxXPpp6YHCzxqwsTDY7Gf79ic340g==",
+            "version": "0.10.0",
+            "resolved": "https://registry.npmjs.org/dockerfile-language-service/-/dockerfile-language-service-0.10.0.tgz",
+            "integrity": "sha512-1EfLBQViRSzRFFet8co2r3CVudniKPbDB/Q595RUouewUs6UKpe4pdzrTuXIc4TzIGdRXUnBZCZdiN0T9fl+8w==",
             "dependencies": {
                 "dockerfile-ast": "0.4.2",
```
bwateratmsft, line 2401: Should `dockerfile-language-service` be updated to use `dockerfile-ast@0.5.0`, to match what is used by `dockerfile-utils@0.11.0`?

package.json
```diff
@@ -2994,7 +2994,7 @@
         "@microsoft/vscode-azext-azureutils": "^1.1.5",
         "@microsoft/vscode-azext-utils": "^1.2.2",
         "dayjs": "^1.11.7",
-        "dockerfile-language-server-nodejs": "^0.9.0",
+        "dockerfile-language-server-nodejs": "^0.10.0",
```
bwateratmsft, line 2997: I'm still hitting the `textDocument/semanticTokens/full` failure

package-lock.json
rcjsuen: Good call. I have made the change and released a new 0.10.1 version.

package.json
rcjsuen: Unfortunately, I am unable to reproduce this problem when I launch the second instance of VS Code. 🤔

bwateratmsft: I'll try restarting completely...

bwateratmsft: Ugh, still happening after I rebuilt the extension into a VSIX and installed it. 😕

[Error - 4:29:50 PM] Request textDocument/semanticTokens/full failed.
  Message: Request textDocument/semanticTokens/full failed with message: Maximum call stack size exceeded
  Code: -32603

rcjsuen: https://github.com/microsoft/vscode-docker/assets/15629116/e196bf06-27e4-4ebb-83cb-7b43e415b1a0

My Extension Development Host window is working well for me.

bwateratmsft: Could it be OS somehow? I'm on Win11

bwateratmsft: Potentially related, if I change the line endings from CRLF to LF in the bottom, the error also goes away. AFAIK there are not actually any carriage returns in the text; `\r` has no results in a regex search.

rcjsuen: > Oh, even weirder, if line 5 is indented by even one space, I do not have the error.

Yeah. That is expected because the problem is that it's a "multiline" argument. The space will break it apart so the parsing doesn't cause any problems.

> Potentially related, if I change the line endings from CRLF to LF in the bottom, the error also goes away.

Thanks for noticing this. I can reproduce the problem after switching to CRLF. :) Will work on a fix.

rcjsuen: @bwateratmsft I've bumped it to version 0.10.2 which should fix the CRLF issue.

bwateratmsft (approved):

bwateratmsft:
Works great, thanks as always Remy!

PR #3960, opened 2023-06-06
Support .NET SDK Container Build

### LINKED ISSUE #3808: Enable debugging support for .NET native container builds
Starting with .NET 7, the .NET SDK enables [containerizing .NET applications without a Dockerfile](https://devblogs.microsoft.com/dotnet/announcing-builtin-container-support-for-the-dotnet-sdk/) via a simple `dotnet publish` command. Subsequent releases of this capability have focused on integrating the package into more project types - the 7.0.200 release of the SDK in February will include the package for Web SDK projects (those that use ASP.NET Core), and we plan to extend this to all .NET project types in the spring release of 7.0.300.

It would be amazing if we could integrate this package into the tooling for this extension, letting developers continue the easy containerization experience all the way up to publishing their app to the cloud provider even inside their IDE.

Conceretely, it would be great if the onboarding experience for .NET debugging support didn't require the use of scaffolded Dockerfiles, and could generate `publish` commands instead of `docker build` commands for the prerequisite build tasks.

alexyaang:
This PR closes #3808

Tests
- [x] Debug a .NET project with .NET SDK Container
- [x] Debug a .NET project with Dockerfile (check for regression)
- [x] Debug a python project with Dockerfile (check for regression)


package.json
```diff
@@ -636,6 +636,11 @@
             {
                 "type": "docker",
                 "label": "Docker: Debug in Container",
+                "languages": [
+                    "csharp",
+                    "razor",
+                    "aspnetcorerazor"
+                ],
```
alexyaang, line 643: should we add python & node.js here too? so when you're on a .py file or .js file and presses `f5`, you'll be able to see our debugger. I was thinking we can do that in a separate PR.

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -42,21 +44,14 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
...
-            .then((result) => {
-                if (result === add) {
-                    /* eslint-disable-next-line @typescript-eslint/no-floating-promises */
-                    commands.executeCommand('vscode-docker.configure');
-                }
-            });
 
-        return [];
+        return callWithTelemetryAndErrorHandling(
+            'docker-provideDebugConfigurations', //TODO: change this later
```
alexyaang, line 49: what should we name this? is `docker-provideDebugConfigurations` sufficient?

src/utils/StateUtils.ts
```diff
@@ -0,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ext } from '../extensionVariables';
+
+export async function getFromWorkspaceState<T>(key: string): Promise<T> {
```
alexyaang, line 8: do we think having a util class for workplace momentos is needed? I thought it would be nice to have a util class that handles it, but we can go either way.

package.json
bwateratmsft: I think another PR is fine

resources/netCore/GetProjectProperties.targets
```diff
@@ -1,11 +1,16 @@
 <Project>
-    <!-- All the relevant info is in root-level PropertyGroups, so there are no dependent targets to make this work -->
-    <Target Name="GetProjectProperties">
+    <PropertyGroup>
+        <GetProjectPropertiesDependsOn Condition=" '$(SDKContainerSupportEnabled)' == 'true' ">$(GetProjectPropertiesDependsOn);ComputeContainerConfig;</GetProjectPropertiesDependsOn>
```
bwateratmsft, line 3: This file still works when SDK container support is not enabled, right?

src/debugging/DockerDebugConfigurationProvider.ts
bwateratmsft: The telemetry name will be prepended with `vscode-docker.` automatically so I think we can at least remove the `docker-`.

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -75,10 +70,15 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
                     }
                 }
 
-                if (debugConfiguration.type === undefined) {
-                    // If type is undefined, they may be doing F5 without creating any real launch.json, which won't work
-                    // VSCode subsequently will call provideDebugConfigurations which will show an error message
-                    return null;
+                if (debugConfiguration.type === undefined &&
```
bwateratmsft, line 73: Should we look specifically for `undefined` or should any falsy value trigger this?

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -170,4 +170,51 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
             }
         }
     }
+
+    /**
+     * Copies the first debug configuration from the array to the second one.
+     */
+    private copyDebugConfiguration(fromArray: DockerDebugConfiguration[], to: DockerDebugConfiguration): void {
```
bwateratmsft, line 177: Does it not work to just do something like `to = fromArray[0]`?

bwateratmsft: (Or, the equivalent in the above function)

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -170,4 +170,51 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
...
+        if (csProjUris) {
+            return await netSdkDebugHelper.provideDebugConfigurations(
+                {
+                    actionContext,
+                    dockerfile: undefined,
+                    folder: folder
+                },
+                {
+                    appProject: csProjUris[0]?.absoluteFilePath || '',
+                });
```
bwateratmsft, line 211: Very minor nit

```suggestion
                }
            );
```

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -170,4 +170,51 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
...
+        }
+        else {
```
bwateratmsft, line 213: Very minor nit

```suggestion
        } else {
```

src/debugging/DockerDebugConfigurationProvider.ts
```diff
@@ -170,4 +170,51 @@ export class DockerDebugConfigurationProvider implements DebugConfigurationProvi
...
+            // for now, we scaffold docker files
+            await commands.executeCommand('vscode-docker.configure');
+            return [];
```
bwateratmsft, line 216: Should we throw a `UserCancelledError` instead?

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -0,0 +1,51 @@
+import { commands } from "vscode";
+import { NetChooseBuildTypeContext, netContainerBuild } from "../../scaffolding/wizard/net/NetContainerBuild";
+import { unresolveWorkspaceFolder } from "../../utils/resolveVariables";
+import { DockerDebugScaffoldContext } from "../DebugHelper";
+import { DockerDebugConfiguration } from "../DockerDebugConfigurationProvider";
+import { NetCoreDebugHelper, NetCoreDebugScaffoldingOptions } from "../netcore/NetCoreDebugHelper";
+
+const netContainerSdkBuildSymbol = 'dotnet-container-sdk: debug';
```
bwateratmsft, line 8: Nit: File-level constants usually get PascalCase
```suggestion
const NetContainerSdkBuildSymbol = 'dotnet-container-sdk: debug';
```

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Checks if the launch task is using the .NET SDK Container build
+     * @param preLaunchTask
+     * @returns true if the launch task is using the .NET SDK Container build
+     *          false otherwise
+     */
+    public isDotnetSdkBuild(preLaunchTask: string): boolean {
+        return preLaunchTask && preLaunchTask === netContainerSdkBuildSymbol;
```
bwateratmsft, line 47: This is valid but you could also do this:

```suggestion
        return preLaunchTask === netContainerSdkBuildSymbol;
```

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        const netCoreBuildContext: NetChooseBuildTypeContext = {
+            ...context.actionContext,
+            scaffoldType: 'debugging',
+            workspaceFolder: context.folder,
+        };
+
+        await netContainerBuild(netCoreBuildContext);
+
+        if (netCoreBuildContext?.containerBuildOptions === 'Use .NET SDK') {
```
bwateratmsft, line 23: Use the constant declaration in NetSdkChooseBuildStep.ts rather than cloning a new string

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -0,0 +1,51 @@
...
+import { DockerDebugScaffoldContext } from "../DebugHelper";
+import { DockerDebugConfiguration } from "../DockerDebugConfigurationProvider";
+import { NetCoreDebugHelper, NetCoreDebugScaffoldingOptions } from "../netcore/NetCoreDebugHelper";
+
+const netContainerSdkBuildSymbol = 'dotnet-container-sdk: debug';
+export class NetSdkDebugHelper extends NetCoreDebugHelper {
+
+    public async provideDebugConfigurations(context: DockerDebugScaffoldContext, options?: NetCoreDebugScaffoldingOptions): Promise<DockerDebugConfiguration[]> {
+
+        const providers: DockerDebugConfiguration[] = [];
```
bwateratmsft, line 13:
```suggestion
        const configurations: DockerDebugConfiguration[] = [];
```

src/scaffolding/wizard/net/NetSdkChooseBuildStep.ts
```diff
@@ -0,0 +1,55 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { IAzureQuickPickItem } from '@microsoft/vscode-azext-utils';
+import * as vscode from 'vscode';
+import { getFromWorkspaceState, updateWorkspaceState } from '../../../utils/StateUtils';
+import { TelemetryPromptStep } from '../TelemetryPromptStep';
+import { NetChooseBuildTypeContext } from './NetContainerBuild';
+
+export const NetContainerBuildOptions = [
+    'Use a Dockerfile',
```
bwateratmsft, line 13: These should be localized

src/scaffolding/wizard/net/NetContainerBuild.ts
```diff
@@ -0,0 +1,32 @@
...
+    ];
+
+    const wizard = new AzureWizard<NetChooseBuildTypeContext>(wizardContext as NetChooseBuildTypeContext, {
+        promptSteps: promptSteps,
+        title: vscode.l10n.t('Initialize for Debugging'),
+    });
+
+    await wizard.prompt();
+    await wizard.execute();
+
```
bwateratmsft, line 31: Nit: extra newline
```suggestion
```

src/tasks/netSdk/NetSdkRunTaskProvider.ts
```diff
@@ -0,0 +1,64 @@
...
+import { DockerTaskProvider } from '../DockerTaskProvider';
+import { DockerTaskExecutionContext } from '../TaskHelper';
+import { NET_SDK_RUN_TASK_SYMBOL, NetSdkTaskHelper } from './NetSdkTaskHelper';
+
+const netSdkDebugTaskName = 'debug';
+export class NetSdkRunTaskProvider extends DockerTaskProvider {
+
+    public constructor(protected readonly helper: NetSdkTaskHelper) { super(NET_SDK_RUN_TASK_SYMBOL, undefined); }
+
+    provideTasks(token: CancellationToken): ProviderResult<Task[]> {
```
bwateratmsft, line 17: Nit for clarity
```suggestion
    public provideTasks(token: CancellationToken): Task[] {
```

src/tasks/netSdk/NetSdkTaskHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+    | 'x64'
+    | 'x86'
+    | 'arm64'
+    | 'arm'
+    | 'ppc64le'
+    | 'mips64'
+    | 's390x'
+    | string;
+
+export const NET_SDK_RUN_TASK_SYMBOL = 'dotnet-container-sdk';
```
bwateratmsft, line 35: Would prefer PascalCase for these constants

src/tasks/netSdk/NetSdkRunTaskProvider.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        // we need to initialize a task first so we can pass it into `DockerPseudoterminal`
+        const task = new Task(
```
bwateratmsft, line 20: This feels weird. Is there no way to use the same task?

src/tasks/netSdk/NetSdkTaskHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+const IMAGE_TAG = 'dev'; // intentionally default to dev tag for phase 1 of this feature
+const ERR_MSG_NO_WORKSPACE_FOLDER = l10n.t(`Unable to determine task scope to execute task ${NET_SDK_RUN_TASK_SYMBOL}. Please open a workspace folder.`);
+
+export class NetSdkTaskHelper {
```
bwateratmsft, line 39: Since this class doesn't implement `TaskHelper` it should be named something else

src/tasks/netSdk/NetSdkTaskHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+        const quotedArgs = Shell.getShellOrDefault().quote(command.args);
+        const commandLine = [client.commandName, ...quotedArgs].join(' ');
+
+        return commandLine;
+    }
+
+    public async inferProjPath(context: IActionContext, folder: WorkspaceFolder): Promise<string> {
+        const noProjectFileErrMessage = l10n.t('No .csproj file could be found.');
+        const item = await quickPickProjectFileItem(context, undefined, folder, noProjectFileErrMessage);
+        return item.absoluteFilePath || '';
```
bwateratmsft, line 91: Why the `|| ''`? Is that to convince TypeScript that there's definitely a value being returned?

src/utils/StateUtils.ts
bwateratmsft: Maybe overkill since it's a one-liner that replaces a one-liner, but I don't feel strongly one way or the other.

resources/netCore/GetProjectProperties.targets
alexyaang: yeah,`$(OutputPath)$(AssemblyName).dll` will remain unchanged

src/debugging/DockerDebugConfigurationProvider.ts
alexyaang: that's a good question. we can do `!debugConfiguration.type`  and make it so if `type` is an empty string and it still triggers it? it just depends on how flexible we want the debug experience to be.

src/tasks/netSdk/NetSdkTaskHelper.ts
alexyaang: my intention was that if `item.absoluteFilePath` is `undefined` then at least it'll return a empty string rather than throw an error? but after looking at `quickPickProjectFileItem` closer, it handles the case where a `fileItem` is not found so I'll remove the `|| ''`. good catch!

src/tasks/netSdk/NetSdkRunTaskProvider.ts
alexyaang: I tried a few ways to rewrite it but I wasn't able to...

src/tasks/netSdk/NetSdkTaskHelper.ts
alexyaang: sounds good!

src/tasks/netSdk/NetSdkTaskHelper.ts
bwateratmsft: Or `NetSdkTaskUtils`? They could also be pure functions instead of class members

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: looks good !

src/debugging/DockerDebugConfigurationProvider.ts
alexyaang: good suggestion, it does work actually

src/tasks/netSdk/NetSdkRunTaskProvider.ts
alexyaang: update: was able to do a little magic to avoid this w/ @bwateratmsft 's help

src/tasks/netSdk/NetSdkTaskHelper.ts
alexyaang: @danegsta any preference on this?

src/tasks/netSdk/NetSdkTaskHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+            default:
+                return architecture;
+        }
+    }
+
+    /**
+     * This methods returns the mount for the remote debugger ONLY as the SDK built container will have
+     * everything it needs to run the app already inside.
+     */
+    private async getRemoteDebuggerMount(): Promise<RunContainerBindMount[] | undefined> {
```
danegsta, line 122: Consider adding a shared utility function that can be re-used both here and in NetCoreTaskHelper.ts which takes an OSType and returns the appropriate debugger, since the logic for setting a debugger path should be the same between the two scenarios.

src/tasks/netSdk/NetSdkTaskHelper.ts
danegsta: I like naming the file either `netSdkTaskUtils` or `netSdkUtils` and agree with Brandon that it makes sense to export the individual functions instead of a class.

src/tasks/netSdk/NetSdkTaskHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+        const volumes: DockerContainerVolume[] = [];
+        const isLinux = await getDockerOSType() === 'linux';
```
danegsta, line 124: I'd suggest using isWindows as the primary condition you check against (similar to the behavior in NetCoreTaskHelper.ts), since there could theoretically be additional Docker OS types, such as darwin, but they will generally match Linux conventions. Treating Windows as the exceptional case is usually safer.

src/tasks/netSdk/NetSdkTaskHelper.ts
```diff
@@ -0,0 +1,138 @@
...
+        const command = await client.runContainer({
+            detached: true,
+            publishAllPorts: true,
+            name: getDefaultContainerName(folderName.name, IMAGE_TAG),
+            environmentVariables: {},
+            removeOnExit: true,
+            imageRef: getDefaultImageName(folderName.name, IMAGE_TAG),
+            labels: defaultVsCodeLabels,
+            mounts: await this.getRemoteDebuggerMount(),
+            customOptions: '--expose 8080',
```
danegsta, line 78: We should add a new `expose` property to `RunContainerCommandOptions` rather than falling back to customOptions. Ideally you could specify `expose: [8080],` here.

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -0,0 +1,52 @@
...
+        return configurations;
+    }
+
+    /**
+     * Checks if the launch task is using the .NET SDK Container build
+     * @param preLaunchTask
+     * @returns true if the launch task is using the .NET SDK Container build
+     *          false otherwise
+     */
+    public isDotnetSdkBuild(preLaunchTask: string): boolean {
```
danegsta, line 47:
```suggestion
    public isDotNetSdkBuild(preLaunchTask: string): boolean {
```

src/tasks/netSdk/NetSdkTaskHelper.ts
alexyaang: will do!

src/tasks/netSdk/NetSdkTaskHelper.ts
alexyaang: I like the idea but after playing around with it, I think this might make the code in `NetCoreTaskHelper.ts` more confusing. We could add a method to `TaskHelper` called  `addDebuggerVolumeWithoutConflict()` and it would make `inferVolumes()` look like this. Which IMO is more confusing than what we have now. 

``` TypeScript
...
addVolumeWithoutConflicts(volumes, appVolume);
addDebuggerVolumeWithoutConflicts(volumes, osType);
if (await fse.pathExists(nugetRootVolume.localPath)) {
    addVolumeWithoutConflicts(volumes, nugetRootVolume);
}
...
```

src/debugging/DockerDebugConfigurationProvider.ts
alexyaang: @danegsta any thoughts on this btw?

src/tasks/netSdk/netSdkTaskUtils.ts
```diff
@@ -0,0 +1,133 @@
...
+    const volumes: DockerContainerVolume[] = [];
+    const isWindows = await getDockerOSType() === 'windows';
+
+    const debuggerVolume: DockerContainerVolume = {
+        localPath: vsDbgInstallBasePath,
+        containerPath: isWindows ? 'C:\\remote_debugger' : '/remote_debugger',
+        permissions: 'ro'
+    };
+
+    addVolumeWithoutConflicts(volumes, debuggerVolume);
```
danegsta, line 129: Why not return a RunContainerBindMount directly? You're always starting with an empty set of volumes here, so addVolumeWithoutConflicts/getMounts aren't really doing anything here.

danegsta: Honestly, I don't love that we go back and forth between DockerContainerVolume and RunContainerBindMount here and in inferVolumes; I think I'd like it more if we normalized the volumes to RunContainerBindMount[] early and did all our volume logic with that type instead.

src/runtimes/docker/contracts/ContainerClient.ts
```diff
@@ -744,6 +744,10 @@ export type RunContainerCommandOptions = CommonCommandOptions & {
      * Optional command to use in starting the container
      */
     command?: Array<string> | string;
+    /**
+     * Optional expose ports for the container
+     */
+    exposes?: Array<number>;
```
bwateratmsft, line 750: I don't feel strongly one way or the other, but maybe "expose"? Most of the other array arguments have a plural, but "exposes" isn't a plural...I don't know that there is a plural of "expose".

danegsta: Maybe exposePorts?

alexyaang: good point! and I think`exposePorts` make the most sense to me

src/runtimes/docker/clients/DockerClientBase/withDockerExposesArg.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { withNamedArg } from "../../utils/commandLineBuilder";
+
+export function withDockerExposesArg(ports?: Array<number>) {
+    return withNamedArg('--expose', ports.toString(), { shouldQuote: false });
```
danegsta, line 9: You'll want to keep the ports argument as an array, otherwise you'll get an error if multiple ports are passed in:

```suggestion
    return withNamedArg('--expose', (ports || []).map(port => port.toString()), { shouldQuote: false });
```

alexyaang: that's right... rookie mistake :disappointed_relieved:

src/debugging/DockerDebugConfigurationProvider.ts
danegsta: What's the actual value of debugConfiguration in the event there isn't an existing launch.json? Is it just `{}`?

src/runtimes/docker/clients/DockerClientBase/withDockerExposesArg.ts
danegsta: It's an easy mistake to make with any kind of array data; happens to us all.

src/debugging/DockerDebugConfigurationProvider.ts
alexyaang: just double checked and yeah exactly, debugConfiguration would be `{}`

bwateratmsft (dismissed):
Looks good to me. @danegsta can you also take a final look?

src/tasks/netSdk/netSdkTaskUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+    const noProjectFileErrMessage = l10n.t('No .csproj file could be found.');
+    const item = await quickPickProjectFileItem(context, undefined, folder, noProjectFileErrMessage);
+    return item.absoluteFilePath;
+}
+
+/**
+ * This method normalizes the Docker OS type to match the .NET Core SDK conventions.
+ * {@link https://learn.microsoft.com/en-us/dotnet/core/rid-catalog}
+ */
+async function normalizeOsToRid(): Promise<'linux' | 'win'> {
```
danegsta, line 94:
```suggestion
async function normalizeOsToRidOs(): Promise<'linux' | 'win'> {
```

src/tasks/netSdk/netSdkTaskUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+    const dockerOsType = await getDockerOSType();
+    return dockerOsType === 'windows' ? 'win' : 'linux';
+}
+
+/**
+ * This method normalizes the native architecture to match the .NET Core SDK conventions.
+ * {@link https://learn.microsoft.com/en-us/dotnet/core/rid-catalog}
+ */
+async function normalizeArchitectureToRid(): Promise<RidCpuArchitecture> {
```
danegsta, line 103:
```suggestion
async function normalizeArchitectureToRidArchitecture(): Promise<RidCpuArchitecture> {
```

src/tasks/netSdk/netSdkTaskUtils.ts
alexyaang: good suggestions

src/tasks/netSdk/netSdkTaskUtils.ts
alexyaang: good suggestion as well

src/tasks/netSdk/netSdkTaskUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+    return item.absoluteFilePath;
+}
+
+/**
+ * This method normalizes the Docker OS type to match the .NET Core SDK conventions.
+ * {@link https://learn.microsoft.com/en-us/dotnet/core/rid-catalog}
+ */
+async function normalizeOsToRid(): Promise<'linux' | 'win'> {
+    const dockerOsType = await getDockerOSType();
+    return dockerOsType === 'windows' ? 'win' : 'linux';
```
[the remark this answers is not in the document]
alexyaang: That's a good point. Let's discuss and make a follow up PR to address the issue.

bwateratmsft (approved):

src/tasks/netSdk/netSdkTaskUtils.ts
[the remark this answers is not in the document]
baronfel: We don't currently support inferring musl-based images in the SDK container tasks but would like to for 7.0.400 or 8.0.100 - if we are able to surface that in the Targets/props would you be able to consume it easily?

Issue #3985, opened 2023-06-23
Add multi-project support for .NET SDK Build


alexyaang:
The current .NET SDK Build feature does not work if you're in a project that have multiple .csproj files. 

If there are multiple `.csproj` files within a workspace folder. We should allow user to choose a `.csproj` file and then debug the project that is associated with that particular .csproj file

PR #3986, opened 2023-06-23
Allow user to pick project to debug for .NET SDK Build

alexyaang:
Closes https://github.com/microsoft/vscode-docker/issues/3985

### Mac
- [x] Single .csproj .NET application with SDK Build without scaffolding 
- [x] Multi .csproj .NET application with SDK Build without scaffolding 
- [x] Single .csproj .NET application with SDK Build with scaffolding 
- [x] Single .csproj .NET application with Nomal Docker Build (check for regression)
- [x] Multi .csproj .NET application with SDK Build with scaffolding 
- [x] Multi .csproj .NET application with SDK Build with old scaffolding in a new Workspace folder
- [x] Single .fsproj .NET application with SDK Build without scaffolding 

### Windows 
- [x] Single .csproj .NET application with SDK Build without scaffolding 
- [x] Multi .csproj .NET application with SDK Build without scaffolding 
- [x] Single .csproj .NET application with SDK Build with scaffolding 
- [ ] Single .csproj .NET application with Nomal Docker Build (check for regression)
- [x] Multi .csproj .NET application with SDK Build with scaffolding 
- [x] Multi .csproj .NET application with SDK Build with old scaffolding in a new Workspace folder
- [x] Single .fsproj .NET application with SDK Build without scaffolding 
- [x] Make sure we handle the case where C# extension is not installed


src/utils/quickPickFile.ts
```diff
@@ -178,3 +178,17 @@ export async function quickPickProjectFileItem(context: IActionContext, fileUri:
     }
     return fileItem;
 }
+
+export async function quickPickCsProjFileItem(context: IActionContext, fileUri: vscode.Uri, rootFolder: vscode.WorkspaceFolder, noProjectFileMessage: string): Promise<Item> {
```
alexyaang, line 182: should we make a new function here that only picks .csproj item? there is an existing method (`quickPickProjectFileItem`) that picks both .csproj and .fsproj right above this method (line 168). We could have an optional parameter in `quickPickProjectFileItem` like `csProjOnly: boolean` to avoid declaring a new method.

i don't feel strongly either way

bwateratmsft: FSProj files might work though we never test with them.

alexyaang: thanks for pointing it out. I just checked with Chet, and turned out it does support .fsproj projects. (and also vbproj)

danegsta: We should definitely use the same logic to pick projects that we do for non-SDK builds

src/tasks/netSdk/netSdkTaskUtils.ts
```diff
@@ -61,18 +56,17 @@ export async function getNetSdkBuildCommand(context: DockerTaskExecutionContext)
...
-        name: getDefaultContainerName(folderName.name, NetSdkDefaultImageTag),
+        name: getDefaultContainerName(context.folder.name, NetSdkDefaultImageTag),
         environmentVariables: {},
         removeOnExit: true,
-        imageRef: getDefaultImageName(folderName.name, NetSdkDefaultImageTag),
+        imageRef: getDefaultImageName(context.folder.name, NetSdkDefaultImageTag),
         labels: defaultVsCodeLabels,
         mounts: await getRemoteDebuggerMount(),
-        exposePorts: [8080], // hard coded for now since the default port is 8080
+        exposePorts: [8080, 80], // the default port is 8080 for .NET 8 and 80 for .NET 7
```
danegsta, line 69: We should consider only exposing ports if the project is targeting web/Asp.NET and it doesn't already expose ports using MSBuild config.

src/tasks/netSdk/netSdkTaskUtils.ts
```diff
@@ -33,25 +33,20 @@ export type RidCpuArchitecture =
...
+    const publishFlag = NetCoreTaskHelper.isWebApp(context.folder.uri.fsPath) ? '-p:PublishProfile=DefaultContainer' : '/t:PublishContainer';
 
     const args = composeArgs(
         withArg('dotnet', 'publish'),
         withNamedArg('--os', await normalizeOsToRidOs()),
         withNamedArg('--arch', await normalizeArchitectureToRidArchitecture()),
         withArg(publishFlag),
         withNamedArg('--configuration', configuration),
-        withNamedArg('-p:ContainerImageName', getValidImageName(folderName.name), { assignValue: true }),
+        withNamedArg('-p:ContainerImageName', getValidImageName(context.folder.name), { assignValue: true }),
```
danegsta, line 49: Change to ContainerRepository since ContainerImageName is deprecated in the image builder targets that will ship with the SDK

src/tasks/netSdk/netSdkTaskUtils.ts
```diff
@@ -33,25 +33,20 @@ export type RidCpuArchitecture =
...
 export async function getNetSdkBuildCommand(context: DockerTaskExecutionContext): Promise<string> {
 
     const configuration = 'Debug'; // intentionally default to Debug configuration for phase 1 of this feature
-    const projPath = await inferProjPath(context.actionContext, context.folder);
-
     // {@link https://github.com/dotnet/sdk-container-builds/issues/141} this could change in the future
-    const publishFlag = NetCoreTaskHelper.isWebApp(projPath) ? '-p:PublishProfile=DefaultContainer' : '/t:PublishContainer';
-
-    const folderName = await quickPickWorkspaceFolder(context.actionContext, ErrMsgNoWorkplaceFolder);
+    const publishFlag = NetCoreTaskHelper.isWebApp(context.folder.uri.fsPath) ? '-p:PublishProfile=DefaultContainer' : '/t:PublishContainer';
```
danegsta, line 41: Not specifically related to this PR, but we should use our MSBuild project config targets to identify Web Apps rather than the current isWebApp method.

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -38,14 +46,70 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
         return configurations;
     }
 
+    public async afterResolveDebugConfiguration(context: DockerDebugContext, debugConfiguration: DockerDebugConfiguration): Promise<void> {
+        const runTasks: Task[] = await tasks.fetchTasks({ type: NetSdkRunTaskType }) || [];
+        const runTaskFirstEntry = runTasks.find(t => t.source === NetSdkRunTaskType);
+        if (!runTaskFirstEntry) {
+            throw Error(l10n.t('No .NET SDK run task found '));
+        }
+        const taskExecution = await tasks.executeTask(runTaskFirstEntry);
```
danegsta, line 55: You can execute a task you define here, rather than having to match a pre-existing task. See: https://github.com/microsoft/vscode-extension-samples/blob/668632196f378491a65240463a16ce09629fe6c8/l10n-sample/src/extension.ts#L16 for an example with a generic shell task.

The idea would be to hardcode the appProject here rather than having to rely on mementos/state in the task itself to retrieve the target project.

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -1,12 +1,18 @@
...
+import { NetSdkRunTaskType, inferProjPath } from "../../tasks/netSdk/netSdkTaskUtils";
+import { PlatformOS } from "../../utils/platform";
 import { unresolveWorkspaceFolder } from "../../utils/resolveVariables";
-import { DockerDebugScaffoldContext } from "../DebugHelper";
+import { DockerDebugContext, DockerDebugScaffoldContext } from "../DebugHelper";
 import { DockerDebugConfiguration } from "../DockerDebugConfigurationProvider";
 import { NetCoreDebugHelper, NetCoreDebugScaffoldingOptions } from "../netcore/NetCoreDebugHelper";
 
-const NetSdkTaskFullSymbol = 'dotnet-container-sdk: debug';
+export const NetSdkProjPathMementoKey = 'netSdkProjPath';
```
danegsta, line 14: I'd prefer not to use a memento to store the target project, to avoid a user getting stuck debugging the wrong project; either a static variable or something in the extension context that will be cleared either when vscode closes or the extension is reloaded.

src/utils/quickPickFile.ts
alexyaang: for non-SDK builds, we can only pick .fsproj and .csproj. but i tested SDK Build with .vbproj and it works too

src/tasks/netSdk/netSdkTaskUtils.ts
alexyaang: was not aware of this, good point

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: the current implementation of memento is that we clear it every time one `f5`s. which means if you choose the wrong project, they can just f5 again and they will get prompt the option to choose a .csproj again. Or even if you choose the project you desire, next time you f5, you're going to get prompted again. 

If we want to change the user experience to only prompt to choose a `csproj` if user `f5`'s on a new vscode session, that's when we should use static variables. So it just depends on which user experience we think is best.

src/debugging/DockerPlatformHelper.ts
```diff
@@ -12,11 +12,10 @@ interface DockerPlatformConfiguration {
     netCore?: unknown;
     node?: unknown;
     python?: unknown;
-    preLaunchTask?: string;
 }
 
 export function getPlatform<T extends DockerPlatformConfiguration>(configuration: T): DockerPlatform | undefined {
-    if (netSdkDebugHelper.isDotNetSdkBuild(configuration?.preLaunchTask) && configuration.netCore !== undefined) {
+    if (netSdkDebugHelper.isDotNetSdkBuild(configuration)) {
```
danegsta, line 18: You don't need this check anymore; if `netSdkDebugHelper.isDotNetSdkBuild` returns true, `configuration.platform === 'netCore'` will also be true.

src/debugging/netSdk/NetSdkDebugHelper.ts
danegsta: In that case, we shouldn't actually need a memento at all if you define the task to run when you call execTask, since you can use that to provide the selected proejct.

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -38,14 +46,70 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
...
+            typeof platformConfiguration === 'object' &&
+            platformConfiguration &&
+            'type' in platformConfiguration &&
+            (platformConfiguration as { type: string }).type === NetSdkRunTaskType
+        ) {
+            return true;
+        } else {
+            return false;
+        }
+    }
```
danegsta, line 108: Since you should only need this in the debug helper (see my comment about removing this check from the task platform check logic), you can simplify by just taking the NetCoreDebugOptions object.

```suggestion
    public isDotNetSdkBuild(netCoreDebugOptions?: NetCoreDebugOptions): boolean {
        return netCoreDebugOptions?.buildWithSdk ?? false;
    }
```

danegsta: Or just accept DockerDebugConfiguration and do `debugConfiguration.netCore?.buildWithSdk ?? false`

src/debugging/DockerPlatformHelper.ts
alexyaang: If this was removed, we wouldn't be able to execute the tasks. Since we still need to use the `NetSdkDebugHelper` instead of the `NetCoreDebugHelper`.

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -38,14 +46,70 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
...
+        const taskExecution = await tasks.executeTask(runTaskFirstEntry);
+
+        const taskEndPromise = new Promise<void>((resolve, reject) => {
+            const disposable = tasks.onDidEndTask(e => {
+                if (e.execution === taskExecution) {
+                    disposable.dispose();
+                    resolve();
+                }
+            });
+        });
```
danegsta, line 64: I'm nervous about possible timing issues here, since we might never detect the task had ended if it failed immediately. Another argument for executing a custom task is that we can implement our own promise that is resolved by the task itself rather than depending on the event system.

bwateratmsft: The listener could be set up before the call to `vscode.tasks.executeTask`

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: yeah i think you might be right

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: the implementation of provideTasks has been changed by suggestions via @danegsta, this shouldn't be a concern anymore (at least i hope 😆 )

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: got rid of momento

src/debugging/DockerPlatformHelper.ts
danegsta: At the very least, you shouldn't multi-purpose isDotNetSdkBuild for both build tasks and debug tasks; the types  and checks are different and they should each have their own implementation to avoid using `unknown` in your types.

alexyaang: yeah, that makes sense to me!

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -38,15 +51,52 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
...
+            platformOS: associatedTask?.dockerRun?.os || 'Linux',
+        };
     }
+
+    public inferDotNetSdkContainerName(debugConfiguration: DockerDebugConfiguration): string {
+        const projFileUri = Uri.file(path.dirname(debugConfiguration.netCore.appProject));
+        return getDefaultContainerName(path.basename(projFileUri.fsPath), "dev");
+    }
+
+    public isDotNetSdkBuild(debugConfiguration: DockerDebugConfiguration): boolean {
```
alexyaang, line 96: should we have this here or `NetCoreDebugHelper.ts`? my thought process was since it is related to .Net SDK debug, it should be in NetCoreDebugHelper. But at the same time, the only time it is being called is in `NetCoreDebugHelper.ts`. So.... idk?

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -187,8 +188,8 @@ export class NetCoreDebugHelper implements DebugHelper {
             throw new Error(l10n.t('Unable to determine assembly output path.'));
         }
 
-        if (netSdkDebugHelper.isDotNetSdkBuild(debugConfiguration.preLaunchTask) && projectInfo.length >= 5) { // if .NET has support for SDK Build
-            if (projectInfo[4] === 'true') { // fifth is whether .NET supports SDK Containers
+        if (netSdkDebugHelper.isDotNetSdkBuild(debugConfiguration) && projectInfo.length >= 5) { // if .NET has support for SDK Build
```
danegsta, line 191: It would probably be cleaner to just override inferAppOutput in NetSdkDebugHelper to implement the SDK specific behavior and rely on polymorphism to apply the correct logic, rather than doing a conditional check here in the base class.

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -92,7 +93,7 @@ export class NetCoreDebugHelper implements DebugHelper {
 
         const additionalProbingPathsArgs = NetCoreDebugHelper.getAdditionalProbingPathsArgs(platformOS);
 
-        const containerAppOutput = netSdkDebugHelper.isDotNetSdkBuild(debugConfiguration?.preLaunchTask)
+        const containerAppOutput = netSdkDebugHelper.isDotNetSdkBuild(debugConfiguration)
             ? appOutput
             : NetCoreDebugHelper.getContainerAppOutput(debugConfiguration, appOutput, platformOS);
```
danegsta, line 98: Can you implement an overridable method in NetCoreDebugHelper that can be used to implement different behavior between standard NetCore and the SDK? That would remove the need to explicitly check for SDK build or not here as the appropriate helper logic could be applied.

src/debugging/netSdk/NetSdkDebugHelper.ts
danegsta: Added a couple comments on the existing usage of isDotNetSdkBuild; it would make sense to slightly refactor the code to use polymorphism to implement the SDK specific behavior rather than having SDK conditioned logic in the base NetCoreDebugHelper class. That would remove the need for isDotNetSdkBuild entirely.

alexyaang: I love the idea!

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -38,14 +53,84 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
...
-     * @param preLaunchTask
-     * @returns true if the launch task is using the .NET SDK Container build
-     *          false otherwise
+     * @returns the project path stored in the static variable projPath if it exists,
+     *          otherwise prompts the user to select a .csproj file and stores the path
+     *          in the static variable projPath
      */
-    public isDotNetSdkBuild(preLaunchTask: string): boolean {
-        return preLaunchTask === NetSdkTaskFullSymbol;
+    public async inferProjPath(context: IActionContext, folder: WorkspaceFolder): Promise<string> {
```
danegsta, line 73: Not strictly necessary, but it can be good to mark the overriding method as `override` to make it more obvious it re-implements a base method.

```suggestion
    public override async inferProjPath(context: IActionContext, folder: WorkspaceFolder): Promise<string> {
```

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -38,14 +53,84 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
...
+        }
+
+        throw new Error(l10n.t('Unable to determine assembly output path.'));
+    }
+
+    public static async clearWorkspaceState(): Promise<void> {
+        await ext.context.workspaceState.update(NetContainerBuildOptionsKey, ''); // clear the workspace state
+        NetSdkDebugHelper.projPath = undefined; // clear the static projPath variable
+        return Promise.resolve();
+    }
```
danegsta, line 125: Is this still necessary?

src/tasks/netSdk/netSdkTaskUtils.ts
```diff
@@ -33,60 +28,54 @@ export type RidCpuArchitecture =
...
+        : '/t:PublishContainer';
 
     const args = composeArgs(
         withArg('dotnet', 'publish'),
         withNamedArg('--os', await normalizeOsToRidOs()),
         withNamedArg('--arch', await normalizeArchitectureToRidArchitecture()),
         withArg(publishFlag),
         withNamedArg('--configuration', configuration),
-        withNamedArg('-p:ContainerImageName', getValidImageName(folderName.name), { assignValue: true }),
+        withNamedArg('-p:ContainerRepository', getValidImageName(context.folder.name), { assignValue: true }),
```
danegsta, line 46: The image name should be based on the appProject folder, not the workspace folder

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: yeah. this method is only called if there was an error (e.g. if they user's msbuild does not support .NET SDK or dotnet publish fails during execution) so the user can get prompted again to choose which way they want to build (.NET SDK or Dockerfile). 

my thought process was that if an user accidentally chooses the wrong project or action, they should be able to f5 again and chooses something that works (Build with Dockerfile instead of .NET SDK)

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: makes sense, i dig it

src/tasks/netSdk/NetSdkRunTaskProvider.ts
```diff
@@ -55,4 +51,55 @@ export class NetSdkRunTaskProvider extends DockerTaskProvider {
...
+        let task: Task;
+        const definition = {
+            ...options,
+            type: NetSdkRunTaskType,
+        };
+
+        const promise = new Promise<number>((resolve, reject) => {
+            task = new Task(
+                definition,
+                this.getProjectFolderFromTask(definition),
```
danegsta, line 65: Folder in relation to tasks is confusing, but it specifically refers to Workspace folders, which are the root working folder you have open in vscode. Technically you can have multiple workspaces open as separate roots, but TaskScope.Workspace indicates a task applies to the current active workspace.

```suggestion
                TaskScope.Workspace,
```

src/tasks/netSdk/NetSdkRunTaskProvider.ts
```diff
@@ -55,4 +51,55 @@ export class NetSdkRunTaskProvider extends DockerTaskProvider {
...
+            const folder = {
+                uri: folderUri,
+                name: path.basename(folderUri.fsPath),
+                index: 1,
+            };
+            return folder;
+        } else {
+            return workspace.workspaceFolders[0];
+        }
+    }
```
danegsta, line 102: We shouldn't need the method as the task should have TaskScope.Workspace set, rather than a specific folder.

resources/netCore/GetProjectProperties.targets
```diff
@@ -10,7 +10,8 @@
 $(TargetFramework)$(TargetFrameworks.Split(';')[0])
 $(OutputPath)$(AssemblyName).dll
 $(ContainerWorkingDirectory)/$(AssemblyName).dll
-$(SDKContainerSupportEnabled)"
+$(SDKContainerSupportEnabled)
+$(EnableSdkContainerSupport)"
```
danegsta, line 14: Extra build property?

src/tasks/netSdk/NetSdkRunTaskProvider.ts
```diff
@@ -3,56 +3,91 @@
...
-            NetSdkDebugTaskName,
-            NetSdkRunTaskType
-        );
-
-        task.execution = new CustomExecution(async (resolvedDefinition: TaskDefinition) =>
-            Promise.resolve(new DockerPseudoterminal(this, task, resolvedDefinition))
-        );
-
-        return [task];
+        return [this.createNetSdkRunTask().task];
```
danegsta, line 24: Thinking about the new flow, we can probably just return [] here for now, since the ambient debug task should handle actually invoking build/run, so we shouldn't need to actually provide an ambient build task.

resources/netCore/GetProjectProperties.targets
alexyaang: yeah.. Chet mentioned you'll need to manually add `<EnableSdkContainerSupport>true</EnableSdkContainerSupport>` to your .csproj for .NET Console apps, for now at least.

@baronfel correct me if i'm wrong

src/tasks/netSdk/NetSdkRunTaskProvider.ts
alexyaang: that makes sense to me, let me do some testing

alexyaang: test results: the method does not get called (breakpoint is not hit) and therefore we should just change it to `return []` and add a comment like `// this task is not discoverable this way`

resources/netCore/GetProjectProperties.targets
danegsta: It should be redundant to the SDKContainerSupportEnabled property. If EnableSdkContainerSupport is false, SDKContainerSupportEnabled should be false as well and if it's true, Microsoft.NET.Build.Containers targets would set the SDKContainerSupportEnabled property to true.

src/utils/netCoreUtils.ts
```diff
@@ -11,11 +11,11 @@ import { ext } from '../extensionVariables';
 import { execAsync } from './execAsync';
 import { getTempFileName } from './osUtils';
 
-export async function getNetCoreProjectInfo(target: 'GetBlazorManifestLocations' | 'GetProjectProperties', project: string): Promise<string[]> {
+export async function getNetCoreProjectInfo(target: 'GetBlazorManifestLocations' | 'GetProjectProperties', project: string, additionalProperties?: string): Promise<string[]> {
     const targetsFile = path.join(ext.context.asAbsolutePath('resources'), 'netCore', `${target}.targets`);
     const outputFile = getTempFileName();
 
-    const command = `dotnet build /r:false /t:${target} /p:CustomAfterMicrosoftCommonTargets="${targetsFile}" /p:CustomAfterMicrosoftCommonCrossTargetingTargets="${targetsFile}" /p:InfoOutputPath="${outputFile}" "${project}"`;
+    const command = `dotnet build /r:false /t:${target} /p:CustomAfterMicrosoftCommonTargets="${targetsFile}" /p:CustomAfterMicrosoftCommonCrossTargetingTargets="${targetsFile}" /p:InfoOutputPath="${outputFile}" ${additionalProperties} "${project}"`;
```
danegsta, line 18: We'll need to use ${additionalProperties || ''} to make sure we don't end up with `unknown` in the string output if additionalProperties is undefined.

resources/netCore/GetProjectProperties.targets
alexyaang: yup, that makes sense!

danegsta (approved):

PR #3993, opened 2023-07-03
.NET SDK Build Prompt For Project File Every Time

alexyaang:
This PR addresses two things:
1. It makes the .NET SDK Feature prompt user every time they f5 for a project file. 
2. It gets the image name from msbuild instead of assuming it from the project name

### Mac
- [x] Single .csproj .NET application with SDK Build without scaffolding 
- [x] Multi .csproj .NET application with SDK Build without scaffolding 
- [x] Single .csproj .NET application with SDK Build with scaffolding 
- [x] Single .csproj .NET application with Nomal Docker Build (check for regression)
- [x] Multi .csproj .NET application with SDK Build with scaffolding 
- [x] Multi .csproj .NET application with SDK Build with old scaffolding in a new Workspace folder

### Windows 
- [x] Single .csproj .NET application with SDK Build without scaffolding 
- [x] Multi .csproj .NET application with SDK Build without scaffolding


src/utils/execAsync.ts
```diff
@@ -5,8 +5,8 @@
 
 import * as cp from 'child_process';
 import * as stream from 'stream';
-import { AccumulatorStream, Shell, spawnStreamAsync, StreamSpawnOptions } from '../runtimes/docker';
 import { CancellationToken } from 'vscode';
+import { AccumulatorStream, Shell, spawnStreamAsync, StreamSpawnOptions } from '../runtimes/docker';
```
alexyaang, line 9: this is just vscode reorganizing imports on same haha

src/utils/getValidImageName.ts
```diff
@@ -16,5 +16,9 @@ export function getValidImageNameFromPath(appPath: string, tag?: string): string
 }
 
 export function getValidImageNameWithTag(nameHint: string, tag: string): string {
-    return `${getValidImageName(nameHint)}:${tag}`;
+    return getImageNameWithTag(getValidImageName(nameHint), tag);
+}
+
+export function getImageNameWithTag(nameHint: string, tag: string): string {
+    return `${nameHint}:${tag}`;
```
alexyaang, line 23: Since Dotnet SDK in handling the image naming, we don't have to normalize it anymore. Instead, we'll just convert it to the image name format "name:tag"

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -56,66 +55,53 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
...
+
+        const { task, promise } = netSdkRunTaskProvider.createNetSdkRunTask(runDefinition);
 
         await tasks.executeTask(task);
 
         try {
             await promise;
         } catch (error) {
-            await NetSdkDebugHelper.clearWorkspaceState();
+            await ext.context.workspaceState.update(NetContainerBuildOptionsKey, '');
```
danegsta, line 75: We shouldn't need to set workspace state at all anymore, should we?

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -124,24 +110,46 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
         return appOutput;
     }
 
-    private inferDotNetSdkContainerName(debugConfiguration: DockerDebugConfiguration): string {
-        const projFileUri = Uri.file(path.dirname(debugConfiguration.netCore.appProject));
-        return getDefaultContainerName(path.basename(projFileUri.fsPath), "dev");
+    private async getProjectInfo(debugConfiguration: DockerDebugConfiguration): Promise<string[]> {
+        if (this.projectInfo !== undefined && this.projectInfo.length >= 6) {
```
danegsta, line 114:
```suggestion
        if (this.projectInfo && this.projectInfo.length >= 6) {
```

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -37,7 +36,7 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
 
         await netContainerBuild(netCoreBuildContext);
         if (netCoreBuildContext?.containerBuildOptions === AllNetContainerBuildOptions[1]) {
-            const appProjectAbsolutePath = await this.inferProjPath(context.actionContext, context.folder);
+            const appProjectAbsolutePath = await this.inferProjPath(context, options);
```
danegsta, line 39: Consider removing options from the arguments to inferProjPath and use it like:

```suggestion
            const appProjectAbsolutePath = options?.appProject || await this.inferProjPath(context);
```

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -56,66 +55,53 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
...
     public async afterResolveDebugConfiguration(context: DockerDebugContext, debugConfiguration: DockerDebugConfiguration): Promise<void> {
-        const { task, promise } = netSdkRunTaskProvider.createNetSdkRunTask(
-            {
-                netCore: {
-                    appProject: await this.inferProjPath(context.actionContext, context.folder),
-                }
+        const projectInfo = await this.getProjectInfo(debugConfiguration);
+        const runDefinition: Omit<NetSdkRunTaskDefinition, "type"> = {
+            netCore: {
+                appProject: await this.inferProjPath(undefined, debugConfiguration.netCore),
```
danegsta, line 61: This code is implicitly relying on the fact that debugConfiguration.netCore.appProject will always be set; otherwise this.inferProjPath would throw an exception due to context being null here. Better to just directly set the value from the configuration.

```suggestion
                appProject: debugConfiguration.netCore.appProject,
```

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -124,24 +110,46 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
...
+            await ext.context.workspaceState.update(NetContainerBuildOptionsKey, '');
+            throw error;
+        }
+
+        if (projectInfo.length < 6 || !projectInfo[5]) {
+            await ext.context.workspaceState.update(NetContainerBuildOptionsKey, '');
+            throw new Error(l10n.t("Your current project configuration or .NET SDK version doesn't support SDK Container build. Please choose a compatible project or update .NET SDK."));
+        }
+
+        this.projectInfo = projectInfo;
```
danegsta, line 135: Consider normalizing the projectInfo array into something easier to parse in code (an object or similar) to avoid potential regressions if the layout of the array ever changes.

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -124,24 +110,46 @@ export class NetSdkDebugHelper extends NetCoreDebugHelper {
         return appOutput;
     }
 
-    private inferDotNetSdkContainerName(debugConfiguration: DockerDebugConfiguration): string {
-        const projFileUri = Uri.file(path.dirname(debugConfiguration.netCore.appProject));
-        return getDefaultContainerName(path.basename(projFileUri.fsPath), "dev");
+    private async getProjectInfo(debugConfiguration: DockerDebugConfiguration): Promise<string[]> {
```
danegsta, line 113: Consider making caching/retrieving project info a concept in the base class so that we can cache the info in either a standard .NET project or an SDK image build project.

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: This is for clearing the cache for "Choose .NET SDK Build vs Dockerfile build." We still need to clear the cache if user runs into an error with the  .NET SDK Build  and wants to use the Dockerfile build

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: Like a dictionary / hashmap?

src/debugging/netSdk/NetSdkDebugHelper.ts
danegsta: One of these days I'll remember that; do consider overriding resolveDebugConfiguration in NetSdkDebugHelper to avoid having to sprinkle these calls to reset the value throughout the debug provider logic. This does assume that everywhere you're currently resetting the state you're also throwing an error:

```
public override async resolveDebugConfiguration(context: DockerDebugContext, debugConfiguration: DockerDebugConfiguration): Promise<ResolvedDebugConfiguration | undefined> {
    try {
        return await super.resolveDebugConfiguration(context, debugConfiguration);
    } catch (error) {
        await ext.context.workspaceState.update(NetcontainerBuildOptionsKey, '');
        throw error;
    }
}
```

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: I see, that looks clean

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: update: i made an interface

package.json
```diff
@@ -1524,6 +1524,18 @@
                         "required": [
                             "appProject"
                         ]
+                    },
+                    "dockerRun": {
+                        "description": "%vscode-docker.tasks.dotnet-container-sdk.dotnetContainerSdk.description%",
```
bwateratmsft, line 1529:
```suggestion
                        "description": "%vscode-docker.tasks.dotnet-container-sdk.dockerRun.description%",
```

The package.json property is still `dockerRun`, so I'd leave that part of the ID unchanged.

package.json
```diff
@@ -1524,6 +1524,18 @@
...
+                        "properties": {
+                            "containerName": {
+                                "type": "string",
+                                "description": "%vscode-docker.tasks.dotnet-container-sdk.dotnetContainerSdk.containerName%"
```
bwateratmsft, line 1533:
```suggestion
                                "description": "%vscode-docker.tasks.dotnet-container-sdk.dockerRun.containerName%"
```

package.nls.json
```diff
@@ -123,6 +123,8 @@
     "vscode-docker.tasks.docker-compose.dockerCompose.envFile.description": "File of environment variables read in and applied to the Docker containers.",
     "vscode-docker.tasks.docker-compose.dockerCompose.files.description": "The docker-compose files to include, in order.",
     "vscode-docker.tasks.docker-compose.dockerCompose.projectName.description": "Alternate project name to use when naming and labeling Docker objects. If using an alternate project name when composing up, the same project name must be specified when composing down.",
+    "vscode-docker.tasks.dotnet-container-sdk.dotnetContainerSdk.description": "Options for running the Docker container used for debugging. Learn more at https://aka.ms/vscode-docker-run-task",
+    "vscode-docker.tasks.dotnet-container-sdk.dotnetContainerSdk.containerName": "Name of the container used for debugging.",
```
bwateratmsft, line 127:
```suggestion
    "vscode-docker.tasks.dotnet-container-sdk.dockerRun.description": "Options for running the Docker container used for debugging. Learn more at https://aka.ms/vscode-docker-run-task",
    "vscode-docker.tasks.dotnet-container-sdk.dockerRun.containerName": "Name of the container used for debugging.",
```

package.nls.json
```diff
@@ -123,6 +123,8 @@
     "vscode-docker.tasks.docker-compose.dockerCompose.envFile.description": "File of environment variables read in and applied to the Docker containers.",
     "vscode-docker.tasks.docker-compose.dockerCompose.files.description": "The docker-compose files to include, in order.",
     "vscode-docker.tasks.docker-compose.dockerCompose.projectName.description": "Alternate project name to use when naming and labeling Docker objects. If using an alternate project name when composing up, the same project name must be specified when composing down.",
+    "vscode-docker.tasks.dotnet-container-sdk.dotnetContainerSdk.description": "Options for running the Docker container used for debugging. Learn more at https://aka.ms/vscode-docker-run-task",
```
bwateratmsft, line 126: I guess we ought to remove the hyperlink since `containerName` is the only option here.

alexyaang: yeah you're right

package.json
alexyaang: that's right

src/debugging/netSdk/NetSdkDebugHelper.ts
```diff
@@ -5,28 +5,33 @@
...
 import { quickPickProjectFileItem } from "../../utils/quickPickFile";
 import { unresolveWorkspaceFolder } from "../../utils/resolveVariables";
-import { DockerDebugContext, DockerDebugScaffoldContext } from "../DebugHelper";
+import { DockerDebugContext, DockerDebugScaffoldContext, ResolvedDebugConfiguration } from "../DebugHelper";
 import { DockerDebugConfiguration } from "../DockerDebugConfigurationProvider";
-import { NetCoreDebugHelper, NetCoreDebugScaffoldingOptions } from "../netcore/NetCoreDebugHelper";
+import { NetCoreDebugHelper, NetCoreDebugScaffoldingOptions, NetCoreProjectProperties } from "../netcore/NetCoreDebugHelper";
 
-export class NetSdkDebugHelper extends NetCoreDebugHelper {
+export interface NetSdkProjectProperties extends NetCoreProjectProperties {
```
bwateratmsft, line 24: Does this need to be exported? Unless I'm overlooking one it looks like all the uses are in this file

src/debugging/netcore/NetCoreDebugHelper.ts
```diff
@@ -35,7 +35,16 @@ export interface NetCoreDebugScaffoldingOptions {
...
 
+export interface NetCoreProjectProperties {
+    assemblyName: string;
+    targetFramework: string;
+    appOutput: string;
+}
+
 export class NetCoreDebugHelper implements DebugHelper {
+
+    protected projectProperties: NetCoreProjectProperties | undefined;
```
bwateratmsft, line 46: I'm not sure that having it as a class member will work in some cases. It is possible to have [multiple debug configurations launch at the same time](https://code.visualstudio.com/docs/editor/debugging#_compound-launch-configurations). If two launch configs were set up for debugging multiple containers (say, with compose up as a pre-launch task), whenever the first one got its project info, the second one would assume it's the same.

Additionally, if the project changes in some substantial way (say, the app output path changes), there's no mechanism to reset the cached project properties.

If possible, I would try to cache the values on the stack (i.e. as local variables) rather than as a class member, so that each time debugging happens, the particular debug call gets the information once.

src/debugging/netSdk/NetSdkDebugHelper.ts
alexyaang: no that's right, it does not need to be exported

bwateratmsft (approved):

PR #3999, opened 2023-07-10
Completed Pre-release items for 1.26 release

alexyaang:


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 1.26.0 - 10 July 2023
+### Added
+* Added support for debugging .NET projects with .NET SDK container build. [#3960](https://github.com/microsoft/vscode-docker/pull/3960), [#3986](https://github.com/microsoft/vscode-docker/pull/3986)
```
bwateratmsft, line 3: I don't have a reason other than "tradition", but in the past I've preferred linking an issue rather than a PR, when an issue exists. My hunch is that issues are interesting to a broader audience than PRs.

alexyaang: Yeah that makes more sense. (I also should've looked at the previous release notes)

PR #4029, opened 2023-08-09
Added Back `createAzureRegistry` command

alexyaang:
- [X] createAzureRegistry


src/commands/registries/azure/deleteAzureRegistry.ts
```diff
@@ -26,7 +26,7 @@ export async function deleteAzureRegistry(context: IActionContext, node?: Unifie
         await azureRegistryDataProvider.deleteRegistry(node.wrappedItem);
     });
 
-    ext.registriesTree.refresh();
+    void ext.registriesTree.refresh();
```
alexyaang, line 29: we don't need to wait here right?

bwateratmsft: Nah

bwateratmsft (approved):

PR #4026, opened 2023-08-08
Registries V2

alexyaang:
TODO: (in no particular order)

## Existing commands that needs to be re-written
- [x] Make all generic registries commands work
  - [x] connectRegistry
  - [x] copyRemoteImageDigest
  - [x] copyRemoteFullTag
  - [x] deleteRemoteImage
  - [x] deployImageToAzure
  - [x] deployImageToAca
  - [X] disconnectRegistryProvider
  - [x] registryHelp
  - [x] logInToDockerCli
  - [x] logOutOfDockerCli
  - [X] pullImageFromRepository
  - [X] pullRepository
  - [x] pushImage
  - [x] reconnectRegistry
---
- [x] MakedockerHub commands work
  - [x] openDockerHubInBrowser
---
- [x] Make all Azure commands work
  - [x] buildImageInAzure
  - [x] createAzureRegistry
  - [x] deleteAzureRegistry
  - [x] deleteAzureRepository
  - [x] openInAzurePortal
  - [x] untagAzureImage
  - [x] viewAzureProperties
---
- [x] Generic V2 Registries
  - [x]  add registry
  - [x]  remove tracked registry
  - [x] https://github.com/microsoft/vscode-docker-extensibility/issues/193
---
## New commands
- [x] Make/Confirm these commands work on Github Registries
  - [x] pushImage
  - [x] copyRemoteFullTag
  - [x] copyRemoteImageDigest
  
---
## User Experience
- [x] Make `registryExperience()` better so that users only get prompted if a the command is possible on a certain registry provider
- [x] Add paging in extensibility package
- [x] In AzureRDP, add a "fire soon" analogue
- [x] Simplify context values https://github.com/microsoft/vscode-docker/pull/4026#discussion_r1313331522
- [x] Recognize `CancellationError` as a cancellation error in azext-utils package
- [x] Release azext-utils package v-2.1.1

---
External things:
- [x] ~Have https://github.com/microsoft/vscode-azuretools/issues/1515 fixed~
- [x] Release extensibility package

PR #4031, opened 2023-08-10
Added More Commands Back

alexyaang:
- [ ] logOutOfDockerCli
 ---
  - [ ] untagAzureImage
  - [ ] viewAzureProperties
  - [ ] buildImageInAzure

### DIFF
--- a/resources/azureSubscription.svg
+++ b/resources/azureSubscription.svg
@@ -1,0 +1,16 @@
+<svg id="bc18bade-5481-447e-a959-659d72346474" 
+    xmlns="http://www.w3.org/2000/svg" width="18" height="18" viewBox="0 0 18 18">
+    <defs>
+        <radialGradient id="a445c717-9d75-44c7-ba6b-0d8f2383e560" cx="-36.63" cy="17.12" r="11.18" gradientTransform="translate(41.88 -7.4) scale(0.94 0.94)" gradientUnits="userSpaceOnUse">
+            <stop offset="0.27" stop-color="#ffd70f"/>
+            <stop offset="0.49" stop-color="#ffcb12"/>
+            <stop offset="0.88" stop-color="#feac19"/>
+            <stop offset="1" stop-color="#fea11b"/>
+        </radialGradient>
+    </defs>
+    <title>Icon-general-2</title>
+    <path id="e3d1e58c-f78e-4fb5-9857-0c9331da9979" d="M13.56,7.19a2.07,2.07,0,0,0,0-2.93h0L10,.69a2.06,2.06,0,0,0-2.92,0h0L3.52,4.26a2.09,2.09,0,0,0,0,2.93l3,3a.61.61,0,0,1,.17.41v5.52a.7.7,0,0,0,.2.5l1.35,1.35a.45.45,0,0,0,.66,0l1.31-1.31h0l.77-.77a.26.26,0,0,0,0-.38l-.55-.56a.29.29,0,0,1,0-.42l.55-.56a.26.26,0,0,0,0-.38L10.4,13a.28.28,0,0,1,0-.41L11,12a.26.26,0,0,0,0-.38l-.77-.78v-.28Zm-5-5.64A1.18,1.18,0,1,1,7.37,2.73,1.17,1.17,0,0,1,8.54,1.55Z" fill="url(#a445c717-9d75-44c7-ba6b-0d8f2383e560)"/>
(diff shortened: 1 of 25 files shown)


alexyaang:
Going into https://github.com/microsoft/vscode-docker/pull/4026

resources/azureSubscription.svg
alexyaang, line 1: added icon

src/commands/registries/azure/tasks/scheduleRunRequest.ts
```diff
@@ -68,7 +68,7 @@ export async function scheduleRunRequest(context: IActionContext, requestType: '
         }
 
         const azureRegistryClient = await createAzureContainerRegistryClient(registry.subscription);
-        const uploadedSourceLocation: string = await uploadSourceCode(azureRegistryClient, registry.subscription.subscriptionId, resourceGroup, rootUri, tarFilePath);
+        const uploadedSourceLocation: string = await uploadSourceCode(azureRegistryClient, registry.label, resourceGroup, rootUri, tarFilePath);
```
alexyaang, line 71: this was a typo from earlier 😞

src/tree/registries/Azure/AzureRegistryDataProvider.ts
```diff
@@ -21,10 +21,14 @@ export interface AzureSubscriptionRegistryItem extends CommonRegistryItem {
     readonly type: 'azuresubscription';
 }
 
-export type AzureRegistry = V2Registry & AzureRegistryItem;
+export type AzureRegistry = V2Registry & AzureRegistryItem & {
+    readonly registryProperties: AcrRegistry;
```
alexyaang, line 25: added this so we can add back the view view azure properties command

src/tree/registries/Azure/AzureRegistryDataProvider.ts
```diff
@@ -63,7 +67,8 @@ export class AzureRegistryDataProvider extends RegistryV2DataProvider implements
                     label: sub.name,
                     type: 'azuresubscription',
                     subscription: sub,
-                    additionalContextValues: ['azuresubscription']
+                    additionalContextValues: ['azuresubscription'],
+                    iconPath: vscode.Uri.joinPath(this.extensionContext.extensionUri, 'resources', 'azureSubscription.svg'),
```
alexyaang, line 71: this is where the icon will be used (subscription tree item)

alexyaang: look something like this
image

resources/azureSubscription.svg
bwateratmsft: We already have this icon. As part of build it gets copied into the resources from the Azure utils package. That happens here: https://github.com/microsoft/vscode-docker/blob/main/webpack.config.js#L74

src/tree/registries/Azure/AzureRegistryDataProvider.ts
```diff
@@ -157,6 +173,22 @@ export class AzureRegistryDataProvider extends RegistryV2DataProvider implements
...
+        const reponse = await registryV2Request({
+            method: 'DELETE',
+            registryUri: item.baseUrl,
+            path: ['v2', '_acr', `${item.parent.label}`, 'tags', `${item.label}`],
+            scopes: [`repository:${item.parent.label}:delete`],
+            authenticationProvider: authenticationProvider,
+        });
+
+        if (!reponse.succeeded) {
+            throw new Error(`Failed to delete repository: ${reponse.statusText}`);
```
bwateratmsft, line 188:
```suggestion
            throw new Error(`Failed to delete tag: ${reponse.statusText}`);

```

bwateratmsft (approved):

src/tree/registries/Azure/AzureRegistryDataProvider.ts
alexyaang: good catch!

alexyaang:
> We already have this icon. As part of build it gets copied into the resources from the Azure utils package. That happens here: https://github.com/microsoft/vscode-docker/blob/main/webpack.config.js#L74 That being said, if the icon is now the only thing we're using from the azureutils package, which might be the case after we move the appsvc deployment out into the appsvc extension, I'd rather dump the azureutils package and keep just the icon.

@bwateratmsft Sounds good. Although I do believe we still need the azure-utils package for creating a new azure registry (unless we want to redo all steps)

resources/azureSubscription.svg
alexyaang: Sounds good. Although I do believe we still need the azure-utils package for creating a new azure registry (unless we want to redo all steps)

bwateratmsft: Good point!

PR #4034, opened 2023-08-14
Added Back `pushImage()`

alexyaang:
Main change:
- [X] pushImage()

Minor change that are unrelated to pushImage():
- added logic for `copy full tag` for images in an azure registry
- added back `images.build`and `images.tag` to `registerCommands.ts` that i previously deleted
- changed function names inside `registryTreeUtils.ts` to be more detailed
- added a method in URTDP that gets all existing registries

Going into https://github.com/microsoft/vscode-docker/pull/4026

TODO before ready for review:
- [x] Check registry logic for GitHub
- [x] Check for things I missed


src/tree/registries/UnifiedRegistryTreeDataProvider.ts
```diff
@@ -130,4 +131,32 @@ export class UnifiedRegistryTreeDataProvider implements vscode.TreeDataProvider<
 
         void this.refresh();
     }
+
+    public async getAllConnectedRegistries(): Promise<UnifiedRegistryItem<CommonRegistry>[]> {
```
bwateratmsft, line 135: I'm a little concerned how the performance of this could be but let's give it a try and see how it goes.

src/tree/registries/registryTreeUtils.ts
```diff
@@ -4,33 +4,36 @@
...
-            const authority = registry.baseUrl.authority;
-            return `${authority.toLowerCase()}/${imageName}`;
+    switch (registry.additionalContextValues?.[0] ?? '') {
+        case 'azureContainerRegistry':
+        case 'genericRegistryV2': {
+            return registry.baseUrl.authority.toLowerCase();
         }
-        case 'dockerHubTag':
+        case 'dockerHubRegistry':
         default:
```
bwateratmsft, line 30: This won't work for GitLab

bwateratmsft (approved):

src/commands/registerCommands.ts
```diff
@@ -50,29 +52,14 @@ import { removeNetwork } from "./networks/removeNetwork";
 import { pruneSystem } from "./pruneSystem";
 import { registerWorkspaceCommand } from "./registerWorkspaceCommand";
 import { createAzureRegistry } from "./registries/azure/createAzureRegistry";
-// import { deleteAzureRegistry } from "./registries/azure/deleteAzureRegistry";
-// import { deleteAzureRepository } from "./registries/azure/deleteAzureRepository";
-// import { deployImageToAzure } from "./registries/azure/deployImageToAzure";
-import { openInAzurePortal } from "./registries/azure/openInAzurePortal";
```
alexyaang, line 56: Removed imports I commented out to make file cleaner

src/tree/registries/Azure/ACROAuthProvider.ts
```diff
@@ -105,6 +106,23 @@ export class ACROAuthProvider implements AuthenticationProvider {
         // Registry scopes, i.e. those passed to `getSession()`, are not valid for acquiring this
         // access token--instead, those only need to be passed to `getOAuthTokenFromRefreshToken()`
         const token = await subscription.credential.getToken([]);
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         return token!.token;
     }
+
```
alexyaang, line 112: just copied over the get refresh token logic from `getSession()` and made it a helper method since it's also used in `getLoginInformation()`

src/tree/registries/UnifiedRegistryTreeDataProvider.ts
alexyaang, line 135: this can be pretty slow if the user does not filter their azure subscriptions, i can definitely optimize this but should we?

src/tree/registries/registryTreeUtils.ts
alexyaang: yeah you're right, putting this on my TODOs

src/tree/registries/UnifiedRegistryTreeDataProvider.ts
alexyaang: yeah me as well, before I was able to filter my subscriptions (there were 70+, and i filtered it to 1), it took well over 15 seconds to load. But I think there is space for optimization.

src/tree/registries/UnifiedRegistryTreeDataProvider.ts
```diff
@@ -132,19 +132,46 @@ export class UnifiedRegistryTreeDataProvider implements vscode.TreeDataProvider<
...
+            const authority = imageBaseName.authority;
+
+            if (authority === 'docker.io') {
+                registryRoots = registryRoots.filter(r => (r.wrappedItem as CommonRegistryRoot).label === 'Docker Hub');
+            }
+            else if (authority.endsWith('azurecr.io')) {
+                registryRoots = registryRoots.filter(r => (r.wrappedItem as CommonRegistryRoot).label === 'Azure');
+                findAzureRegistryOnly = true;
+            }
+            else {
```
bwateratmsft, line 156: You could also cover GitHub here (ghcr.io)

alexyaang: good point

bwateratmsft (approved):

PR #4041, opened 2023-08-25
Added Untag Image Command for Azure RDP

alexyaang:
Related to https://github.com/microsoft/vscode-docker/pull/4026
- [X] untagAzureImage

### DIFF
--- a/package.json
+++ b/package.json
@@ -506,7 +506,7 @@
                 },
                 {
                     "command": "vscode-docker.registries.copyRemoteFullTag",
-                    "when": "view == dockerRegistries && viewItem =~ /(dockerHubTag|genericRegistryV2Tag|azureContainerTag|githubRegistryTag)/i",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -506,12 +506,12 @@
                 },
                 {
                     "command": "vscode-docker.registries.copyRemoteFullTag",
-                    "when": "view == dockerRegistries && viewItem =~ /(dockerHubTag|registryV2Tag|azureContainerTag)/i",
+                    "when": "view == dockerRegistries && viewItem =~ /(dockerHubTag|genericRegistryV2Tag|azureContainerTag|githubRegistryTag)/i",
```
alexyaang, line 509: should we change this to (dockerHubTag|registryV2Tag)?

alexyaang: the advantage would be any RDP class that extends v2 registry would be able to have the command show up

bwateratmsft: Yes, I think that's a good idea.

bwateratmsft (approved):

PR #4057, opened 2023-09-05
Fixed Registries Url Changes 

alexyaang:
These changes fixes the extra steps an user has to go through when pushing an image to a registry. In other words, these changes allow users to skip the step to choose which registry they want to push to if the image name matches a connected registry name. (which was our original behavior)

- Reverted this change https://github.com/microsoft/vscode-docker/pull/4026/commits/b49d849b1d2ca391ba2f26c843f942401f3183c2 as it was causing extra steps for push image
- Changed getBaseUrlFromItem() to getImageNameFromTag()

Tested:
- [X] Github
- [X] Docker Hub 
- [X] Generic V2
- [X] Azure

### DIFF
--- a/src/tree/registries/registryTreeUtils.ts
+++ b/src/tree/registries/registryTreeUtils.ts
@@ -25,9 +25,10 @@ export function getBaseImagePathFromRegistry(registry: CommonRegistry): string {
     } else if (isDockerHubRegistry(registry)) {
         registry = registry as CommonRegistry;
         return registry.label.toLowerCase();
+    } else {
+        registry = registry as CommonRegistry;
+        return registry.baseUrl.authority.toLowerCase();
     }
-
-    throw new Error(l10n.t('The type of registry is not supported. Unable to get base image path'));
 }
 
 export function getFullImageNameFromTag(tag: CommonTag): string {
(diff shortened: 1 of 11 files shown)


src/tree/registries/registryTreeUtils.ts
```diff
@@ -17,33 +17,35 @@ export function getImageNameFromRegistryTagItem(tag: CommonTag): string {
...
 
-export function getBaseUrlFromItem(item: CommonRegistryItem): string {
-    if (!isTag(item) && !isRepository(item) && !isRegistry(item)) {
-        throw new Error(l10n.t('Unable to get base URL'));
+export function getBaseImagePathFromRegistry(registry: CommonRegistry): string {
+    if (!isRegistry(registry)) {
+        throw new Error(l10n.t('Unable to get base image path'));
+    } else if (isAzureRegistry(registry) || isGenericV2Registry(registry) || isGitHubRegistry(registry)) {
+        return registry.baseUrl.authority.toLowerCase();
+    } else if (isDockerHubRegistry(registry)) {
```
bwateratmsft, line 25: This will make it impossible to work for any other registry providers (including GitLab). We should still have an `else` case in this if/else block.

alexyaang: yeah you're right, i'm thinking we should do `registry.baseUrl.authority.toLowerCase();` for the else case

alexyaang:
Gonna close this and do something else

Issue #4056, opened 2023-09-04
Bug: Link to Docker Hub gets confused when image is hash-pinned


pnacht:
When I write a `FROM` command, a Docker Hub link is automatically created pointing to the image.

![FROM cirrusci/windowsservercore:2019](https://github.com/microsoft/vscode-docker/assets/15221358/48f3f10d-d03c-4415-8792-c41c34c57921)

Clicking on the underlined section in the image above, for example, takes me to https://hub.docker.com/r/cirrusci/windowsservercore/.

However, if I decide to hash-pin the image, the parser seems to get confused:

![Screenshot 2023-09-04 at 13 56 21](https://github.com/microsoft/vscode-docker/assets/15221358/3799fa42-9a44-4f23-ad17-c749e99dba83)

We can see the underlined section has moved, and the link now takes me to https://hub.docker.com/_/windowsservercore:2019/, which 404's.

Hash-pinning without the tag works correctly, as before:

![Screenshot 2023-09-04 at 13 56 03](https://github.com/microsoft/vscode-docker/assets/15221358/91e52da0-63d4-4060-b931-429b42cec6e3)

bwateratmsft:
@rcjsuen can you look at this? Thanks!

pnacht:
Not really, but I'm fairly certain it's precisely what you think it is.

Docker allows using both a tag and a digest, but that effectively turns the tag into a "comment". 

When docker actually goes to fetch the image, it'll only care about the digest. [It won't even validate that the tag exists](https://www.mikenewswanger.com/posts/2020/docker-image-digests/#lets-see-it-in-action). But having the tag there makes it easier for humans to understand what exactly is being fetched, and it gives automated tools such as dependabot a better idea of what to track.

rcjsuen:
> Docker allows using both a tag and a digest, but that effectively turns the tag into a "comment".
```Dockerfile
FROM [--platform=<platform>] <image> [AS <name>]
FROM [--platform=<platform>] <image>[:<tag>] [AS <name>]
FROM [--platform=<platform>] <image>[@<digest>] [AS <name>]
```
The documentation mentions the three forms above. I wonder why it does not mention this "fourth" form which is the format in your file.
```Dockerfile
FROM [--platform=<platform>] <image>[:<tag>][@<digest>] [AS <name>]
```
As it is right now your example does not conform to any of the three forms but perhaps that is just an oversight that was missed in the documentation.

rcjsuen:
[...]

Issue #4054, opened 2023-09-02
Dockerfile schema missing support for HEALTHCHECK --start-interval switch


lonix1:
The [`HEALTHCHECK`](https://docs.docker.com/engine/reference/builder/#healthcheck) allows a `--start-interval` switch, which is unrecognised:
> Unknown flag: start-interval dockerfile-utils

vscode 1.80.2
extension 1.26.0
docker 24.0.5

bwateratmsft:
@rcjsuen can you look at this? Thanks!

rcjsuen:
Interestingly enough, this feature is actually not live yet (https://github.com/docker/cli/issues/4486). However, it can be enabled with `#syntax=docker/dockerfile:1` and I have confirmed that to be the case.

I have opened https://github.com/rcjsuen/dockerfile-utils/issues/115 to track this. Thank you for reporting this issue, @lonix1.

Issue #4051, opened 2023-08-31
`ADD --checksum` in `Dockerfile` is not recognized


datorr2:
### Issue

When editing a `Dockerfile`, using the `--checksum` option with `ADD` directive results in the following error:

![image](https://github.com/microsoft/vscode-docker/assets/28566413/ebe767ee-3117-4ccd-9c17-11db54f84429)

The `--checksum` flag was added in Dockerfile v1.6.0 on 2023-06-13.

https://docs.docker.com/build/dockerfile/release-notes/#160

### Applicability

This issue affects `ms-azuretools.vscode-docker` up to v1.26.0 (2023-08-30)

bwateratmsft:
@rcjsuen can you look at this? Thanks!

rcjsuen:
Thank you for reporting this issue, @datorr2. I have verified that this can be enabled with `#syntax=docker/dockerfile:1` and have opened https://github.com/rcjsuen/dockerfile-utils/issues/116 to track this.

Issue #4012, opened 2023-07-20
FROM ${TARGETARCH} wrongly detected as "base name should not be blank"


timon-schelling:
Dockers build in automatic platform args in the global scope should be recognized and should not be annotated as missing.
https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

Example:
```dockerfile
FROM alpine:latest as amd64
COPY /appbin/x86_64-unknown-linux-gnu/bin /app
RUN chmod +x /app

FROM alpine:latest as arm
COPY /appbin/armv7-unknown-linux-gnueabihf/bin /app
RUN chmod +x /app

FROM ${TARGETARCH} as build
FROM scratch
COPY --from=build /app /app
ENTRYPOINT ["/app"]
```
![image](https://github.com/microsoft/vscode-docker/assets/36821505/cf539a96-8f8e-4b46-8d96-430d8718c2ef)

bwateratmsft:
@rcjsuen can you take a look?

rcjsuen:
@timon-schelling Interesting. I guess you are using the `AS` to set something "equal to" `${TARGETPLATFORM}` so that `FROM ${TARGETPLATFORM}` works. Is this correct?

timon-schelling:
@rcjsuen, No ${TARGETPLATFORM} is provided by docker.

See https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

Docker predefines a set of ARG variables with information on the platform of the node performing the build (build platform) and on the platform of the resulting image (target platform). The target platform can be specified with the --platform flag on docker build.

The following ARG variables are set automatically:

- TARGETPLATFORM - platform of the build result. Eg linux/amd64, linux/arm/v7, windows/amd64.
- TARGETOS - OS component of TARGETPLATFORM
- TARGETARCH - architecture component of TARGETPLATFORM
- TARGETVARIANT - variant component of TARGETPLATFORM
- BUILDPLATFORM - platform of the node performing the build.
- BUILDOS - OS component of BUILDPLATFORM
- BUILDARCH - architecture component of BUILDPLATFORM
- BUILDVARIANT - variant component of BUILDPLATFORM

rcjsuen:
> @rcjsuen, No ${TARGETPLATFORM} is provided by docker.

@timon-schelling Right. But to clarify, you can't just use that variable to build, right?

```Dockerfile
FROM ${TARGETARCH} as build
```
```
[...]

PR #4065, opened 2023-09-10
Update Dockerfile language server to 0.11.0

rcjsuen:
This update adds support for a few new flags that have been added to Dockerfile so that they no longer get flagged as errors. The other big change is the introduction of a comment that can be inserted into the Dockerfile to instruct the linter to not flag something as an error. While most errors can be ignored this way, clearly fatal errors that suggest the Dockerfile is completely broken cannot be ignored.

This release fixes the following issues. As always, thank you all for your bug reports and suggestions!
- #4012
- #4051
- #4054
- #4056

Please use the following Dockerfile to test the new features and bug fixes.
```Dockerfile
FROM scratch
# 1. Use code completion at the end. You should now see --start-interval and --start-period.
HEALTHCHECK --start-
# 2. --start-interval used to be flagged as an error. This has been fixed.
# 3. (Issue #4054) Hover over --start-interval and you will get hover documentation help.
HEALTHCHECK --start-interval=5s CMD ls

# 4. You can now use a comment like the above to stop flagging an error.
# Now you can rid yourself of errors without having to wait for me to push out a fix first!
# dockerfile-utils: ignore
UNKNOWN instruction

# Note that the comment must be before the instruction itself rather than the line immediately before the error itself.
# In this case the comment being on line 20 means it will not take effect. Move it to up to before the HEALTHCHECK and it will work.
# Most errors can be ignored but clearly fatal errors (missing FROM) cannot be ignored.

HEALTHCHECK \
    --interval=30s \
    --retries=3 \
    # dockerfile-utils: ignore
    --start-interval-typo=5s \
    --start-period=5s \
    --timeout=30s \
    CMD ls

# 5. (Issue #4051) --checksum is now supported and no longer flagged as an error.
ADD --checksum=sha256:24454f830cdb571e2c4ad15481119c43b3cafd48dd869a9b2945d1036d1dc68d https://mirrors.edge.kernel.org/pub/linux/kernel/Historic/linux-0.01.tar.gz /
# 6. --keep-git-dir is now supported and no longer flagged as an error.
ADD --keep-git-dir=true https://github.com/moby/buildkit.git#v0.10.1 /buildkit

# 7. (Issue #4012) Predefined ARG variables are now assumed to be okay for FROM instructions.
FROM ${TARGETARCH}

# 8. (Issue #4056) The parser now parses tags with digests better. The link should now open correctly instead of going to a 404.
FROM microsoft/dotnet:non-existent-tag@sha256:5483e2b609c0f66c3ebd96666de7b0a74537613b43565879ecb0d0a73e845d7d
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
                 "@microsoft/vscode-azext-utils": "^1.2.2",
                 "@microsoft/vscode-container-client": "^0.1.0",
                 "dayjs": "^1.11.7",
-                "dockerfile-language-server-nodejs": "^0.10.2",
+                "dockerfile-language-server-nodejs": "^0.11.0",
                 "fs-extra": "^11.1.1",
                 "gradle-to-js": "^2.0.1",
                 "handlebars": "^4.7.7",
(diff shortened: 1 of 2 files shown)


alexyaang (approved):
@rcjsuen Awesome work, Remy! :) Thank you for your contribution as always.

rcjsuen:
@alexyaang Thanks for the review and for merging this in! Have a nice day! 👍

PR #4066, opened 2023-09-11
Added fire soon function to AzureRDP

alexyaang:
Related to https://github.com/microsoft/vscode-docker/pull/4026


src/tree/registries/Azure/AzureRegistryDataProvider.ts
```diff
@@ -215,4 +215,14 @@ export class AzureRegistryDataProvider extends RegistryV2DataProvider implements
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         return this.authenticationProviders.get(registryString)!;
     }
+
+    private async fireSoon(element?: CommonRegistryItem | undefined): Promise<void> {
+        return new Promise<void>(resolve => {
```
bwateratmsft, line 220: Doesn't need to be async. Also, store and clear the timeout when it is finished, i.e.
```typescript
const timeout = setTimeout(() => {
    clearTimeout(timeout);
   // ...
}, 5);
```

bwateratmsft (approved):

PR #4073, opened 2023-09-14
Improve User Experience when running registry commands

alexyaang:


src/utils/registryExperience.ts
```diff
@@ -3,22 +3,105 @@
...
+    const pickWithoutProvider = await runQuickPickWizard(context, {
         hideStepCount: true,
         promptSteps: promptSteps,
     });
+
+    // add the provider back to the registry item
+    const pickWithProvider: UnifiedRegistryItem<RegistryItem> = {
+        wrappedItem: pickWithoutProvider,
+        provider: (unifiedRegistryItem as UnifiedRegistryItem<RegistryItem>).provider,
+        parent: undefined, // TODO: figure out how to get the parent
```
alexyaang, line 61: Not exactly sure how to do this. An idea is we could add a method that takes an wrappedItem and the provider and returns the parent of the this wrappedItem. (It creates a new `UnifiedRegistryItem` by keep setting the parent field until it reaches wrappedItem.parent is undefined)

alexyaang: logic similar to this for in a while loop: 

```TS
  return {
      provider: element.provider,
      wrappedItem: e,
      parent: element
  };
```
inspiration:https://github.com/microsoft/vscode-docker/blob/d3da812b31860dc3b9606a92a7e7717d7341efe8/src/tree/registries/UnifiedRegistryTreeDataProvider.ts#L41

src/commands/registries/deleteRemoteImage.ts
```diff
@@ -13,7 +13,11 @@ import { registryExperience } from '../../utils/registryExperience';
 
 export async function deleteRemoteImage(context: IActionContext, node?: UnifiedRegistryItem<CommonTag>): Promise<void> {
     if (!node) {
-        node = await registryExperience(context, ext.registriesTree, { include: 'commontag', exclude: ['githubRegistryTag', 'dockerHubTag'] }, false);
+        node = await registryExperience(context, ext.registriesTree,
+            {
```
bwateratmsft, line 17:
```suggestion
        node = await registryExperience(context, ext.registriesTree, {
```

bwateratmsft: This will also adjust the indentation of the next three lines, should be automatic when you save

src/utils/registryExperience.ts
bwateratmsft: I think it would be better to run this as one single quick pick wizard instead of two separate ones. That would also solve this issue because you'd be picking the URTDP item instead of the wrapped BDP item.

alexyaang:
>I think it would be better to run this as one single quick pick wizard instead of two separate ones. That would also solve this issue because you'd be picking the URTDP item instead of the wrapped BDP item.

Yeah that makes sense

src/commands/registries/azure/viewAzureProperties.ts
```diff
@@ -11,7 +11,10 @@ import { registryExperience } from "../../../utils/registryExperience";
 
 export async function viewAzureProperties(context: IActionContext, node?: UnifiedRegistryItem<AzureRegistry>): Promise<void> {
     if (!node) {
-        node = await registryExperience(context, ext.registriesTree, { contextValueFilter: { include: 'azureContainerRegistry' }, registryFilter: { exclude: [ext.genericRegistryV2DataProvider.label, ext.dockerHubRegistryDataProvider.label, ext.githubRegistryDataProvider.label] } });
+        node = await registryExperience(context, ext.registriesTree, {
```
alexyaang, line 14: example of calling registryExperience() for azure registries

src/commands/registries/deleteRemoteImage.ts
```diff
@@ -13,7 +13,10 @@ import { registryExperience } from '../../utils/registryExperience';
 
 export async function deleteRemoteImage(context: IActionContext, node?: UnifiedRegistryItem<CommonTag>): Promise<void> {
     if (!node) {
-        node = await registryExperience(context, ext.registriesTree, { include: 'commontag', exclude: ['githubRegistryTag', 'dockerHubTag'] }, false);
+        node = await registryExperience(context, ext.registriesTree, {
```
alexyaang, line 16: example for all tags but github and docker tags

src/tree/registries/Azure/AzureRegistryDataProvider.ts
```diff
@@ -79,13 +79,18 @@ export class AzureRegistryDataProvider extends RegistryV2DataProvider implements
...
+                registry.additionalContextValues = ['azure'];
+            });
+            return registries;
         } else {
             const children = await super.getChildren(element);
 
             if ((element as AzureRegistryItem)?.subscription) {
                 children.forEach(e => {
                     e.subscription = (element as AzureRegistryItem).subscription;
+                    e.additionalContextValues = ['azure'];
```
alexyaang, line 93: need to do this for all providers

alexyaang:
Closing this because @bwateratmsft's version is way better

PR #4076, opened 2023-09-18
Improve Registry Experience & Simplify Context Values

alexyaang:
- [x] Make `registryExperience()` better so that users only get prompted if a the command is possible on a certain registry provider
- [x] Simplify context values https://github.com/microsoft/vscode-docker/pull/4026#discussion_r1313331522

### DIFF
--- a/package.json
+++ b/package.json
@@ -542,7 +542,7 @@
                 },
                 {
                     "command": "vscode-docker.registries.azure.viewProperties",
-                    "when": "view == dockerRegistries && viewItem == azure;commonregistry;",
+                    "when": "view == dockerRegistries && viewItem == azure;commonregistry",
                     "group": "regs_zzz_common@2"
                 },
                 {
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -492,7 +492,7 @@
                 },
                 {
                     "command": "vscode-docker.registries.copyImageDigest",
-                    "when": "view == dockerRegistries && viewItem =~ /registryV2Tag/i",
+                    "when": "view == dockerRegistries && viewItem =~ /commontag/i && !(viewItem =~ /commontag;dockerhub/i)",
```
bwateratmsft, line 495:
```suggestion
                    "when": "view == dockerRegistries && viewItem =~ /commontag/i && !(viewItem =~ /commontag;.*dockerhub/i)",
```

bwateratmsft: This applies also to line 515, 520, 525, 530, 540

alexyaang: looks good. we currently know that there isn't any context value between commontag and dockerhub, so is this for if we add something in the future?

bwateratmsft: Yep, exactly

bwateratmsft (approved):

Issue #4087, opened 2023-09-22
[Suggestion] It would be better to add a prompt "No matching resources found" when no registry can be connected


v-ruizh:
**OS:** Win 10
**Build Version:** 20230922.1

**Repro Steps:**
1. Connect all registries (Azure/Docker Hub/GitHub/Generic Registry V2) in REGISTRIES view.
2. Click "Connect Registry..." button again.
3. Check whether there is a prompt "No matching resources found" or not.

**Actual:**
There is no prompt, it keeps in "Select a registry provider to use" step.
![image](https://github.com/microsoft/vscode-docker/assets/45614674/46f8cbd4-316c-42e2-84be-6e96ef3f0685)

**Suggestion:**
It would be better to add a prompt "No matching resources found" when no registry can be connected.

alexyaang:
I like the suggestion. We can add something like this when there's 0 things present. @bwateratmsft, how do you feel?

bwateratmsft:
Agreed, should be easy to add an error or something like that.

Issue #4083, opened 2023-09-22
There is no tooltip for the registry list when pushing image to registry


v-ruizh:
**OS:** Win 10
**Build Version:** 20230922.1
**Regression:** Previous Release

**Repro Steps:**
1. Navigate to Image view -> Right click one image and select "Push..." action.
2. Check whether there is a tooltip for the registry list.

**Expect:**
There is a tooltip for the registry list, such as the below one.
![image](https://github.com/microsoft/vscode-docker/assets/45614674/73e34bef-20e6-4d93-998a-8ae88bb6728f)

**Actual:**
There is no tooltip for the registry list.
![image](https://github.com/microsoft/vscode-docker/assets/45614674/85bc0bea-2c8d-4e8c-b936-e0b2ad957ba5)

**More Info:**
1. This issue also reproduces for the next steps.
2. This issue does not reproduce on the current released Docker extension v1.26.1.

alexyaang:
I feel indifferent about this one as well, @bwateratmsft any preferences?

bwateratmsft:
We should add the tooltip, I don't think it's much work.

alexyaang:
Sounds good

v-ruizh:
This issue also reproduces for the steps after selecting the registry provider, such as "select subscription" option, "select registry" option, "select namespace" and so on.
![image](https://github.com/microsoft/vscode-docker/assets/45614674/9ab79922-934d-4165-a537-6b724abc1068)

alexyaang:
I don't feel strongly about this either way. Having the prompt would be nice. @bwateratmsft what do you think?

dbreshears:
Can revisit if we get compelling feedback,  but seems fairly low priority since it is self explanatory

Issue #4082, opened 2023-09-22
The registries are not listed in alphabetical order


v-ruizh:
**OS:** Win 10
**Build Version:** 20230922.1
**Regression:** Previous Release

**Repro Steps:**
1. Connect all registries (Azure/Docker Hub/GitHub/Generic Registry V2) in REGISTRIES view.
2. Check whether the registries are listed in alphabetical order or not.

**Expect:**
The registries are listed in alphabetical order.

**Actual:**
The registries are not listed in alphabetical order.
![image](https://github.com/microsoft/vscode-docker/assets/45614674/2a3f7cf6-84e7-4961-8a29-af58ade442af)

**More Info:**
This issue does not reproduce on the current released Docker extension v1.26.1.

alexyaang:
I feel indifferent about this but it should be an easy fix if we decide to do this. What are your thoughts? @bwateratmsft

bwateratmsft:
Easy fix, let's do it.

bwateratmsft:
I'd suggest doing it in URTDP.getChildren so that everything is alphabetized

alexyaang:
yeah sounds good

PR #4089, opened 2023-09-22
Minor Registry UX enhancements

alexyaang:
Closes https://github.com/microsoft/vscode-docker/issues/4082
Closes https://github.com/microsoft/vscode-docker/issues/4083
Closes https://github.com/microsoft/vscode-docker/issues/4087

### DIFF
--- a/src/tree/registries/UnifiedRegistryTreeDataProvider.ts
+++ b/src/tree/registries/UnifiedRegistryTreeDataProvider.ts
@@ -127,7 +127,8 @@ export class UnifiedRegistryTreeDataProvider implements vscode.TreeDataProvider<
                 });
             }
 
-            const picked = await vscode.window.showQuickPick(picks, { placeHolder: vscode.l10n.t('Select a registry provider to use') });
+            const pickPrompt = picks && picks.length > 0 ? vscode.l10n.t('Select a registry provider to use') : vscode.l10n.t('No registry providers are available for connection');
+            const picked = await vscode.window.showQuickPick(picks, { placeHolder: pickPrompt });
             if (!picked) {
                 return;
             }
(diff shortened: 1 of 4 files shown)


src/tree/registries/UnifiedRegistryTreeDataProvider.ts
```diff
@@ -74,17 +70,17 @@ export class UnifiedRegistryTreeDataProvider implements vscode.TreeDataProvider<
...
                         wrappedItem: r,
                         parent: undefined
                     };
                 }));
             }
-
-            return unifiedRoots;
         }
+
+        return unifiedRegistryItems.sort((a, b) => (a.wrappedItem as CommonRegistryItem).label.localeCompare((b.wrappedItem as CommonRegistryItem).label));
```
bwateratmsft, line 83: I think we need a more advanced locale compare function. Even though all of our current registry providers use `CommonRegistryItem`, it's not a strict requirement. So what I'd do is try getting their labels if they are `CommonRegistryItems`, but if not, sort by their `toString` results.

alexyaang: yeah i like that!

alexyaang: should we add a `isCommonRegistryItem()` in the package then?

alexyaang: or can we do `instanceof`

bwateratmsft: I'm surprised we haven't already got an `isCommonRegistryItem` method but that would be the way to go.

bwateratmsft (approved):

Issue #4099, opened 2023-09-26
There is no prompt "No matching resources found" when executing "Docker Registries: Disconnect" command from command palette


v-chenzha:
**OS:** All
**Build Version:** 20230926.1
**Regression:** Previous Release

**Repro Steps:**

1. Disconnect all registries (Azure/Docker Hub/GitHub/Generic Registry V2) in REGISTRIES view.
2. F1 -> Docker Registries: Disconnect.
3. Check whether there is a prompt "No matching resources found" or not.

**Expect:**
There is a prompt "No matching resources found".
![image](https://github.com/microsoft/vscode-docker/assets/107091828/fc7be4d8-4c6d-49fb-a694-e0fb4926d23a)

**Actual:**
There is "Connect Registry..." command.
![image](https://github.com/microsoft/vscode-docker/assets/107091828/2d20b12d-02de-4953-b4cc-58f62811474c)

alexyaang:
I actually like the current behavior more. @bwateratmsft what do you think?

bwateratmsft:
It doesn't really make sense to offer a "Connect registry" option when you're in the process of disconnecting

alexyaang:
![image](https://github.com/microsoft/vscode-docker/assets/59073590/b6cdb339-b58f-4dde-b833-ee8c30fffad5)

Would this be okay when you run `Docker Registries: Disconnect.` when no registries are connected?

bwateratmsft:
I think it's fine--not perfect but not a big deal.

Issue #4098, opened 2023-09-26
An error occurs when executing "Docker Registries: Copy Image Digest" command from command palette


v-chenzha:
**OS:** All
**Build Version:** 20230926.1
**Regression:** Previous Release

**Repro Steps:**

1. Disconnect all registries (Azure/Docker Hub/GitHub/Generic Registry V2) in REGISTRIES view.
2. F1 -> Docker Registries: Copy Image Digest -> Select "Connect Registry...".
3. Check whether the next step is "Select the provider for your registry".

**Expect:**
The next step is "Select the provider for your registry".
![image](https://github.com/microsoft/vscode-docker/assets/107091828/7f18539f-3cd0-4273-8bbc-12b7c6d1256d)

**Actual:**
An error occurs.
![image](https://github.com/microsoft/vscode-docker/assets/107091828/965449e8-a30a-4713-a6c3-413caa7cea4a)

**More Info:**
This issue also reproduces when executing "Docker Registries: Copy Full Tag/Delete Image/Deploy Image to Azure App Service.../Deploy Image to Azure Container Apps.../Pull Repository/Pull Image/Log Out of Docker CLI/Log In to Docker CLI" command from command palette.

alexyaang:
That would make sense since these commands need at least one registry to be connected to work. I think we can either 1. throw a better error, or 2.  run the connect registry command.

bwateratmsft:
Let's throw a better error here.

Issue #4097, opened 2023-09-26
There is no connected registry in REGISTRIES view after executing "Docker Registries: Connect to Generic Docker Registry..." command from command palette


v-chenzha:
**OS:** All
**Build Version:** 20230926.1
**Regression:** Previous Release

**Repro Steps:**

1. Disconnect Generic Registry V2 registry in REGISTRIES view.
2. F1 -> Docker Registries: Connect to Generic Docker Registry... ->Enter registry url -> Enter registry username -> Enter registry password.
3. Check whether there is a connected registry in REGISTRIES view.

**Expect:**
There is a connected registry in REGISTRIES view.
![image](https://github.com/microsoft/vscode-docker/assets/107091828/29d4e351-b707-4580-a805-8d1a13bae765)

**Actual:**
There is no connected registry in REGISTRIES view.
![image](https://github.com/microsoft/vscode-docker/assets/107091828/e0ad3c2a-e0d1-41ab-b46c-4b8fbc4f8d2a)

bwateratmsft:
Maybe a missed refresh?

Issue #4094, opened 2023-09-25
There is no action "Connect Registry..." in REGISTRIES view when no registry connected in REGISTRIES view


v-xinda:
**OS:** All
**Build Version:** 20230923.1
**Regression:** Previous Release

**Repro Steps:**
1. Disconnect all registries (Azure/Docker Hub/GitHub/Generic Registry V2) in REGISTRIES view.
2. Check whether there is an action "Connect Registry..." in REGISTRIES view.

**Expect:**
There is an action "Connect Registry..." in REGISTRIES view.
<img width="356" alt="Screenshot 2023-09-25 at 10 11 48" src="https://github.com/microsoft/vscode-docker/assets/57662131/63489f5d-64ee-4d62-bb9b-04201f77c9cf">

**Actual:**
There is no action "Connect Registry..." in REGISTRIES view.
<img width="466" alt="Screenshot 2023-09-25 at 10 06 50" src="https://github.com/microsoft/vscode-docker/assets/57662131/a6a3c236-4d4e-4619-a678-dbb204d0d874">

alexyaang:
@bwateratmsft  We should add this back.

bwateratmsft:
Agreed, I think this is a good feature. Shouldn't be too difficult to add something for this to `URTDP.getChildren()`

PR #4101, opened 2023-09-26
Throw Better Error when running registries commands when No Registries are Connected

alexyaang:
Closes https://github.com/microsoft/vscode-docker/issues/4098
Closes https://github.com/microsoft/vscode-docker/issues/4097
Closes https://github.com/microsoft/vscode-docker/issues/4099

image

Also closes these as we have updated the registries package
Closes https://github.com/microsoft/vscode-docker/issues/4094
Closes https://github.com/microsoft/vscode-docker/issues/4092

TODO: 
- [X] Release new registries package

### DIFF
--- a/src/commands/registries/genericV2/addTrackedGenericV2Registry.ts
+++ b/src/commands/registries/genericV2/addTrackedGenericV2Registry.ts
@@ -8,11 +8,8 @@ import { ext } from "../../../extensionVariables";
 import { UnifiedRegistryItem } from "../../../tree/registries/UnifiedRegistryTreeDataProvider";
 
 export async function addTrackedGenericV2Registry(context: IActionContext, node?: UnifiedRegistryItem<unknown>): Promise<void> {
-    const root = ext.genericRegistryV2DataProvider.getRoot();
-    const registries = await ext.genericRegistryV2DataProvider.getRegistries(root);
-
-    if (registries?.length > 0) {
-        // if there are already registries, add a new registry to the existing root node
+    // if there are already registries, add a new registry to the existing root node
+    if (ext.genericRegistryV2DataProvider.hasTrackedRegistries()) {
         await ext.genericRegistryV2DataProvider.addTrackedRegistry();
     } else {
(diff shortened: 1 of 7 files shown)


src/commands/registries/genericV2/addTrackedGenericV2Registry.ts
```diff
@@ -8,7 +8,17 @@ import { ext } from "../../../extensionVariables";
...
-    await ext.genericRegistryV2DataProvider.addTrackedRegistry();
+    const root = ext.genericRegistryV2DataProvider.getRoot();
+    const registries = await ext.genericRegistryV2DataProvider.getRegistries(root);
+
+    if (registries?.length > 0) {
+        // if there are already registries, add a new registry to the existing root node
+        await ext.genericRegistryV2DataProvider.addTrackedRegistry();
+    } else {
+        // if there are no registries, connect as usual
+        await ext.registriesTree.connectRegistryProvider(ext.genericRegistryV2DataProvider);
```
bwateratmsft, line 19: Is `connectRegistryProvider` idempotent? If so we could just unconditionally call it. If not then maybe it should be...

alexyaang: yeah i do like that more. we can do this by adding a condition here that checks if `provider` is `GenericV2RegistryProvider`, which is the edge case here. for all other cases, we shouldn't call `onConnect` again, but for GenericV2, we can. 

https://github.com/microsoft/vscode-docker/blob/f53de2c6b1cb47c563a81ece4b398a15d07b6324/src/tree/registries/UnifiedRegistryTreeDataProvider.ts#L164

another approach is to just remove the if statement entirely, which would rely on the quickpicks above to not show registry providers that are already connected, (which i think is reliable enough)

bwateratmsft (dismissed):
Looks good. I'll have to reapprove once we get the registries package updated on NPM and in this PR.

bwateratmsft (approved):

PR #4134, opened 2023-10-23
Add a way to view image manifest

### LINKED ISSUE #3996: Displaying remote image manifest similarly to inspecting local images
Issue Type: **Feature Request**

Issue is about adding an additional option to RMB context menu when clicking a specific tag of image on remote repository.

For example, under "Copy Image Digest" might be option called "Manifest", there could also be command for that - "Docker Registries: Manifest".

image
_(Current context menu)_

The idea is, that for now it is impossible to _inspect_ a image without pulling it first. This command would allow getting more information about remote images without doing that.

After running the command the manifest would open similarly as the result of running "Inspect" on local image:

image
_(Visualisation how this could look like)_

When it comes to implementing it, self-hosted repositories allow it simply by doing a specific GET request to API v2, i do not know how it works in Azure, GitLab and Docker HUb, but I believe they also allow is one way or another.

bwateratmsft:
Closes #3996

Depends on:
- [x] https://github.com/microsoft/vscode-docker-extensibility/pull/213
- [x] Release above to NPM
- [x] Update `package-lock.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -500,9 +500,14 @@
                     "group": "regs_tag_1_general@3"
                 },
                 {
+                    "command": "vscode-docker.registries.inspectRemoteImageManifest",
+                    "when": "view == dockerRegistries && viewItem =~ /commontag/i && !(viewItem =~ /commontag;.*dockerhub/i)",
+                    "group": "regs_tag_1_general@4"
+                },
+                {
                     "command": "vscode-docker.registries.deployImageToAzure",
                     "when": "view == dockerRegistries && viewItem =~ /commontag/i",
-                    "group": "regs_tag_1_general@4"
+                    "group": "regs_tag_1_general@5"
                 },
                 {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -499,10 +499,15 @@
                     "when": "view == dockerRegistries && viewItem =~ /commontag/i && !(viewItem =~ /commontag;.*dockerhub/i)",
                     "group": "regs_tag_1_general@3"
                 },
+                {
+                    "command": "vscode-docker.registries.inspectRemoteImageManifest",
+                    "when": "view == dockerRegistries && viewItem =~ /commontag/i && !(viewItem =~ /commontag;.*dockerhub/i)",
```
bwateratmsft, line 504: This is not working for GitHub. Either needs to be disabled for GitHub or fixed.

bwateratmsft: Nevermind...weird...

PR #4177, opened 2023-11-27
Fix `streamToFile` implementation

### LINKED ISSUE #4176: Debugging busted by update to Node 18
Debugging was busted by the update to Node 18 in #4167. The `streamToFile` function never resolves, it looks like the write stream doesn't even get closed.

bwateratmsft:
Fixes #4176. I'm not sure why this worked when I originally tried it but it works for sure now.


danegsta (approved):

Ross1774:
Thanks

PR #4234, opened 2024-02-07
Scaffold Python Dockerfile with a floatier tag

bwateratmsft:
Uses a floatier tag for scaffolding Python apps. So we don't have to update this every year when a new Python release comes out.


karolz-ms (approved):

bwateratmsft:
This should be released now in version Docker extension version 1.29.1.

PR #4268, opened 2024-03-23
Ignore commented <UserSecretsId> tag in .csproj

FiftyNine:
Currently the extension employs this simple regex /UserSecretsId/i to detect if .NET project file uses values stored in Secret Manager. If this string is found in the project file, than the "docker-run" task generates a command that attempts to mount the secrets folder onto the container. If this folder does not exist than the command fails.
Problem is that this regex doesn't consider that the tag may be commented out.
I propose removing all commented sections from the .csproj contents before testing UserSecretsId regex against it.


bwateratmsft (approved):
This looks good, thank you @FiftyNine!

bwateratmsft:
This should be released now in version Docker extension version 1.29.1.

PR #4342, opened 2024-07-17
Update dependencies

### LINKED ISSUE #4330: Update `@vscode/extension-telemetry` dependency
Requested by telemetry team to do this

danegsta:
Update several outdated dependencies to latest versions. Resolves #4330.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,10 +13,10 @@
                 "@azure/arm-containerregistry": "^10.1.0",
                 "@azure/storage-blob": "^12.14.0",
                 "@microsoft/compose-language-service": "^0.2.0",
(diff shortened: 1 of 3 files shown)


bwateratmsft (approved):
Did the VSIX grow or shrink with the new packages? Probably grew but how much would be interesting to know.

danegsta:
> Did the VSIX grow or shrink with the new packages? Probably grew but how much would be interesting to know.

Seems like it might be as much as 1MB more, which isn't great, but that's comparing the vsix from a pipeline build to a local dev build, so take with a grain of salt.

danegsta:
In the official build it actually shrunk by about 0.01 MB, so looks to be basically the same size as before.

Issue #2128, opened 2020-07-05
Reconsideration of Dockerfile.* file association


simpleapples:
I have a file named `dockerfile.go`, which is recognised as a `Dockerfile` because of #1908 .

I know this decision has a point #1907 , but I think it's unresonable. When a `Dockerfile` is recognised as `Unknown`, nothing will happen. But when a `go` file is recoginsed as `Dockerfile`, VSCode will format it as `Dockerfile`. (Especially when I choose `Format on save`, I think it's widely used.). As you can see on the screenshot below.

When it happend, I have to revert the code and add a *.go file association. It's a heavier action than just adding a dockerfile associaiton.

So, I think it necessary to remove `Dockerfile.*`  file association.

![Annotation 2020-07-05 114338](https://user-images.githubusercontent.com/1655036/86525198-a46aa800-beb6-11ea-86c6-10b6fdf52991.jpg)

bwateratmsft:
There were overwhelmingly (about 50x) more people with actual Dockerfiles named `Dockerfile.***`, than non-Dockerfiles named that, so we plan to keep #1908. You can override this in your project with something like this in settings.json:

```json
    "files.associations": {
        "dockerfile.go": "go"
    }
```

simpleapples:
I read issue #1907 and I know there are a lot of people use `Dockerfile.***`, but it can be covered by adding `Dockerfile.prod`, `Dockerfile.dev` and so on. You can cater the large group by adding `Dockerfile.xxx` as more as possible. But it does not mean you you can hurt the small group.

Besides, the pattern `Dockerfile.*` mess up the file association of other extension and cause conflicts.

bwateratmsft:
> it can be covered by adding `Dockerfile.prod`, `Dockerfile.dev` and so on.

I saw hundreds of different extension endings on searches in GitHub and in telemetry; it is not sustainable to try to add them case-by-case.
 
> Besides, the pattern `Dockerfile.*` mess up the file association of other extensions and cause conflicts.

Can you elaborate? How does setting a language association affect other extensions?

simpleapples:
> Can you elaborate? How does setting a language association affect other extensions?

For example, I install `go extension` and `dockerfile.go` is recognized as `go file`, but it will be recognized as a `Dockerfile` after installing `docker extension`.

bwateratmsft:
I see what you mean. Can you try the language association? That should make `dockerfile.go` be treated as a Go file.

[...]

PR #4363, opened 2024-08-14
fix: incorrect dockerfile pattern

Yeuoly:
When there is a normal code file named dockerfile.go dockerfile.py etc. it will be considered as a dockerfile instead of code file

### reproduce
As shown below, a dockerfile.go was detected as dockerfile, but actually it's not.
image


karolz-ms:
This has been discussed in the past https://github.com/microsoft/vscode-docker/issues/2128

dbreshears:
Please check out this comment as a workaround.  https://github.com/microsoft/vscode-docker/issues/2128#issuecomment-654237998.   We don't have plans to take this PR based on the past investigations.

PR #4380, opened 2024-09-10
Add release pipeline

alexweininger:
To make a release, you'll manually run the pipeline and need to fill out the parameters:
image

A version string is required at pipeline run time that will be matched against the version present in the package.json downloaded from the build pipeline artifacts. The "Dry Run" parameter will skip the last step that actually publishes the extension.

Under Resources you can change target build pipeline. It defaults to the latest successful build.

Releases will need to be approved by one member of the deployment environment. View the environment I created: https://dev.azure.com/devdiv/DevDiv/_environments/1258?view=checks. For now I've left the "allow self approval" setting enabled.

### DIFF
--- a/.azure-pipelines/release.yml
+++ b/.azure-pipelines/release.yml
@@ -1,0 +1,36 @@
+trigger: none # Only run this pipeline when manually triggered
+
+parameters:
+  - name: publishVersion
+    displayName: Version to publish
+    type: string
+  - name: dryRun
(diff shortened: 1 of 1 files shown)


alexweininger:
Pipeline has been created https://dev.azure.com/devdiv/DevDiv/_build?definitionId=23635

bwateratmsft:
Thanks so much @alexweininger!

PR #4418, opened 2024-11-20
Add tsaoptions.json

bwateratmsft:


dbreshears (changes requested):
I would think we want to remove the codebaseAdmins.  I don't see that used in some of the other dotnet repos?  Is it required?

dbreshears (approved):

Issue #4458, opened 2025-01-09
dockerCompose down task does not have a profiles argument


kpaxton:
My docker-compose file uses profiles to load group related services.In VSCode I'm able to create a task using the `docker-compose` type and can configure the `up` command to use a specific profile. However, the `down` task doesn't allow for profiles and therefore the `docker compose down` command doesn't stop the services that were started under the profile. 

Can profile support be added to the docker-compose down task configuration?

PR #4467, opened 2025-01-14
add profile support for docker-compose down

PotatoZhou:
for issue #4458


PotatoZhou:
@microsoft-github-policy-service agree

bwateratmsft:
Thanks for the contribution @PotatoZhou! We're right in the middle of releasing 1.29.4 so we're going to hold off on this PR until that release is done, but at a glance everything looks good.

bwateratmsft:
So this is different from what #4458 is asking for--that issue is asking for the `docker-compose` `down` VSCode task to support profiles. Your PR is still valid but is adding a customizable command instead of enhancing the `docker-compose` `down` task.

Also, we've made some changes in this area recently that would affect your PR--can you merge main and update `compose.ts`' `compose()` function? It now references `'upSubset'` directly.

PotatoZhou:
sure, will do

bwateratmsft (changes requested):
I don't want to change the existing Compose Down command. If your goal is to fix #4458, you want to do the VSCode task, not the customizable command. The code for that is partly in package.json's `contributes.taskDefinitions -> docker-compose` and partly in src/tasks.

PotatoZhou:
correct me if I am wrong :) I believe that the client.down (in DockerComposeTaskProvider.ts) does not support profiles, I checked [ContainerOrchestratorClient.ts of vscode-container-client](https://github.com/microsoft/vscode-docker-extensibility/blob/main/packages/vscode-container-client/src/contracts/ContainerOrchestratorClient.ts) and in type DownCommandOptions, "profiles" is not presented, Should I report this as an issue over there?

saltman007web (approved):

saltman007web (approved):
commit

bwateratmsft:
@PotatoZhou you are right, good catch. We will need to add support for profiles in the compose down function in `ContainerOrchestratorClient`. Can you file an issue on https://github.com/microsoft/vscode-docker-extensibility/issues?

bwateratmsft (commented):
These changes look good. I'll try to cut a release of the `@microsoft/vscode-container-client` package soon.

sovtig (approved):

bwateratmsft:
@PotatoZhou `@microsoft/vscode-container-client` version 0.2.0 is available with the changes to support profiles in `compose down` commands. You can update your PR to take advantage of that.

bwateratmsft (approved):
Looks good, thank you for the contribution @PotatoZhou!

Issue #4408, opened 2024-10-29
Wrong heredoc kills Docker LS


danCrespo:
Type: <b>Performance Issue</b>

### Steps to reproduce.
- You can start writing a *Dockerfile*
- Use a `RUN` instruction (not sure if the issue also happens with any other instruction)
- Start the command using a here-string `<<<CRASH `, as if it were a heredoc. (I already know this sintax is wrong and invalid)
- Enter a new line and type anything...
- Close the wrong heredoc `CRASH` and...
that's when the LS is killed by this wrong sintax.

Extension version: 1.29.3
VS Code version: Code 1.94.2 (384ff7382de624fb94dbaf6da11977bba1ecd427, 2024-10-09T16:08:44.566Z)
OS version: Windows_NT x64 10.0.22635
Modes:
Remote OS version: Linux x64 5.15.153.1-microsoft-standard-WSL2

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|11th Gen Intel(R) Core(TM) i9-11900 @ 2.50GHz (16 x 2496)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|31.21GB (4.34GB free)|
|Process Argv|--folder-uri=vscode-remote://wsl+Ubuntu-22.04/home/dancr/repos/panacea-lg-scast --remote=wsl+Ubuntu-22.04 --crash-reporter-id ce7ffe95-f363-4111-947c-31b32712f148 --crash-reporter-id ce7ffe95-f363-4111-947c-31b32712f148|
|Screen Reader|no|
|VM|0%|

|Item|Value|
|---|---|
|Remote|WSL: Ubuntu-22.04|
|OS|Linux x64 5.15.153.1-microsoft-standard-WSL2|
|CPUs|11th Gen Intel(R) Core(TM) i9-11900 @ 2.50GHz (16 x 0)|
[...]

PR #4553, opened 2025-04-11
Update the Dockerfile Language Server to 0.14.0

rcjsuen:
This update includes some general fixes as well as one to prevent the language server crashing (#4408). It also includes an update to how we handle links so that we no longer just blindly route everything to Docker Hub.

Please use the Dockerfile below to reproduce the crash and verify the fix.
```Dockerfile
FROM scratch
RUN <<<CRASH
CRASH

FROM scratch
```
Please use the Dockerfile below to test the other fixes and the update to the URL links.
```Dockerfile
# Ctrl+Clicking on alpine on line 3 should not jump to line 4,
# they're not referencing the same thing
FROM alpine
FROM hello AS alpine

# click on "base" on lines 11, 12, and 14
# only lines 11 and 14 will be highlighted,
# the new release highlights line 12 as well,
# then try repeating the steps with renaming,
# same problem, same fix
FROM busybox AS base
FROM base
FROM busybox
COPY --from=base . .

# links should go to the right place instead of all jumping to Docker Hub
# Docker Hub
FROM busybox
# GitHub Container Registry
FROM ghcr.io/super-linter/super-linter
FROM ghcr.io/super-linter/super-linter:latest-buildcache
# Microsoft Artifact Repository
FROM mcr.microsoft.com/powershell
FROM mcr.microsoft.com/powershell:3.17
FROM mcr.microsoft.com/windows/servercore
FROM mcr.microsoft.com/windows/servercore:ltsc2025
# Quay.io
FROM quay.io/prometheus/node-exporter:v1.9.1
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
                 "@microsoft/vscode-container-client": "^0.2.1",
                 "@microsoft/vscode-docker-registries": "^0.2.0",
                 "dayjs": "^1.11.7",
-                "dockerfile-language-server-nodejs": "^0.13.0",
+                "dockerfile-language-server-nodejs": "^0.14.0",
                 "fs-extra": "^11.1.1",
                 "gradle-to-js": "^2.0.1",
                 "handlebars": "^4.7.7",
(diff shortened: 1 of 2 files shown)


karolz-ms:
Thanks Remy 🤗

rcjsuen:
Thanks for the review, Karol! This release also fixes another crash. I don't believe it was reported here but I thought I would mention it anyway. :)
```Dockerfile
RUN <<'END_OF_PATCH
END_OF_PATCH
RUN <<'SCRIPT'
SCRIPT
```

PR #4562, opened 2025-04-22
Add a toast to direct users to Container Tools

bwateratmsft:


package.json
```diff
@@ -2990,9 +2990,6 @@
         "vscode.docker",
         "vscode.yaml"
     ],
-    "extensionPack": [
-        "docker.docker"
-    ],
```
bwateratmsft, line 2995: @rcjsuen I wanted to call your attention to this. Removing Docker DX from the extension pack will mean that, once this is released, Docker DX will no longer be automatically installed alongside Microsoft's Docker extension. _However_, if we were to leave it in the extension pack, then when the Docker extension gets uninstalled (as part of this toast notification), it takes Docker DX with it. We anticipate a lot of users opting in to Container Tools, so to leave it would mean a lot of Docker DX uninstalls.

src/utils/migrateToContainerTools.ts
```diff
@@ -0,0 +1,74 @@
...
+        } else {
+            // User closed the notification
+            context.telemetry.properties.choice = 'closed';
+            return false;
+        }
+    });
+
+    if (shouldInstall) {
+        // Install the Container Tools extension
+        await vscode.commands.executeCommand('workbench.extensions.installExtension', 'ms-azuretools.vscode-azureresourcegroups'); // TODO: Update to the correct extension ID for Container Tools
```
bwateratmsft, line 66: TODO: Do

karolz-ms (approved):

fiveisprime:
I think the first two sentences can be combined to make it a little more concise. How about this?

>The Docker extension is becoming the Container Tools extension, which you can install and try now. Installing the Container Tools extension will uninstall the Docker extension and the window will be reloaded. Make sure to save your changes before proceeding.

fiveisprime (approved):

PR #4691, opened 2026-07-16
Add Dockerfile for grok2api container image

valitacrossdress-oss:
Add a Dockerfile that uses ghcr.io/chenyme/grok2api:latest as a base image and copies a local configuration file into the container.


bwateratmsft:
Not relevant and probably spam

